PR #21, opened 2017-07-27
Yr annotations only

YanayRosen:
Adding user annotation creation, no sharing yet.


app/models/fire_cloud_client.rb
```diff
@@ -17,6 +17,8 @@ class FireCloudClient < Struct.new(:access_token, :api_root, :storage, :expires_
 	PORTAL_NAMESPACE = 'single-cell-portal'
 	# location of Google service account JSON (must be absolute path to file)
 	SERVICE_ACCOUNT_KEY = File.absolute_path(ENV['SERVICE_ACCOUNT_KEY'])
+	#location of Google service account JSON (must be absolute path to file)
```
bistline: This is a regression and should not be committed - please remove.  It should load from the environment variable.

app/controllers/user_annotations_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+    #If the labels sued to include undefined, make sure they do again
+    if @user_annotation.values.include? 'Undefined'
+      new_labels.push('Undefined')
+    end
+
+    #Remeber the old values
+    old_labels = @user_annotation.values
+
+    #Remeber the old annotations
+    annotation_arrays = @user_annotation.user_data_arrays.where(array_type: 'annotations').to_a
```
bistline: You can call @user_annotation.user_data_arrays.by_name_and_type(name, array_type) instead.

app/models/user_annotation.rb
```diff
@@ -0,0 +1,291 @@
...
+    y_arrays = y_array.each_slice(UserDataArray::MAX_ENTRIES).to_a
+    cell_name_arrays = cell_name_array.each_slice(UserDataArray::MAX_ENTRIES).to_a
+
+    #create each data array
+    annotation_array.each_slice(UserDataArray::MAX_ENTRIES).each_with_index do |val, i|
+      #if threshold exists then the subsample annotation and threshold need to be set on the created data array
+      if !threshold.nil?
+        #Create annotation array
+        UserDataArray.create(name: name, array_type: 'annotations', values: val, cluster_name: cluster.name, array_index: i+1, subsample_threshold: threshold, subsample_annotation: sub_an, user_id: self.user_id, cluster_group_id: self.cluster_group_id, study_id: self.study_id, user_annotation_id: id)
+        logger.info('created annotation')
```
bistline: These logging statements are uninformative - you should always have a timestamp with Time.now and say with much more detail what you are doing.  This will be needed if we are trying to debug issues in production.  Look at the types of logging statements in the Study class when we are parsing uploaded files to get an idea of how to reformat these.

app/models/user_annotation.rb
```diff
@@ -0,0 +1,291 @@
...
+  end
+
+  #figure out what level this annotation was created at
+  #if you update an annotation's 'Undefined' level, than it will be registered as created at full data
+  def subsampled_at
+    #get all the annotation arrays
+    annotations = user_data_arrays.find_by(array_type: 'annotations').to_a
+
+    #get the labels of this annotation
+    vals = values
```
bistline: This association is ambiguous - is 'values' the attribute 'values' for the instance of UserAnnotation?  If so, this needs to be written as self.values.

app/models/user_annotation.rb
```diff
@@ -0,0 +1,291 @@
...
+
+    #so far we don't think undefined exists at these subsamplings
+    #undefined can never exist at a subsampling level of 1k, because it is the smallest level possible
+    undefined_exists_at_10k = false
+    undefined_exists_at_20k = false
+
+    #check were iundefined exists
+    if max_length > 10000
+      #because max_length > 10k, it's possible that undefined exists
+      #check if the annotation array has undefined
```
bistline: Same thing here - this should be self.user_data_arrays

app/models/user_annotation.rb
```diff
@@ -0,0 +1,291 @@
...
+      elsif max_length < 20000 and max_length > 10000
+        if undefined_exists_at_10k
+          #undefined exists at 10k and 20k, so must have been created at 1k
+          created_at = 'Created at a subsample of 1,000 Cells'
+        else
+          #undefined exists at 20k but not 10k, so was created at 10k
+          created_at = 'Created at a subsample of 10,000 Cells'
+        end
+      elsif max_length < 10000 and max_length > 1000
+        #max length is less than 10k. Max length > 1k. Undefined exists, so must be created at 1k.
```
bistline: More unhelpful logging statements - please refine.

app/models/study.rb
```diff
@@ -89,6 +89,10 @@ def are_synced
     end
   end
 
+  #User annotations are per study
```
bistline: We should add delete statements for user_annotations and user_data_arrays into Study.delete_queued_studies.  This will improve study deletion performance.

app/controllers/site_controller.rb
```diff
@@ -863,8 +1007,9 @@ def load_gene_set_expression_boxplot_scores(annotation, consensus, subsample_thr
         end
       end
       # remove any empty values as annotations may have created keys that don't exist in cluster
-      values.delete_if {|key, data| data[:y].empty?}
+      logger.info("Look: #{annotation[:scope]}")
```
bistline: Please remove

app/controllers/site_controller.rb
```diff
@@ -925,8 +1080,9 @@ def load_gene_set_expression_data_arrays(annotation, consensus, subsample_thresh
   def initialize_plotly_objects_by_annotation(annotation)
     values = {}
     annotation[:values].each do |value|
-      values["#{value}"] = {y: [], name: "#{value}" }
+      values["#{value}"] = {y: [], cells: [], annotations: [], name: "#{value}" }
     end
+    logger.info("Values: #{values}")
```
bistline: Another logging statement to remove.

app/views/site/_search_options.html.erb
```diff
@@ -1,173 +1,188 @@
...
-							$('#expression').change(function() {
-								  if ($(this).val() != '') {
+<div id="search-parent" class="panel-group">
+  <div class="panel panel-info" id="search-options-panel">
+    <div class="panel-heading">
+      <div class="panel-title">
+        <h4><a href="#panel-1" data-toggle="collapse" data-parent="#search-parent">Search Genes <span class='fa fa-search'></span></a><%= link_to "<span class='fa fa-compress'></span>&nbsp;<span class='fa fa-search'></span>".html_safe, 'javascript:toggleSearch();', 'data-toggle' => 'tooltip', title: 'Collapse Search Options', 'data-placement' => 'left', class: 'btn btn-sm btn-danger pull-right', style: 'margin-top: -5px;' %></h4>
+      </div>
+    </div>
+    <div class="panel-body collapse in" id="panel-1">
```
bistline: Give the panels better ids than 'panel-1' and 'panel-2'

app/views/site/_search_options.html.erb
```diff
@@ -1,173 +1,188 @@
...
+          </div>
+          <div class="modal-footer">
+            <button class="close" data-dismiss="modal">×</button>
+          </div>
         </div>
       </div>
     </div>
   </div>
+  <% if action_name !~ /heatmap/ %>
+    <div class="panel panel-info" id="search-options-panel">
```
bistline: search-options-panel is a duplicate ID here - this needs to be changed to something different and informative.

app/views/site/_selection_submit.html.erb
```diff
@@ -0,0 +1,25 @@
...
+$('#selection-submit').click(function(){
+    var needText = $('.need-text');
+    var numFields = needText.toArray().length;
+    var values = [];
+    var cont = true;
+    for(i = 0; i < numFields; i++){
+        var text = needText.eq(i).val();
+        values.push(text);
+        if(text === ""){cont = false}
+    }
```
bistline: You shouldn't be making a model query directly in a form like this - if you need these values, load them in the corresponding controller action into a variable to access here.

app/views/site/_selection_well.html.erb
```diff
@@ -0,0 +1,225 @@
+<% if user_signed_in? %>
+    <script type="text/javascript">
```
bistline: Is there a reason why this script tag is separate from the lower one?  You're declaring variables here that are referenced lower, so ideally they should be in the same block.

app/views/site/update_user_annotations.js.erb
```diff
@@ -0,0 +1,11 @@
+//reload the menu after saving a user annotaion
+closeModalSpinner('#save-modal-spinner', '#save-modal', function () {
+    $("#annotation").html("<%= escape_javascript(select_tag :annotation, grouped_options_for_select(@cluster_annotations, params[:annotation]), class: 'form-control' )%>");
+    $('#notices-target').html("<%= escape_javascript(render partial: '/layouts/notices', locals: {notice: @notice, alert: @alert}) %>");
+    <% if @user_annotation.errors.any? %>
+
```
bistline: You ideally should be checking to see where the errors are occurring rather than assuming they're in the name.

app/views/user_annotations/_form.html.erb
```diff
@@ -0,0 +1,79 @@
...
+			</ul>
+		</div>
+	<% end %>
+  <div class="form-group row">
+    <div class="col-sm-3">
+      <%= f.label 'Name:' %>
+      <%= f.text_field :name, class: 'form-control need-text', autocomplete: 'off' %>
+    </div>
+    <div class="col-sm-3">
+      <%= f.label 'Cluster:' %>
```
bistline: While this works to disable the field to prevent an error, really what you should do is follow the paradigm you set below of creating an anonymous input and setting the value rather than putting an incorrect value inside the _id association fields.

app/controllers/user_annotations_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+  before_action :check_permission, except: :index
+  # GET /user_annotations
+  # GET /user_annotations.json
+  def index
+    #get all this user's annotations
+    @user_annotations = current_user.user_annotations.owned_by(current_user)
+  end
+
+  # GET /user_annotations/1
+  # GET /user_annotations/1.json
```
bistline: We don't need the show action so you can remove this.

app/views/user_annotations/show.html.erb
```diff
@@ -0,0 +1,4 @@
+<p id="notice"><%= notice %></p>
```
bistline: You should remove this file since we're removing the show action.

app/views/user_annotations/show.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.extract! @user_annotation, :id, :created_at, :updated_at
```
bistline: This can be deleted too.

config/environments/test.rb
```diff
@@ -33,7 +33,7 @@
   config.action_mailer.delivery_method = :test
 
   # Randomize the order test cases are executed.
-  config.active_support.test_order = :random
```
bistline: This should set back to random since we're calling test classes in order now.

config/routes.rb
```diff
@@ -38,7 +38,12 @@
         get 'load_annotation_options', to: 'studies#load_annotation_options', as: :load_annotation_options
         post 'update_default_options', to: 'studies#update_default_options', as: :update_default_options
 			end
-		end
+    end
+
+    # user annotation actions
```
bistline: You can take show out of the resources declaration here.

test/ui_test_suite.rb
```diff
@@ -2732,6 +2971,454 @@ def compare_within_rounding_error(error, a1, a2)
...
+		annotation_labels.each_with_index do |annot, i|
+			annot.send_keys("group#{i}")
+		end
+
+		sleep 0.5
+
+		#create the annotation
+		submit_button = @driver.find_element(:id, 'selection-submit')
+		submit_button.click
+
```
bistline: You should actually call wait_for_render and look for the confirmation modal here rather than the sleep statement.  While the sleep may work most of the time, it's better to have a boolean condition to govern how long to wait.

test/ui_test_suite.rb
```diff
@@ -2732,6 +2971,454 @@ def compare_within_rounding_error(error, a1, a2)
...
+		# choose the user annotation
+		@wait.until {wait_for_plotly_render('#cluster-plot', 'rendered')}
+		annotation_dropdown = @driver.find_element(:id, 'annotation')
+		annotation_dropdown.send_keys("user-#{$random_seed}")
+		@wait.until {wait_for_plotly_render('#cluster-plot', 'rendered')}
+
+		#make sure the new annotation still renders a plot for plotly
+		annot_rendered = @driver.execute_script("return $('#cluster-plot').data('rendered')")
+		assert annot_rendered, "cluster plot did not finish rendering on annotation change, expected true but found #{annot_rendered}"
+
```
bistline: Again, you have random sleep statements without a clear reason why.  If you're waiting for something to load, then check for it in the DOM rather than waiting for a half second.

test/ui_test_suite.rb
```diff
@@ -2732,6 +2971,454 @@ def compare_within_rounding_error(error, a1, a2)
...
+		# wait until violin plot renders, at this point all 3 should be done
+		@wait.until {wait_for_plotly_render('#expression-plots', 'box-rendered')}
+		violin_rendered = @driver.execute_script("return $('#expression-plots').data('box-rendered')")
+		assert violin_rendered, "violin plot did not finish rendering, expected true but found #{violin_rendered}"
+		scatter_rendered = @driver.execute_script("return $('#expression-plots').data('scatter-rendered')")
+		assert scatter_rendered, "scatter plot did not finish rendering, expected true but found #{scatter_rendered}"
+		reference_rendered = @driver.execute_script("return $('#expression-plots').data('reference-rendered')")
+		assert reference_rendered, "reference plot did not finish rendering, expected true but found #{reference_rendered}"
+
+		#Create an annotation from the gene page
```
bistline: More sleep statements that I'm not sure why they're here.

test/ui_test_suite.rb
```diff
@@ -2732,6 +2971,454 @@ def compare_within_rounding_error(error, a1, a2)
...
+		# load annotation panel
+		annot_path = @base_url + '/user_annotations'
+		@driver.get annot_path
+
+		#Click download and get filenames
+		filename = (@driver.find_element(:class, "user-#{$random_seed}_cluster").attribute('innerHTML') + "_user-#{$random_seed}.txt").gsub(/ /, '_')
+		basename = filename.split('.').first
+
+		@driver.find_element(:class, "user-#{$random_seed}-download").click
+		# give browser 5 seconds to initiate download
```
bistline: This is a place where the sleep is actually necessary because we just really don't know how long it will take to download the file...

test/ui_test_suite.rb
```diff
@@ -2732,6 +2971,454 @@ def compare_within_rounding_error(error, a1, a2)
...
+		file_exists = Dir.entries($download_dir).select {|f| f =~ /#{basename}/}.size >= 1 || File.exists?(File.join($download_dir, filename))
+		assert file_exists, "did not find downloaded file: #{filename} in #{Dir.entries($download_dir).join(', ')}"
+
+		# delete file
+		File.delete(File.join($download_dir, filename))
+
+		puts "Test method: #{self.method_name} successful!"
+	end
+
+	#check user annotation deletion
```
bistline: This test has no assertions, so therefore it is invalid.  You should grab a DOM id for an element, delete it, and then confirm it has been removed.  You should also check that it's been removed from the list of available annotations on the view_study page.

app/models/user_annotation.rb
YanayRosen: Updated

app/models/user_annotation.rb
YanayRosen: Updated

app/models/user_annotation.rb
YanayRosen: Changed to give time and name of annotation and subsample if any

app/controllers/user_annotations_controller.rb
YanayRosen: Updated

app/models/fire_cloud_client.rb
YanayRosen: I think this may be a leftover from when firecloud crashed and I was trying to get things working. Removed

app/models/study.rb
YanayRosen: Added

app/models/user_annotation.rb
YanayRosen: That statement is a remnant from testing, removed.

app/controllers/site_controller.rb
YanayRosen: Removed

app/controllers/site_controller.rb
YanayRosen: Removed

app/views/site/_search_options.html.erb
YanayRosen: Updated to be more informative

app/views/site/_search_options.html.erb
YanayRosen: I think these might be used for the collapse button, which I still wanted to work without modifying.

app/views/site/_selection_submit.html.erb
YanayRosen: Removed

app/views/site/_selection_well.html.erb
YanayRosen: Updated.

app/views/site/update_user_annotations.js.erb
YanayRosen: Updated

app/views/user_annotations/_form.html.erb
YanayRosen: Updated

app/controllers/user_annotations_controller.rb
YanayRosen: Removed

app/views/user_annotations/show.html.erb
YanayRosen: Removed

app/views/user_annotations/show.json.jbuilder
YanayRosen: Removed

config/environments/test.rb
YanayRosen: Updated

config/routes.rb
YanayRosen: Removed

test/ui_test_suite.rb
YanayRosen: Added assertions to test that annotation is not present in page.

test/ui_test_suite.rb
YanayRosen: Ok

test/ui_test_suite.rb
YanayRosen: Updated

test/ui_test_suite.rb
YanayRosen: Updated

test/ui_test_suite.rb
YanayRosen: Updated

PR #22, opened 2017-08-02
Yr user annotation share

YanayRosen:
User annotation sharing features and tests.

### DIFF
--- a/app/controllers/user_annotations_controller.rb
+++ b/app/controllers/user_annotations_controller.rb
@@ -8,7 +8,6 @@ class UserAnnotationsController < ApplicationController
   def index
     #get all this user's annotations
     @user_annotations = current_user.user_annotations.owned_by(current_user)
-    logger.info("Look: #{@user_annotations}")
     views = UserAnnotation.viewable(current_user)
     edits = UserAnnotation.editable(current_user)
     @user_annotations.concat(views).concat(edits).uniq!
@@ -33,7 +32,6 @@ class UserAnnotationsController < ApplicationController
     else
       @share_changes = false
     end
-    logger.info("Params: #{user_annotation_params}")
     #update the annotation's defined labels
     new_labels = user_annotation_params.to_h['values']
     #If the labels sued to include undefined, make sure they do again
@@ -52,7 +50,7 @@ class UserAnnotationsController < ApplicationController
       if @user_annotation.update(user_annotation_params)
         changes = []
         if @share_changes
-          changes << 'Study shares'
+          changes << 'Annotation shares'
         end
         if @user_annotation.user_annotation_shares.any?
           SingleCellMailer.annot_share_update_notification(@user_annotation, changes, current_user).deliver_now
@@ -75,7 +73,6 @@ class UserAnnotationsController < ApplicationController
           end
 
           #index of values remembers what the order of the old annotations was
-
           index_of_values.each_with_index do |old_index, i|
             old_index.each do |index|
               old_values[index] = new_labels[i]
@@ -140,7 +137,6 @@ class UserAnnotationsController < ApplicationController
       end
 
     end
-    logger.info(rows)
     @data = [headers.join("\t"), types.join("\t"), rows].join"\n"
 
     send_data @data, type: 'text/plain', filename: filename, disposition: 'attachment'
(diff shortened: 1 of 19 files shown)


app/controllers/user_annotations_controller.rb
```diff
@@ -8,6 +8,9 @@ class UserAnnotationsController < ApplicationController
   def index
     #get all this user's annotations
     @user_annotations = current_user.user_annotations.owned_by(current_user)
+    views = UserAnnotation.viewable(current_user)
+    edits = UserAnnotation.editable(current_user)
```
bistline: We actually don't need to concatenate the views & edits as edits is a subgroup of views.  You can just pull all the annotations viewable by the user, and then in the index view, depending on whether the user can edit or not, render the correct buttons.

app/models/user_annotation.rb
```diff
@@ -2,24 +2,47 @@ class UserAnnotation
   include Mongoid::Document
   field :name, type: String
   field :values, type: Array
+  field :queued_for_deletion, type: Boolean, default: false
```
bistline: You need to implement the deletion queue functionality (use StudyFile as the example - there are attributes that have to be renamed, some can be cleared out, and some dependent models can be deleted right away)

app/models/user_annotation.rb
```diff
@@ -2,24 +2,47 @@ class UserAnnotation
...
 
-  #user data arrays belong to user annotations
+  # user data arrays belong to user annotations
   has_many :user_data_arrays do
     def by_name_and_type(name, type, subsample_threshold=nil, subsample_annotation=nil)
       where(name: name, array_type: type, subsample_threshold: subsample_threshold, subsample_annotation: subsample_annotation).order_by(&:array_index).to_a
     end
   end
 
+  has_many :user_annotation_shares, dependent: :delete do
```
bistline: Is there a reason you're only returning the emails in these association queries?  Traditionally you'd return the entire object, not just an attribute.

app/views/user_annotations/edit.html.erb
```diff
@@ -1,6 +1,8 @@
-<h1>Editing '<em><%= @user_annotation.name %></em>'</h1>
-<em><h4><%= @user_annotation.subsampled_at %></h4></em>
```
bistline: This edit permission check is unnecessary as you're checking edit permission both in the index view (to render the button) and in a before_filter to redirect if the user doesn't have permission.

test/ui_test_suite.rb
```diff
@@ -3504,6 +3515,234 @@ def open_new_page(url)
...
+		share_permission = @driver.find_element(:class, 'share-permission')
+		share_permission.send_keys('Edit')
+
+		#update the annotation
+		submit = @driver.find_element(:id, 'submit-button')
+		submit.click
+
+		wait_until_page_loads('user annotation path')
+
+		# logout
```
bistline: It's actually not necessary to call wait_for_render before closing modals anymore as close_modal does this now.

test/ui_test_suite.rb
```diff
@@ -3504,6 +3515,234 @@ def open_new_page(url)
...
+		search_box = @driver.find_element(:id, 'search_genes')
+		search_box.send_key(gene)
+		wait_for_render(:id, 'perform-gene-search')
+		search_genes = @driver.find_element(:id, 'perform-gene-search')
+		search_genes.click
+
+		#make sure the new annotation still renders plots for plotly
+		assert element_present?(:id, 'box-controls'), 'could not find expression violin plot'
+		assert element_present?(:id, 'scatter-plots'), 'could not find expression scatter plots'
+
```
bistline: rather than calling sleep, use wait_for_render or some other boolean declaration to hold execution.

app/controllers/user_annotations_controller.rb
YanayRosen: Updated

app/models/user_annotation.rb
YanayRosen: Implemented

app/models/user_annotation.rb
YanayRosen: Only emails are returned in the study model as well

app/views/user_annotations/edit.html.erb
YanayRosen: Removed

test/ui_test_suite.rb
YanayRosen: Removed

test/ui_test_suite.rb
YanayRosen: Updated to wait_for_render(:class, 'queried_gene')

PR #24, opened 2017-08-03
Yr persist user annotations

YanayRosen:
Added persistence and test


app/controllers/site_controller.rb
```diff
@@ -126,7 +126,7 @@ def update_study_settings
   def render_cluster
     subsample = params[:subsample].blank? ? nil : params[:subsample].to_i
     @coordinates = load_cluster_group_data_array_points(@selected_annotation, subsample)
-
```
bistline: Please remove debugging logger statements.

app/controllers/user_annotations_controller.rb
```diff
@@ -152,6 +152,15 @@ def download_user_annotation
     send_data @data, type: 'text/plain', filename: filename, disposition: 'attachment'
   end
 
+  def persist_to_study
+    respond_to do |format|
+      # redirect back and say success
+      format.html { redirect_to user_annotations_path, notice: "User Annotation '#{@user_annotation.name}' will be added to the study. You will receive an email upon completion or error." }
+      format.json { render :index, status: :ok, location: user_annotations_path }
```
bistline: This method call is running in the foreground - you should run it in the background by calling @user_annotation.delay.persist_to_study(current_user)

app/controllers/site_controller.rb
YanayRosen: Removed

app/controllers/user_annotations_controller.rb
YanayRosen: Added

app/models/user_annotation.rb
```diff
@@ -366,4 +366,116 @@ def self.delete_queued_annotations
     end
   end
 
+  def persist_to_study(current_user)
```
bistline: This method looks good, but it should have a few 'progress' logger statements so that if we encounter errors, we will have more information as to where exactly the error occurred (since we won't have a stack trace).  Particularly, you should do this right before and right after creating a new record.

app/views/site/_search_options.html.erb
```diff
@@ -184,27 +184,25 @@
         </div>
       </div>
     </div>
-  <% end %>
-</div>
-<script type="text/javascript">
-    <% if user_signed_in? %>
-      <% if params[:gene_set].nil? and params[:gene].nil?%>
+    <script type="text/javascript">
+        <% if user_signed_in? %>
```
bistline: use && instead of 'and'

test/ui_test_suite.rb
```diff
@@ -567,7 +580,7 @@ def open_new_page(url)
 	end
 
 	#create a 2d scatter study for use in user annotation testing
-	test 'admin: create a 2d scatter study' do
```
bistline: I like what you did here with the namespacing, but they should still have front-end: as part of the name so that we can run all front end tests as needed.  We can have as many namespaces as we want, and we can nest them.

app/models/user_annotation.rb
YanayRosen: Updated

app/views/site/_search_options.html.erb
YanayRosen: Updated

test/ui_test_suite.rb
YanayRosen: Updated

PR #25, opened 2017-08-08
Yr last week

YanayRosen:
Labels and errors changes

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -553,7 +553,7 @@ class SiteController < ApplicationController
         @cluster_annotations = load_cluster_group_annotations
         @options = load_cluster_group_options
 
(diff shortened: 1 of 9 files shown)


app/controllers/site_controller.rb
```diff
@@ -565,7 +565,7 @@ def create_user_annotations
         @options = load_cluster_group_options
         @notice = nil
         @alert = 'The following errors prevented the annotation from being saved: ' + @user_annotation.errors.full_messages.join(',')
-        logger.error "#{Time.now}: Creating user annotation of params: #{user_annotation_params}, unable to save user annotation #{e.message}"
```
bistline: rather than just calling @user_annotation.errors, we should call @user_annotation.errors.full_messages.join(', ') so that it prints everything to the log (otherwise we just see an object and not the messages)

PR #42, opened 2017-11-02
Multithread signed URL generation

eweitz:
This optimizes our curl bulk data download feature by multithreading calls to generate signed URLs for a collection of files.  It speeds up the method from O(n) to O(1), e.g. 18 seconds to 1-2 seconds for 30 files.


app/controllers/site_controller.rb
```diff
@@ -605,31 +607,45 @@ def download_bulk_files
...
-        user_quota += file_size
+        curl_files.push(file)
       end
     end
 
+    start_time = Time.now
+
+    # Get signed URLs for all files in the requested download objects, and update user quota
+    Parallel.map(curl_files, in_threads: 100) do |file|
+      fc_client = FireCloudClient.new(current_user, @study.firecloud_project)
```
bistline: We actually probably shouldn't instantiate the FireCloudClient with the user's credentials.  Its safer to use the service account credentials as we know that they'll always have access to the requested data.

eweitz: Fixed!

PR #47, opened 2017-11-17
Compress all uncompressed files on upload

eweitz:
This change compresses all uncompressed files on upload.  This has a few project-wide savings:

* Time to upload from SCP to FirecCloud / Google Cloud Storage (GCS)
* Costs for storage on GCS
* Costs for egress from GCS
* Time to download from GCS to user

If my local testing is any indication, this optimization will cut costs and download time roughly in half for applicable files.

Mechanistically, this gzip-compresses any file that isn't a gzipped file, *.bam, or *.cram in our Portal middleware layer.  Then, by setting the {{Content-encoding}} header to {{gzip}}, we enable any downloads via HTTP to pull down the compressed file, then transparently decompress it before writing to file.


app/models/fire_cloud_client.rb
```diff
@@ -1144,19 +1144,39 @@ def get_workspace_file(workspace_namespace, workspace_name, filename)
...
+			# This saves time on upload and download, and money on egress and storage.
+			gzip_filepath = filepath + '.tmp.gz'
+			Zlib::GzipWriter.open(gzip_filepath) do |gz|
+				File.open(filepath).each do |line|
+					gz.write line
+				end
+				gz.close
+			end
+			File.rename gzip_filepath, filepath
+			bucket.create_file filepath, filename, content_encoding: 'gzip'
```
bistline: Technically, right here the call should be:
bucket.create_file filepath, filename, opts.merge(content_encoding: 'gzip')
We want to preserve any other options that were provided with the method call.

app/models/fire_cloud_client.rb
```diff
@@ -1144,19 +1144,39 @@ def get_workspace_file(workspace_namespace, workspace_name, filename)
...
+	# * *return*
+	#   - +Google::Cloud::Storage::File+
 	def create_workspace_file(workspace_namespace, workspace_name, filepath, filename, opts={})
 		bucket = self.get_workspace_bucket(workspace_namespace, workspace_name)
-		bucket.create_file filepath, filename, opts
+		first_two_bytes = File.open(filepath).read(2)
+		gzip_signature = "\x1F\x8B".force_encoding(Encoding::ASCII_8BIT) # per IETF
+		file_is_gzipped = (first_two_bytes == gzip_signature)
+		if file_is_gzipped or filepath.last(4) == '.bam' or filepath.last(5) == '.cram'
+			Rails.logger.info "Direct upload"
```
bistline: I like to add timestamps on the logger calls with "#{Time.now}: " and the message.  Since this gets called as a part of study.send_to_firecloud, I don't think it needs more info than that.

app/models/fire_cloud_client.rb
```diff
@@ -1144,19 +1144,39 @@ def get_workspace_file(workspace_namespace, workspace_name, filename)
...
+			# Directly upload data if it's already compressed
+			bucket.create_file filepath, filename, opts
+		else
+			Rails.logger.info "Gzipping, then upload"
```
bistline: Same comment - add timestamp.

PR #119, opened 2018-06-15
Yr plotly violins

YanayRosen:
Replacing custom implementation of violin plots with Plotly's implementation.


YanayRosen:
I added a jstat dependency and js for stats-- when we merge with the branch that has the extended morpheus package we can get rid of this.

eweitz:
Looks good to me as well.  The updated violin plots look good on my local machine, and code style is consistent with what we have in place.  Kudos, Yanay!

eweitz:
I'll defer to @bistline on final review and merge.

bistline:
So we found a problem with data that essentially has no distribution:

screen shot 2018-06-15 at 3 47 50 pm

In this instance, we should plot nothing, or just a standard box, if the violin is going to be misleading.

PR #143, opened 2018-07-23
Jb csp updates

bistline:


bistline:
Merged locally in development and pushed commit to test before merge was pushed.

PR #144, opened 2018-07-23
auto-select MM coordinate matrix when .mtx file is added

bistline:


app/views/studies/_initialize_expression_form.html.erb
```diff
@@ -72,6 +72,11 @@
 					fileUploading = true;
 					var that = this;
 					var fileName = data.files[0].name.replace(/ /g, '_');
+					// auto-select MM Coordinate Matrix if applicable
+          fileExt = fileName.split('.').slice(-1)[0];
+          if (fileExt == 'mtx') {
+              $('#expression_form_<%= study_file._id %> .file-type').val('MM Coordinate Matrix');
```
eweitz: Nit: I'd use strict equality (`===`) here and wherever possible.

As an aside, the intermixed Rails code and JavaScript here is an example of what I'd like to decouple, per our chats. 
 But I don't see that as in scope here, so no action recommended here.

bistline: Agreed on both points.

PR #158, opened 2018-08-01
refactoring sparse matrix parse to use gene-sorted matrices

bistline:
ParseUtils.cell_ranger_expression_parse no longer reads the entire sparse matrix into memory to convert into a Hash for reading.  With sparse matrices sorted by their gene indices first (then barcode indices), we can read one line at a time like Study#initialize_gene_expression_data does, which is memory efficient and will not run into issues due to file size.  Unsorted sparse matrices will throw and error and abort the parse job.


bistline:
Agreed re: UX concerns.  It may be that we move away from sparse matrices altogether for hdf5 if that is possible, but for now this at least solves the memory constraints of MM coordinate matrix parsing.

PR #161, opened 2018-08-08
Global gene search performance

bistline:
* Paginating global gene search results to stagger results
* Better browser performance for larger numbers of genes/studies requested


app/views/site/index.html.erb
```diff
@@ -15,3 +15,46 @@
...
+        $("#pagination").toggle();
+    }
+
+    $("#pagination").click(function(e) {
+        e.preventDefault();
+        $(window).unbind('scroll');
+        paginateStudies(totalPages, order, search, project);
+    });
+
+    if ( search != "" ) {
```
eweitz: Two nits: we should use single- or double-quotes consistently in our JavaScript, and use strict comparison operators.  (I'd opt for using single-quotes everywhere in JS.)

PR #169, opened 2018-09-26
Species integration with Study, StudyFile

bistline:
- Adding Taxon, GenomeAssembly, and GenomeAnnotation models/views for collecting species information about Study and StudyFile objects
- Includes admin panel for ease in adding new entries


app/assets/javascripts/application.js
```diff
@@ -667,23 +667,32 @@ function renderMorpheus(dataPath, annotPath, selectedAnnot, selectedAnnotType, t
...
-    $(fastqField).toggleClass('hidden');
-    // toggle disabled status by returning inverse of current state
-    $(fastqField).find('input').attr('disabled', !$(fastqField).find('input').is('[disabled=disabled]'));
-    // set human data attr to true
     var humanData = $(fastqField).find('input[type=hidden]');
-    $(humanData).val($(humanData).val() === 'true' ? 'false' : 'true' );
-    // enable name field & update button to allow saving
     var saveBtn = selector.find('.save-study-file');
-    $(saveBtn).attr('disabled', !$(saveBtn).is('[disabled=disabled]'));
     var nameField = selector.find('.filename');
```
eweitz: It looks like `nameField` (and `fastqField`, `humanData`, `fileField` and `saveBtn`) are only used as jQuery objects.  If that's true, then perhaps consider getting each's jQuery object as part of the initial assignment.

E.g.:
`var nameField = $(selector.find('.filename'));`

That would enable cleaner references to those variables later, e.g. `nameField.attr('readonly', false);` instead of `$(nameField).attr('readonly', false);`.

Not a blocking issue; just a possible minor readability enhancement.

app/assets/javascripts/application.js
```diff
@@ -667,23 +667,32 @@ function renderMorpheus(dataPath, annotPath, selectedAnnot, selectedAnnotType, t
...
-    $(nameField).attr('readonly', !$(nameField).is('[readonly=readonly]'));
-    $(nameField).attr('placeholder', '');
+    if (state === 'true') {
```
eweitz: It looks like `state` is intended to represent a Boolean.  If its only values should be `'true'` or `'false'`, consider casting it as a Boolean before passing it into this function -- even it's parsed from a DOM value, as I imagine it is.

app/helpers/taxons_helper.rb
```diff
@@ -0,0 +1,2 @@
+module TaxonsHelper
```
eweitz: Is this module intentionally empty?

app/views/site/_study_download_data.html.erb
```diff
@@ -187,8 +193,9 @@
       <table class="table table-striped tabled-condensed study-data-table" id="fastq-table">
         <thead>
         <tr>
-          <th class="col-sm-5">Filename</th>
-          <th class="col-sm-5">Description</th>
+          <th class="col-sm-4">Filename</th>
+          <th class="col-sm-4">Description</th>
+          <th class="col-sm-2">Species</th>
```
eweitz: I see "Species/Assembly" as the table header in other tables that have this column.  Should "Species" be updated to "Species/Assembly" here as well, for consistency?

app/helpers/taxons_helper.rb
bistline: Yes - it's auto-generated when creating new controllers/models

app/views/site/_study_download_data.html.erb
bistline: Yup - will do

app/assets/javascripts/application.js
bistline: Sounds reasonable, and it does cut down on the proliferation of $()

PR #172, opened 2018-10-03
REST API

bistline:


app/controllers/api/v1/directory_listings_controller.rb
```diff
@@ -0,0 +1,322 @@
...
+            key :type, :string
+          end
+          response 200 do
+            key :description, 'DirectoryListing object'
+            schema do
+              key :title, 'DirectoryListing'
+              key :'$ref', :DirectoryListing
+            end
+          end
+          response 401 do
```
eweitz: How feasible would it be to have a method like "handle_4xx_errors(response)" to reduce duplicate code?  Most endpoints seem to have very similar code to handle 401-406 errors.

Not blocking; just to consider as a possible future improvement for maintainability.

bin/run_unit_tests.sh
```diff
@@ -1,20 +1,24 @@
...
 echo "Seeding test database..."
 rake RAILS_ENV=test db:seed
 echo "Database initialized, launching unit & integration tests..."
 ruby -I test test/integration/fire_cloud_client_test.rb
 ruby -I test test/integration/cache_management_test.rb
+ruby -I test test/api/studies_controller_test.rb
+ruby -I test test/api/study_files_controller_test.rb
+ruby -I test test/api/study_file_bundles_controller_test.rb
+ruby -I test test/api/study_shares_controller_test.rb
+ruby -I test test/api/directory_listings_controller_test.rb
```
eweitz: Hooray for API tests!

app/views/studies/_initialize_primary_data_form.html.erb
```diff
@@ -108,24 +109,23 @@
           if (selectedTaxon !== '') {
               $.getJSON('<%= get_taxon_path %>?taxon=' + selectedTaxon, function(taxon) {
                   var humanDataSelect = $('#primary_data_form_<%= study_file.id %>').find('.human-data');
-                  if ( restrictedTaxons.includes(taxon.taxon_identifier) ) {
+                  if ( taxon.restricted ) {
```
eweitz: Nit: `if ( taxon.restricted )` should be `if (taxon.restricted)`.

app/controllers/api/v1/directory_listings_controller.rb
bistline: As far as I know, Swagger needs each response detailed individually.  However, there may be a way to extract these into a common library to reduce duplication.

app/views/studies/_initialize_primary_data_form.html.erb
bistline: I'll fix that.

PR #173, opened 2018-10-03
Display "Genome" tab whenever a BAM is available

eweitz:
This improves the discoverability of the "Genomes" tab. 

Previously, if a user wanted to view data in a genome browser, they would need to click the Download tab, then click "Browse in genome" for a BAM.  Now, the Genome tab is displayed by default if the study has a BAM file.

Also, users must now sign in to view genome visualizations.


app/assets/javascripts/scp-igv.js
```diff
@@ -35,15 +35,19 @@ $(document).on('click', '.bam-browse-genome', function(e) {
...
 
 $(document).on('click', '#genome-tab-nav', function (e) {
-  initializeIgv();
+  window.location.hash = '#genome-tab';
+
+  // Go to Google sign-in page, redirect to this view after authenticating.
+  if (accessToken === null) {
+    ga('send', 'event', 'genome', 'sign_in_start');
+    localStorage.setItem('signed-in-via-genome-tab', true);
+    window.location = '/single_cell/users/auth/google_oauth2';
```
bistline: This isn't something that needs to be change, but since we can't interpolate Ruby in here since this is pure javascript, this is something that we'll have to make a note of if we ever change our URL.

eweitz: Good point, agreed.

PR #175, opened 2018-10-05
Updating Docker container to run on Ruby 2.5.1

bistline:
### Enhancements/Refactoring
- Updating Docker container Ruby version to 2.5.1
- Refactored Devise & Swagger controller patches to use ActiveSupport::Concern rather than complete rewrite of controllers
- Updating Rubyzip gem re: [CVE-2018-1000544](https://nvd.nist.gov/vuln/detail/CVE-2018-1000544)
- Removed billing UI regression tests that never passed due to time it takes to provision new FireCloud billing projects

### Fixes
- Editing user annotations also edits subsampled user_data_arrays that had erroneously been excluded from query before
- Added validation on Study model to prevent changing FireCloud project/workspace values (was only blocked at UI level before, now would be regression with REST API)


app/controllers/user_annotations_controller.rb
```diff
@@ -54,7 +54,7 @@ def update
     old_labels = @user_annotation.values
 
     # Remeber the old annotations
```
eweitz: Remeber -> Remember

PR #183, opened 2018-10-30
Process genome references

eweitz:
This Python data pipeline downloads, transforms, and uploads reference data for genome assemblies and genome annotations needed for Single Cell Portal.

Raw data is fetched from NCBI and Ensembl, processed to transform it as needed, then uploaded to a Google Cloud Storage (GCS) bucket for SCP.  The uploaded data is used to display interactive genome visualizations, e.g. igv.js and Ideogram.js, and (in the future) to run biological workflows / analysis pipelines, e.g. Cell Ranger and inferCNV.

Installation instructions and usage examples are in `process_genome_references.py`.


eweitz:
The commits in this PR were originally in the [`single_cell_portal`](https://github.com/broadinstitute/single_cell_portal) repo.  But the code belongs in this repo, as the audience is SCP administrators and developers.  To ensure I can go back to previous states of this code if needed (and not lose weeks of history), I patched my relevant commits from the other repo into this repo.

Steps I took to effectively cherry-pick a range of commits from a different repo:
```
# Set the single_cell_portal repo as a remote of single_cell_portal_core
git remote add other_scp https://github.com/broadinstitute/single_cell_portal.git

# Create a new branch `ew-genome-references` in this repo, 
# using the same branch from the other repo as a starting point
git checkout -b ew-genome-references other_scp/ew-genome-references

# Prepare patches for the last 35 commits in that other repo's branch
git format-patch -35

# Apply those 35 patches to this local branch, using a three-way merge as needed
git am -3 0001-Create-script-to-gather-genome-metadata-relevant-to-.patch
git am -3 0002-Ignore-cached-assembly-report-and-other-genome-data.patch 
...
git am -3 0035-Uncomment-all-organisms-add-Drosophila.patch

# Success.  Push patched branch to single_cell_portal_core.
git push origin ew-genome-references
```

bin/python/genomes/parse_genome_annotations.py
```diff
@@ -0,0 +1,204 @@
...
+        if (taxid == '10090' and strain != 'reference (CL57BL6)'):
+            # Mouse has multiple annotated assemblies; only use reference assembly
+            continue
+
+        ensembl_metadata[taxid] = {
+            'organism': name,
+            'taxid': taxid,
+            'assembly_name': assembly,
+            'assembly_accession': species['accession'],
+            'release': str(species['release'])
```
bistline: Wherever we're getting a release date, we need to check the format of that date.  YYYY-MM-DD is fine, as is YYYY/MM/DD, or similar formats.  But just YYYY-MM won't save - we have to make sure we have a day.  If the day isn't present, we should default to '01'.

eweitz: Fixed!

eweitz:
> bin/python/genomes/organisms.tsv should probably be moved

I'll do that and add the needed CLI parameter for input directory right after demo.

PR #189, opened 2018-11-21
API retry backoffs, StudyFileBundle bugfixes, Docker updates

bistline:
* Adding incremental retry backoffs to FireCloud & GCS API errors (5 times, adds 15 sec. per retry)
  * Also includes blacklist of methods/URLs to not increment backoff in cases of UI blocking
* Refactoring UploadCleanupJob to retry pushing files to FireCloud on failure (up to 3 times over 10 minutes)
  * Study.storage_sanity_check will now also push missing files using UploadCleanupJob
* Fixing bug with StudyFileBundles not cleaning up after themselves on deletion of parent file (either during or after upload has completed)
* Updates
  * Updated rack to 2.0.6 re: CVE-2018-16470/1
  * Updated Docker base image version to 1.6
    * Updated yarn to 1.12.3

### DIFF
--- a/test/ui_test_helper.rb
+++ b/test/ui_test_helper.rb
@@ -348,6 +348,7 @@ class Test::Unit::TestCase
   # load file either in browser or download and check for existence
   def download_file(link, basename)
     link.click
+    sleep(3)
     if @driver.current_url.include?('https://storage.googleapis.com/')
       assert @driver.current_url =~ /#{basename}/, "Downloaded file url incorrect, did not find #{basename}"
       @driver.navigate.back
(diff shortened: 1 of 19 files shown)


test/ui_test_helper.rb
```diff
@@ -348,6 +348,7 @@ def accept_alert
   # load file either in browser or download and check for existence
   def download_file(link, basename)
     link.click
+    sleep(3)
```
eweitz: I assume using a "waitUntil" construct is infeasible here.  Not a significant concern for a one-off, but we should revisit if this apparent need becomes more widespread or ends up causing false-positive failures.

bistline: I agree - I hate adding sleep(n) calls, but the issue is that if there is a problem and the file doesn't load, the test will hang indefinitely.  I think a future refactor can have a while loop with a counter, but it just didn't seem like that much of an issue.

PR #222, opened 2019-03-19
Analysis Submission fixes

bistline:
* Automatically formatting input parameters based on type
* Sorting input parameters alphabetically

### DIFF
--- a/app/models/analysis_parameter.rb
+++ b/app/models/analysis_parameter.rb
@@ -159,7 +159,10 @@ class AnalysisParameter
   def get_instances_by_associations(study)
     model = self.associated_model_class
     instances = []
-    if study.present?
+    if model == Study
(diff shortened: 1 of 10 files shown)


app/models/analysis_parameter.rb
```diff
@@ -101,6 +101,22 @@ def value_method_type
...
   end
 
+  # helper to format a user-supplied value based in parameter type
+  def formatted_user_value(value)
+    if value.blank?
+      ''
+    else
+      case self.parameter_type
+      when /String/
+        !value.start_with?('"') ? "\"#{value}\"" : value
```
eweitz: Can this be cleaner like so?
```suggestion
        !value.start_with?('"') ? '"#{value}"' : value
```

PR #229, opened 2019-03-29
REST API Site Actions

bistline:
* Adding endpoints for browsing public/shared studies, and downloading/streaming data


app/controllers/api/v1/concerns/current_api_user.rb
```diff
@@ -0,0 +1,53 @@
...
+                  Rails.logger.error "Unable to retrieve user info from access token: #{api_access_token}"
+                end
+              rescue => e
+                error_context = {
+                    auth_response_body: response.present? ? response.body : nil,
+                    auth_response_code: response.present? ? response.code : nil,
+                    auth_response_headers: response.present? ? response.headers : nil
+                }
+                ErrorTracker.report_exception(e, user, error_context)
+                Rails.logger.error "Error retrieving user api credentials: #{e.class.name}: #{e.message}"
```
eweitz: Fix case:

```suggestion
                Rails.logger.error "Error retrieving user API credentials: #{e.class.name}: #{e.message}"
```

PR #230, opened 2019-03-29
User session timeouts

bistline:
* Adding 30 minute timeout on idle user sessions
* Supports catching 401s on AJAX calls and prompts user to reauthenticate

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -967,6 +967,8 @@ window.clearGeneSearchLoading = function() {
 // force login on ajax 401
 $(document).ajaxError(function (e, xhr, settings) {
     if (xhr.status == 401) {
-        location.reload();
+        alert('You are not signed in or your session has expired - please login to continue.');
+        var url = 'https://' + window.location.hostname + '/single_cell/users/auth/google_oauth2';
+        location.href = url;
     }
 });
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -962,4 +962,13 @@ window.clearGeneSearchLoading = function() {
     console.log('Clearing global gene search message');
     $('#wrap').data('spinner').stop();
     $('#gene-search-results-count').html($('.gene-panel').length);
-};
\ No newline at end of file
+};
+
+// force login on ajax 401
+$(document).ajaxError(function (e, xhr, settings) {
+    if (xhr.status == 401) {
```
eweitz: Strict equality would be a bit better:

```suggestion
    if (xhr.status === 401) {
```

app/controllers/application_controller.rb
```diff
@@ -104,4 +105,20 @@ def is_valid_signed_url?(signed_url)
...
+      format.js {render js: "alert('Your session has expired; Please log in again')"}
+    end
+  end
+
+  protected
+
+  # patch to supply CSRF token on all ajax requests if it is not present
+  def set_csrf_headers
+    if request.xhr?
+      cookies['XSRF-TOKEN'] = form_authenticity_token if cookies['XSRF-TOKEN'].blank?
```
eweitz: Could you explain what service is expecting `XSRF-TOKEN`?  Knowing that would help me remember which services call this a "CSRF" and which call it a "XSRF".

bistline: That's a great question - what I've been able to deduce about this is that it's a Rails-specific header for CSRF tokens on AJAX requests.

PR #231, opened 2019-03-29
Analysis submission report

bistline:
* Restoring monthly analysis submission graph to reports page
* Adding complete submission report for all successful submissions for admins (email, analysis, completion date)

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -127,88 +127,35 @@ class ReportsController < ApplicationController
 
     # get pipeline statistics
     @pipeline_submissions = {}
-    @pipeline_runtimes = {}
     @has_pipeline_stats = false
+    submissions = []
     # Deferring until Q2 2019; investigating better ways of calculating runtimes and costs
     #
-    # AnalysisMetadatum.all.each do |analysis|
-    #   @has_pipeline_stats = true
-    #   study = analysis.study
-    #   comp_method = analysis.payload['computational_method']
-    #   parts = comp_method.split('/')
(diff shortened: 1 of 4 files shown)


app/controllers/reports_controller.rb
```diff
@@ -225,6 +172,34 @@ def submit_report_request
     redirect_to reports_path, notice: 'Your request has been submitted.' and return
   end
 
+  def export_submission_report
+    if current_user.admin?
+      @submission_stats = []
+      Parallel.map(AnalysisMetadatum.all.to_a, in_threads: 3) do |analysis|
```
eweitz: Nice use of multi-threading on an I/O-bound task.

app/views/reports/index.html.erb
```diff
@@ -417,4 +391,16 @@
         });
     });
 
+    $('#export-submission-report').on('click', function() {
+        var that = $(this);
+        var text = that.text();
+        var btn = that.button();
+        btn.button('disable');
+        btn.button({'label': 'Exporting... file will download momentarily'});
+        setTimeout(function() {
```
eweitz: Using a time-out seems reasonable for expediency, but it might be worth recording our hope to refactor.

```suggestion
        setTimeout(function() {
        // TODO: Investigate ways to detect start of download, and eliminate explicit wait
```

bistline: Agreed, but given that we are responding to the request with a text file, there's no response to render that can update the DOM.  This is mainly just to prevent the admin from clicking the button multiple times and getting a whole bunch of files.

PR #241, opened 2019-04-10
Denoise genomic heatmap

eweitz:
This enables interactively adjusting expression thresholds for the genomic heatmap.  Using this new Ideogram functionality, users can explore data output by inferCNV.

These changes pass local functional tests for genome visualizations.

### DIFF
--- a/app/assets/javascripts/scp-ideogram.js
+++ b/app/assets/javascripts/scp-ideogram.js
@@ -111,43 +111,103 @@ function updateTracks() {
   ideogram.updateDisplayedTracks(selectedTracks);
 }
 
-function addMarginControls() {
-  console.log('addMarginControls');
-
-  if (document.querySelector('#chrMarginContainer')) return;
-
-  chrMargin = (typeof chrMargin === 'undefined' ? 10 : chrMargin)
-  marginSlider =
-    `<label id="chrMarginContainer">
-      Chromosome margin:
-    <input type="range" id="chrMargin" list="chrMarginList" value="` + chrMargin + `">
-    </label>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/scp-ideogram.js
```diff
@@ -111,43 +111,103 @@ function updateTracks() {
...
+      </datalist>`;
+    d3.select('#_ideogramLegend').node().innerHTML += marginSlider;
+}
+
+function updateThreshold(event) {
+  var thresholds, newThreshold, numThresholds, i;
+
+  window.expressionThreshold = parseFloat(event.target.value);
+
+  window.adjustedExpressionThreshold = expressionThreshold/10 - 4;
```
bistline: Just out of curiosity, what's the reasoning for this formula?  Is the `expressionThreshold/10 - 4;` just so the margins line up right?

eweitz: Good question.  This formula is independent of the chromosome margins.

It is the way that the range of "normal" gene expression changes -- e.g. restricting "gain" or "loss" to outliers.  More concretely, `adjustedExpressionThreshold` is used to adjust which array indices for the list of inferCNV's heatmap thresholds are considered high or low expression.

The underlying values being pulled in `event.target.value` (0 - 100, steps of 10) is needed because Chrome does not seem to render useful tick marks in the `datalist` UI control otherwise.  

So, in summary, the formula is needed as a consequence of the UI.  A more refined control would be nice in the future, and that'd allow us to smooth out this obscure formula.

PR #250, opened 2019-04-24
Persist gene search in distribution tab after login

bistline:
* Persisting gene search after a user logs in when clicking the 'Genome' tab 
* Bugfixes
  * Restoring javascript console in production environment
  * bin/job_monitor.rb correctly notices missing job workers that no longer have a pid file

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -23,13 +23,6 @@
       <%= render '/layouts/ga_script.js' %>
   </script>
 
-
-	<% if Rails.env == 'production' %>
-    <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
-        // suppress logging in production
-        var console = { log: function() {} };
-    </script>
-	<% end %>
   <% if ENV['TCELL_AGENT_APP_ID'].present? && ENV['TCELL_AGENT_API_KEY'].present? %>
     <script src="https://jsagent.tcell.io/tcellagent.min.js" nonce="<%= content_security_policy_script_nonce %>"
             tcellappid="<%= ENV['TCELL_AGENT_APP_ID'] %>" tcellapikey="<%= ENV['TCELL_AGENT_API_KEY'] %>"></script>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -5,6 +5,14 @@
   <%= csrf_meta_tags %>
   <meta http-equiv="Content-Type" content="text/html;charset=utf-8">
 	<meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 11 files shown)


app/views/layouts/application.html.erb
```diff
@@ -15,13 +23,6 @@
       <%= render '/layouts/ga_script.js' %>
   </script>
 
-
-	<% if Rails.env == 'production' %>
-    <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
```
eweitz: Thanks for this!

app/views/site/view_gene_expression.html.erb
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+
+<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+    // re-open a tab on reload if necessary
+    var urlParts = window.location.href.split('#');
+    if (urlParts.length > 1) {
+        openedTab = urlParts[1];
```
eweitz:
```suggestion
        var openedTab = urlParts[1];
```

app/views/site/view_gene_expression_heatmap.html.erb
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+
+<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+    // re-open a tab on reload if necessary
+    var urlParts = window.location.href.split('#');
+    if (urlParts.length > 1) {
+        openedTab = urlParts[1];
```
eweitz:
```suggestion
        var openedTab = urlParts[1];
```

app/views/site/view_gene_expression_heatmap.html.erb
```diff
@@ -0,0 +1,90 @@
...
+        <div class="spinner-target" id="update-study-settings-spinner"></div>
+      </div>
+      <div class="modal-footer">
+        <button class="close" data-dismiss="modal">×</button>
+      </div>
+    </div>
+  </div>
+</div>
+
+<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
```
eweitz: I won't remember this code needs to be changed in two places, if that's ever needed.  

De-duplicating this code and [its other copy](https://github.com/broadinstitute/single_cell_portal_core/pull/250/files#diff-44ceb12e4c5a5d3a76b268ad3ab0c872R82) added here would fix that lurking maintainability issue.  If a quicker solution is desired, a code comment noting as much would also mitigate this.

bistline: Yes, you're correct.  Also, there's a much better way to write this function.  But I'll abstract it out and just make a function call so we're not duplicating code.

PR #251, opened 2019-04-30
adding unique user_id cookie for GA tracking

bistline:
* Adding unique userId for differentiating users from sessions
* Creating custom events for 'engaged' user actions

### DIFF
--- a/app/views/site/_selection_submit.html.erb
+++ b/app/views/site/_selection_submit.html.erb
@@ -21,6 +21,7 @@
             setErrorOnBlank(needText);
         } else {
             $('#generic-modal-title').html("Saving... Please Wait");
+            ga('send', 'event', 'Custom Cell Annotations', 'create_annotation');
             launchModalSpinner('#generic-modal-spinner', '#generic-modal', function() {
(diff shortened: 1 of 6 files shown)


app/views/site/_selection_submit.html.erb
```diff
@@ -21,6 +21,7 @@
             setErrorOnBlank(needText);
         } else {
             $('#generic-modal-title').html("Saving... Please Wait");
+            ga('send', 'event', 'custom_cell_annotation', 'create_annotation');
```
eweitz: I have minor qualms about mixing snake case (for SCP event category and label strings, e.g. `create_annotation`) with camel case (for built-in Google Analytics properties, e.g. `userId`), as that inconsistency could frustrate users if they query.  

But per our chat, it seems the internal user would be exploring this information in a GUI, and queries that combined built-in GA properties with SCP event labels would be rare.  So we might be inconsistent in edge cases, but our readability would be better in almost all cases.  That seems better than perfect consistency but lower general readability.

So overall I think your case choice is the best option.

PR #254, opened 2019-04-30
Prevent plotting extremely many points

eweitz:
This fixes a bug in which searching a gene with "All Cells" in the Subsampling threshold dropdown list breaks the Single Cell Portal UI if the cluster has a very large number of cells.  It does so by omitting the "All Cells" option if the number of points is 250,000 or more.

### DIFF
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -40,7 +40,7 @@ class ClusterGroup
   # fixed values to subsample at
   SUBSAMPLE_THRESHOLDS = [1000, 10000, 20000].freeze
 
-  MAX_THRESHOLD = 250000
+  MAX_THRESHOLD = 100000
 
   COLORBREWER_SET = %w(#e41a1c #377eb8 #4daf4a #984ea3 #ff7f00 #a65628 #f781bf #999999
     #66c2a5 #fc8d62 #8da0cb #e78ac3 #a6d854 #ffd92f #e5c494 #b3b3b3 #8dd3c7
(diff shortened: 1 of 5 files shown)


app/models/cluster_group.rb
```diff
@@ -40,6 +40,8 @@ def by_name_and_type(name, type, subsample_threshold=nil)
   # fixed values to subsample at
   SUBSAMPLE_THRESHOLDS = [1000, 10000, 20000].freeze
 
+  MAX_THRESHOLD = 250000
```
knapii-developments: How did you come up with this number?

eweitz: Great question!  Estimation.  We know the search fails with 1.3 million cells, so I figured 25% of that would be safe.

Further testing reveals that 250K cells throws the same `Maximum call stack size exceeded` error on production in `createTracesAndLayout`.  This PR's optimization in `arrayMin` and `arrayMax` should mitigate that; it enabled me to get past the previous line of failure with the 1.3M-cell study, but too slowly for decent UX.  The current UI cannot handle gene search results for a 153,539-cell study, but can for a 118,861-cell study.

So I've lowered `MAX_THRESHOLD` to 100,000 cells.  This is 5x our highest enumerated threshold of 20,000 cells, and comfortably below our empirically verified limit.  More substantial performance engineering this quarter and next should allow to us to raise the threshold.

bistline: I think this is a really good plan.

PR #271, opened 2019-05-31
Adding user identifier to FireCloud API calls

bistline:
Adding the Current module to extract the current_user from any requests made by users and passing that along to all FireCloud API calls.  Database identifier in lieu of email.


bistline:
> Looks good.
> 
> Is this `x-user-id` included in logs, as well?

It will be in the logs both for Single Cell Portal and for FireCloud/Terra.  See line 287 in fire_cloud_client.rb

PR #275, opened 2019-06-14
Adding analysis configuration description to API responses

bistline:
All API responses for listing/configuring analyses now include the configuration 'description' to allow for API users to have the same contextual information as GUI users.

### DIFF
--- a/app/views/site/_study_settings_form.html.erb
+++ b/app/views/site/_study_settings_form.html.erb
@@ -129,6 +129,7 @@
             <div class="col-sm-5">
               <%= share.label :permission %><br />
               <%= share.select :permission, options_for_select(StudyShare::PERMISSION_TYPES, share.object.permission), {}, class: 'form-control share-permission' %>
+              <span class="help-block share-description"><%= StudyShare::PERMISSION_DESCRIPTION_MAP[share.object.permission] %></span>
             </div>
             <div class="col-sm-1">
               <%= share.label :delete, "&nbsp;".html_safe %><br />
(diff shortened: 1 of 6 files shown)


db/seeds.rb
```diff
@@ -116,4 +116,4 @@
 # Analysis Configuration seeds
 AnalysisConfiguration.create(namespace: 'single-cell-portal', name: 'split-cluster', snapshot: 1, user_id: user.id,
                              configuration_namespace: 'single-cell-portal', configuration_name: 'split-cluster',
-                             configuration_snapshot: 2, description: '<p>This is an test description.</p>')
\ No newline at end of file
+                             configuration_snapshot: 2, description: 'This is an test description.')
```
eweitz:
```suggestion
                             configuration_snapshot: 2, description: 'This is a test description.')
```

app/views/site/_study_settings_form.html.erb
```diff
@@ -160,4 +161,11 @@
             updateForm.submit();
         });
     });
+
+    $("#update-study-settings-form").on('change', '.share-permission', function() {
+        var newPermission = $(this).val();
+        var permissionText = <%= raw StudyShare::PERMISSION_DESCRIPTION_MAP.to_json %>;
```
eweitz: `permissionText` seems to be a constant.  So it would be better to assign it once (e.g. right above this event handler) rather than every time this callback is invoked.  This is a minor structuring issue that doesn't affect performance, though, so I don't consider it blocking.

bistline: I take the point, but we're doing the same thing in multiple places, and given that all models/constants are held in memory in Rails, this doesn't involve a query.  It's just accessing the object.  So I don't think it makes much of a substantive difference in terms of performance, and it makes it a little more complicated (one more chance for something to break).

db/seeds.rb
bistline: Good catch - fixing

PR #276, opened 2019-06-18
Enable deployment notification banner dismissal

knapii-developments:
This PR deals with [SCP-1725](https://broadworkbench.atlassian.net/browse/SCP-1725). This will enable users to close banner deployment notifications. 

A close 'X' button was added to the deployment notification banner. When clicked, the banner is removed from the site for that user. This is done by setting a cookie when the banner is closed.

### DIFF
--- a/app/views/layouts/_nav.html.erb
+++ b/app/views/layouts/_nav.html.erb
@@ -64,35 +64,11 @@
 			<% end %>
 		</ul>
 	</div>
-  <% if @deployment_notification.present?%>
-    <% if Time.zone.now >= @deployment_notification.display_time%>
-    <div class="notification-banner">
-      <button type="button" class="close" id='close-notification-banner' aria-label="Close" style="padding-right: 10px;">
-        <span aria-hidden="true">&times;</span>
-      </button>
-      <div class="col-sm-3">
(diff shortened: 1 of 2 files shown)


app/views/layouts/_nav.html.erb
```diff
@@ -65,16 +65,34 @@
...
+      <div >
+        <span class="align-middle"><%= @deployment_notification.message %></span>
       </div>
+
     </div>
     <%end %>
+    <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+        $('#close-notification-banner').click(function() {
+            $(".notification-banner").hide();
+            if ( typeof Cookies.get('hide_deployment_notification') === 'undefined' ) {
```
eweitz: We'll need to ensure that hiding one deployment's banner doesn't hide the next deployment's banner.  

Deleting the `hide_deployment_notification` cookie upon creating a banner or including a string representation of `@deployment_notification.deployment_time` in the cookie name seem like two ways to achieve that.

bistline: We discussed this, and the plan is to use the ID of the deployment notification as the value for the cookie.  That's the simplest check to make (no date math required).

eweitz: Excellent!

PR #277, opened 2019-06-25
Cluster/Metadata parsing & compute permissions bugfixes

bistline:
* Addressing corner case when parsing clusters/cell metadata of incorrectly matching group-based annotations based on case of TYPE header lines
* Eliminating need for checking billing account members when determining compute permissions (checking for existence of project & role instead)

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -1741,7 +1741,7 @@ class Study
       raw_header_data = cluster_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
       raw_type_data = cluster_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
       header_data = self.sanitize_input_array(raw_header_data)
-      type_data = self.sanitize_input_array(raw_type_data)
+      type_data = self.sanitize_input_array(raw_type_data).map(&:downcase)
 
       # determine if 3d coordinates have been provided
       is_3d = header_data.include?('Z')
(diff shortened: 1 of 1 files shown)


app/models/study.rb
```diff
@@ -1741,7 +1743,7 @@ def initialize_cluster_group_and_data_arrays(ordinations_file, user, opts={local
       raw_header_data = cluster_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
       raw_type_data = cluster_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
       header_data = self.sanitize_input_array(raw_header_data)
-      type_data = self.sanitize_input_array(raw_type_data)
+      type_data = self.sanitize_input_array(raw_type_data).map(&:downcase)
```
knapii-developments: What is this syntax? &:downcase

app/models/study.rb
```diff
@@ -2281,10 +2283,10 @@ def initialize_cell_metadata(metadata_file, user, opts={local: true})
       Rails.logger.info "#{Time.zone.now}: Validating metadata file headers for #{metadata_file.name}:#{metadata_file.id} in #{self.name}"
       headers = m_file.readline.split(/[\t,]/).map(&:strip)
       @last_line = "#{metadata_file.name}, line 1"
-      second_header = m_file.readline.split(/[\t,]/).map(&:strip)
+      second_header = m_file.readline.split(/[\t,]/).map {|entry| entry.downcase.strip}
```
knapii-developments: Can you explain this?
|entry| entry.downcase.strip

app/models/study.rb
bistline: This is Ruby shorthand for passing a block.  If we have an array of values called `array`:

`array.map(&:downcase) == array.map {|element| element.downcase}`

The `&` just means "for every element in this iterable object, call the following method", which is passed as a symbol.

app/models/study.rb
bistline: Like the above example of using the `&` block shorthand, this is the expanded version.  Since I need to call two methods (`downcase` to lower-case the string, and then `strip` to remove leading/trailing whitespace).  We're doing this to be more permissive in matching types - so a column in a cluster/metadata file labeled 'Group' instead of 'group' won't cause the values to be stored incorrectly.

PR #278, opened 2019-06-26
Show dot plots for multi-gene search results

eweitz:
This adds dot plots to Single Cell Portal.  

Dot plots are similar to heatmaps, and better for summarizing expression across many cells.  Upon searching for two or more genes, a dot plot is now shown by default, and users can explore a heatmap of the same data in another prominent tab.

This extends and productionizes great work in SCP done by Yanay Rosen (@YanayRosen).  A new UI test verifies it.  I'll update our documentation separately.

See [SCP-1103](https://broadworkbench.atlassian.net/browse/SCP-1103) for more detail.


app/assets/javascripts/scp-dot-plot.js
```diff
@@ -0,0 +1,296 @@
...
+  $.ajax({
+    url: dotPlotAnnotValuesPath + '?' + renderUrlParams,
+    dataType: 'JSON',
+    success: function(annotations) {
+      renderMorpheusDotPlot(dataPath, newAnnotPath, annotName, annotType, '#dot-plot', annotations, fit, height);
+    }
+  });
+}
+
+$('#annotation').change(function() {
```
bistline: How is this not clobbering the event listener that's already in the page? See https://github.com/broadinstitute/single_cell_portal_core/blob/master/app/views/site/_study_visualize.html.erb#L102 and https://github.com/broadinstitute/single_cell_portal_core/blob/master/app/views/site/_view_gene_expression_heatmap.html.erb#L103 for reference.  I tried it out locally and it's doing the right thing, but looking at the code it seems to me that it shouldn't (unless I'm missing something intrinsic about event listeners).

app/assets/javascripts/scp-dot-plot.js
```diff
@@ -0,0 +1,296 @@
...
+    var an = $(this).val();
+    // Keep track for search purposes
+    $('#search_annotation').val(an);
+    $('#gene_set_annotation').val(an);
+  }
+  drawDotplot();
+});
+
+// When changing cluster, re-render annotation options and call render function
+$('#cluster').change(function(){
```
bistline: See previous comment about event listeners.  I understand why you want to abstract this out to different .js files, but I worry about how that's going to impact what we already have (co-mingled JS in HTML views that define event listeners as needed).  Not necessarily saying this has to change, but I'd like a little more insight on this.

app/assets/javascripts/scp-dot-plot.js
eweitz: Good catch!  It is clobbering the event listener.

However, it turns out none of the event listeners  in `scp-dot-plot.js` are actually used, and their functionality is all handled in the event listeners attached via `_view_gene_expression_heatmap.html.erb`.

So I've removed that dead code.  It looks like it was started as a way to do more refined event handling before I moved dot-plot-specific code out of `_view_gene_expression_heatmap.html.erb`, but the refinement was never completed.  In other words, the event handler in https://github.com/broadinstitute/single_cell_portal_core/blob/master/app/views/site/_view_gene_expression_heatmap.html.erb#L103 renders both dot plots and heatmaps when an annotation changes.  The unrefined event handler there is chattier than strictly necessary, but seems fine.

app/assets/javascripts/scp-dot-plot.js
eweitz: My [reply](https://github.com/broadinstitute/single_cell_portal_core/pull/278#discussion_r298204507) to your previous comment addresses this case, too.

eweitz:
Thanks Jon.

Regarding Codacy issues: we discussed these at standup, and none were a blocking concern.  I plan to revise the rule sets to align with our team's quality standards so we have fewer false positive failures.

PR #284, opened 2019-07-09
Checking firecloud registration before compute permissions/billing projects check

bistline:
Skipping checking compute permissions in workspaces for users who have not registered for FireCloud.  Also, redirect to FireCloud profile page when user tries to load billing projects if they haven't registered yet.  Both will throw unnecessary 401s that get reported to Sentry.


app/controllers/billing_projects_controller.rb
```diff
@@ -217,6 +218,14 @@ def load_service_account
   #
   ##
 
+  # make sure a user is registered for firecloud before showing billing information
```
knapii-developments: Great fix!

PR #288, opened 2019-07-22
Handling for 'detached' studies

bistline:
Adding 'detached' attribute for studies (indicates the underlying workspace/bucket has been deleted)
* Will disable all edits/downloads but allow visualization to continue
* Cannot be set directly, only gets set as part of the Study.storage_sanity_check cron
* Updates to portal & API controllers to support this

### DIFF
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -970,7 +970,7 @@ module Api
       end
 
       def check_study_detached
-        if @study.detached
+        if @study.detached?
           head 410 and return
         end
       end
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -7,6 +7,9 @@ module Api
       before_action :set_current_api_user!
       before_action :set_study, except: [:studies, :analyses, :get_analysis]
       before_action :set_analysis_configuration, only: [:get_analysis, :get_study_analysis_config]
+      before_action :check_study_detached, only: [:download_data, :stream_data, :get_study_analysis_config,
+                                                  :submit_study_analysis, :get_study_submissions,
+                                                  :get_study_submission, :sync_submission_outputs]
       before_action :check_study_view_permission, except: [:studies, :analyses, :get_analysis]
       before_action :check_study_compute_permission,
                     only: [:get_study_analysis_config, :submit_study_analysis, :get_study_submissions,
@@ -127,6 +130,9 @@ module Api
           response 406 do
             key :description, 'Accept or Content-Type headers missing or misconfigured'
           end
+          response 410 do
+            key :description, 'Study workspace is not found, cannot complete action'
+          end
         end
       end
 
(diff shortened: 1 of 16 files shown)


app/controllers/api/v1/site_controller.rb
```diff
@@ -127,6 +130,9 @@ def view_study
           response 406 do
             key :description, 'Accept or Content-Type headers missing or misconfigured'
           end
+          response 410 do
```
eweitz: For others' reference: we went with a 410 error mostly due to its caching semantics.  From https://tools.ietf.org/html/rfc7231#section-6.5.9:


> The 410 (Gone) status code indicates that access to **the target
  resource is no longer available at the origin server and that this
  condition is likely to be permanent**.  If the origin server does not
  know, or has no facility to determine, whether or not the condition
  is permanent, the status code 404 (Not Found) ought to be used
  instead.

>  The 410 response is primarily intended to assist the task of web
  maintenance by notifying the recipient that the resource is
  intentionally unavailable and that the server owners desire that
  remote links to that resource be removed.  Such an event is common
  for limited-time, promotional services and for resources belonging to
  individuals no longer associated with the origin server’s site.  It
  is not necessary to mark all permanently unavailable resources as
  “gone” or to keep the mark for any length of time -- that is left to
  the discretion of the server owner.

  > **A 410 response is cacheable by default**; i.e., unless otherwise
  indicated by the method definition or explicit cache controls (see
  Section 4.2.2 of [RFC7234]).

app/controllers/api/v1/site_controller.rb
```diff
@@ -127,6 +130,9 @@ def view_study
...
+            key :description, 'Study workspace is not found, cannot complete action'
```
eweitz: The string "Study workspace is not found, cannot complete action" is duplicated some 23 times in this PR.  That seems like a lurking maintainability issue -- a developer could easily change the string in one place, but overlook changing it in another.  

Would it be worth extracting the message into a separate file and referencing it here by a variable, to ease future updates?

We've discussed ideas for DRY in the controller code previously; some related ideas are e.g. [here](https://github.com/broadinstitute/single_cell_portal_core/pull/201#discussion_r248796792) and [here](https://github.com/broadinstitute/single_cell_portal_core/pull/172#discussion_r222387268).

knapii-developments: I agree. I think this should be saved somehow as a variable.

app/controllers/api/v1/studies_controller.rb
```diff
@@ -533,6 +548,8 @@ def set_study
         @study = Study.find_by(id: params[:id])
         if @study.nil? || @study.queued_for_deletion?
           head 404 and return
+        elsif @study.detached?
+          head 410 and return
```
knapii-developments: I enjoy this check being here

app/controllers/api/v1/site_controller.rb
bistline: I think it's worth noting that is only the "display" when looking at the swagger UI.  This is not the response returned by the API itself.  Having said that, you're right - this is very much not DRY.  I'll make a commit to extract this out somehow, and put in a framework for handling common response descriptions like this.

PR #292, opened 2019-08-02
fixing issues with local dev & service account credentials

bistline:
* Addresses issue with pulling service account credentials out of vault and writing JSON files when using bin/load_env_secrets.sh

### DIFF
--- a/test/integration/fire_cloud_client_test.rb
+++ b/test/integration/fire_cloud_client_test.rb
@@ -290,8 +290,7 @@ class FireCloudClientTest < ActiveSupport::TestCase
   def test_create_configuration_template
     puts "#{File.basename(__FILE__)}: '#{self.method_name}'"
 
-    workflow_methods = @fire_cloud_client.get_methods(entityType: 'workflow')
-    method = workflow_methods.sample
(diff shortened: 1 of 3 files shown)


test/integration/fire_cloud_client_test.rb
```diff
@@ -290,8 +290,7 @@ def test_get_configurations
   def test_create_configuration_template
     puts "#{File.basename(__FILE__)}: '#{self.method_name}'"
 
-    workflow_methods = @fire_cloud_client.get_methods(entityType: 'workflow')
-    method = workflow_methods.sample
+    method = @fire_cloud_client.get_method('single-cell-portal', 'split-cluster', 1)
```
eweitz: Out of curiosity, how does this address the flaky test?

bistline: Because that is a method/configuration that is part of portal source that is run in another integration test, so we know that it will always exist and always be available.  Before, we just queried for a method and _hoped_ that it had available configurations.  Most do, but not all, so that is why occasionally this test would fail at the next assertion.

PR #294, opened 2019-08-09
Fixing Jenkins deployments

bistline:
* Addressing issues with environment variable quoting that broke deployments
* Also adding a final check to ensure portal is available at homepage (looking for HTTP status 200)


bin/remote_deploy.sh
```diff
@@ -58,4 +79,11 @@ function run_command_in_deployment {
     cd $DESTINATION_BASE_DIR ; $COMMAND
 }
 
+# make a HEAD request on URL and return HTTP status code
+function ensure_portal_is_available {
```
eweitz: `get_http_status_code` might be a more accurate name.

PR #299, opened 2019-08-21
Upgrade Webpack, Webpacker, and other dependencies

eweitz:
This enhances security by upgrading Webpack, Webpacker, IGV, Ideogram, and other dependencies.  It satisfies [SCP-1608](https://broadworkbench.atlassian.net/browse/SCP-1608).


package.json
```diff
@@ -1,10 +1,11 @@
 {
   "dependencies": {
-    "@ckeditor/ckeditor5-build-classic": "^11.0.1",
-    "@rails/webpacker": "3.5",
+    "@ckeditor/ckeditor5-build-classic": "12.3.1",
+    "@rails/webpacker": "4.0.7",
+    "core-js": "3",
     "fstream": "1.0.12",
```
bistline: This was a transient dependency that we manually declared & updated due to an issue reported in Sourceclear - ideally I'd like to remove these and have webpack manage its own transient dependencies.

package.json
```diff
@@ -1,10 +1,11 @@
...
-    "ideogram": "1.8.0",
-    "igv": "2.1.0",
+    "ideogram": "1.11.0",
+    "igv": "2.2.15",
     "jquery": "3.4.0",
     "jquery-ui": "1.12.1",
     "js-yaml": "3.13.1",
```
bistline: Same thing here - I'd like to remove these and see if by upgrading to webpack 4 we can get onto later versions that will remove the need to declare these directly.

eweitz: Resolved in 87d7ac1.

package.json
eweitz: Makes sense, resolved in 87d7ac1.  I manually confirmed it compiles, and visualizations still work.

eweitz:
It's somewhat indirect, but I plan to add a `.codacy.yaml` file to ignore failing Codacy style checks.  The violating files are essentially third-party boilerplate.

PR #300, opened 2019-08-22
Release 0.35.0

bistline:


bistline:
The codacy check failure is a red herring:
* Possible XSS vulnerability in rails-html-sanitizer (CVE-2015-7580)

This only affected versions prior to 1.0.3, we are using 1.2.0 (latest release)

PR #305, opened 2019-09-04
updates re CVE-2019-10746 & 10747

bistline:
* Addressing mixin-deep [CVE-2019-10746](https://nvd.nist.gov/vuln/detail/CVE-2019-10746) and set-value [CVE-2019-10747](https://nvd.nist.gov/vuln/detail/CVE-2019-10747)


package.json
```diff
@@ -7,7 +7,6 @@
     "igv": "2.1.0",
     "jquery": "3.4.0",
     "jquery-ui": "1.12.1",
-    "lodash": "4.17.11",
```
eweitz: I was curious where we use Lodash, and per https://github.com/broadinstitute/single_cell_portal_core/commit/77a867413a5ed8aefb6c1f68a2941605fd036aaf it seems we introduced it to address a separate CVE.

bistline: correct - just removing an unnecessary peer dependency.

PR #306, opened 2019-09-04
Fix "Send an Email to all Users" admin function

eweitz:
This fixes the "Send an Email to all Users" admin function.  It enables emails to be easily sent to all registered user accounts in the portal, which could otherwise be quite tedious.  The bug was caused by an overloooked breaking change in CKEditor.

I've verified the fix manually.  This satisfies [SCP-1863](https://broadworkbench.atlassian.net/browse/SCP-1863).

### DIFF
--- a/app/views/admin_configurations/compose_users_email.html.erb
+++ b/app/views/admin_configurations/compose_users_email.html.erb
@@ -33,18 +33,18 @@
 
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
 
-    function validateEmail() {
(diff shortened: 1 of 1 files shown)


vickylihorst:
Thanks Eric! Could we add a test to the test suite sometime after this PR so future breaking change in CKEditor won't be overlooked like this?

eweitz:
Good idea Vicky.  I created [SCP-1865](https://broadworkbench.atlassian.net/browse/SCP-1865) to track that task.

PR #308, opened 2019-09-05
PAPI Integration V1

bistline:
* New classes for visualizing Firestore data
  * FirestoreGene
  * FirestoreCellMetadatum
  * FirestoreCluster
* New classes for launching Ingest pipeline runs in Google Pipelines API (PAPI)
  * PapiClient (wrapper around PAPI calls)
  * IngestJob (wrapper for polling PAPI for completion and updating state)

TODO: 
* Testing through Selenium & end-to-end integration test that can run in Travis
* Handling existing file types/classes that maintained associations to now defunct classes (Coordinate Labels, User Annotations)
* Optimization of deletes in Firestore to prevent user blocking
* Optimization of DelayedJob queueing


app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+        commands: commands,
+        environment: environment,
+        flags: flags,
+        image_uri: INGEST_DOCKER_IMAGE,
+        labels: labels,
+        timeout: timeout
+    )
+  end
+
+  # Instatiate a resources object to tell where to run a pipeline
```
eweitz:
```suggestion
  # Instantiate a resources object to tell where to run a pipeline
```

app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+  # * *params*
+  #   - +study_file+ (StudyFile) => StudyFile to be ingested
+  #   - +action+ (String/Symbol) => Action to perform on ingest
+  #
+  # * *return*
+  #   - (Array) Command Line, in Docker "exec" format
+  #
+  # * *raises*
+  #   - (ArgumentError) => The requested StudyFile and action do not correspond with each other, or cannot be run yet
+  def get_command_line(study_file:, action:)
```
eweitz: Succinct!

I'll keep an eye on this -- I'm interested to see how our forthcoming integration validation commands will relate to this method.

app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+        command_line += " --matrix-file #{study_file.gs_url} --matrix-file-type dense"
+      elsif study_file.file_type === 'MM Coordinate Matrix'
+        bundled_files = study_file.bundled_files
+        genes_file = bundled_files.detect {|f| f.file_type == '10X Genes File'}
+        barcodes_file = bundled_files.detect {|f| f.file_type == '10X Barcodes File'}
+        command_line += " --matrix-file #{study_file.gs_url} --matrix-file-type mtx" \
+                      " --gene-file #{genes_file.gs_url} --barcode-file #{barcodes_file.gs_url}"
+      end
+    when 'ingest_cell_metadata'
+      command_line += " --cell-metadata-file #{study_file.gs_url} --ingest-cell-metadata"
```
eweitz: It doesn't concern this PR per se, but I'm still unsure why `--ingest-cell-metadata` is needed.  It seems like it should already be entailed by the subparser argument `ingest_cell_metadata`.

(And similarly for `ingest_cluster` / `--ingest-cluster`, and `ingest_subsample` / `--subsample`.)

app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+  #   - +study_file+ (StudyFile) => File to be ingested
+  #
+  # * *returns*
+  #   (Array) => Array representation of optional arguments (Docker exec form), based on file type
+  def get_command_line_options(study_file)
+    opts = []
+    case study_file.file_type
+    when /Matrix/
+      taxon = study_file.taxon
+      opts += ["--taxon-name", "#{taxon.scientific_name}", "--taxon-common-name", "#{taxon.common_name}",
```
eweitz: Could you add a comment in the code explaining what is passed for these three fields when their value is empty / null?  (I forget if it's an empty string; a code comment would ease recall.)

app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+               "--ncbi-taxid", "#{taxon.ncbi_taxid}"]
+      if taxon.current_assembly.present?
+        assembly = taxon.current_assembly
+        opts += ["--genome-assembly-accession", "#{assembly.accession}"]
+        if assembly.current_annotation.present?
+          opts += ["--genome-annotation", "#{assembly.current_annotation.name}"]
+        end
+      end
+    when 'Cluster'
+      opts += ["--name", "#{study_file.name}"]
```
eweitz: For a cluster named "Infected" in a file named "my_cluster.txt", is "Infected" the value of `study_file.name`?

app/models/papi_client.rb
```diff
@@ -0,0 +1,307 @@
...
+  extend ErrorTracker
+
+  # Service account JSON credentials
+  SERVICE_ACCOUNT_KEY = !ENV['SERVICE_ACCOUNT_KEY'].blank? ? File.absolute_path(ENV['SERVICE_ACCOUNT_KEY']) : ''
+  # Google authentication scopes necessary for running pipelines
+  GOOGLE_SCOPES = %w(https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/datastore)
+  # GCP Compute project to run pipelines in
+  COMPUTE_PROJECT = ENV['GOOGLE_CLOUD_PROJECT'].blank? ? '' : ENV['GOOGLE_CLOUD_PROJECT']
+  # Docker image in GCP project to pull for running ingest jobs
+  INGEST_DOCKER_IMAGE = 'gcr.io/broad-singlecellportal-staging/ingest-pipeline:0.3.1_2959221'
```
eweitz: Is it critical to make `INGEST_DOCKER_IMAGE` configurable before launch?

If so, what kind of interface would be best to update values like "gcr.io/broad-singlecellportal-staging/ingest-pipeline:0.3.1_2959221" -- an environment variable set via CLI argument, an admin GUI, or perhaps something else?

app/models/concerns/firestore_documents.rb
```diff
@@ -72,13 +79,30 @@ def self.by_study_and_name(accession, name)
...
 
+    ##
+    # Delete methods
+    ##
+
+    def self.delete_by_study(accession, threads=10)
+      delete_documents(self.query_by(study_accession: accession), threads)
+    end
+
+    def self.delete_by_study_and_file(accession, file_id, threads=10)
```
knapii-developments: Whats the reason for self.delete_by_study and this one?

app/models/papi_client.rb
knapii-developments: This has to deal with how we tell ingest what to ingest. Sub-parsers are not captured in arguments. Cluster, cell metadata, and subsampling sub-parsers share argument names that are the same. We simply cannot check to see if an argument exist to see what function to invoke.  If you don't want --ingest-cell-metdata we should think of other ways to invoke ingest of cell-metadata, cluster and subsampling.

eweitz: > Sub-parsers are not captured as arguments from argparse

That's the piece I was missing.  Thanks for helping me understand!

bistline: Yeah, I have the same concern, and while I think that there _might_ be a better way of doing this, it does make sense in terms of the hierarchy of the parsers/parameters.

app/models/papi_client.rb
bistline: Can do.

app/models/papi_client.rb
bistline: The `name` field for Cluster files actually corresponds to the name of the cluster represented in the file contents.  The name of the file in this case is preserved in `study_file.upload_file_name`.  It's just the case that for most other file types, `study_file.name === study_file.upload_file_name`.

app/models/papi_client.rb
bistline: I agree that this should be configurable eventually.  There are a variety of ways that we could achieve this.  I think the most robust would be to use the `AdminConfiguration` class and create a new entry type for "Ingest Pipeline Docker Image".  For now, the constant can suffice.

I've created [SCP-1872](https://broadworkbench.atlassian.net/browse/SCP-1872) to account for this.

app/models/concerns/firestore_documents.rb
bistline: Mainly, we would use `delete_by_study_and_file` when a user deletes a single file from their study.  However, in the case of a user deleting their entire study without first clearing out files, we could use `delete_by_study` instead of having to loop through each file first.

app/models/papi_client.rb
eweitz: Thanks for clarifying!  (For my convenience, yes, "Infected" would be the value of `study_file.name` in that hypothetical case.)

PR #311, opened 2019-09-26
Changing references of FireCloud to Terra

bistline:
* Includes UI, API, and regression test updates to support transition to Terra

This fulfills [SCP-1886](https://broadworkbench.atlassian.net/browse/SCP-1886)


bistline:
Given that the FISMA boundary still refers to "FireCloud" instead of Terra, and the API is still api.firecloud.org, I don't see any reason to change internal references to Terra.

eweitz:
That makes sense!

PR #313, opened 2019-10-01
Study Accession refinements

bistline:
* Allows searching for studies via accession values
* Accessions are now assigned before study creation rather than after to allow for API responses to include accession values

This PR satisfies [SCP-1828](https://broadworkbench.atlassian.net/browse/SCP-1828)

### DIFF
--- a/app/models/study_accession.rb
+++ b/app/models/study_accession.rb
@@ -7,8 +7,6 @@ class StudyAccession
 
   validates_uniqueness_of :accession
 
-  after_create :set_study_accession_value
-
   # is this accession currently assigned to an existing study?
   def assigned?
     self.study.present?
@@ -25,13 +23,5 @@ class StudyAccession
       study.assign_accession
       puts "Accession for #{study.name} assigned: #{study.accession}"
     end
-  end
-
-  private
-
-  # once an accession is persisted, set the accession value in the associated study
-  def set_study_accession_value
-    study = self.study
-    study.update(accession: self.accession)
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/study_accession.rb
```diff
@@ -27,11 +30,15 @@ def self.assign_accessions
...
-  # once an accession is persisted, set the accession value in the associated study
-  def set_study_accession_value
-    study = self.study
-    study.update(accession: self.accession)
+  # sanitize an list of terms to format as a StudyAccession
+  def self.sanitize_accessions(terms)
+    possible_accessions = []
+    terms.each do |term|
+      accession_string = term.strip.gsub(ACCESSION_SANITIZER, '')
+      if accession_string.match(ACCESSION_FORMAT)
```
eweitz: Per our chat, this approach is permissive and seems like it could match more cases than expected (e.g. "SCPBRCA1", "SCP BRCA1"), but these counterexamples are totally contrived and I think actual cases of false positive matches for accession search would be vanishingly rare.

If we ever need a more precise pattern, simply matching on `/^SCP[:_]?\d+$/` might work.

bistline: I agree, but the use case is a little more refined than that.  We need to trim out all the things that don't match SCP123 so that we can search on them, so if we match on SCP:123, the search will then fail.  What we would like to have happen:

User enters SCP123 - search directly on this value
User enters SCP:123 - portal views this as a possible accession, strips out the :, and searches on SCP123
User enters SCPBRCA1 - portal does NOT view this as a possible accession and ignores.

So totally possible, but requires some regex gymnastics that I don't think are really very useful right now for this feature.

PR #323, opened 2019-10-10
MVP for confusing early failure--add log message

skiggety:
This PR is a bit of a straw-man, I just wanted to post up an MVP change so we definitely have something to address the false failure in the last prod deploy jenkins job. Feel free to reword or add on better logic.


skiggety:
Also FWIW: https://github.com/broadinstitute/mmrf-researcher-portal/blob/qa/workflow/run_webapp_container#L80

PR #327, opened 2019-10-25
Add favicon, for better findability and branding (SCP-1469)

eweitz:
This adds a [favicon](https://en.wikipedia.org/wiki/Favicon) to Single Cell Portal.  It enables users to more easily find SCP among many other web browser tabs, and enhances SCP's branding.

Here's how it looks:
scp_favicon_screenshot

This satisfies SCP-1469.


eweitz:
The failing check seems to be a bug in Codacy: "Codacy was unable to analyse your pull request".  Their suggested remedy -- clicking "Re-try" -- also failed.  I think the failure is a false positive.

PR #333, opened 2019-11-15
Reporting error messages from failed ingest runs

bistline:
Retrieve all error/warning messages that are delocalized back to the workspace bucket and report to the end user via email.

This satisfies [SCP-1944](https://broadworkbench.atlassian.net/browse/SCP-1944)


app/models/papi_client.rb
```diff
@@ -195,7 +195,7 @@ def create_resources_object(regions:)
   #   - +preemptible+ (Boolean) => Indication of whether VM can be preempted (defaults to false)
   # * *return*
   #   - (Google::Apis::GenomicsV2alpha1::VirtualMachine)
-  def create_virtual_machine_object(machine_type: 'n1-standard-1', preemptible: false)
+  def create_virtual_machine_object(machine_type: 'n1-highmem-4', preemptible: false)
```
eweitz: Good first step to help ensure we can ingest non-trivial files.

PR #335, opened 2019-11-19
Docker base image update

bistline:
* Updating to using new Docker image singlecellportal/rails-baseimage:0.2.3, which incorporates the marketplace.gcr.io/google/ubuntu1804:latest managed base image.  This image is regularly updated and patched by Google to include all necessary security patches/fixes.

This satisfies [SCP-1999](https://broadworkbench.atlassian.net/browse/SCP-1999)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # use KDUX base Rails image, configure only project-specific items here
-FROM singlecellportal/rails-baseimage:0.2.3
+FROM singlecellportal/rails-baseimage:1.0.0
 
 # Set ruby version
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # use KDUX base Rails image, configure only project-specific items here
-FROM broadinstitute/kdux-rails-baseimage:1.9
+FROM singlecellportal/rails-baseimage:0.2.3
```
skiggety: version 1.0.0 is now available
```suggestion
FROM singlecellportal/rails-baseimage:1.0.0
```

PR #337, opened 2019-12-02
End-to-end integration with Ingest Pipeline

bistline:
This adds end-to-end integration with the [SCP Ingest Pipeline](https://github.com/broadinstitute/scp-ingest-pipeline) for ingesting parseable files (expression, metadata, and clustering).  Data is written back to MongoDB via jobs submitted to the Google Pipelines API (PAPI).

This satisfies [SCP-1940](https://broadworkbench.atlassian.net/browse/SCP-1940)


test/ui_test_suite.rb
```diff
@@ -273,20 +273,22 @@ def teardown
...
-    coords_form = @driver.find_element(:class, 'initialize_labels_form')
-    cluster_select = coords_form.find_element(:id, 'study_file_options_cluster_group_id')
-    cluster_select.send_keys('Test Cluster 1')
-    upload_coords = coords_form.find_element(:class, 'upload-labels')
-    upload_coords.send_keys(@test_data_path + 'coordinate_labels_1.txt')
-    upload_btn = coords_form.find_element(:id, 'start-file-upload')
-    sleep(0.5)
-    upload_btn.click
-    close_modal('upload-success-modal')
+    # wait_for_render(:class, 'add-coordinate-labels')
```
eweitz: Does this branch of SCP not support upload coordinate label files?  If so, is such support critical for launch?

bin/boot_docker
```diff
@@ -141,7 +145,8 @@ then
 	# generate random secret key for secure cookies
 	docker run -d --name "$CONTAINER_NAME" -p 80:80 -p 443:443 -h "$PROD_HOSTNAME" -v "$PROJECT_DIR:/home/app/webapp:rw" \
 	    -v "$PROJECT_DIR/data:/home/app/webapp/data:rw" -v "$PROJECT_DIR/log/nginx:/var/log/nginx:rw" -e PASSENGER_APP_ENV="$PASSENGER_APP_ENV" \
-	    -e MONGO_LOCALHOST="$MONGO_LOCALHOST" -e PROD_DATABASE_PASSWORD="$PROD_DATABASE_PASSWORD" -e SERVICE_ACCOUNT_KEY="$SERVICE_ACCOUNT_KEY" \
+	    -e MONGO_LOCALHOST="$MONGO_LOCALHOST" -e MONGO_INTERNAL_IP="$MONGO_INTERNAL_IP" -e PROD_DATABASE_PASSWORD="$PROD_DATABASE_PASSWORD" \
+	    -e GCP_NETWORK_NAME=$GCP_NETWORK_NAME -e GCP_SUB_NETWORK_NAME=$GCP_SUB_NETWORK_NAME -e SERVICE_ACCOUNT_KEY="$SERVICE_ACCOUNT_KEY" \
```
eweitz: Could you point me to where these two new environment variables -- GCP_NETWORK_NAME and GCP_SUB_NETWORK_NAME -- are defined? I don't see them among the command-line options defined above, so my hunch is this is a bug.

Assuming that's so, once those CLI options are added, would it also be worth documenting them in the README?

test/ui_test_suite.rb
bistline: It will support them, but there is an inherent race condition right now where the cluster object itself is not created quickly enough for the UI test to pass - we'd have to wait the minute or two for that parse to complete for this particular test to work.  In a real-world scenario, users will still be able to upload coordinate labels as always, so it shouldn't be a change in production.  It's just this particular test can't do it in a timeframe that I consider tractable.

bin/boot_docker
bistline: Yes, they will need to be added to our vault configurations.  As far as the README is concerned, it would only affect installations where the MongoDB VM is not on the default network, which is likely not something other places would do.  I'll see if there's a place where it makes sense to call this out, regardless.

eweitz:
Please add a security risk label before merging.

PR #338, opened 2019-12-04
(SCP-1997) ingest integration testing, continued from Jon's similarly named branch

skiggety:


test/integration/study_creation_test.rb
```diff
@@ -30,400 +31,45 @@ class StudyCreationTest < ActionDispatch::IntegrationTest
...
     # expression matrix #1
     file_params = {study_file: {file_type: 'Expression Matrix', study_id: @study.id.to_s}}
     perform_study_file_upload('expression_matrix_example.txt', file_params, @study.id)
     assert_response 200, "Expression matrix upload failed: #{@response.code}"
     assert @study.expression_matrix_files.size == 1, "Expression matrix failed to associate, found #{@study.expression_matrix_files.size} files"
     expression_matrix_1 = @study.expression_matrix_files.first
-    @study.initialize_gene_expression_data(expression_matrix_1, @test_user)
-    assert @study.genes.size > 0, 'Did not parse any genes from expression matrix'
+    assert_equal 'unparsed', expression_matrix_1.parse_status
+    initiate_study_file_parse('expression_matrix_example.txt', @study.id)
```
skiggety: @bistline , you had mentioned passing in file_params from the outside like with perform_study_file_upload, but I'm not sure what they would be, if anything. In fact, it looks like the way perform_study_file_upload is actually used, @study.id is passed in twice, creating the possibility of inconsistency. Thoughts?

PR #343, opened 2019-12-16
SCP-2020: Sl ingest validation tests

skiggety:


skiggety:
This should be done as soon as I plug in a version of the ingest pipeline that it likes, hence putting it up as a draft.

PR #344, opened 2019-12-17
Add BigQuery parameters, log Ingest request (SCP-1978)

eweitz:
This update will:
* Add BigQuery parameters needed to load to metadata backend
* Log parameters sent to Ingest Pipeline
* Remove extraneous Firestore permissions

This interstitial merge relates to SCP-1978, but more work is needed to resolve that issue.


app/models/papi_client.rb
```diff
@@ -97,6 +97,8 @@ def run_pipeline(study_file: , user:, action:)
     action = self.create_actions_object(commands: command_line, environment: environment)
     pipeline = self.create_pipeline_object(actions: [action], environment: environment, resources: resources)
     pipeline_request = self.create_run_pipeline_request_object(pipeline: pipeline, labels: labels)
+    Rails.logger.info "Request object sent to Google Pipelines API (PAPI):"
+    Rails.logger.info pipeline_request.to_yaml
```
bistline: We need to remove the `environment` attribute of the pipeline request as it contains the database password.  We should not be logging this.  Since you can't remove attributes from a YAML representation, you'll probably need to call `to_h` to cast it to a Hash and then remove the `environment` section of the pipeline request.

eweitz: Good catch!  Fixed in 9f94b15.

PR #350, opened 2020-01-03
Upload Wizard updates, yarn compilation

bistline:
Updates to upload wizard:
* Adding link to metadata convention
* Removing text regarding sorting sparse matrices

Yarn compilation in local development now invokes `yarn upgrade` rather than `yarn:install`.  This is to force upgrades and continually update the `yarn.lock` file to avoid compilation issues when deploying if a package registry URL has changed.  This is preferable to removing the yarn.lock file altogether as it will still enforce consistency on deployments, and will still be validated in CI before deployment.

This PR fulfills [SCP-1986](https://broadworkbench.atlassian.net/browse/SCP-1986) and [SCP-2048](https://broadworkbench.atlassian.net/browse/SCP-1986).


app/views/studies/_sort_warning.html.erb
```diff
@@ -1 +1,5 @@
-<p class='text-danger sort-warning'><strong>Coordinate matrices must be sorted by gene indices (first column) in order to be parsed.</strong>  Please refer to our <a href='https://github.com/broadinstitute/single_cell_portal/wiki/Sorting-Matrix-Market-Coordinate-Matrices' target='_blank' style='text-decoration: underline;'>wiki</a> for more information.</p>
```
eweitz: This small change is a big deal!  

I added [a note](https://github.com/broadinstitute/single_cell_portal/wiki/Sorting-Matrix-Market-Coordinate-Matrices) to remove the linked wiki page after launching Ingest Pipeline.

bistline: Yes - good catch.  I'm glad we wont have to force users to do that anymore.

PR #351, opened 2020-01-03
Gating subsampling, adding cleanup

bistline:
Adding the `subsampled` flag to ClusterGroups to gate invocation of subsampling via ingest pipeline to avoid a race condition when both a cluster and cell metadata file finish parsing at roughly the same time.

Also, when deleting a cell metadata file, all subsampled data will be deleted as it will need to be regenerated once a new cell metadata file is added.

This PR satisfies [SCP-2050](https://broadworkbench.atlassian.net/browse/SCP-2050).

### DIFF
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -316,15 +316,20 @@ class ClusterGroup
     SUBSAMPLE_THRESHOLDS.select {|sample| sample < self.points}
   end
 
+  # find all 'subsampled' data arrays
+  def find_subsampled_data_arrays
+    DataArray.where(study_id: self.study_id, study_file_id: self.study_file_id, linear_data_type: 'ClusterGroup',
+                    linear_data_id: self.id, :subsample_threshold.nin => [nil],
+                    :subsample_annotation.nin => [nil])
+  end
+
   # control gate for invoking subsampling
   def can_subsample?
     if self.points < SUBSAMPLE_THRESHOLDS.min || self.subsampled
       false
     else
       # check if there are any data arrays belonging to this cluster that have a subsample threshold & annotation
-      !DataArray.where(study_id: self.study_id, study_file_id: self.study_file_id, linear_data_type: 'ClusterGroup',
-                       linear_data_id: self.id, :subsample_threshold.nin => [nil],
-                       :subsample_annotation.nin => [nil]).any?
+      !self.find_subsampled_data_arrays.any?
     end
   end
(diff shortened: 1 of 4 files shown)


app/models/cluster_group.rb
```diff
@@ -315,6 +316,18 @@ def subsample_thresholds_required
     ClusterGroup::SUBSAMPLE_THRESHOLDS.select {|sample| sample < self.points}
   end
 
+  # control gate for invoking subsampling
+  def can_subsample?
+    if self.points < 1000 || self.subsampled
```
eweitz: Setting `MIN_THRESHOLD = 1000` and using that both here and in `SUBSAMPLE_THRESHOLDS` would avoid potential bugs caused by updating `1000` here but not in `SUBSAMPLE_THRESHOLDS` (or vice versa).

bistline: How about `SUBSAMPLE_THRESHOLDS.min` instead?  Then it's only one place that needs updating, which I think is the goal of your suggestion.

eweitz: Sounds good!

app/models/cluster_group.rb
```diff
@@ -312,7 +313,19 @@ def generate_subsample_arrays(sample_size, annotation_name, annotation_type, ann
...
+    SUBSAMPLE_THRESHOLDS.select {|sample| sample < self.points}
+  end
+
+  # control gate for invoking subsampling
+  def can_subsample?
+    if self.points < SUBSAMPLE_THRESHOLDS.min || self.subsampled
+      false
+    else
+      # check if there are any data arrays belonging to this cluster that have a subsample threshold & annotation
+      !DataArray.where(study_id: self.study_id, study_file_id: self.study_file_id, linear_data_type: 'ClusterGroup',
```
skiggety: seeing this pattern repeat itself made me debate with myself if it would be worth method extracting something like this:
`
!matching_data_arrays(self).any?
`
`
def self.matching_data_arrays(cluster_group)
  DataArray.where(study_id: self.study_id, study_file_id: self.study_file_id, linear_data_type: 'ClusterGroup',
    linear_data_id: self.id, :subsample_threshold.nin => [nil],
                       :subsample_annotation.nin => [nil])
end
`
..with delete_queue_job taking advantage as well.

bistline: Good call - I'll go ahead and use that.

PR #352, opened 2020-01-08
Limiting multiple/global gene searches to 50 genes

bistline:
For performance reasons (e.g. MongoDB running out of memory), we are now limiting multi-gene searches to a maximum of 20 genes, both client- and server-side.

This PR satisfies AC 1 & 3 from [SCP-2007](https://broadworkbench.atlassian.net/browse/SCP-2007).


app/assets/javascripts/application.js
```diff
@@ -37,6 +37,9 @@ var PAGE_RENDERED = false;
 var OPEN_MODAL = '';
 var CLUSTER_TYPE = '3d';
 var UNSAFE_CHARACTERS = /[\;\/\?\:\@\=\&\'\"\<\>\#\%\{\}\|\\\^\~\[\]\`]/g;
+var MAX_GENE_SEARCH = 20;
+var GENE_SEARCH_LIMIT_MSG = 'For performance reasons, gene search is limited to ' + MAX_GENE_SEARCH + ' genes.  ' +
```
eweitz: Having two copies of these lines -- one here and in one in `app/models/gene.rb` -- will require vigilance to keep in sync.  Ideas to mitigate that:

* Assign these two variables in a `<script>` tag atop the Rails view template [application.html.erb](https://github.com/broadinstitute/single_cell_portal_core/blob/035a8250d9d8354ddd21115fd30532bba0f92ead/app/views/layouts/application.html.erb#L9), using a value defined in server code.  E.g. `window.MAX_GENE_SEARCH = <%= @my_object.MAX_GENE_SEARCH %>`. Assign them in `app/models/gene.rb` by referencing the same server code.

* Add a comment here and in `app/models/gene.rb` to change the other place when updating either value.

(If they're intended to be independent, then having names that suggest they're different -- e.g. incorporating `GLOBAL` and `STUDY` into the variable names -- would help maintainers.)

bistline: They're not intended to be independent, I just couldn't think of an easy way to use the value from Ruby in `application.js`.  But your idea of using `window` in `application.html.erb` will solve that nicely.  I'll make that change.

app/assets/javascripts/application.js
```diff
@@ -258,11 +258,18 @@ function initializeAutocomplete(selector, entities) {
             },
             select: function(event, ui) {
                 var terms = split(this.value);
-                // remove the current input
-                terms.pop();
-                // add the selected item
-                terms.push(ui.item.value);
-                terms.push("");
+                // check if user has added more that 20 genes, in which case alert and remove the last term
```
skiggety: looks like terms.pop() could go above the conditional if the comparator becomes a ">=". It's not obvious to me why there is an "extra" pop (result not saved) and push (blank input), which may be just my own lack of context.

bistline: Oh yeah - I don't know why I didn't think of that.  I just need to adjust the math on the `if` statement.

PR #353, opened 2020-01-09
delete cell metadata in bigquery (for SCP-2039 and SCP-2043)

skiggety:


app/controllers/application_controller.rb
```diff
@@ -1,3 +1,5 @@
+require "google/cloud/bigquery"
```
bistline: Do we need this require statement here?  I though including it in the Gemfile was sufficient.

test/integration/study_creation_test.rb
```diff
@@ -108,23 +118,48 @@ class StudyCreationTest < ActionDispatch::IntegrationTest
...
-    study.reload
     gene_count = study.gene_count
     cluster_count = study.cluster_groups.size
     metadata_count = study.cell_metadata.size
     cluster_annot_count = study.cluster_annotation_count
     study_file_count = study.study_files.non_primary_data.size
     share_count = study.study_shares.size
 
-    assert_equal 19, gene_count, "did not find correct number of genes"
+    # I don't need this, perhaps, because this happens a little later: assert_equal 19, gene_count, "did not find correct number of genes"
```
bistline: I would remove this if we in fact don't need it anymore.

test/integration/study_creation_test.rb
```diff
@@ -108,23 +118,48 @@ class StudyCreationTest < ActionDispatch::IntegrationTest
...
-    assert_equal 3, metadata_count, "did not find correct number of metadata objects"
+    assert_equal 22, metadata_count, "did not find correct number of metadata objects"
     assert_equal 2, cluster_annot_count, "did not find correct number of cluster annotations"
     assert_equal 3, study_file_count, "did not find correct number of study files"
     assert_equal 1, share_count, "did not find correct number of study shares"
 
+    assert_equal initial_bq_row_count + 30, get_bq_row_count(bq_dataset, study)
+
+    # request delete
+    study.study_files.each do |sf|
```
bistline: Do we need to delete all the study files here?  Don't we just want to delete the metadata file and confirm that it cleans up?  After looking at this, this almost seems like this portion (the delete and BQ validation) should be its own  validation test that would run after this test completes and ingests data correctly.

skiggety: I'll change this. I had been thinking about checking that all the files got deleted while I was at it, but ended up not wanting to more wait time to the test.

app/controllers/application_controller.rb
skiggety: Whoops, should have just done it in the gemfile...

test/integration/study_creation_test.rb
skiggety: all set

test/integration/study_creation_test.rb
skiggety: all set

app/controllers/application_controller.rb
skiggety: all set now

skiggety:
The test that I worked on for this has exposed an intermittent issue, captured in SCP-2064. It's not an issue with this code or functionality, but we may want to **delay merging this** for now nevertheless so we don't have travis intermittently failing for any branch.

bistline:
After speaking with the OLS team, we've discovered the issue is likely on our end - too many concurrent requests that flood the outgoing TCP/IP socket and cause some to time out or get refused.  We can address this in the ingest pipeline codebase with sleeps and exponential backoffs.

As a result, there isn't anything wrong with this branch, so I'm going to go ahead and merge it.

PR #354, opened 2020-01-10
Limiting gene search to 50 genes

bistline:
For performance reasons (e.g. MongoDB running out of memory), we are now limiting multi-gene searches to a maximum of 20 genes, both client- and server-side.

This PR satisfies AC 1 & 3 from [SCP-2007](https://broadworkbench.atlassian.net/browse/SCP-2007), but is for non scp-ingest-pipeline branches.

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -136,6 +136,11 @@ class SiteController < ApplicationController
   # will redirect to appropriate method as needed
   def search_genes
     @terms = parse_search_terms(:genes)
+    # limit gene search for performance reasons
+    if @terms.size > Gene::MAX_GENE_SEARCH
+      @terms = @terms.take(Gene::MAX_GENE_SEARCH)
+      search_message = Gene::MAX_GENE_SEARCH_MSG
+    end
     # grab saved params for loaded cluster, boxpoints mode, annotations, consensus and other view settings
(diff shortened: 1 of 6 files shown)


app/controllers/site_controller.rb
```diff
@@ -205,7 +210,10 @@ def search_all_genes
       delim = params[:search][:genes].include?(',') ? ',' : ' '
       raw_genes = params[:search][:genes].split(delim)
       @genes = sanitize_search_values(raw_genes).split(',').map(&:strip)
-
+      # limit gene search for performance reasons
+      if @genes.size > Gene::MAX_GENE_SEARCH
+        @genes = @genes.take(Gene::MAX_GENE_SEARCH)
+      end
```
devonbush: I'll bet there's a good reason, but I'm not following why here we don't set a message while in search_genes we do--is that becasue search_genes forwards to here?

devonbush: don't hold anything up on my account -- I assume this comment is  just a result of seeing rails paths that I have yet to fully trace myself

bistline: it's done at the UI level for the global gene search.  There's a way you can short-circuit it for the autocomplete in a single study, but not the global one.  So the alert always shows on the global gene search.

PR #356, opened 2020-01-14
Reverting expression matrix parsing to local Rails parsers

bistline:
Reverting parsing of expression matrices to using the local Rails parsers, rather than scp-ingest-pipeline.  Also optimizes searching for single/multiple genes by doing presence checks first, rather than loading gene data multiple times (expression data is only pulled from the database at violin/box/heatmap render time).

This PR satisfies [SCP-2066](https://broadworkbench.atlassian.net/browse/SCP-2066)

### DIFF
--- a/app/models/gene.rb
+++ b/app/models/gene.rb
@@ -68,6 +68,11 @@ class Gene
   # CLASS INSTANCE METHODS
   ##
 
+  # find if a study has a given gene quickly for search gating
+  def self.study_has?(study_id, expr_matrix_ids, gene_name)
+    Gene.where(study_id: study_id, :study_file_id.in => expr_matrix_ids).any_of({name: gene_name},{searchable_name: gene_name}).exists?
+  end
+
   # calculate a mean value for a given gene based on merged expression scores hash
   def self.mean(scores, cells)
     values = cells.map {|c| scores[c].to_f}
--- a/app/models/gene.rb
+++ b/app/models/gene.rb
@@ -69,7 +69,7 @@ class Gene
   ##
 
   # find if a study has a given gene quickly for search gating
-  def self.study_has?(study_id, expr_matrix_ids, gene_name)
+  def self.study_has_gene?(study_id:, expr_matrix_ids:, gene_name:)
     Gene.where(study_id: study_id, :study_file_id.in => expr_matrix_ids).any_of({name: gene_name},{searchable_name: gene_name}).exists?
   end
 
--- a/app/views/studies/initialize_study.html.erb
+++ b/app/views/studies/initialize_study.html.erb
@@ -52,7 +52,8 @@
           </div>
           </br>
           <div class="row">
-            <p class="col-sm-12">* Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
+            <p class="col-sm-12">* Before upload, <strong>MM Coordinate Matrix file must be <%= link_to 'sorted', 'https://github.com/broadinstitute/single_cell_portal/wiki/Sorting-Matrix-Market-Coordinate-Matricesx', target: :_blank %></strong>.
+              Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
               <%= link_to 'barcodes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.tsv or .csv) files separately.</p>
             <p class="col-sm-12">** <%= link_to 'Gzipped', 'https://www.gnu.org/software/gzip/manual/gzip.html', target: :_blank %>  files of this type (e.g. .txt.gz) are accepted as well</p>
           </div>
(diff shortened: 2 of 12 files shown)


app/views/studies/initialize_study.html.erb
```diff
@@ -102,7 +103,7 @@
                 <li>The header of <strong>“TYPE”</strong>to declare metadata types (see below).</li>
                 <li>A value for each metadata column declaring its datatype</li>
                 <ul>
-                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong>(continuous scores). *</li>
+                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong> (continuous scores). *</li>
```
eweitz: A similar spacing nit:

```suggestion
                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong> (continuous scores).*</li>
```

app/views/studies/initialize_study.html.erb
```diff
@@ -132,15 +133,15 @@
...
+              <li>A header row containing the value <strong>“NAME”, “X”, “Y”,</strong> optionally <strong>“Z”</strong>, and columns containing cell-level annotations
               </li>
               <li>A second row with:</li>
               <ul >
                 <li>The header of <strong>“TYPE”</strong>to declare metadata types (see below).</li>
                 <li>A value for each metadata column declaring its datatype
                 </li>
                 <ul>
-                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong>(continuous scores). *</li>
+                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong> (continuous scores). *</li>
```
eweitz:
```suggestion
                  <li>The two accepted values are <strong>“group”</strong> (set membership) or <strong>“numeric”</strong> (continuous scores).*</li>
```

test/ui_test_suite.rb
```diff
@@ -1361,6 +1361,16 @@ def teardown
     assert primary_data_count == 1, "did not find correct number of primary data files, expected 1 but found #{primary_data_count}"
     assert other_data_count == 19, "did not find correct number of other data files, expected 19 but found #{primary_data_count}"
 
+    # give time for parses to complete
```
eweitz: For easier reading:

```suggestion
    # give 2 minutes time for parses to complete
```

I like that this serves not only as a functional test, but also a performance test.

app/models/gene.rb
```diff
@@ -68,6 +68,11 @@ def species
   # CLASS INSTANCE METHODS
   ##
 
+  # find if a study has a given gene quickly for search gating
+  def self.study_has?(study_id, expr_matrix_ids, gene_name)
```
devonbush: Nice, I like how you pulled this into a separate method.  How about renaming to `study_has_gene?` for clarity?

Also, I could see using named parameters here so that mixing up argument order is less likely

app/models/gene.rb
```diff
@@ -68,6 +68,11 @@ def species
...
+    Gene.where(study_id: study_id, :study_file_id.in => expr_matrix_ids).any_of({name: gene_name},{searchable_name: gene_name}).exists?
+  end
+
```
devonbush: The post-rails 4 preferred syntax for searching where something is in an array is just `Gene.where(study_id: study_id, study_file_id: expr_matrix_ids) ` .   is that applicable for Mongoid here?

bin/cycle_logs.rb
```diff
@@ -17,15 +17,23 @@
     end
   else
 		all_logs.select {|l| l =~ /#{i}/}.each do |log|
-      basename = log.split('.').first
+      if log.split('.').first == 'delayed_job' && log.split('.').size == 3
```
devonbush: this could use a comment explaining the intent

test/ui_test_suite.rb
bistline: [ad9f42c](https://github.com/broadinstitute/single_cell_portal_core/pull/356/commits/ad9f42cf41c378362223edbf639e821c2eb265b2)

app/views/studies/initialize_study.html.erb
bistline: [ad9f42c](https://github.com/broadinstitute/single_cell_portal_core/pull/356/commits/ad9f42cf41c378362223edbf639e821c2eb265b2)

app/views/studies/initialize_study.html.erb
bistline: [ad9f42c](https://github.com/broadinstitute/single_cell_portal_core/pull/356/commits/ad9f42cf41c378362223edbf639e821c2eb265b2)

test/ui_test_suite.rb
```diff
@@ -2373,9 +2383,10 @@ def teardown
     assert rendered, "cluster plot did not finish rendering, expected true but found #{rendered}"
 
     # check for coordinate labels
-    labels = @driver.execute_script("return layout.scene.annotations;")
-    assert_not_nil labels, 'Did not return coordinate labels'
-    assert labels.size == 8, "Did not find coorect number of coordinate labels, expected 8 but found #{labels.size}"
+    # TODO: figure out how to get these back into first create-study: test so we can assert this
```
devonbush: what's the SCP team norm around committing TODOs?  My preferred approach is that TODOs require creation of a corresponding ticket, and then the TODO has to reference the ticket -- otherwise the comment should use more neutral language like 'MAYBE' or 'Would be nice if'.  But I'll defer to whatever SCP usually does

bin/cycle_logs.rb
bistline: Sure - it has to do with Scott renaming the `delayed_job` logfiles to have the environment name in them (made testing easier locally).

test/ui_test_suite.rb
bistline: We don't have a team norm related to TODOs in code that I'm aware of.  Also there was no ticket relating to this, but I just created it ([SCP-2072](https://broadworkbench.atlassian.net/browse/SCP-2072)).  These files are rarely uploaded by SCP users, so I considered this minor regression to be of low priority.  But I can at least put in a reference to the ticket for this.

app/models/gene.rb
bistline: Mongoid has different syntax for array queries.

PR #357, opened 2020-01-14
New Relic monitoring for performance and JS errors (SCP-1902)

eweitz:
This adds New Relic (NR) monitoring for application performance and JavaScript errors, as required by FISMA.

The large JavaScript blob in the `<head>` is recommended by NR.  `NEWRELIC_AGENT_ID` corresponds to the current deployment -- development, staging, or production.  It is not a secret (it's [public in DataBiosphere/terra-ui](https://github.com/DataBiosphere/terra-ui/blob/65ec4ed4ed2a3e869330fae08707e1720c857b44/public/index.html#L45)), but it's stored in Vault as a convenient place for environment variables.

I verified that NR monitors performance and JS errors for development.  After this PR is approved, I will write `NEWRELIC_AGENT_ID` to Vault SCP configurations for production and other developers.

This satisfies SCP-1902.  For details on how to access this monitoring data, [see here](https://broadworkbench.atlassian.net/browse/SCP-1902?focusedCommentId=25814&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25814).


app/views/layouts/application.html.erb
```diff
@@ -18,6 +18,13 @@
           window.location = searchUrl;
       }
   </script>
```
devonbush: I'd pull this into a separate erb file (not because it will get referenced anywhere else, just because it will declutter this file)

eweitz: Good point, done in beb0ff9.

devonbush: oh nice!  I was thinking of pulling out the whole script tag, but I actually like what you did better, since the separate file is now pure js.

eweitz:
@bistline, done in ef8431a, thanks for the catch.

PR #362, opened 2020-01-22
Autocomplete bugfixes

bistline:
Stripping whitespace from gene search tokens to prevent returning all results due to an empty string, which will result in severely degraded performance, or crashes in some instances.  Also refactoring references to unique gene list to optimize searching/autocomplete.

This PR addresses [SCP-2084](https://broadworkbench.atlassian.net/browse/SCP-2084).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -214,7 +214,6 @@ function split(val) {
 // extract last term from a string of autocomplete entries
 function extractLast(term) {
     sanitizedTerm = term.trim(); // remove trailing whitespace to prevent returning all results
-    console.log('attempting to autocompelte on "' + sanitizedTerm + '"');
     return split(sanitizedTerm).pop();
 }
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -224,7 +224,7 @@ var keydownIsFromAutocomplete = false;
  * @param selector: DOM selector for form element
  * @param entities: Array of pre-populated values to search
  **/
-function initializeAutocomplete(selector, entities, maxGenes, errorMsg) {
+function initializeAutocomplete(selector) {
 
     var jqObject = $(selector);
     jqObject.on("keydown", function(event) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/application.js
```diff
@@ -213,7 +213,9 @@ function split(val) {
 
 // extract last term from a string of autocomplete entries
 function extractLast(term) {
-    return split(term).pop();
+    sanitizedTerm = term.trim(); // remove trailing whitespace to prevent returning all results
+    console.log('attempting to autocompelte on "' + sanitizedTerm + '"');
```
eweitz: If this `console.log` is indeed needed:

```suggestion
    console.log('attempting to autocomplete on "' + sanitizedTerm + '"');
```

bistline: It's not, actually - that was for debugging purposes.  I'll remove it.

bistline: 64e4e5c

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,7 @@
       // load the max number of gene searches and corresponding error message.
       window.MAX_GENE_SEARCH = <%= Gene::MAX_GENE_SEARCH %>;
       window.MAX_GENE_SEARCH_MSG = '<%= Gene::MAX_GENE_SEARCH_MSG %>';
+      window.uniqueGenes = <%= @unique_genes.present? ? raw(@unique_genes) : [] %>;
```
devonbush: I'd like to start scoping all our scp-specific window variables that we use into a global 'SCP' object so they're easier to track and split out from other global.  So we could have a policy that all new window variables have to be inside window.SCP, and any time you mod code that uses an existing raw window variable, you move it inside window.SCP (assuming the migration isn't unduly difficult)

devonbush: since this is the first one, in application.js (or whatever you can guarantee will load first), you'll just add a `window.SCP = window.SCP || {}`

PR #363, opened 2020-01-22
Bugfixes for deployments re: yarn, subsampling cleanup

bistline:
When running deploy.sh, the script now checks out a fresh copy of the `yarn.lock` file in case it was changed after deployment and differs from what is in source.  Also, when deleting metadata files and cleaning up subsampled data, clusters now have their `subsampled` flag correctly set to `false` after the cleanup to allow future subsampling when new metadata is provided.

### DIFF
--- a/app/models/delete_queue_job.rb
+++ b/app/models/delete_queue_job.rb
@@ -71,7 +71,6 @@ class DeleteQueueJob < Struct.new(:object)
         # once a user adds another metadata file
         ClusterGroup.where(study_id: study.id).each do |cluster_group|
           delete_subsampled_data(cluster_group)
-          cluster_group.update(subsampled: false)
         end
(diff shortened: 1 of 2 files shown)


app/models/delete_queue_job.rb
```diff
@@ -71,6 +71,7 @@ def perform
         # once a user adds another metadata file
         ClusterGroup.where(study_id: study.id).each do |cluster_group|
           delete_subsampled_data(cluster_group)
+          cluster_group.update(subsampled: false)
```
devonbush: these two lines seem like they'd be better placed in a delete_subsampled_data function on the cluster_group model

bistline: Fair point.

PR #365, opened 2020-01-23
Fixing parse error email contents

bistline:
Correctly reading the contents of remote error files when determining what to print into error emails delivered to users on parse failures.

This PR helps address issues as a part of [SCP-2076](https://broadworkbench.atlassian.net/browse/SCP-2076).


bistline:
The one failing test in the build has been addressed in dacb729, so I'm merging this without the passed build due to time constraints.

PR #370, opened 2020-01-24
Removing mandatory yarn upgrade from setup script

devonbush:
After discussion with Jon, monthly sounds like the right frequency for running yarn upgrade, rather than every build.  I've created a recurring monthly meeting to remind me to run the yarn upgrade and test what results . (Jon is also on that invitation as backup).


bistline:
We have Dependabot already running on this repository, but it only does automatic PRs for vulnerabilities - not regular updates.  This would not have addressed the issues we were seeing in CI or deployments.

PR #372, opened 2020-01-27
Db react integration

devonbush:
This is an extension of the PR Jon & Eric already approved, now updated to have the react dependencies and pipeline included. I've hidden the new search UI by just removing the div where it would be attached (and leaving the old search UI in its place). But I want all the react dependencies and compilation steps to get merged to development ASAP so we can make sure they don't cause problems anywhere in our pipeline.


eweitz:
(Please add a `security-risk` GitHub label.)

PR #373, opened 2020-01-29
enabling local development with tcell turned off

devonbush:


bin/load_env_secrets.sh
```diff
@@ -69,6 +69,9 @@ if [[ -z $SERVICE_ACCOUNT_PATH ]] && [[ -z $VAULT_SECRET_PATH ]] ; then
   exit 1
 fi
 
+#  clear this evironment variable just in case this terminal was used for local development
+export NOT_DOCKERIZED=
```
eweitz: [`unset`](https://stackoverflow.com/a/6877747) seems like best practice for deleting environment variables.
```suggestion
unset NOT_DOCKERIZED
```

bin/load_env_secrets.sh
```diff
@@ -69,6 +69,9 @@ if [[ -z $SERVICE_ACCOUNT_PATH ]] && [[ -z $VAULT_SECRET_PATH ]] ; then
   exit 1
 fi
 
+#  clear this evironment variable just in case this terminal was used for local development
```
eweitz:
```suggestion
#  clear this environment variable just in case this terminal was used for local development
```

rails_local_setup.rb
```diff
@@ -46,6 +58,13 @@
 puts 'Setting secret_key_base'
 source_file_string += "export SECRET_KEY_BASE=#{`openssl rand -hex 64`}\n"
 
+if !exclude_tcell
+  puts 'Configuring TCell'
+  `bin/configure_tcell.rb`
+else
+  puts "****\nYou are developing with TCell disabled--be sure to test that any third-party scripts or connection sources you add are also added to TCell\n****\n"
```
eweitz: I can see this biting us without revised practices.  E.g. engineer adds a third-party script, doesn't test with tCell enabled, then something breaks in staging or production.  Simple proposal to avoid that:

Engineers iterate without tCell, but at least manually test _with tCell enabled_ before opening PR.

Ideally this would be caught by UI tests, but given we don't run them all for every PR, I think this slightly more tedious manual process would lower risk.  I'm open to other ideas, just thought this lurker worth noting.

PR #374, opened 2020-01-29
Adding SearchFacet to cache BigQuery data for faceted search

bistline:
Adds the SearchFacet model to allow for for caching unique values from BigQuery columns to use to populate faceted search UI components under development.  These facets will update automatically after the successful parse or deletion of a convention cell metadata file.

Currently, SearchFacet instances will have to be created by hand, but future work could be done to automate this process based off of the convention JSON schema file.

Also abstracts out the BigQueryClient into a separate class for consistency with the FireCloudClient and PapiClient models.  In addition, all test fixtures files have been removed as they are unused and do not work with MongoDB (we currently use db/seeds.rb in place of fixtures).

This PR satisfies [SCP-2044](https://broadworkbench.atlassian.net/browse/SCP-2044).


app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
...
+    end
+  end
+
+  # generate a single query to get DISTINCT values from an array-based column, preserving order
+  # this way we can do two queries, and stitch them together in a hash as the order will be the same
+  # e.g. IDs will line up with NAMEs
+  def generate_array_query(column_name, identifier)
+    "SELECT DISTINCT #{identifier} FROM(SELECT array_col AS #{identifier}, " + \
+    "FROM #{CellMetadatum::BIGQUERY_TABLE}, UNNEST(#{column_name}) AS array_col " + \
+    "WITH OFFSET AS offset ORDER BY offset)"
```
devonbush: nice!

app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
...
+      id_array.each_with_index do |filter_hash, index|
+        filters_array << filter_hash.merge(name_array[index])
+      end
+      filters_array
+    end
+  end
+
+  private
+
+  # determine if this facet references array-based data in BQ as data_type will look like "ARRAY<STRING>"
```
devonbush: maybe rename to 'set_is_array_based_from_bq'?

app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
...
+      updated = facet.update_filter_values!
+      if updated
+        Rails.logger.info "Update to #{facet.name} complete!"
+      else
+        Rails.logger.error "Update to #{facet.name} failed"
+      end
+    end
+  end
+
+  # return a formatted object to use to render search UI component
```
devonbush: Nice, we'll just want to move this method to the controller once we make the API

test/models/search_facet_test.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  # should return expected filters list
+  # mocks call to BigQuery to avoid unnecessary overhead
+  test 'should update filters list' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    # mock call to BQ, until a better library can be found
+    mock = Minitest::Mock.new
+    mock.expect :query, @filter_results, [String]
+
```
devonbush: oooh, that's slick how you made the bq_dataset method which made this test easier to write.  Nice!

test/models/search_facet_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    SearchFacet.stub :big_query_dataset, mock do
+      filters = @search_facet.get_unique_filter_values
+      assert_equal @filter_results, filters
+      mock.verify
```
devonbush: This is super-nit-picky, but I would reorder these, so that the checks mirror execution order.  First you mock.verify to confirm the method got called, then you assert_equal to confirm it returned what you expected

test/models/search_facet_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} complete!"
+  end
+
+  # should return facet configuration object with name of facet for UI and filter options of id/name
```
devonbush: cool, likewise, this test will move a a controller evenutally

test/models/search_facet_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    mock = Minitest::Mock.new
+    mock.expect :query, @filter_results, [String]
+
+    SearchFacet.stub :big_query_dataset, mock do
+      filters = @search_facet.get_unique_filter_values
+      assert_equal @filter_results, filters
+      mock.verify
+    end
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} complete!"
```
devonbush: Are these puts necessary?  I would expect that whatever test running library we would use would have a --verbose option to print stuff like this

app/controllers/application_controller.rb
```diff
@@ -24,27 +24,14 @@ class ApplicationController < ActionController::Base
   rescue_from ActionController::InvalidAuthenticityToken, with: :invalid_csrf
 
   @papi_client = PapiClient.new
+  @big_query_client = BigQueryClient.new
 
   def self.papi_client
```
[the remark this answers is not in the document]
bistline: So we had been doing that in the past, and Codacy always complains with the standard "Don't use class variables in Ruby".  I'm totally fine to change that back though as I feel like we are _correctly_ using a class variable in this case.

app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
+class SearchFacet
```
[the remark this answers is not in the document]
bistline: Can do

app/models/search_facet.rb
bistline: That makes sense.

test/models/search_facet_test.rb
bistline: Also makes sense.

test/models/search_facet_test.rb
bistline: This is something that I do all over the place... I just like seeing which tests are being run and from where in the output (either in the console or in CI).  Let's chat about this because if we take them out, there's about a dozen other test suites that need to get changed.

app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
...
+  field :convention_name, type: String
+  field :convention_version, type: String
+
+  validates_presence_of :name, :identifier, :big_query_id_column, :big_query_name_column, :convention_name, :convention_version
+  validates_uniqueness_of :big_query_id_column, scope: [:convention_name, :convention_version]
+  validate :ensure_ontology_url_format, if: proc {|attributes| attributes[:is_ontology_based]}
+  before_create :set_is_array_based, if: proc {|attributes| attributes[:is_array_based].nil?}
+
+  def self.big_query_dataset
+    ApplicationController.big_query_client.dataset(CellMetadatum::BIGQUERY_DATASET)
```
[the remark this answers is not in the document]
bistline: Makes sense, but let's table that for a future enhancement as it requires changing `FireCloudClient` and `PapiClient` as well for consistency.

app/models/search_facet.rb
bistline: 4d685ac

app/models/search_facet.rb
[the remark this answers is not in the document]
bistline: 4d685ac

app/controllers/application_controller.rb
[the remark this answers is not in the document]
bistline: 4d685ac

test/models/search_facet_test.rb
bistline: 4d685ac

app/models/search_facet.rb
```diff
@@ -0,0 +1,165 @@
...
+      if updated
+        Rails.logger.info "Update to #{facet.name} complete!"
+      else
+        Rails.logger.error "Update to #{facet.name} failed"
+      end
+    end
+  end
+
+  # return a formatted object to use to render search UI component
+  def facet_config
```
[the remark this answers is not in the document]
bistline: Yeah - if we decide to change that, let's address it in future work.

app/models/search_facet.rb
devonbush: I think you forgot to make the change

test/models/search_facet_test.rb
devonbush: good idea, it's a larger convo

bistline:
> Looks good to me, except I think you accidentally left off a method rename that you said you were going to do

You're right - fixed in c567821

PR #377, opened 2020-01-31
Adding API endpoints for keyword, faceted search

bistline:
Adding API endpoints to drive keyword and faceted search (to be called from new React-based search front end).  Provides the following endpoints:
* GET /single_cell/api/v1/search/facets - Get all available facets & filter values
* GET /single_cell/api/v1/search/facet_filters - Search for matching filter values with a given facet
* GET /single_cell/api/v1/search - Perform a keyword or faceted search against MongoDB & BigQuery

API Swagger documentation has been updated with new endpoints to allow for easier development.  This PR also includes an integration test that performs searches against BigQuery (does not mock requests).  Existing test order has been modified to allow usage of upstream inserts in queries to optimize CI runtime, rather than seeding data in BQ just for this test.

This PR satisfies [SCP-2086](https://broadworkbench.atlassian.net/browse/SCP-2086).


app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+          key :summary, 'Faceted & keyword search for studies & cells'
+          key :description, 'Search studies or cells using facets and keywords'
+          key :operationId, 'search_studies'
+          parameter do
+            key :name, :type
+            key :in, :query
+            key :description, 'Type of query to perform (study- or cell-based)'
+            key :required, true
+            key :type, :string
+            key :enum, ['study', 'cell']
```
eweitz: I double-taked the lack of `'gene'`, but I suppose omitting it makes sense.

That said, I don't think we track yet the task of integrating the existing global gene search endpoint into the upcoming React front-end.

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+            key :description, 'Accept or Content-Type headers missing or misconfigured'
+          end
+          response 500 do
+            key :description, 'Server error or malformed query'
+          end
+        end
+      end
+
+      def index
+        order = [:view_order.asc, :name.asc]
```
devonbush: I'd move this logic to a "self.get_viewable_studies(user)" method in study.rb (which returns the public studies if user is nil)

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+          end
+          response 406 do
+            key :description, 'Accept or Content-Type headers missing or misconfigured'
+          end
+          response 500 do
+            key :description, 'Server error or malformed query'
+          end
+        end
+      end
+
```
devonbush: this method should only be responsible for sanitizing the params and  identifiying the current user (if there is one), and returning the result. All actual search logic should live somewhere else.  Either in a static method on this controller or the study model

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+                  key :'$ref', :SiteStudyWithFiles
+                end
+              end
+            end
+          end
+          response 406 do
+            key :description, 'Accept or Content-Type headers missing or misconfigured'
+          end
+          response 500 do
+            key :description, 'Server error or malformed query'
```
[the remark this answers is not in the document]
bistline: That's a very good point about the `400 Bad Request`.  I'll update the text to remove that and think about how we would handle bad queries with more appropriate codes.

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+          sanitized.join(',')
+        else
+          view_context.sanitize(terms)
+        end
+      end
+
+      # generate query string for BQ
+      # array-based columns need to set up data in WITH clauses to allow for a single UNNEST(column_name) call,
+      # otherwise UNNEST() is called multiple times for each user-supplied filter value and could impact performance
+      def generate_bq_query_string(facets)
```
devonbush: make this method static

test/api/search_controller_test.rb
```diff
@@ -0,0 +1,78 @@
+require 'api_test_helper'
+
```
devonbush: once you refactor more of the logic into static methods, this test class will  hopefully get simpler and faster :)

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,293 @@
...
+            key :in, :query
+            key :description, 'Type of query to perform (study- or cell-based)'
+            key :required, true
+            key :type, :string
+            key :enum, ['study', 'cell']
+          end
+          parameter do
+            key :name, :facets
+            key :in, :query
+            key :description, 'User-supplied list facets and filters, formatted as: "facet_id:filter_value+facet_id_2:filter_value_2,filter_value_3"'
```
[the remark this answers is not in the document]
bistline: Let me see if I can figure out how to get large example blocks into the top-level description.  The `swagger-blocks` gem doesn't have any advice on how to do this but I might be able to glean some info directly from the Swagger spec.

app/controllers/api/v1/search_controller.rb
bistline: Yes - global gene search functionality is totally separate from keyword/faceted search and should continue to be treated as such.  It requires a totally different query/response flow with different methods and views.

devonbush:
Overall, looks good -- I'm excited at how far we (you) have gotten after just the first week of working on this

app/controllers/api/v1/search_controller.rb
bistline: This method actually already exists (Study.viewable) - I just need to use it here.

app/controllers/api/v1/search_controller.rb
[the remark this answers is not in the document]
bistline: So I was not able to figure this out, but I want to keep this idea alive, so let's open a ticket for this.  Apparently the `swagger-blocks` gem is a bit of a defunct project, so we may need to switch to a different implementation for our Swagger documentation, which is a larger undertaking.

app/controllers/api/v1/search_controller.rb
bistline: cdf206c

app/controllers/api/v1/search_controller.rb
bistline: cdf206c

app/controllers/api/v1/search_controller.rb
bistline: cdf206c

app/controllers/api/v1/search_controller.rb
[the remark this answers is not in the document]
bistline: cdf206c

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,289 @@
...
+          # determine if search values contain possible study accessions
+          possible_accessions = StudyAccession.sanitize_accessions(search_terms.split)
+          @studies = @viewable.any_of({:$text => {:$search => search_terms}}, {:accession.in => possible_accessions})
+        else
+          @studies = @viewable
+        end
+
+        # only call BigQuery if list of possible studies is larger than 0 and we have matching facets to use
+        if @studies.count > 0 && @facets.any?
+          @big_query_search = Api::V1::SearchController.generate_bq_query_string(@facets)
```
devonbush: I'd use 'self.class.generate_bq_query_string(@facets)'  to call the method, it's shorter and robust to file renames

app/controllers/api/v1/search_controller.rb
devonbush: nice!

PR #378, opened 2020-01-31
Prototype UI for facets and filters (SCP-2047)

eweitz:
This adds a rough prototype for facets and filters in the new advanced search UI.  It's a start.  Note this PR targets merging into a new long-lived topic branch (`launch-advanced-search`), making an already-big PR very large and partly redundant with recent PRs merged into `development`.

That said, the code seems to be at a reasonable point for review, to split up a very large task.  

Next steps include setting up JavaScript test infrastructure (SCP-2107), and incorporating API endpoints, more interactions, refining polish, and improving abstraction (SCP-2109).

This satisfies SCP-2047.


app/javascript/components/FacetsAccordion.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Accordion from 'react-bootstrap/Accordion';
+import Card from 'react-bootstrap/Card';
+
+import Filters from './Filters';
+
+export default function FacetsAccordion(props) {
+  return (
+    // Accordions provide a way to restrict Card components to only open one at a time.
+    // https://react-bootstrap.github.io/components/accordion/
```
knapii-developments: For our code based are we supposed to mention where we get each component from if its not apart of the standard library?

app/javascript/components/Facet.js
```diff
@@ -0,0 +1,37 @@
...
+import FiltersBox from './FiltersBox';
+
+/**
+ * Converts string value to lowercase, hyphen-delimited version
+ * e.g. "Cell type" -> "cell-type"
+ */
+function slug(value) {
+  return value.toLowerCase().replace(/ /g, '-');
+}
+
```
devonbush: I might rename this FacetControl, but maybe that's redundant

app/javascript/components/FacetsAccordion.js
devonbush: I think it's useful in places like this one where it's non-obvious

app/javascript/components/FiltersBox.js
```diff
@@ -0,0 +1,137 @@
+import React, { useState, useEffect } from 'react';
+import Button from 'react-bootstrap/Button';
+import InputGroup from 'react-bootstrap/InputGroup';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons';
+
+import FiltersSearchBar from './FiltersSearchBar';
+
```
devonbush: I would use lodash here instead of rolling our own array compare (it looks like it's already in our yarn.lock file) .  Lodash also supports partials imports so if it's not already in our package, we only have to bring in the functions we need


>  __.isEqual(array1, array2)

app/javascript/styles/_global.scss
```diff
@@ -480,6 +480,100 @@ h4, .h4, h5, .h5, h6, .h6 {
   padding-left: 15px;
 }
```
devonbush: let's pull these out into a separate _search.scss file and import it into application.scss

devonbush: also, you can make this SCSS more concise/robust by wrapping everything in  a single
#search-panel {} rather than repeating it for each rule

app/javascript/components/FacetsAccordion.js
eweitz: That summarizes my thinking.  My rationale here was to link for maintainers' convenience.

app/javascript/styles/_global.scss
eweitz: [Nesting](https://sass-lang.com/guide#topic-3), wow!  Done in 9576866 and 0873e98.

app/javascript/components/FiltersBox.js
eweitz: Good point, done in 69208ca.

app/javascript/components/Facet.js
eweitz: Done in cd210e6.

PR #380, opened 2020-01-31
Db feature flag search

devonbush:
this should allow us to merge launch-advanced-search into development.  Note that the only new things in this PR that aren't just pulled from the already-merged feature flag branch are the 'feature_flag_for_user' helper method in user.rb, and the change to index.html.erb

### DIFF
--- a/app/controllers/admin_configurations_controller.rb
+++ b/app/controllers/admin_configurations_controller.rb
@@ -306,12 +306,28 @@ class AdminConfigurationsController < ApplicationController
     @user = User.find(params[:id])
     respond_to do |format|
       if @user.update(user_params)
+        self.class.process_feature_flag_form_data(@user, params)
         format.html { redirect_to admin_configurations_path, notice: "User: '#{@user.email}' was successfully updated." }
       else
         format.html { render :edit_user }
         format.json { render json: @user.errors, status: :unprocessable_entity }
       end
     end
+  end
+
+  # Update the user with the form data on their feature flags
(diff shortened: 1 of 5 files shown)


app/views/site/index.html.erb
```diff
@@ -18,7 +18,11 @@
             Ensure any updates to this ID are reflected in the ReactDOM.render call
             in app/javascript/packs/application.js
           -->
-          <div id="scp-search-studies-root-element"></div>
+          <% if User.feature_flag_for_user(current_user, "faceted_search") %>
```
eweitz: Simple enough!

app/controllers/admin_configurations_controller.rb
```diff
@@ -314,6 +315,21 @@ def update_user
...
   end
 
+  # Update the user with the form data on their feature flags
+  # because of the true/false/default structure, we have to have custom logic to handle the values
+  def self.process_feature_flag_form_data(user, params)
+    updated_flag_data = {}
+    User::DEFAULT_FEATURE_FLAGS.each do |key, _value|
+      # for each possible feature flag, check if the vaule was set on the form via a field name feature_flag_<<flag name>>
+      flag_form_val = params[('feature_flag_' + key).to_sym]
+      # if it was set, and not set to '-' (default), include it in the updated values'
```
eweitz:
```suggestion
      # if it was set, and not set to '-' (default), include it in the updated values
```

PR #383, opened 2020-02-04
Discard extraneous long-lived feature branch

eweitz:
The long-lived feature branch `launch-advanced-search` is extraneous now that we have features flags (#379, #380).

After this merge, work on that feature can simply use the `development` branch as usual.


app/javascript/components/FacetControl.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default function FacetControl(props) {
+  
+  const [showFilters, setShowFilters] = useState(false);
+  
+  const facetName = props.facet.name;
+  const facetID = `facet-${slug(facetName)}`;
+
+  function handleClick() {
+    setShowFilters(!showFilters);
```
knapii-developments: What is this?

eweitz: It's code from #378.

This PR basically merges code from that PR #378 into `development` instead of `launch-advanced-search`.  Then we'll delete `launch-advanced-search` and use `development` -- for more streamlined merging.

PR #385, opened 2020-02-05
fixing webpack for staging deploys

devonbush:
these are suboptimal changes, but after 3 hours of trying to get webpacker to do what it claims it should (not do an extra compilation step and inline the css with the js), I'm happy to just go with the simple-but-works of keeping the css include separate

### DIFF
--- a/rails_startup.bash
+++ b/rails_startup.bash
@@ -10,10 +10,10 @@ echo "*** COMPLETED ***"
 if [[ $PASSENGER_APP_ENV = "production" ]] || [[ $PASSENGER_APP_ENV = "staging" ]]
 then
     echo "*** PRECOMPILING ASSETS ***"
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE yarn:install
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE assets:clean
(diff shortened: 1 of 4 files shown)


rails_startup.bash
```diff
@@ -10,10 +10,10 @@ echo "*** COMPLETED ***"
 if [[ $PASSENGER_APP_ENV = "production" ]] || [[ $PASSENGER_APP_ENV = "staging" ]]
 then
     echo "*** PRECOMPILING ASSETS ***"
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE yarn:install
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE assets:clean
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE assets:precompile
-    sudo -E -u app -H bundle exec rake RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE webpacker:compile
+    sudo -E -u app -H bundle exec rake NODE_ENV=$PASSENGER_APP_ENV RAILS_ENV=$PASSENGER_APP_ENV SECRET_KEY_BASE=$SECRET_KEY_BASE yarn:install
```
bistline: Totally not a blocking concern, but we could probably clean this up by setting RAILS_ENV and NODE_ENV to PASSENGER_APP_ENV beforehand.

PR #386, opened 2020-02-06
Ea keyword search

knapii-developments:
This PR adds components for keyword search and for results.  Please take note that the results that are shown are from static text and styling is not included. 

What current UI looks like:
Screen Shot 2020-02-07 at 11 00 16 AM

This satisfies [SCP-2012](https://broadworkbench.atlassian.net/browse/SCP-2012)


app/javascript/components/KeyWordSearch.js
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
+import Button from 'react-bootstrap/lib/Button';
+import InputGroup from 'react-bootstrap/lib/InputGroup';
+import Form from 'react-bootstrap/lib/Form';
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+
```
devonbush: I think the more traditional capitalization is just 'Keyword' and not KeyWord'

app/javascript/components/KeyWordSearch.js
```diff
@@ -0,0 +1,63 @@
...
+class KeyWordSearch extends React.Component{
+  constructor(props){
```
devonbush: Eric might have some suggestions around using 'useState' rather than a constructor for this case

app/javascript/components/KeyWordSearch.js
```diff
@@ -0,0 +1,63 @@
...
+    this.state = {
+      searchTerms: undefined,
+    };
+  };
+
+  handleSubmit(e){
+    // Prevent full page reload
+    e.preventDefault();
+
+    const searchTerm = e.target.elements.searchText.value.trim();
```
devonbush: does this logic correctly handle the case where a user sumbits a keyword search, then clears the field and hits enter again to see unfiltered results?

app/javascript/components/KeyWordSearch.js
```diff
@@ -0,0 +1,63 @@
...
+    if(searchTerm){
+      // Need to check if search terms are empty and if prevstate is different
+      // From current state
+      this.setState(()=>{
+        return {
+          searchTerms:searchTerm
+        };
+
+      });
```
devonbush: my react must be too rusty, I don't understand what this second setState call is supposed to do -- it looks like you're undoing the previous setState call?

app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
...
+class ResultsPanel extends React.Component{
+  constructor(props){
+    super(props);
+    this.state = {
+      results: undefined,
+      facets : []
+    };
+    this.handleStudyLabel = this.handleStudyLabel.bind(this);
+    };
+
```
devonbush: maybe rename to handleStudyLabelClick?

app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
...
+StudyResults.defaultProps = {
+  // This may move up to the homepage as a property that's passed into the 
+  // ResultsPanel component as the property results
+  results: results
+
+}
+
+const Study =(props)=>{
+   //There needs to be a handler for when a study is clicked it takes
+  // a user to the corresponding study page
```
devonbush: does it need to be a handler?  Can it just be a plain `<a href="/single_cell/study/{studyId}/{studyName}">`?

app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
...
+    <Tab.Content>
+       { props.results.length &&
+          props.results.map((result)=>(
+            <Study 
+              key={result.acession} 
+              study={result}
+              handleStudyLabel = {props.handleStudyLabel}
+              />
+          )
+
```
devonbush: nice.  Do we have a message to display for when results.length is 0?  Something like 'no studies matched your criteria'?

app/javascript/components/ResultsPanel.js
devonbush: Also, I think this method belongs in the Study component, rather than here.  (assuming the study click function does a expansion of the study block similar  to the functionality  in the current site)

app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
...
+  },
+];
+
+class ResultsPanel extends React.Component{
+  constructor(props){
+    super(props);
+    this.state = {
+      results: undefined,
+      facets : []
+    };
```
devonbush: see previous comment about replacing bind with alternate delcaration syntax

app/javascript/components/KeyWordSearch.js
```diff
@@ -0,0 +1,63 @@
...
+import Button from 'react-bootstrap/lib/Button';
+import InputGroup from 'react-bootstrap/lib/InputGroup';
+import Form from 'react-bootstrap/lib/Form';
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+
+class KeyWordSearch extends React.Component{
+  constructor(props){
+    super(props);
```
devonbush: you don't have to call bind if you instead declare your component functions with the syntax

handleStudyLabel = (studyUrl) => {
...
}

app/javascript/components/SearchPanel.js
```diff
@@ -0,0 +1,57 @@
...
+const facets = facetsResponseMock;
+
+const defaultFacetIDs = ['disease', 'organ', 'species', 'cell_type'];
+const moreFacetIDs = ['sex', 'race', 'library_preparation_protocol', 'organism_age'];
+
+const defaultFacets = facets.filter(facet => defaultFacetIDs.includes(facet.id));
+const moreFacets = facets.filter(facet => moreFacetIDs.includes(facet.id));
+
+window.searchFiltersResponse = searchFiltersResponseMock;
+
```
devonbush: I'd rather not have css declared in javascript components except as a last resort

app/javascript/packs/application.js
```diff
@@ -7,10 +7,9 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
-import 'styles/application.scss'
-
 import React from 'react';
 import ReactDOM from 'react-dom';
```
devonbush: was there a reason you reordered this import?   we should figure out a standard import ordering scheme

app/views/site/index.html.erb
```diff
@@ -2,34 +2,36 @@
 <div class="container-fluid home-page-fix" id="wrap">
   <div class="row section-pad" id="main-body">
     <div class="col-md-12">
```
devonbush: these should stick to 2-space indentation

app/javascript/components/KeyWordSearch.js
knapii-developments: That logic would be controlled by the parent class. In the event where there's keyword search is empty, it should show the original set of studies AKA the default state.

app/javascript/components/KeyWordSearch.js
knapii-developments: This has been taken care of in the new PR

app/javascript/components/ResultsPanel.js
knapii-developments: This will actually just be an <a>. This has been reflected in the new PR.

app/javascript/components/ResultsPanel.js
knapii-developments: This can be taken care of in Keyword search and the Parent wrapper component. There are a couple of ways to approach this:

1.  In the event where the search API doesn't return results, then ResultsPanel shouldn't actually be rendered. Instead maybe a `<span> No Results</span>` or a `<p> No results</`>
2. When the `<ResultsPanel>` has not results to show it can render `<span> No Results</span>` or a `<p> No results</`>

app/javascript/components/ResultsPanel.js
knapii-developments: This is correct and has been reflected in the new PR

app/javascript/components/KeyWordSearch.js
knapii-developments: SetState is asynchronous and React will batch multiple setState() calls into one for performance reasons. So to ensure state has actually change and avoid async issues, you can call it twice. 

Read more about it [here](https://reactjs.org/docs/state-and-lifecycle.html).

PR #387, opened 2020-02-06
Prototype Bulk Download UI for files matching study search (SCP-2088)

eweitz:
This implements a rough prototype of UI components for Bulk Download in the new search UI.

It also refactors code to use the previous major of `react-bootstrap`, so the new UI uses Bootstrap 3 (not 4) like the rest of Single Cell Portal.

I have tested these changes manually, and also manually verified that the related Bulk Download modal in the study detail view works.  Follow-on work to integrate API endpoints, incorporate demo feedback, automate tests, etc. is tracked in SCP-2121.

Here's what the prototype Bulk Download currently looks like:
prototype_bulk_download_ui_2020-02-06

This satisfies SCP-2088.


app/javascript/components/DownloadButton.js
```diff
@@ -0,0 +1,133 @@
...
+  const url = (
+    window.origin +
+    '/api/v1/bulk_download?auth_code=' +
+    authCode +
+    searchQuery
+  );
+  const flag = (window.location.host === 'localhost') ? 'k' : ''; // "-k" === "--insecure"
+  
+  // This is what the user will run in their terminal to download the data.
+  const downloadCommand = (
```
bistline: I had a conversation with Vicky yesterday about this - we may want to append `; rm cfg.txt` to the download command that the user runs.  Otherwise, they can _technically_ re-run the `curl -K cfg.txt` multiple times over the next day and download the data again as the URLs are valid for 24 hours.

Not a blocking concern, but possibly something we would like to think about.

app/assets/javascripts/scp-react-rails.js
```diff
@@ -0,0 +1,34 @@
+/**
```
devonbush: I would call this file 'bulkDownloadClipboard.js' and add a line at the beginning of the file like 
`window.SCP = window.SCP || {}`
that way we don't have to declare all our SCP globals in the same file, but also don't have to worry about loading order

app/javascript/components/DownloadButton.js
```diff
@@ -0,0 +1,133 @@
...
+
+  // Gets a curl configuration ("cfg.txt") containing signed
+  // URLs and output names for all files in the download object.
+  const url = (
+    window.origin +
+    '/api/v1/bulk_download?auth_code=' +
+    authCode +
+    searchQuery
+  );
+  const flag = (window.location.host === 'localhost') ? 'k' : ''; // "-k" === "--insecure"
```
devonbush: rename variable to curlSecureFlag?  Also think about checking the node environment (either at compile or runtime) instead of the hostname

app/javascript/components/DownloadButton.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  const handleClose = () => setShow(false);
+  const handleShow = () => setShow(true);
+
+  window.SCP.initBulkDownloadClipboard();
+
+  return (
+      <>
+      <span
```
devonbush: I generally prefer to put tag openers all on one line if there are < 3 attributes.  But if you and Eno prefer attributes always on separate lines, I'm down with it

app/views/site/_study_download_data.html.erb
```diff
@@ -64,21 +64,7 @@
...
-            .tooltip('fixTitle')
-            .tooltip('show');
-        });
-
-        $('body').on('click', '.btn-refresh', function(event) {
-          var commandContainer = $(this).parentsUntil('.command-container').parent();
-          var downloadObject = commandContainer.attr('id').split('-').slice(-1)[0];
-          writeDownloadCommand(downloadObject);
-        });
+        window.SCP.initBulkDownloadClipboard();
```
devonbush: nice!

app/javascript/components/DownloadButton.js
```diff
@@ -0,0 +1,133 @@
...
+          </p>
+          <p className='lead command-container' id='command-container-all'>
+            <DownloadCommandContainer />
+          </p>
+        </Modal.Body>
+
+      </Modal>
+      </>
+    );
+}
```
devonbush: You did a really good job of scoping and structuring this file so it was easy to read -- thanks!

app/assets/javascripts/scp-react-rails.js
eweitz: Per our chat abouth this, in 619e805 I did a `git mv` of this file from `scp-react-rails.js` to `react_rails/bulkDownloadClipboard.js` and updated to use that technique.

app/javascript/components/DownloadButton.js
eweitz: Done in 619e805.

app/javascript/components/DownloadButton.js
eweitz: Worth a try!  Done in 619e805.

app/javascript/components/DownloadButton.js
eweitz: Good point, done in 619e805.  I manually verified it with single-study Bulk Download.

PR #388, opened 2020-02-06
API endpoint for bulk download of search results

bistline:
Enabling bulk download of files from multiple studies as returned by keyword and faceted search from the search API.  Results are returned via the current `curl` configuration file feature, and downloads files in the following directory structure: <study_accession>/<file_type>/<filename>.  Also includes API endpoint for generating one-time authentication tokens (OTACs) to use in download requests.

This PR satisfies [SCP-2046](https://broadworkbench.atlassian.net/browse/SCP-2046).

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -5,6 +5,7 @@ module Api
       include Swagger::Blocks
 
       before_action :set_current_api_user!
+      before_action :authenticate_api_user!, only: [:create_auth_code, :bulk_download]
       before_action :set_search_facet, only: :facet_filters
       before_action :set_search_facets_and_filters, only: :index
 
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -207,6 +208,164 @@ def facet_filters
...
+          end
+          response 200 do
+            key :description, 'Curl configuration file with signed URLs for requested data'
+            key :type, :string
+          end
+          response 400 do
+            key :description, 'Invalid study accessions or requested file types'
+          end
+          response 401 do
+            key :description, 'User not signed in'
```
eweitz: Thanks for fixing my `403`/`401` bug here!

app/controllers/api/v1/search_controller.rb
```diff
@@ -207,6 +208,164 @@ def facet_filters
...
+            key :type, :string
+            key :in, :query
+            key :description, 'Comma-delimited list of Study accessions'
+            key :required, true
+          end
+          parameter do
+            key :name, :file_types
+            key :type, :string
+            key :in, :query
+            key :description, "Comma-delimited list of file types (including 'all' for all files)"
```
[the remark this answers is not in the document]
bistline: Yeah - Swagger can do enums.  I'll add that in.  And I like that approach for the parameter- if there are no specified file types, all will be returned.

app/controllers/api/v1/search_controller.rb
```diff
@@ -207,6 +208,164 @@ def facet_filters
...
+          key :description, 'Create and return a One-Time Authorization Code to identify a user for bulk downloads'
+          key :operationId, 'search_auth_code_path'
+          response 200 do
+            key :description, 'One-time auth code and time interval, in seconds'
+            schema do
+              property :totat do
+                key :type, :integer
+                key :description, 'One-time auth code'
+              end
+              property :ti do
```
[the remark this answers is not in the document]
bistline: So the reason I hadn't changed that was because the `User` model returns the fields `totat` and `ti`.  But having said that, I realize now I can manually render some JSON with the fields we agreed upon, so I'll change that.

app/controllers/api/v1/search_controller.rb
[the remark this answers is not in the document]
bistline: So I'm hesitant to do the snake_case representation of the file types in the parameter, because it will require converting back and forth between this format and the format that is stored in the `file_type` attribute for each `StudyFile` in order to do the query.  If the only reason for this is to make the query parameters easier to pass, I would argue it's just as easy to pass the normal form than snake case, and it makes the Mongo query easier.

app/controllers/api/v1/search_controller.rb
```diff
@@ -207,6 +208,164 @@ def facet_filters
...
+        # validate request parameters
+        if totat.blank? || valid_totat == false
+          render json: {error: 'Invalid authorization token'}, status: 403 and return
+        elsif sanitized_accessions.blank? || sanitized_file_types.blank?
+          render json: {error: 'Invalid request parameters; study accessions or file types not found'}, status: 400 and return
+        end
+
+        # load the user from the auth token
+        requested_user = valid_totat
+
```
devonbush: I think everything below here belongs in a static method or, better yet, a new BulkDownloadService class in app/lib

app/controllers/api/v1/search_controller.rb
```diff
@@ -317,6 +476,35 @@ def self.match_studies_by_facet(query_results, search_facets)
         end
         matches
       end
+
+      # Helper method for generating a curl command to download a file from a bucket.  Returns file's curl config, size.
```
devonbush: move this method to the BulkDownloadService :)

app/controllers/application_controller.rb
```diff
@@ -156,4 +150,15 @@ def set_csrf_headers
       cookies['XSRF-TOKEN'] = form_authenticity_token if cookies['XSRF-TOKEN'].blank?
     end
   end
+
+  # protected method to allow access from other controllers
```
devonbush: nice, I could also see moving this to BulkDownloadService

app/models/study_file.rb
```diff
@@ -33,6 +33,8 @@ class StudyFile
   ASSEMBLY_REQUIRED_TYPES = ['BAM', 'Ideogram Annotations']
   GZIP_MAGIC_NUMBER = "\x1f\x8b".force_encoding(Encoding::ASCII_8BIT)
   REQUIRED_ATTRIBUTES = %w(file_type name)
+  BULK_DOWNLOAD_TYPES = ['Expression', 'Metadata', 'Cluster', 'Coordinate Labels','Fastq', 'BAM', 'Documentation',
```
devonbush: move to BulkDownloadService

app/models/study_file.rb
```diff
@@ -653,10 +655,36 @@ def local_location
     end
   end
 
+  # generate a download path to use with bulk_download
```
devonbush: move to BulkDownloadService, and have it take the study as an argument

test/api/search_controller_test.rb
```diff
@@ -80,4 +80,45 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
+
```
devonbush: sweet, move to new file bulk_download_service_test.rb

app/controllers/api/v1/search_controller.rb
bistline: f5a85e5

app/controllers/api/v1/search_controller.rb
bistline: f5a85e5

app/controllers/application_controller.rb
bistline: f5a85e5

app/models/study_file.rb
bistline: f5a85e5

app/models/study_file.rb
bistline: I need this here to use in other places (like views and responses)

test/api/search_controller_test.rb
bistline: f5a85e5

PR #389, opened 2020-02-07
Ea keyword search new

knapii-developments:
This PR adds components for keyword search and for results. Please take note that the results that are shown are from static text and styling is not included.

What current UI looks like:
Screen Shot 2020-02-07 at 11 00 16 AM
 This satisfies [SCP-2012](https://broadworkbench.atlassian.net/browse/SCP-2012).


app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
...
+StudyResults.defaultProps = {
+  // This may move up to the homepage as a property that's passed into the 
+  // ResultsPanel component as the property results
+  results: results
+
+}
+
+const Study =(props)=>{
+   //There needs to be a handler for when a study is clicked it takes
+  // a user to the corresponding study page
```
eweitz: I think this comment can be removed, as we won't need a handler until the study page uses React (which isn't likely in the near term).  Until then, the traditional anchor tag (`<a href...`) is the way to go.

app/javascript/packs/application.js
```diff
@@ -25,15 +25,21 @@ import Ideogram from 'ideogram';
 // Per https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/advanced-setup.html#scenario-1-integrating-existing-builds
 import ClassicEditor from '@ckeditor/ckeditor5-build-classic';
 
-// Note 'components/ScpSearchStudies' is '/app/javascript/components/ScpSearchStudies.js'
-import ScpSearchStudies from 'components/ScpSearchStudies';
+// Note 'components/SearchPanel' is '/app/javascript/components/SearchPanel.js'
+import ScpSearchStudies from 'components/SearchPanel';
```
eweitz: I like the new name!

app/javascript/components/SearchPanel.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import KeyWordSearch from './KeywordSearch';
```
eweitz: For consistency:

```suggestion
import KeywordSearch from './KeywordSearch';
```

Or change the module name to `KeyWordSearch`.  If you go with the class name `KeywordSearch`, please ensure that's updated everywhere.

I lean toward `KeywordSearch` as we naturally write it as "keyword" instead of "key word".  But I defer to you, just as long as the capitalization is internally consistent throughout the code.

app/javascript/components/KeywordSearch.js
```diff
@@ -0,0 +1,62 @@
+import React, { Component } from 'react';
+import Button from 'react-bootstrap/lib/Button';
+import InputGroup from 'react-bootstrap/lib/InputGroup';
+import Form from 'react-bootstrap/lib/Form';
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+
+class KeyWordSearch extends React.Component{
```
eweitz: We should probably migrate to function components and manage state with [hooks](https://reactjs.org/docs/hooks-overview.html).  They're more modern and considerably more succinct.  Check out [FiltersBox.js](https://github.com/broadinstitute/single_cell_portal_core/blob/2857ce9af09d961e25ac30048446d0b42a6dee79/app/javascript/components/FiltersBox.js#L13) for an example.

There's an open question on when we migrate, so I don't consider this style blocking.

app/javascript/components/SearchPanel.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import FacetControl from './FacetControl';
 import MoreFacetsButton from './MoreFacetsButton';
 import DownloadButton from './DownloadButton';
+import KeywordSearch from './KeywordSearch';
```
eweitz: There seems to be a bug here: 

`KeyWordSearch` is exported from `KeywordSearch.js`, but this imports `KeywordSearch`.  The class `KeywordSearch` is undefined.

I'd update the class name to `KeywordSearch`, as we naturally write this thing as "keyword" and not "key word".

app/javascript/components/KeywordSearch.js
knapii-developments: Your examples have been a great help! I will be migrating to function components for [SCP-2321](https://broadworkbench.atlassian.net/browse/SCP-2123) and [SCP-2124](https://broadworkbench.atlassian.net/browse/SCP-2124).

app/javascript/components/SearchPanel.js
knapii-developments: Addressed in 8e23d95

app/javascript/components/ResultsPanel.js
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
+import Tab from 'react-bootstrap/lib/Tab';
+import Tabs from 'react-bootstrap/lib/Tabs';
+
+const results = [
```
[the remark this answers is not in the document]
knapii-developments: Thanks for for the FYI. For [SCP-2321](https://broadworkbench.atlassian.net/browse/SCP-2123) I will be rendering the results via an API call so I don't believe I'll need mock data. But in the case that I do, I will follow suit.

app/javascript/components/ResultsPanel.js
knapii-developments: I agree!

PR #390, opened 2020-02-07
Adding branding group filter to search API

bistline:
Preserving behavior that if a user is inside a branding group, that filter will be applied to all search requests.  Only studies inside that branding group matching the search criteria will be returned.

This also includes fixes for pagination of search results, and returns all file types for search results to allow the UI to filter down what files the user would like to download.

This PR satisfies [SCP-2129](https://broadworkbench.atlassian.net/browse/SCP-2129).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -31,6 +31,7 @@
 //= require scp-igv
 //= require scp-ideogram
 //= require scp-dot-plot
+//= require react_rails/bulkDownloadClipboard
 
 var fileUploading = false;
 var PAGE_RENDERED = false;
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/application.js
```diff
@@ -31,7 +31,6 @@
 //= require scp-igv
 //= require scp-ideogram
 //= require scp-dot-plot
-//= require react_rails/bulkDownloadClipboard
```
eweitz: Let's restore this `require` after merging #389.  I mistakenly omitted a file in #387, then [added it](https://github.com/broadinstitute/single_cell_portal_core/pull/389/files#diff-644c19332e6dfb234e3eaf974eff11cfR1) in #389, which fixes the page load error that likely prompted you to comment this out.

It's needed for Bulk Download to work in the study detail page.

bistline: Yeah, I had to take it out because it was throwing an error preventing the asset pipeline from rendering and thus blocking development.  I figured that it would get restored in another PR.

PR #391, opened 2020-02-10
Updating delete button text/messages

bistline:
Updating the button text & alert/modal messages for study delete buttons in the 'My Studies' page to be more understandable and consistent.

This PR satisfies [SCP-2103](https://broadworkbench.atlassian.net/browse/SCP-2103).

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -419,7 +419,7 @@ class StudiesController < ApplicationController
 
       # revoke all study_shares
       @study.study_shares.delete_all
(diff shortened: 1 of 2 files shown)


bistline:
> I'm not sure people will know what database the 'database records' refers to. maybe 'single-cell portal records' or just 'portal records'?
> 
> I would also add something about how the terra workspace is what holds the underlying study files, in case they've not looked into Terra particularly

So the information about what a Terra workspace is and how it's used is information that is in the "New Study" form page - there's a large description about the workspace and why it is important.   So I don't think trying to recapitulate that in short form in the message will help much.  Having said that, I think the point about the database is valid so I'll change that.

PR #393, opened 2020-02-12
Adding synthetic study populator to programatically add studies to a portal

devonbush:
Satisfies SCP-2100.  We'll update the populator to not suffix the study names next week once we create unique namespaces for everyone


app/lib/file_parse_service.rb
```diff
@@ -0,0 +1,79 @@
+# collection of methods involved in parsing files
```
eweitz: Nice modularization!

app/lib/synthetic_study_populator.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    puts("Populating synthetic study from #{synthetic_study_folder}")
+    study = create_study(synthetic_study_folder, study_config, user)
+    add_files(study, study_config, synthetic_study_folder, user)
+  end
+
+  private
+
+  def self.create_study(synthetic_study_folder, study_config, user)
+    user_suffix = '-' + (`git config user.email`.strip[0, 2] || 'xx')
```
eweitz: I can see this approach yielding name collisions as new folks are onboarded.  Splitting `user.email` on `@` would avoid that.

I assume this approach is temporary given our chat about separating Terra namespaces (and I'd suggest incorporating this refinement there), so I don't see this foreseeable issue as blocking.

db/seed/synthetic_studies/bronchial_lymphoid/study_info.json
```diff
@@ -0,0 +1,10 @@
+{
+  "study": {
+    "name": "Mouse Bronchial Lymphoid and GERD",
+    "description": "<p>Mouse tuberculosis and GERD (AA).  Jean Chang, Eric Weitz, Jon Bistline, Eno-Abasi Augustine-Akpan, Devon Bush, Vicky Horst.  Journal of Nonexistent Mock Studies.  Volume 1, p1, February 2020.  Adding mock study data to development and staging environments has been shown to improve testability of new code.  We extend those findings to the Single Cell Portal by adding this mock data.</p>",
+    "data_dir": "test"
+  },
+  "files": [
```
eweitz: Future enhancements:
* Other file types
* Multiple files

db/migrate/20200213205538_add_initial_facets.rb
```diff
@@ -0,0 +1,61 @@
+class AddInitialFacets < Mongoid::Migration
```
bistline: A couple things about these initial facets: technically, we are referring to the convention as "alexandria_convention" - that's the name of the table in BigQuery, so we should be consistent in that.  Also, the current version is 1.1.3, not 1.0.  We should be using the JSON schema to manage all of these values:  https://github.com/broadinstitute/scp-ingest-pipeline/blob/master/tests/data/AMC_v1.1.3.json.

db/migrate/20200213205538_add_initial_facets.rb
```diff
@@ -0,0 +1,61 @@
...
+  def self.up
+    SearchFacet.find_or_create_by!(name: "species") do |facet|
+      facet.identifier = "species"
+      facet.is_ontology_based = true
+      facet.is_array_based = false
+      facet.big_query_id_column = 'species'
+      facet.big_query_name_column = 'species__ontology_label'
+      facet.convention_name = 'alexandria'
+      facet.convention_version = '1.0'
+      facet.ontology_urls = [{name: 'todo', url: 'https://to.do/ontology'}]
```
bistline: We can pull the ontology URLs from the [JSON](https://github.com/broadinstitute/scp-ingest-pipeline/blob/master/tests/data/AMC_v1.1.3.json) schema for all of these.

db/migrate/20200213205538_add_initial_facets.rb
```diff
@@ -0,0 +1,61 @@
...
+      facet.big_query_name_column = 'organ__ontology_label'
+      facet.convention_name = 'alexandria'
+      facet.convention_version = '1.0'
+      facet.ontology_urls = [{name: 'todo', url: 'https://to.do/ontology'}]
+    end
+
+    SearchFacet.find_or_create_by!(name: "disease") do |facet|
+      facet.identifier = "disease"
+      facet.is_ontology_based = true
+      facet.is_array_based = false
```
bistline: Disease is array based as per the schema definition.

app/lib/synthetic_study_populator.rb
bistline: Personally, whenever I see backticks for the `system` command, I worry that this is the kind of thing that can easily break or lead to unintended return values.  Also, since we have a `user` object already here, we can strip off the username before `@` easily enough to get the same value.

devonbush: agree on this being hacky, the idea was it would be short-lived while we work out the separate workspace issues.  But since 'short-lived' is always a goal rather than guarantee, I'll pull from the user like you suggest :)

PR #396, opened 2020-02-14
Annotation dropdown timing/logic refactoring

bistline:
Refactoring the timing and logic of how the `get_new_annotations` is handled by moving the code to update the value of the dropdown menu itself into the JS response view, and then moving the subsequent call to re-render a cluster/heatmap/violin/dot plot into the `complete` handler, rather than the `success` handler.

Also addressing lingering issues with % encoding of parameters in URLs.

This PR satisfies [SCP-2118](https://broadworkbench.atlassian.net/browse/SCP-2118).

### DIFF
--- a/app/views/site/_study_visualize.html.erb
+++ b/app/views/site/_study_visualize.html.erb
@@ -130,7 +130,7 @@
             method: 'GET',
             dataType: 'script',
             complete: function (jqXHR, textStatus) {
-                renderWithNewCluster(textStatus, renderScatter());
+                renderWithNewCluster(textStatus, renderScatter);
             }
         });
     });
--- a/app/views/site/_study_visualize.html.erb
+++ b/app/views/site/_study_visualize.html.erb
@@ -123,25 +123,20 @@
         $('#search_cluster').val(newCluster);
         $('#gene_set_cluster').val(newCluster);
         // grab currently selected annotation
-        var currAnnot = $('#annotation').val();
         var currSubsample = $('#subsample').val();
         // get new annotation options and re-render
         $.ajax({
             url: "<%= get_new_annotations_path(accession: @study.accession, study_name: @study.url_safe_name)%>?cluster=" + encodeURIComponent(newCluster) + '&subsample=' + encodeURIComponent(currSubsample),
             method: 'GET',
             dataType: 'script',
-            success: function (data) {
-                // parse response as a string and see if currently selected annotation exists in new annotations
-                if (data.indexOf(currAnnot) >= 0) {
-                    $('#annotation').val(currAnnot);
-                }
-                $(document).ready(function () {
+            complete: function (jqXHR, textStatus) {
+                if (textStatus === 'success') {
                     // since we now have new annotations, we need to set them in the search form for persistence
                     var an = $('#annotation').val();
                     $('#search_annotation').val(an);
                     $('#gene_set_annotation').val(an);
                     renderScatter();
-                });
+                }
             }
         });
     });
--- a/app/views/site/_study_visualize.html.erb
+++ b/app/views/site/_study_visualize.html.erb
@@ -130,13 +130,7 @@
             method: 'GET',
             dataType: 'script',
             complete: function (jqXHR, textStatus) {
-                if (textStatus === 'success') {
-                    // since we now have new annotations, we need to set them in the search form for persistence
-                    var an = $('#annotation').val();
-                    $('#search_annotation').val(an);
-                    $('#gene_set_annotation').val(an);
-                    renderScatter();
-                }
+                renderWithNewCluster(textStatus, renderScatter());
             }
         });
     });
--- a/app/views/site/_view_gene_expression_heatmap.html.erb
+++ b/app/views/site/_view_gene_expression_heatmap.html.erb
@@ -153,7 +153,7 @@
             method: 'GET',
             dataType: 'script',
             complete: function (jqXHR, textStatus) {
-                renderWithNewCluster(textStatus, drawHeatmap());
+                renderWithNewCluster(textStatus, drawHeatmap);
             }
         });
     });
--- a/app/views/site/_view_gene_expression_heatmap.html.erb
+++ b/app/views/site/_view_gene_expression_heatmap.html.erb
@@ -152,18 +152,14 @@
             url: "<%= get_new_annotations_path(study: params[:study])%>?cluster=" + newCluster,
             method: 'GET',
             dataType: 'script',
-            success: function (data) {
-                // parse response as a string and see if currently selected annotation exists in new annotations
-                if (data.indexOf(currAnnot) >= 0) {
-                    $('#annotation').val(currAnnot);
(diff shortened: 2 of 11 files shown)


app/views/site/_study_visualize.html.erb
```diff
@@ -123,25 +123,20 @@
...
             dataType: 'script',
-            success: function (data) {
-                // parse response as a string and see if currently selected annotation exists in new annotations
-                if (data.indexOf(currAnnot) >= 0) {
-                    $('#annotation').val(currAnnot);
-                }
-                $(document).ready(function () {
+            complete: function (jqXHR, textStatus) {
+                if (textStatus === 'success') {
                     // since we now have new annotations, we need to set them in the search form for persistence
```
devonbush: that's interesting that the document.ready used to be there, presumably to solve a race condition where this ajax could complete before the html page finished loading.  It looks like that's no longer applicable

app/views/site/_view_gene_expression_heatmap.html.erb
```diff
@@ -152,18 +152,14 @@
...
-            success: function (data) {
-                // parse response as a string and see if currently selected annotation exists in new annotations
-                if (data.indexOf(currAnnot) >= 0) {
-                    $('#annotation').val(currAnnot);
-                }
-                $(document).ready(function () {
+            complete: function (jqXHR, textStatus) {
+                if (textStatus === 'success') {
                     // since we now have new annotations, we need to set them in the search form for persistence
                     var an = $('#annotation').val();
```
devonbush: seems like these 3 lines occur in 3 different places.  Is there a sensible place to pull these out into an 'updateAnnotationValues' method or somesuch?

app/views/site/get_new_annotations.js.erb
```diff
@@ -1,3 +1,10 @@
+// dump out annotation dropdown values to see if we have a match for the existing selected annotation
+var annotationData = <%= raw @cluster_annotations.values.map {|coll| coll.map(&:last)}.flatten %>;
```
devonbush: As part of the long trek towards SCP UI 2.0, I'd like to gradually start separating javascript code from ruby dynamic templates as much as possible.   So this line can be split into 
```
var selectedAnnotationElement = <%= .. %>
var selectedAnnotation = selectedAnnotationElement.val()
```

This will slowly make it clearer exactly what data we need from the server so we can build the appropriate APIs

(For bonus credit, the above line could be rewritten to put the ruby computation of the value.map on a controller, so that the template can just say `var annotationData = <%= @flattenedAnnotationValues %>
`

app/views/site/get_new_annotations.js.erb
```diff
@@ -1,3 +1,10 @@
...
+var selectedAnnotation = $("#<%= @target %>annotation").val();
 $("#<%= @target %>annotation").html("<%= escape_javascript(select_tag :annotation, grouped_options_for_select(@cluster_annotations, params[:annotation]), class: 'form-control' )%>");
 // also reload subsampling options as they may have changed
```
devonbush: same -- disentangle dynamic template from JS where practical

app/views/site/render_global_gene_expression_plots.js.erb
```diff
@@ -78,22 +78,16 @@ $("#<%= @target %>-annotation, #<%= @target %>-subsample").change(function() {
...
     $.ajax({
         url: "<%= get_new_annotations_path(study_name: @study.url_safe_name)%>?cluster=" + encodeURIComponent(newCluster) + '&target=<%= @target %>',
         method: 'GET',
         dataType: 'script',
-        success: function (data) {
+        complete: function (jqXHR, textStatus) {
             // parse response as a string and see if currently selected annotation exists in new annotations
-            if (data.indexOf(currAnnot) >= 0) {
-                $('#<%= @target %>-annotation').val(currAnnot);
+            if (textStatus === 'success') {
```
devonbush: same disentangling

app/views/site/get_new_annotations.js.erb
bistline: Makes sense - I can do that.

app/views/site/_view_gene_expression_heatmap.html.erb
bistline: I had wanted to do that, but there was enough difference between all the various instances of this callback that it seemed like abstraction would be messy.  But I'll take another look at it to see if there's a clean way of doing it.

app/views/site/_study_visualize.html.erb
bistline: True, but `complete` makes it even more explicit for the client to wait until all DOM updates have completed.  Every now and then I've seen some flakiness in UI tests with cluster/violin plot rendering and I'm thinking that this might be an underlying cause.

app/views/site/render_global_gene_expression_plots.js.erb
```diff
@@ -78,22 +78,13 @@ $("#<%= @target %>-annotation, #<%= @target %>-subsample").change(function() {
...
-            if (data.indexOf(currAnnot) >= 0) {
-                $('#<%= @target %>-annotation').val(currAnnot);
-            }
-            $(document).ready(function () {
-                // since we now have new annotations, we need to set them in the search form for persistence
-                var an = $('#annotation').val();
-                renderDistribution<%= @identifier %>()
-            });
+        complete: function (jqXHR, textStatus) {
+            renderWithNewCluster(textStatus, renderDistribution<%= @identifier %>(), false);
```
eweitz: This `renderDistribution<%= @identifier %>()` is a holdover so I don't consider it blocking, but it'd be nice to refactor in the future to not construct the function name from a Rails variable.

Accepting `identifier` as a function parameter, e.g.:

```
renderDistribution(identifier);
```

Would enable disentangling JS and Rails.

bistline: I think the reason we did that initially is because there is a dynamic function generated in the response to every call to load a gene result for global gene search, and each function needed to be specific to that response as there was a great deal of Ruby interpolation in setting the URLs, the cluster & annotation dropdowns, and other things - it wasn't just the value of `identifier`.  Those concerns aside, it would be possible to refactor into a generic `renderDistribution()` function with a little planning.  I think a refactoring is probably warranted since we will be likely mucking around in this code sooner rather than later as a part of advanced/faceted search work.

app/views/site/_study_visualize.html.erb
```diff
@@ -123,25 +123,14 @@
...
-                    $('#annotation').val(currAnnot);
-                }
-                $(document).ready(function () {
-                    // since we now have new annotations, we need to set them in the search form for persistence
-                    var an = $('#annotation').val();
-                    $('#search_annotation').val(an);
-                    $('#gene_set_annotation').val(an);
-                    renderScatter();
-                });
+            complete: function (jqXHR, textStatus) {
```
devonbush: this is fantastic -- even cleaner than I had expected

PR #401, opened 2020-02-19
 fix facet updating

devonbush:
find_or_create_by! didn't do what I thought it did...


eweitz:
Please add a GitHub `security-risk` label.

PR #402, opened 2020-02-21
Adding data_type to SearchFacet

bistline:
Adding the data_type attribute to SearchFacets for controlling query & UI rendering behavior for faceted search.  Currently supported data_types are `string`, `number`, and `boolean`.  SearchFacets now have the ability to auto-detect data type upon creation.  Also, the behavior of updating facet filter values is changed so that facets automatically updated their filters upon creation.

This PR satisfies [SCP-2151](https://broadworkbench.atlassian.net/browse/SCP-2151).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem 'bootsnap', '>= 1.1.0', require: false
 gem 'minitest-rails'
 gem 'minitest-reporters'
 
-group :development do
+group :development, :test do
   # Access an IRB console on exception pages or by using <%= console %> in views
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
@@ -46,16 +46,12 @@ group :development do
   gem 'byebug'
 end
 
-group :test do
-  gem 'byebug'
-end
-
 gem 'devise'
 gem 'omniauth-google-oauth2'
 gem 'omniauth-rails_csrf_protection'
 gem 'googleauth'
 gem 'google-cloud-storage', require: 'google/cloud/storage'
-gem 'google-cloud-bigquery', require: "google/cloud/bigquery"
+gem 'google-cloud-bigquery', require: 'google/cloud/bigquery'
 gem 'selenium-webdriver'
 gem 'jquery-ui-rails'
 gem 'bootstrap-sass', :git => 'https://github.com/twbs/bootstrap-sass'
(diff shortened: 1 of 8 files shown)


Gemfile
```diff
@@ -33,7 +33,7 @@ gem 'bootsnap', '>= 1.1.0', require: false
 gem 'minitest-rails'
 gem 'minitest-reporters'
 
-group :development do
+group :development, :test do
```
eweitz: Glad this will do away with the recent warning I saw!

test/models/search_facet_test.rb
```diff
@@ -69,4 +72,31 @@ def setup
...
+  test 'should set data_type and is_array_based on create' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    mock = Minitest::Mock.new
+    mock.expect :query, @column_schema, [String]
+
+    SearchFacet.stub :big_query_dataset, mock do
+      age_facet = SearchFacet.create(
+          name: 'Organism Age',
+          identifier: 'organism_age',
```
devonbush: so it's less confusing, I'd use 'sex' as the example facet here instead of 'organism_age' here, since you're saying it's not ontology based.

app/lib/search_facet_populator.rb
```diff
@@ -25,32 +27,43 @@ def self.populate_facet_by_name(facet_name, schema_object)
...
     updated_facet.big_query_name_column = is_ontology_based ? ontology_label_field_name : facet_name
     updated_facet.convention_name = schema_object['title']
     updated_facet.convention_version = alexandria_convention_config[:version]
-    # for now, just set the ontology name to the URL, since IIRC we're delegating display responsibility to the front end
-    updated_facet.ontology_urls = is_ontology_based ? [{name: field_def['ontology'], url: field_def['ontology']}] : []
+    if is_ontology_based
+      url = field_def['ontology']
+      ontology = fetch_json_from_url(url)
+      # check if response has expected keys; if not, default to URL for name value
+      ontology_name = ontology.dig('config', 'title') ? ontology['config']['title'] : url
```
devonbush: this is pretty slick!  If it's not too much of a hassle, you might as well add an assert for it around line 14 of search_facet_populator_test to check that we got the right ontology name for diseases

bistline: Can do.

bistline: fbc10ea

PR #405, opened 2020-02-21
Bulk Download size preview

bistline:
Adding a preview of total number of files requested and total bytes to bulk download requests.  This can be used as a pre-flight request before generating an auth_code for a user to make an informed decision before initiating a bulk download request.

This also fixes an issue with the Swagger documentation for `/single_cell/api/v1/search` not containing all the possible file types in the response documentation, and also not filtering search results on valid accessions (as opposed to possible accessions).

This PR satisfied [SCP-2150](https://broadworkbench.atlassian.net/browse/SCP-2150).


app/controllers/api/v1/search_controller.rb
```diff
@@ -493,6 +559,25 @@ def self.match_studies_by_facet(query_results, search_facets)
         end
         matches
       end
+
+      # find valid StudyAccessions from query parameters
+      # only returns accessions currently in use
+      def self.find_matching_accessions(raw_accessions)
+        accessions = split_query_param_on_delim(parameter: raw_accessions)
+        sanitized_accessions = StudyAccession.sanitize_accessions(accessions)
+        Study.where(:accession.in => sanitized_accessions).map(&:accession)
```
devonbush: in cases like this where you're just getting ids it's good to be in the habit of using pluck rather than map for performance and clarity reasons.  (although for this particular case it probably doesn't make much of a difference)

test/integration/lib/bulk_download_service_test.rb
```diff
@@ -37,6 +37,29 @@ def setup
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
```
devonbush: nice test!

PR #392, opened 2020-02-11
Jest test infrastructure (SCP-2107), and SCP API JavaScript client

eweitz:
This adds Jest test infrastructure and a JavaScript client for the [Single Cell Portal REST API](https://singlecell.broadinstitute.org/single_cell/api).

### Jest test infrastructure
Infrastructure for automated tests with [Jest](https://jestjs.io/), a popular JS testing framework, is now available here.  This enables React UI components to be tested in a manner that is isolated from Rails and the REST API.

You can run it via `yarn test`. 

<details><summary>Example test output, with coverage metrics</summary>
This output comes from code as of commit 6cdec0
<p>

```bash

yarn test
yarn run v1.16.0
$ jest --coverage
 PASS  test/js/scp-api.test.js
  JavaScript client for SCP REST API
    ✓ should return `authCode` and `timeInterval` from fetchAuthCode (38ms)
    ✓ should return 10 filters from fetchFacetFilters (19ms)

  console.log app/javascript/lib/scp-api.js:114
    in scpApi, path:  /search/auth_code

  console.log app/javascript/lib/scp-api.js:134
    json { auth_code: 123456, time_interval: 1800 }

  console.log app/javascript/lib/scp-api.js:114
    in scpApi, path:  /search/facets_filters_disease_tuberculosis

  console.log app/javascript/lib/scp-api.js:134
    json {
      facet: 'disease',
      query: 'tuberculosis',
      filters: [
        { name: 'Tubercolosis', id: 'DOID0000123' },
        { name: 'Ocular tubercolosis', id: 'DOID0000124' },
        { name: 'Tuberculosis, spinal', id: 'DOID0000125' },
        { name: 'Endocrine tuberculosis', id: 'DOID0000126' },
        { name: 'Inactive tuberculosis', id: 'DOID0000127' },
        { name: 'Tubercolosis, bovine', id: 'DOID0000128' },
        { name: 'Tuberculosis, avian', id: 'DOID0000129' },
        { name: 'Esophageal tubercolosis', id: 'DOID0000130' },
        { name: 'Intestinal tuberculosis', id: 'DOID0000131' },
        { name: 'Abdominal tuberculosis', id: 'DOID0000132' }
      ]
    }

(node:60341) UnhandledPromiseRejectionWarning: TypeError: Only absolute URLs are supported
[...]

PR #407, opened 2020-02-24
Refresh button for Bulk Download UI (SCP-2121)

eweitz:
This adds a Refresh button for the Bulk Download UI.  It also implements [a suggestion](https://github.com/broadinstitute/single_cell_portal_core/pull/392#discussion_r379563599) from from PR #392, in which the `/search/auth_code` API endpoint was integrated.

This satisfies SCP-2121.  

Follow ups to integrate more upcoming API updates: SCP-2169, polish: SCP-2167.

### DIFF
--- a/app/javascript/components/DownloadButton.js
+++ b/app/javascript/components/DownloadButton.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faDownload } from '@fortawesome/free-solid-svg-icons';
 import Modal from 'react-bootstrap/lib/Modal';
-import Tooltip from 'react-bootstrap/lib/Tooltip';
+// import Tooltip from 'react-bootstrap/lib/Tooltip'; // We'll need this when refining onClipboardCopySuccess
 import Clipboard from 'react-clipboard.js';
 
 import { fetchAuthCode } from 'lib/scp-api';
--- a/app/javascript/components/DownloadButton.js
+++ b/app/javascript/components/DownloadButton.js
@@ -12,8 +12,10 @@ import { fetchAuthCode } from 'lib/scp-api';
  *
  * @returns {Object} Object for auth code, time interval, and download command
  */
-async function generateDownloadConfig() {
-  const searchQuery = '&file_types=metadata,expression&accessions=SCP1,SCP2';
+async function generateDownloadConfig(matchingStudies) {
+
+  const accessions = matchingStudies.join(',');
+  const searchQuery = `&file_types=metadata,expression&accessions=${accessions}`;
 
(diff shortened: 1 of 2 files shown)


app/javascript/components/DownloadButton.js
```diff
@@ -37,20 +40,25 @@ async function generateDownloadConfig() {
...
     fetchData();
+  }
+
+  useEffect(() => {
+    updateDownloadConfig(props.matchingStudies);
   }, []);
 
-  function onSuccess() {
-    console.log('TODO (SCP-2121): Show "Copied!" tooltip');
+  function onClipboardCopySuccess(event) {
```
knapii-developments: Nice rename!

app/javascript/components/DownloadButton.js
```diff
@@ -2,6 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faDownload } from '@fortawesome/free-solid-svg-icons';
 import Modal from 'react-bootstrap/lib/Modal';
+import Tooltip from 'react-bootstrap/lib/Tooltip';
```
devonbush: is this import directly used in this file?  Or just needed so the js for data-toggle='tooltip' is loaded?  If the latter, this might be better imported in application.js since we will probably use it in other places too.

app/javascript/components/SearchPanel.js
```diff
@@ -23,20 +23,24 @@ export const SearchContext = React.createContext(searchContexts.study);
...
  * See related integration at /app/javascript/packs/application.js
  */
 export default function SearchPanel() {
   // Note: This might become  a Higher-Order Component (HOC).
   // This search component is currently specific to the "Studies" tab, but
   // could possibly also enable search for "Genes" and "Cells" tabs.
 
+  // TODO (SCP-2169): Remove this crude mock data with a prop when
+  // 1) HomePage component is merged and and 2) /search response is updated.
+  const matchingStudies = ['SCP1', 'SCP2'];
```
bistline: For reference, we're going to call this `matching_accessions` in search API responses.

app/javascript/components/DownloadButton.js
eweitz: This will be used directly used in this file -- but later, when refining `onClipboardCopySuccess` to briefly display a tooltip.  I've also added a comment to this effect in 68fc531.

In other words, this will get directly used when the TODO (referencing SCP-2167) is done.

app/javascript/components/SearchPanel.js
eweitz: I've added a note clarifying this in 68fc531.

eweitz: Thanks for the note, I've added a comment to that effect in 68fc531.

PR #409, opened 2020-02-25
Allowing case-sensitive gene searching

bistline:
From: https://github.com/broadinstitute/single_cell_portal/issues/90

Now honoring case-sensitive searching of genes both within and across studies.  The portal will return case-sensitive searches first, and then do case-insensitive matching only if no exact matches are found.  This also fixes a bug when two genes have the same searchable_name (e.g. down-cased gene symbol) and are then incorrectly merged together as the same gene when visualizing violin plots or heatmaps.

This update also addresses an issue with the Docker build context by excluding unnecessary files (compiled assets, Git branches, etc), updates a broken integration test due to changes to the MONDO ontology configuration in OLS, and adds a missing parameter to the `search` API endpoint for pagination control.

This PR satisfies [SCP-2170](https://broadworkbench.atlassian.net/browse/SCP-2170).


.dockerignore
```diff
@@ -1,7 +1,8 @@
...
 tmp/*
 data/*
 log/*
 data_bak
-/config/*.json
-node_modules/*
\ No newline at end of file
+config/*.json
+node_modules/*
+.git
```
devonbush: don't forget to add newline-at-end-of-file to your IDE when you get a chance

app/controllers/site_controller.rb
```diff
@@ -428,9 +428,8 @@ def view_gene_expression
 
   # re-renders plots when changing cluster selection
   def render_gene_expression_plots
-    matches = @study.genes.by_name_or_id(params[:gene], @study.expression_matrix_files.map(&:id))
     subsample = params[:subsample].blank? ? nil : params[:subsample].to_i
```
devonbush: @study.expression_matrix_files.pluck(:id)  
:)

app/controllers/site_controller.rb
```diff
@@ -475,9 +474,8 @@ def render_gene_expression_plots
   # renders gene expression plots, but from global gene search. uses default annotations on first render, but takes URL parameters after that
   def render_global_gene_expression_plots
     if check_xhr_view_permissions
-      matches = @study.genes.by_name_or_id(params[:gene], @study.expression_matrix_files.map(&:id))
       subsample = params[:subsample].blank? ? nil : params[:subsample].to_i
```
devonbush: pluck again

app/models/study.rb
```diff
@@ -100,18 +100,28 @@ def by_type(file_type)
 
   has_many :genes, dependent: :delete do
     def by_name_or_id(term, study_file_ids)
```
devonbush: oh, this is cool, I'd never seen the any_of method before

test/models/study_test.rb
```diff
@@ -3,5 +3,57 @@
...
+      DataArray.find_or_create_by!(name: @genes[gene].score_key, cluster_name: @exp_matrix.name, array_type: 'expression',
+                                   array_index: 0, study_file: @exp_matrix, values: @values,
+                                   linear_data_type: 'Gene', linear_data_id: @genes[gene].id, study: @study)
+      upcased_gene = gene.upcase
+      # do not insert data for upcased genes
+      @genes[upcased_gene] = Gene.find_or_create_by!(name: upcased_gene, searchable_name: upcased_gene.downcase,
+                                                     study: @study, study_file: @exp_matrix)
+    end
+  end
+
```
devonbush: great test

app/controllers/site_controller.rb
bistline: The return value of `@study.expression_matrix_files` is actually an array, not a query, so that's why I'm using `map` here.

app/controllers/site_controller.rb
bistline: Same as above.

PR #414, opened 2020-02-27
Apply and Clear for "More Facets" (SCP-2177)

eweitz:
This adds the Apply and Clear buttons to facets within the accordion shown upon clicking the "More Facets" button.  Per demo chat, these buttons now only apply to specific facets within the accordion, and are not applied across such facets.

This satisfies SCP-2177.


app/javascript/components/FiltersBoxSearchable.js
```diff
@@ -0,0 +1,90 @@
...
+export default function FiltersBoxSearchable(props) {
+
+  // State that is specific to FiltersBox
+  const [matchingFilters, setMatchingFilters] = useState(props.facet.filters);
+  const [hasFilterSearchResults, setHasFilterSearchResults] = useState(false);
+
+  // TODO: Get opinions, perhaps move to a UI code style guide.
+  //
+  // Systematic, predictable IDs help UX research and UI development.
+  //
```
devonbush: Concur.  We also need a good sense of what ids should and shouldn't be used for.  Specifically, I think that we should be looking to minimize as much as possible the use of ids in our javascript handlers--using getElementById is usually a sign that the code isn't as modular as it could be (since ids are the DOM equivalent of global vars).  Ids are probably best reserved for CSS and/or instrumentation

app/javascript/components/FacetsAccordion.js
```diff
@@ -1,32 +1,36 @@
...
-          );
-        })
-      }
-    </Accordion>
+    <div id='facets-accordion'>
+      <PanelGroup accordion>
+        {
+          props.facets.map((facet, i) => {
+            return (
+              <Panel key={i} eventKey={i}>
```
knapii-developments: Glad I saw this component! I can use this in my ResultsPanel. This will help make the look and feel more consistent.

app/javascript/components/Filters.js
```diff
@@ -5,6 +5,7 @@ import { Slider, Rail, Handles, Tracks, Ticks } from "react-compound-slider";
  * Component for a list of string-based filters, e.g. disease, species
  */
 function FilterList(props) {
+  if (typeof props.filters === 'undefined') return (<></>); // TODO: Remove this once /search/facets response is fixed
```
knapii-developments: What are you trying to do here?

eweitz: Good catch, this was a temporary stop-gap for an API issue that no longer exists.  I just removed it in c1e32ec.

app/javascript/components/FiltersBoxSearchable.js
eweitz: Agreed!

PR #416, opened 2020-02-28
Wiring up faceted and keyword search, and style polishes

devonbush:
Uses a central StudySearchContext to manage all search state and params.  this should give us an easy entry point for adding features like url-linkable searches, and a data-browser style text string representation of the search

image

image

image


app/javascript/components/FiltersSearchBar.js
```diff
@@ -31,7 +31,7 @@ export default function FiltersSearchBar(props) {
           id={filtersSearchBarId}
           type='text'
           autoComplete='false'
-          placeholder='Search'
+          placeholder='Search for a filter'
```
eweitz: Nice!  This refines understandability for users.

app/javascript/components/KeywordSearch.js
```diff
@@ -1,39 +1,41 @@
...
  * Component to search using a keyword value
- *
+ * optionally takes a 'keywordValue' prop with the initial value for the field
  */
 export default function KeywordSearch(props) {
-  const handleSubmit = event => {
+  const searchContext = useContext(StudySearchContext)
+  const [keywordValue, setKeywordValue] = useState('');
+
+  const handleSubmit = submitValue => {
```
eweitz: The above style is more common in class components, but in function components it would conventionally be:

```suggestion
  function handleSubmit(submitValue) {
```

At least as I see in React docs, e.g. https://reactjs.org/docs/handling-events.html.  Its traditional form is also slightly more readable, at least for me.

app/javascript/components/search/StudySearchProvider.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+  results: [],
+  isLoaded: false,
+  isError: false,
+  updateSearch: () => { throw new Error('You are trying to use this context outside of a Provider container') }
+}
+
+export const StudySearchContext = React.createContext(emptySearch)
+
+export default class StudySearchProvider extends React.Component {
```
eweitz: A function component would be more consistent with other components, and likely more succinct.

app/javascript/components/search/StudySearchProvider.js
```diff
@@ -0,0 +1,49 @@
...
+    initialState.updateSearch = this.updateSearch
+    this.state = initialState
+  }
+
+  componentDidMount() {
+    this.updateSearch();
+  }
+
+  updateSearch = async (searchParams) => {
+    const effectiveParams = _assign(this.state.params, searchParams)
```
eweitz: I prefer standard JavaScript when an equivalent exists, e.g. `Object.assign` here.  But I consider this a matter of taste.

app/javascript/lib/scp-api.js
```diff
@@ -136,10 +137,21 @@ export async function fetchFacetFilters(facet, query, mock=false) {
  */
 export async function fetchSearch(type, terms, facets, mock=false){
   // Needs to be edited to include facets
-  const searchPathAndQueryString = `/search?type=${type}&terms=${terms}`
+  const searchPathAndQueryString = `/search?type=${type}&terms=${terms}&facets=${buildFacetQueryString(facets)}`
   return await scpApi(searchPathAndQueryString, defaultInit, mock);
 }
 
+function buildFacetQueryString(facets) {
```
eweitz: Elegant!

app/javascript/components/FacetControl.js
```diff
@@ -10,25 +13,52 @@ function slug(value) {
...
+    };
+  }, []);
 
   const [showFilters, setShowFilters] = useState(false);
 
   const facetName = props.facet.name;
   const facetId = `facet-${slug(facetName)}`;
+  const searchContext = useContext(StudySearchContext)
+  const facetParams = searchContext.params.facets[props.facet.id]
+  let selectedFilterString = undefined
```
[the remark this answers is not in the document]
devonbush: I prefer 'let' for these cases where function and block scope are equivalent, and you're declaring a variable somewhere in the middle.  I like that it means a referenceError will be thrown if you accidentally try to use it before it was declared, as opposed to var which gets automatically initialized to undefined at the top of the function

app/javascript/components/search/StudySearchProvider.js
devonbush: concur

app/javascript/components/search/StudySearchProvider.js
devonbush: Yeah, I just spent some time attempting that, and it was beyond my abilities.  Hopefully soon, though!

PR #419, opened 2020-02-28
filter selections prepopulate from search url

devonbush:
This also has the benefit of making the filter boxes less reliant on unique DOM ids.


eweitz:
Please add `security-risk` GitHub label.

bistline:
I looked at the failed build, and the error was actually a timeout in waiting for a file to fail parsing.  This happens occasionally if it takes longer than usual to spin up the VM in PAPI.  So there is no regression on this build.

PR #420, opened 2020-03-02
fixing bug with search with no keywords

devonbush:
fixes a nil error on the @search_terms.split that would happen if no search terms were specified.  

@bistline this seems like a case that we should have automated testing to catch (sorry I missed it on peer review).  Can you (1) pull the business logic of the search into a static method, and then (2) add a few tests so that we've got coverage for the various permutations of empty keyword, empty facets, etc... (and we can write them without requiring http requests in the tests)

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -140,6 +140,7 @@ module Api
         @viewable = Study.viewable(current_api_user)
         # variable for determining how we will sort search results for relevance
         sort_type = :keyword
+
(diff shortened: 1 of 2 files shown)


bistline:
@devonbush so this brings up a question: why would we ever make a search request with no terms/facets?  I've seen that on page load the React front end makes a call to /search with no parameters.  Why wouldn't we use API endpoint to load visible studies for a user (/single_cell/api/v1/site)?  That's what that endpoint is for, and I should have made it more apparent as we were working on search that it should be used for that.

PR #422, opened 2020-03-02
Include accessions in download command, support contexts in tests (SCP-2169)

eweitz:
This incorporates study accessions into the bulk download `curl` command, so users can download only studies that match their search.  

It also refines the button UI to consider user's sign-in state, and adds Jest / Enzyme test support for code that uses the React `useContext` hook.

This satisfies SCP-2169.


app/javascript/components/DownloadButton.js
```diff
@@ -108,17 +110,35 @@ function DownloadCommandContainer(props) {
  */
 export default function DownloadButton(props) {
 
+  const searchContext = useStudySearchContext()
```
devonbush: this is an interesting pattern I haven't seen before.  I could see renaming the method to useContextStudySearch just so that it's really clear that it's a useContext hook and we can be careful of all the usual hook concerns, but I'll defer to you.

app/javascript/components/DownloadButton.js
```diff
@@ -1,20 +1,22 @@
...
 
+import { useStudySearchContext } from 'components/search/StudySearchProvider'
+import { useUserContext } from './UserProvider'
 import { fetchAuthCode } from 'lib/scp-api';
 
 /**
  * Fetch auth code, build download command, return configuration object
  *
  * @returns {Object} Object for auth code, time interval, and download command
  */
```
devonbush: good rename

app/javascript/components/DownloadButton.js
```diff
@@ -108,17 +110,35 @@ function DownloadCommandContainer(props) {
...
 
+  const searchContext = useStudySearchContext()
+  const userContext = useUserContext()
+
+  const matchingAccessions = searchContext.results.matchingAccessions || [];
+
+  /**
+   * Reports whether Download button be active,
+   * i.e. user is signed in and has search results
+   */
```
devonbush: I don't think this needs to be a function -- if userContext or searchContext changes, the component will rerender anyway.  So this can just be `active = userContext.accessToken !== '' && matchingAccessions.length > 0;`

app/javascript/components/DownloadButton.js
```diff
@@ -108,17 +110,35 @@ function DownloadCommandContainer(props) {
...
+  function shouldBeActive() {
+    return userContext.accessToken !== '' && matchingAccessions.length > 0;
+  }
+
```
devonbush: I don't see why this is state -- it's an entirely derived value

app/javascript/components/DownloadButton.js
```diff
@@ -108,17 +110,35 @@ function DownloadCommandContainer(props) {
...
+  const [active, setActive] = useState(shouldBeActive());
   const [show, setShow] = useState(false);
```
devonbush: Likewise, I don't think you need this useEffect

app/javascript/components/FacetsAccordion.js
```diff
@@ -7,31 +7,27 @@ import FiltersBox from './FiltersBox'
 export default function FacetsAccordion(props) {
 
   return (
```
devonbush: nice, I like getting rid of the surrounding div

app/javascript/styles/_search.scss
```diff
@@ -68,6 +68,20 @@
     font-weight: 500;
   }
 
+  #download-button.active {
```
devonbush: we now have the $action-color variable colors.scss for this color

app/javascript/styles/_search.scss
```diff
@@ -68,6 +68,20 @@
...
+    background-color: #4d72aa;
+  }
+
+  #download-button.active > span {
+    color: #fff;
+  }
+
+  .facet.active > span,
+  #more-facets-button.active > span {
+    color: #fff;
```
devonbush: $action-color

test/js/download.test.js
```diff
@@ -1,13 +1,28 @@
...
+import * as UserProvider from '../../app/javascript/components/UserProvider';
+import * as StudySearchProvider from '../../app/javascript/components/search/StudySearchProvider';
 
 describe('Download components for faceted search', () => {
   beforeAll(() => {
     global.fetch = fetch;
+
+    const userContext = {accessToken: 'test'};
+    const studySearchContext = {results: {matchingAccessions: ['SCP1', 'SCP2']}}
+
```
devonbush: this is cool

app/javascript/styles/_search.scss
eweitz: Addressed in b72b604.

app/javascript/styles/_search.scss
eweitz: Addressed in b72b604.

app/javascript/components/DownloadButton.js
eweitz: Addressed in b72b604.

app/javascript/components/DownloadButton.js
eweitz: Addressed in b72b604.

app/javascript/components/DownloadButton.js
eweitz: Addressed in b72b604.

app/javascript/components/DownloadButton.js
eweitz: Good readability suggestion, done in b72b6048.

This pattern enables testing components that use `useContext` in Jest / Enzyme.  It builds atop the [approach described here](https://stackoverflow.com/a/56380444).  It works, but feels kludgy.

PR #423, opened 2020-03-02
Changing message_modal behavior, fixing gene search submit

bistline:
The default behavior of the main "message_modal" has changed to always render into the page, whether there is any content to display or not.  Showing/hiding the modal has been abstracted out into the `showMessageModal()` function.  This fixes the issue where removing the open modal from the DOM to show a new message wedges the UI, and cannot be recovered.

Also, fixing a regression with submitting a gene search by pressing enter, rather than clicking the "Search Genes" button.

This PR satisfies [SCP-1272](https://broadworkbench.atlassian.net/browse/SCP-1272).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -546,7 +546,6 @@ function closeModalSpinner(spinnerTarget, modalTarget, callback) {
 
 // handles showing/hiding main message_modal and cleaning up state on full & partial page renders
 function showMessageModal(notice=null, alert=null) {
-    console.log('calling showMessageModal');
     // close any open modals
     if (OPEN_MODAL) {
         var modalTarget = $('#' + OPEN_MODAL);
@@ -574,7 +573,7 @@ function showMessageModal(notice=null, alert=null) {
     }
 
     // don't timeout alert messages
-    if (alert) {
+    if (!alert) {
         setTimeout(function() {
             $("#message_modal").modal("hide");
         }, 3000);
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -544,6 +544,43 @@ function closeModalSpinner(spinnerTarget, modalTarget, callback) {
     $(modalTarget).modal('hide');
 }
 
+// handles showing/hiding main message_modal and cleaning up state on full & partial page renders
+function showMessageModal(notice=null, alert=null) {
+    console.log('calling showMessageModal');
+    // close any open modals
+    if (OPEN_MODAL) {
+        var modalTarget = $('#' + OPEN_MODAL);
+        var modalData = modalTarget.data('bs.modal');
+        if ( typeof modalData !== 'undefined' && modalData.isShown) {
+            modalTarget.modal("hide");
+        }
+    }
+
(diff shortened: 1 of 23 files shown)


app/assets/javascripts/application.js
```diff
@@ -544,6 +544,43 @@ function closeModalSpinner(spinnerTarget, modalTarget, callback) {
     $(modalTarget).modal('hide');
 }
 
+// handles showing/hiding main message_modal and cleaning up state on full & partial page renders
+function showMessageModal(notice=null, alert=null) {
+    console.log('calling showMessageModal');
```
eweitz: [`console.log` in production](https://www.google.com/search?q=console.log+in+production) is generally discouraged.  It's fine in development, but in production omitting such debugging code is good form.  Breakpoints in DevTools are a cleaner and more robust solution if we need to debug in production.

```suggestion
```

app/assets/javascripts/application.js
```diff
@@ -544,6 +544,43 @@ function closeModalSpinner(spinnerTarget, modalTarget, callback) {
...
+        $("#message_modal").modal("show");
+    }
+
+    // don't timeout alert messages
+    if (alert) {
+        setTimeout(function() {
+            $("#message_modal").modal("hide");
+        }, 3000);
+    }
+}
```
eweitz: Does this code mean "if there is an alert, hide the message modal after 3 seconds"?  If so:

```suggestion
    // timeout alert messages
    if (alert) {
        setTimeout(function() {
            $("#message_modal").modal("hide");
        }, 3000);
    }
}
```

Elsewhere, I see `if (<%= alert.nil? %>) {` with another copy of this code snippet, which hints that maybe `if (!alert) {` is intended here.

Either way, either the code or the comment seems incorrect.

eweitz:
Please add `security-risk` GitHub label.

app/assets/javascripts/application.js
bistline: Yeah, that's a debug statement I forgot to take out.  Will remove.

app/assets/javascripts/application.js
bistline: You're correct - if there is an alert message present, it should not call `setTimeout()`.  I'll update that as well.

app/controllers/api/v1/search_controller.rb
```diff
@@ -176,14 +178,19 @@ def index
...
         when :keyword
-          @studies = @studies.to_a.sort_by {|study| -study.search_weight(@search_terms.split) }
+          @studies = @studies.sort_by {|study| -study.search_weight(@search_terms.split) }
         when :accession
-          @studies = @studies.to_a.sort_by {|study| possible_accessions.index(study.accession) }
+          @studies = @studies.sort_by {|study| possible_accessions.index(study.accession) }
         when :facet
-          @studies = @studies.to_a.sort_by {|study| -@studies_by_facet[study.accession][:facet_search_weight]}
+          @studies = @studies.sort_by {|study| -@studies_by_facet[study.accession][:facet_search_weight]}
+        else
```
devonbush: nice, I like what you've done here better a lot better than the way I had it

PR #425, opened 2020-03-03
Support for numeric facet range queries in search API

bistline:
Adding support for number-based facets and performing range queries via search API (e.g. values between a minimum and maximum threshold).  This includes server-side unit conversion for temporal facets (like organism_age) via a normalized conversion column (in this case, organism_age__seconds, from [scp-ingest-pipeline#80](https://github.com/broadinstitute/scp-ingest-pipeline/pull/80)).

This also includes a bugfix to ensure BigQuery stays consistent with the portal database when users delete studies containing convention metadata.

This PR satisfies SCP-2125.


app/controllers/api/v1/search_controller.rb
```diff
@@ -503,9 +503,25 @@ def set_search_facets_and_filters
             facet = SearchFacet.find_by(identifier: facet_id)
             if facet.present?
               matching_filters = []
-              facet.filters.each do |filter|
-                if filter_values.include?(filter[:id])
-                  matching_filters << filter
+              if facet.is_numeric?
```
devonbush: this method is getting long, extract this part into a helper method?

app/controllers/api/v1/search_controller.rb
```diff
@@ -559,6 +575,19 @@ def self.generate_bq_query_string(facets)
             from_clause += ", #{filter_arr_name}, UNNEST(#{filter_arr_name}.#{filter_val_name}) AS #{filter_where_val}"
             where_clauses << "(#{filter_where_val} IN UNNEST(#{column_name}))"
             base_query += ", #{filter_where_val}"
+          elsif search_facet.is_numeric?
```
devonbush: likewise, this might be a good method extract

app/controllers/api/v1/search_controller.rb
bistline: yeah that makes sense

app/models/search_facet.rb
```diff
@@ -255,13 +283,40 @@ def is_numeric?
     self.data_type == 'number'
   end
 
+  # know if a facet needs unit conversion
+  def must_convert?
```
eweitz: Nice foresight in supporting unit-less numeric metadata, e.g. `bmi` and `number_of_reads`.

app/controllers/api/v1/search_controller.rb
```diff
@@ -503,9 +503,25 @@ def set_search_facets_and_filters
             facet = SearchFacet.find_by(identifier: facet_id)
             if facet.present?
               matching_filters = []
-              facet.filters.each do |filter|
-                if filter_values.include?(filter[:id])
-                  matching_filters << filter
+              if facet.is_numeric?
+                # first filter value is range, e.g. 20-40, second value is unit, e.g. years
+                min_value, max_value = self.class.split_query_param_on_delim(parameter: filter_values.first, delimiter: '-').map(&:to_f)
+                requested_unit = filter_values.last
```
eweitz: We haven't defined URL parameters syntax for unit-less numeric facets.  But you could imagine `number_of_reads:150-500`.  Note the lack of units.

In that case, `filter_values.last` might not work.  Something to consider for later, at least.

app/controllers/api/v1/search_controller.rb
```diff
@@ -503,9 +503,25 @@ def set_search_facets_and_filters
             facet = SearchFacet.find_by(identifier: facet_id)
             if facet.present?
               matching_filters = []
-              facet.filters.each do |filter|
-                if filter_values.include?(filter[:id])
-                  matching_filters << filter
+              if facet.is_numeric?
+                # first filter value is range, e.g. 20-40, second value is unit, e.g. years
+                min_value, max_value = self.class.split_query_param_on_delim(parameter: filter_values.first, delimiter: '-').map(&:to_f)
```
eweitz: As mentioned in our departing chat, using `-` here precludes easy range filters involving negative numbers.  

Practically speaking, all `number` fields in [the metadata convention](https://storage.googleapis.com/broad-singlecellportal-public/AMC_v1.1.3.json) are non-negative.  But negative numbers are seen e.g. with differential expression metrics like log<sub>2</sub>(fold change), and it'd be nice to not preclude that.

Let's chat tomorrow about possible solutions!

app/controllers/api/v1/search_controller.rb
bistline: That was the thought I had yesterday evening.  I need logic to specifically check for units by looking in the `SearchFacet::TIME_UNITS` constant.

PR #429, opened 2020-03-04
Backfill organism age data

bistline:
Back filling data for the `organism_age__seconds` column in BigQuery to use with an Organism Age facet.  This is handled via a reversible Rails database migration.

Also fixing an issue with test schemas not being updated with migrations.

This PR satisfies SCP-2206.

### DIFF
--- a/db/migrate/20200304203650_create_organism_age_facet_and_backfill_data.rb
+++ b/db/migrate/20200304203650_create_organism_age_facet_and_backfill_data.rb
@@ -4,7 +4,7 @@ class CreateOrganismAgeFacetAndBackfillData < Mongoid::Migration
     bq_dataset = SearchFacet.big_query_dataset
     update_command = "UPDATE #{CellMetadatum::BIGQUERY_TABLE} "
     update_command += "SET organism_age__seconds = CAST((organism_age * #{SearchFacet::TIME_MULTIPLIERS['years']}) AS NUMERIC) "
-    update_command += "WHERE organism_age IS NOT NULL and organism_age__seconds IS NULL"
+    update_command += "WHERE organism_age IS NOT NULL and organism_age__seconds IS NULL AND organism_age__unit_label = 'year'"
     bq_dataset.query update_command
     SearchFacet.create(name: 'Organism Age', identifier: 'organism_age', big_query_id_column: 'organism_age',
                        big_query_name_column: 'organism_age', big_query_conversion_column: 'organism_age__seconds',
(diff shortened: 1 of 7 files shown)


db/migrate/20200304203650_create_organism_age_facet_and_backfill_data.rb
```diff
@@ -0,0 +1,21 @@
+class CreateOrganismAgeFacetAndBackfillData < Mongoid::Migration
+  def self.up
+    SearchFacet.where(identifier: 'organism_age').destroy
+    bq_dataset = SearchFacet.big_query_dataset
+    update_command = "UPDATE #{CellMetadatum::BIGQUERY_TABLE} "
+    update_command += "SET organism_age__seconds = CAST((organism_age * #{SearchFacet::TIME_MULTIPLIERS['years']}) AS NUMERIC) "
```
devonbush: how about adding a "AND organism_age__unit_label = 'year'" to this where clause just to be on the safe side?

bistline: 4c3b606

db/seeds.rb
```diff
@@ -115,16 +115,24 @@
...
+
+age = SearchFacet.find_or_initialize_by(identifier: 'organism_age')
+age.name = 'Organism Age'
+age.data_type = 'number'
+age.is_ontology_based = false
+age.is_array_based = false
+age.big_query_id_column = 'organism_age'
+age.big_query_name_column = 'organism_age'
+age.big_query_conversion_column ='organism_age__seconds'
+age.convention_name = 'Alexandria Metadata Convention'
```
eweitz: In [app/lib/search_facet_populator.rb line 76](https://github.com/broadinstitute/single_cell_portal_core/pull/429/files#diff-d2e81b6d4ff0bd3a2b5210c769a513a0R76), I see `facet.convention_name = 'Alexandria Convention'`.  So it seems either that should be "Alexandria Metadata Convention", or this should be "Alexandria Convention".

bistline: Long story short - you are correct, but in `SearchFacetPopulator.populate_from_schema`, we use the `title` attribute from the convention JSON file.  That value is `Alexandria Metadata Convention`.  `SearchFacetPopulator.populate_sample_facets` is only referenced in a migration that gets clobbered completely by the next migration that runs.  So it really isn't used anywhere, and all the code that does get used has the value set to `Alexandria Metadata Convention`.

Having said that, I agree it's totally Department of Redundancy Department, and not even consistently so.  But if those values don't match, it causes some tests to fail, and the JSON is in source control in a different repository.

bistline: And now that I've said all that, I agree that the inconsistency in the codebase (even in unused or deprecated methods) is overly confusing, so I've updated everything to be the same.

bistline:
Build failure is another false positive - I've been seeing issues today with parse jobs running a little slow, and the lone error was a job taking longer than the 3 minutes we allow for integration tests.

PR #431, opened 2020-03-05
Facet control styling updates

devonbush:
This updates the facet control styling to match the UI mockups, including handling when multiple filters are selected, styling the active facet(s), and including an 'X' button to clear facets

### DIFF
--- a/app/javascript/components/FacetControl.js
+++ b/app/javascript/components/FacetControl.js
@@ -82,7 +82,7 @@ export default function FacetControl(props) {
         id={facetId}
         className={`facet ${showFilters ? 'active' : ''} ${selectedFilterString ? 'selected' : ''}`}>
         <a onClick={handleButtonClick}>
-          { selectedFilterString ? controlContent : facetName  }
+          { controlContent }
         </a>
         <FiltersBoxSearchable
           show={showFilters}
--- a/app/javascript/components/FacetControl.js
+++ b/app/javascript/components/FacetControl.js
@@ -18,13 +18,16 @@ export default function FacetControl(props) {
 
   const [showFilters, setShowFilters] = useState(false);
 
+
+
   const facetName = props.facet.name;
   const facetId = `facet-${slug(facetName)}`;
   const searchContext = useContext(StudySearchContext)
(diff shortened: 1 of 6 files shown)


app/javascript/components/FacetControl.js
```diff
@@ -23,11 +25,26 @@ export default function FacetControl(props) {
...
     let selectedFilters = props.facet.filters.filter(filter => facetParams.includes(filter.id))
-    selectedFilterString = selectedFilters.map(filter => filter.name).join(', ')
+    if (selectedFilters.length > 1) {
+      selectedFilterString = `${facetName} (${selectedFilters.length})`
+    } else {
+      selectedFilterString = selectedFilters[0].name
+    }
+  }
+
+  const clearNode = useRef()
```
knapii-developments: I didn't know this was a function till this review!

app/javascript/components/FacetControl.js
```diff
@@ -51,13 +68,17 @@ export default function FacetControl(props) {
     };
   }, []);
 
+  let controlContent = facetName
+  if (selectedFilterString) {
+    controlContent = <>{ selectedFilterString } <button ref={clearNode} className="facet-clear" onClick={ clearFacet }><FontAwesomeIcon icon={faTimes}/></button></>
```
knapii-developments: I made the same comment on another review:

> I'm not a fan of reassigning variables. I personally avoid them. I think it can introduce problems such as:

> 1. You can change the external state on accident. Since it defined in the global context it's possible for another developer to reassign the value inside of a function if they use the same name. Or it's possible to repurpose the use of the variable on accident. This may cause unexpected behavior. 
> 2. It can make the code more complex. When we see 'let' I know the value is going to change. But it's hard to keep track of. 
> 
> I think if there could be a way to decide the final value of the variable it would make things easier. But the choice is up to you! And I misunderstood this code, please disregard.

devonbush: Yeah, I agree that in React, reassigning is something that should be done with great care.  I actually tried using a ternary operator with this logic at first, but couldn't get the jsx syntax quite right.  I'll take another stab at it

PR #434, opened 2020-03-09
Fixing more facets functionality and adding tests

devonbush:
Also refactors StudySearchProvider to have routable and non-routable components for easier testing without having to mock scpAPI. 

image

### DIFF
--- a/app/javascript/components/FiltersBox.js
+++ b/app/javascript/components/FiltersBox.js
@@ -77,7 +77,7 @@ export default function FiltersBox(props) {
     if (!value) {
       _remove(newSelection, id => {return id === filterId})
     }
-    setSelection(newSelection);
+    setSelection(newSelection)
   }
 
   function updateSelectionForFilterSlider(ranges) {
--- a/app/javascript/components/FiltersBox.js
+++ b/app/javascript/components/FiltersBox.js
@@ -77,7 +77,7 @@ export default function FiltersBox(props) {
     if (!value) {
       _remove(newSelection, id => {return id === filterId})
     }
-    setSelection(newSelection)
+    setSelection(newSelection);
   }
 
   function updateSelectionForFilterSlider(ranges) {
--- a/test/js/search/more-facets.test.js
+++ b/test/js/search/more-facets.test.js
@@ -38,7 +38,7 @@ const testFacets = [{
     allUnits: ["years", "months", "weeks", "days", "hours"]
   }]
 
-describe('Basic More facets capability for faceted search', () => {
+describe('Basic "More Facets" capability for faceted search', () => {
   it('the More Facets Button should correctly render when facets are selected', async () => {
     const routerNav = jest.spyOn(Reach, 'navigate')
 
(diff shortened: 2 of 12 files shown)


app/javascript/components/FiltersBox.js
```diff
@@ -78,6 +78,7 @@ export default function FiltersBox(props) {
     if (!value) {
       _remove(newSelection, (id) => { return id === filterId; })
     }
+    debugger
```
knapii-developments: Does this need to stay here?

test/js/search/more-facets.test.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import * as ReactAll from 'react';
+import { mount } from 'enzyme';
+import * as Reach from '@reach/router'
+
+const fetch = require('node-fetch');
+
+import MoreFacetsButton from 'components/MoreFacetsButton';
+import StudySearchProvider from 'components/search/StudySearchProvider';
+import * as ScpAPI from 'lib/scp-api'
```
eweitz: Imports like these, that rely on Rails `webpacker.yml`, prevent developers from using out-of-the-box IDE features like `Look up definition`, `Rename symbol`, and displaying docstrings on mouse-hover.

Relative imports enable such developer productivity features by default.  Specifically:
```suggestion
import MoreFacetsButton from '../../app/javascript/components/MoreFacetsButton';
import StudySearchProvider from '../../app/javascript/components/search/StudySearchProvider';
import * as ScpAPI from '../../app/javascript/lib/scp-api'
```

test/js/search/more-facets.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+const testFacets = [{
```
eweitz: Capturing our discussion on test strategy w.r.t. this approach of having mock data directly within the test and using [`const routerNav = jest.spyOn(Reach, 'navigate')`](https://github.com/broadinstitute/single_cell_portal_core/pull/434/files#diff-127e766ff6b693241ea844277b907d1dR43):

Upside:
* Tests are more self-contained in terms of readability

Downside:
* Will have false negatives upon breaking changes in other parts of our JS code, e.g. `scp-api.js`

In other words, this is towards to the unit test end of the spectrum.  

Tests in e.g. [`download.test.js`](https://github.com/broadinstitute/single_cell_portal_core/blob/b72b6048ef95adb7ae3622043b2070e48ab194c0/test/js/download.test.js) are closer to integration tests in that they cover other parts of the JS code (e.g. code in `scp-api.js`).  They're more complex to maintain, but slightly less susceptible to false negatives.

A diversity of approaches is worthwhile!

app/javascript/components/FiltersBox.js
eweitz:
```suggestion
    debugger
```

devonbush: heh, good catch!

test/js/search/more-facets.test.js
devonbush: per our chat, I'll keep these for now, and investigate ways to enable IDE discovery of them

PR #436, opened 2020-03-10
SCP-2219 - updating home page search style

devonbush:
This adds space between studies, adds a 750-character limit on descriptions (the current home page has a 100-word limit, but 750 characters is about the same length, and will be more consistent across studies.), and strips html tags.  

image

### DIFF
--- a/app/javascript/components/Study.js
+++ b/app/javascript/components/Study.js
@@ -1,0 +1,36 @@
+import React from 'react'
+
+const descriptionCharLimit = 750
+
+export function formatDescription(rawDescription) {
+
+  let textDescription = stripTags(rawDescription)
+  let suffixString = ''
+  if (textDescription.length > descriptionCharLimit) {
+    suffixString = '... (continued)'
+  }
+  return `${textDescription.substring(0, descriptionCharLimit)}${suffixString}`
+}
+
+function stripTags(rawString) {
+  let tempDiv = document.createElement('div');
+  // Set the HTML content with the providen
+  tempDiv.innerHTML = rawString;
+  // Retrieve the text property of the element
+  return tempDiv.textContent || '';
+}
+
+export default function Study(props) {
+  const studyDescription = { __html: formatDescription(props.study.description) }
+
+  return (
+    <div key={props.study.accession}>
+      <label htmlFor={props.study.name} id= 'result-title'>
+        <a href={props.study.study_url}>{props.study.name} </a></label>
+      <div>
+        <span id='cell-count' className='badge badge-secondary'>{props.study.cell_count} Cells </span>
+      </div>
+      <span dangerouslySetInnerHTML={studyDescription} id='descrition-text-area' disabled accession = {props.study.name}></span>
+    </div>
+  )
+}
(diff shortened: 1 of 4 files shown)


app/javascript/components/Study.js
```diff
@@ -0,0 +1,36 @@
...
+
+function stripTags(rawString) {
+  let tempDiv = document.createElement('div');
+  // Set the HTML content with the providen
+  tempDiv.innerHTML = rawString;
+  // Retrieve the text property of the element
+  return tempDiv.textContent || '';
+}
+
+export default function Study(props) {
```
eweitz: Please add a brief docstring to exported functions, for developer convenience.  E.g.:

```suggestion
/**
* Component for study search result.  Reports study name, accession, and brief description.
*/
export default function Study(props) {
```

(I consider `@param` docs for `props` extraneous, but I wouldn't object to including them.)

app/javascript/components/Study.js
```diff
@@ -0,0 +1,36 @@
...
+  const studyDescription = { __html: formatDescription(props.study.description) }
+
+  return (
+    <div key={props.study.accession}>
+      <label htmlFor={props.study.name} id= 'result-title'>
+        <a href={props.study.study_url}>{props.study.name} </a></label>
+      <div>
+        <span id='cell-count' className='badge badge-secondary'>{props.study.cell_count} Cells </span>
+      </div>
+      <span dangerouslySetInnerHTML={studyDescription} id='descrition-text-area' disabled accession = {props.study.name}></span>
```
eweitz: Naive question: why use `dangerouslySetInnerHTML` instead of putting content between tags, i.e.:

```suggestion
      <span id='description-text-area' disabled accession = {props.study.name}>{studyDescription}</span>
```

Also note `descrition` -> `description`.

app/javascript/components/Study.js
```diff
@@ -0,0 +1,36 @@
...
+
+const descriptionCharLimit = 750
+
+export function formatDescription(rawDescription) {
+
+  let textDescription = stripTags(rawDescription)
+  let suffixString = ''
+  if (textDescription.length > descriptionCharLimit) {
+    suffixString = '... (continued)'
+  }
```
eweitz: This is more readable to me than its "const-and-ternary" equivalent.

app/javascript/components/Study.js
knapii-developments: The returned value for studyDescription has inner html tags. Taken from [ReactJS](https://reactjs.org/docs/dom-elements.html),  "dangerouslySetInnerHTML is React’s replacement for using innerHTML in the browser DOM"

eweitz: Helpful reference, thank you!

Given that, it seems the new `stripTags` function makes `dangerouslySetInnerHTML` unnecessary.

devonbush: great point

app/javascript/components/Study.js
devonbush: yeah, I tend to overuse ternaries, so I'm glad I have you as a counterweight.  Ternaries are really only good when the condition and assignments are really simple and short

PR #437, opened 2020-03-10
Supporting keyword and exact phrase searching

bistline:
Previously, searching only supported keyword or a single phrase when performing searches.  We will now be able to do multiple phrases, or a combination of phrase and keyword.  Also, study matches will report which terms matched during search (before it just returned all search terms).

This PR satisfies SCP-2217.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -151,7 +151,7 @@ module Api
           # for keywords only, we can use the text index
           # for exact phrases (or combination) we have to run a regex search manually on names/descriptions
           if @search_terms.include?("\"")
-            @term_list = self.class.split_query_string_by_context(query_string: @search_terms)
+            @term_list = self.class.extract_phrases_from_search(query_string: @search_terms)
             study_regex = /(#{@term_list.join('|')})/i
             @studies = @viewable.any_of({name: study_regex}, {description: study_regex},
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -525,6 +536,21 @@ def sanitize_search_values(terms)
         end
       end
 
+      # extract any "quoted phrases" from query string and tokenize terms
+      def self.extract_phrases_from_search(query_string:)
+        terms = []
+        query_string.split("\"").each do |substring|
+          # when splitting on double quotes, phrases will not have any leading/trailing whitespace
```
devonbush: do we need to escape regex special characters that are inside terms?

bistline: That's already happened in https://github.com/broadinstitute/single_cell_portal_core/blob/jb-keyword-phrase-search/app/controllers/api/v1/search_controller.rb#L147 - `sanitize_search_terms` escapes any unsafe characters.

bistline: I actually just realized that this isn't true in this case - it's removing some unsafe characters, but not ones specific to regular expressions.  We want to use `Regexp.quote` for that.  I'll make that change.

bistline: c9a63d2

PR #441, opened 2020-03-11
UX Updates for Convention Metadata

bistline:
Addressing some UX paper cuts with uploading convention metadata files:

1. Dropdown disappearing after upload
2. Unable to specify convention metadata on sync
3. Adding indication in "Study Info" page of convention metadata
4. Adding line to parse success email indicating this was a convention metadata file (failure emails have the command line, which will say `--validate-convention`)

This PR satisfies SCP-2225 and SCP-2226

### DIFF
--- a/app/views/studies/_initialize_metadata_form.html.erb
+++ b/app/views/studies/_initialize_metadata_form.html.erb
@@ -20,13 +20,13 @@
 			<%= f.text_field :file_type, readonly: true, class: 'form-control file-type' %>
 		</div>
 		<div class="col-sm-5 upload-field">
+      <div class="col-sm-5">
+        <%= f.label :use_metadata_convention, "Use Metadata Convention?" %> <%= render partial: 'metadata_convention_help_popover', locals: {id: f.object.id.to_s} %><br />
+        <%= f.select :use_metadata_convention, options_for_select([['Yes', true],['No', false]], study_file.use_metadata_convention), {}, class: 'form-control use-metadata-convention', disabled: !study_file.new_record? %>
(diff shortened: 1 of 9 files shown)


app/views/studies/_initialize_metadata_form.html.erb
```diff
@@ -20,13 +20,13 @@
 			<%= f.text_field :file_type, readonly: true, class: 'form-control file-type' %>
 		</div>
 		<div class="col-sm-5 upload-field">
+      <div class="col-sm-5">
+        <%= f.label :use_metadata_convention, "Use Metadata Convention?" %> <%= render partial: 'metadata_convention_help_popover', locals: {id: f.object.id.to_s} %><br />
+        <%= f.select :use_metadata_convention, options_for_select([['Yes', true],['No', false]], study_file.use_metadata_convention), {}, class: 'form-control use-metadata-convention', disabled: !study_file.new_record? %>
```
devonbush: is there a reason you use !study_file.new_record? here instead of f.object.new_record like you do in the partial?  I could see standardizing, and  then having the _metadata_file_fields partial just be the label and select, so that you could use the partial here too.

bistline: I tried to stay consistent within whatever partial/form I was in and not mix usage.  Most of the views in the upload wizard say `study_file`, whereas the ones in the sync page and other places use `f.object`.  I agree it would be good to standardize, but that would mean editing a large number of partials to make the nomenclature consistent.  Seemed out-of-scope for this particular PR.

PR #443, opened 2020-03-12
Removing running migrations in test environment

bistline:
Removing command to run database migrations in the Rails `test` environment via `bin/run_tests.sh`.  This is to address issues with inconsistent test results via CI.

This PR satisfies SCP-2232.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -115,24 +115,20 @@ AnalysisConfiguration.create(namespace: 'single-cell-portal', name: 'split-clust
                              configuration_snapshot: 2, description: 'This is a test description.')
 
 # SearchFacet seeds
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -115,24 +115,20 @@
                              configuration_snapshot: 2, description: 'This is a test description.')
 
 # SearchFacet seeds
```
devonbush: I'd add a comment here explaining why we chose these three facets.  Or if it's not particularly important to have these three particular facets, maybe use populate_from_schema instead?

bistline: The reason: one is string based, but not arrays.  The other is string & array based.  The last is numeric.  This covers all of our facet types.  I'll add a comment to that effect.

bistline: 8b721b0

PR #445, opened 2020-03-13
Fixing issue re: deleting studies w/o metadata files

bistline:
Addressing an issue in `DeleteQueueJob` where deleting a study without a metadata file throws an error, and does not delete the study.

This PR satisfies SCP-2234.


bistline:
> If it were me, I'd do
> `if metadata_file.try(:use_metadata_convention)`
> on line 161. But either way works for a hotfix

Interesting - I'm actually not familiar with the `try` method.  That would work too, but I feel like this approach is more readable/explicit.

PR #447, opened 2020-03-13
Allowing inferred searching using facets as keywords

bistline:
Non-numeric faceted searches will now attempt to "infer" results by converting filter values into exact phrases (using filter display names, e.g. "tuberculosis", or "Homo sapiens") and running a secondary search after all exact matches are found.  These studies are labeled with `inferred_match: true` in the response, and ordered after all exact matches have returned.  They will also list what terms matched using the standard `term_matches` parameter.  These results are weighted together, so higher confidence inferred matches are returned first (but still after all exact matches).  This PR also DRYs up the search logic somewhat by extracting MongoDB-based searches into a static method.

This PR satisfies SCP-2205.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -171,7 +171,7 @@ module Api
 
         # filter results by branding group, if specified
         if @selected_branding_group.present?
-          @studies = @viewable.where(branding_group_id: @selected_branding_group.id)
+          @studies = @studies.where(branding_group_id: @selected_branding_group.id)
         end
 
         # only call BigQuery if list of possible studies is larger than 0 and we have matching facets to use
@@ -207,9 +207,9 @@ module Api
         # save list of study accessions for bulk_download/bulk_download_size calls, as well as caching query results
         @matching_accessions = @studies.map(&:accession)
 
-        # if a user ran a faceted search, also run a "fuzzy" search by converting filter display values to keywords
+        # if a user only ran a faceted search, attempt to infer results by converting filter display values to keywords
         if params[:terms].blank? && @facets.any?
-          @filter_keywords = self.class.convert_filter_names_to_search(facets: @facets)
+          @filter_keywords = self.class.convert_filters_for_inferred_search(facets: @facets)
           filter_regex = self.class.escape_terms_for_regex(term_list: @filter_keywords)
           base_studies = @selected_branding_group.present? ? @viewable.where(branding_group_id: @selected_branding_group.id) : @viewable
           inferred_studies = base_studies.any_of({name: filter_regex}, {description: filter_regex}).
@@ -626,8 +626,8 @@ module Api
         with_statement + base_query + from_clause + " WHERE " + where_clauses.join(" AND ")
       end
 
-      # convert a list of facet filters into a keyword search for inferred/fuzzy matching
-      def self.convert_filter_names_to_search(facets:)
+      # convert a list of facet filters into a keyword search for inferred matching
+      def self.convert_filters_for_inferred_search(facets:)
         filter_terms = []
         facets.each do |facet|
           search_facet = SearchFacet.find(facet[:object_id])
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -138,6 +138,11 @@ module Api
 
       def index
         @viewable = Study.viewable(current_api_user)
+
+        # filter results by branding group, if specified
+        if @selected_branding_group.present?
+          @viewable = @viewable.where(branding_group_id: @selected_branding_group.id)
+        end
         # variable for determining how we will sort search results for relevance
         sort_type = :none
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -557,6 +574,27 @@ def self.escape_terms_for_regex(term_list:)
         /(#{escaped_terms.join('|')})/i
       end
 
+      # generate a Mongoid::Criteria object to perform a keyword/exact phrase search based on use case
+      # supports the following cases: :keyword (individual terms), :phrase (quoted phrases & keywords)
+      # and :inferred (converting a facet-based query to keywords)
+      # will scope the query based off of :base_studies, and include/exclude studies matching
+      # :accessions based on the :query_case (included by default, but excluded in :inferred to avoid duplicates)
+      def self.generate_mongo_query_by_case(terms:, base_studies:, accessions:, query_case:)
```
devonbush: hmmm.. I'm trying to think of a better word to use than 'case', since I at first thought this was referring to whether the search was case-sensitive. 'type'? 'term_type'? 'goal'? I'm not sure any of those are better though

app/controllers/api/v1/search_controller.rb
```diff
@@ -612,6 +650,19 @@ def self.generate_bq_query_string(facets)
         with_statement + base_query + from_clause + " WHERE " + where_clauses.join(" AND ")
       end
 
+      # convert a list of facet filters into a keyword search for inferred matching
+      def self.convert_filters_for_inferred_search(facets:)
+        filter_terms = []
+        facets.each do |facet|
+          search_facet = SearchFacet.find(facet[:object_id])
+          # only use non-numeric facets
```
devonbush: nice catch :). That's definitely an edge case I would have missed

app/controllers/api/v1/search_controller.rb
```diff
@@ -612,6 +650,19 @@ def self.generate_bq_query_string(facets)
...
+          if !search_facet.is_numeric?
+            filter_terms += search_facet[:filters].map {|filter| filter[:name]}
```
devonbush: I think this can be rewritten as 
```suggestion
            filter_terms += search_facet.filters.map(&:name)
```

test/api/search_controller_test.rb
```diff
@@ -245,4 +245,38 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
+                 "Did not find expected accessions after inferred search, expected #{inferred_accessions} but found #{json['matching_accessions']}"
+    inferred_study = json['studies'].last # inferred matches should be at the end
+    assert inferred_study['inferred_match'],
+           "Did not mark last search results as inferred_match: #{inferred_study['inferred_match']} != true"
+
+    # reset description so other tests aren't broken
+    other_study.update(description: original_description)
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
+  end
```
devonbush: This is good.  Can you also add a test for 
1. Searching for multiple facets/filters behaves as expected
2. Searching for an exact phrase + facet behaves as expected.  
I expect both of these cases to work, but given that this is home-page behavior, an extra layer of testing would be good

app/controllers/api/v1/search_controller.rb
bistline: Actually it cannot.  The shorthand block notation does not work with hash keys (I wish it did).

test/api/search_controller_test.rb
bistline: Can do.

bistline: To be clear, we already have a test for the first case.  The existing test `'should return search results using facets'` uses multiple facets at the same time.  But I can add a test for the second case.

app/controllers/api/v1/search_controller.rb
bistline: I landed on `generate_mongo_query_by_context`, and changed the last parameter to `query_context`.

test/api/search_controller_test.rb
bistline: 70f04a9

app/controllers/api/v1/search_controller.rb
devonbush: nice

bistline:
> I don't know what our conventions/log setups are, but this feels more like a trace/debug log than INFO

I hear you, but to be perfectly honest the logic behind searches now is somewhat confusing.  The presence of keyword + facets + inferred search led to some corner cases that I didn't expect.  So those statements started out as debug, but then I decided they will be useful for addressing issues in production where we think we should be getting results but might not.

devonbush:
> I don't see the upside to that, really. We're trying to test the corner cases of inferred search, so I don't see the difference between extra assertions in a test that is already targeting that feature and extracting it out to its own test. I'd have to do the same setup that I did in this test.

the benefit is that when something breaks, you get a full picture of what's broken and still working, rather than only learning what the first failure was.  Concur that it's non-essential in this case

bistline:
> > I don't see the upside to that, really. We're trying to test the corner cases of inferred search, so I don't see the difference between extra assertions in a test that is already targeting that feature and extracting it out to its own test. I'd have to do the same setup that I did in this test.
> 
> the benefit is that when something breaks, you get a full picture of what's broken and still working, rather than only learning what the first failure was. Concur that it's non-essential in this case

Fair enough - I'll abstract it out to a separate test

PR #449, opened 2020-03-13
Bugfix for updating array-based facet filters

bistline:
Fixing a bug where querying for distinct IDs and names from array-based columns in BigQuery returned inconsistent results - sometimes IDs and names would line up, but not always, as this was done across two separate queries.  This has been updated to a single query which unnests both columns with offsets for each column, and specifically checks for both offsets to be equal before returning results.  This has an added benefit of simplifying the code or SearchFacet#get_unique_filter_values.

This PR satisfies SCP-2236.

### DIFF
--- a/app/models/search_facet.rb
+++ b/app/models/search_facet.rb
@@ -311,25 +311,14 @@ class SearchFacet
   # retrieve unique values from BigQuery and format an array of hashes with :name and :id values to populate :filters attribute
   def get_unique_filter_values
     Rails.logger.info "Updating filter values for SearchFacet: #{self.name} using id: #{self.big_query_id_column} and name: #{self.big_query_name_column}"
-    queries = []
-    if self.is_array_based
-      queries << self.generate_array_query(self.big_query_id_column, 'id')
-      queries << self.generate_array_query(self.big_query_name_column, 'name')
-    elsif self.is_numeric?
-      queries << self.generate_minmax_query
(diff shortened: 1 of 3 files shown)


app/models/search_facet.rb
```diff
@@ -349,13 +338,22 @@ def update_filter_values!
     end
   end
 
+  # return the correct query string for updating filter values from BQ based on facet type
+  def generate_query_string_by_type
```
devonbush: I would just call this method `generate_bq_query_string` since you're not passing in the type

app/models/search_facet.rb
```diff
@@ -349,13 +338,22 @@ def update_filter_values!
...
+  end
+
   # generate a single query to get DISTINCT values from an array-based column, preserving order
-  # this way we can do two queries, and stitch them together in a hash as the order will be the same
-  # e.g. IDs will line up with NAMEs
-  def generate_array_query(column_name, identifier)
-    "SELECT DISTINCT #{identifier} FROM(SELECT array_col AS #{identifier}, " + \
-    "FROM #{CellMetadatum::BIGQUERY_TABLE}, UNNEST(#{column_name}) AS array_col " + \
-    "WITH OFFSET AS offset ORDER BY offset)"
+  def generate_array_query
```
devonbush: love how much cleaner this query got!

PR #450, opened 2020-03-13
Include "Authorization: Bearer" via SCP API JS client (SCP-2231)

eweitz:
This ensures the `Authorization: Bearer` HTTP header is included in requests from the SCP API JS client when the user is signed in.  It enables faceted search results to include private studies.

I've manually verified this, in addition to the new automated test.

This satisfies SCP-2231.

### DIFF
--- a/app/javascript/components/UserProvider.js
+++ b/app/javascript/components/UserProvider.js
@@ -1,8 +1,8 @@
 import React, { useContext } from 'react'
 
 // window.SCP is not available when running via Jest tests,
-// so default such cases to a blank string
-export const accessToken = 'SCP' in window ? window.SCP.userAccessToken : ''
+// so default such cases to a string "test"
+export const accessToken = 'SCP' in window ? window.SCP.userAccessToken : 'test'
 
 const user = {
   accessToken
--- a/app/javascript/components/UserProvider.js
+++ b/app/javascript/components/UserProvider.js
@@ -2,7 +2,7 @@ import React, { useContext } from 'react'
 
 // window.SCP is not available when running via Jest tests,
 // so default such cases to a blank string
-const accessToken = 'SCP' in window ? window.SCP.userAccessToken : ''
+export const accessToken = 'SCP' in window ? window.SCP.userAccessToken : ''
 
 const user = {
   accessToken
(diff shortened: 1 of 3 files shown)


app/javascript/components/UserProvider.js
```diff
@@ -1,8 +1,8 @@
 import React, { useContext } from 'react'
 
 // window.SCP is not available when running via Jest tests,
-// so default such cases to a blank string
-const accessToken = 'SCP' in window ? window.SCP.userAccessToken : ''
+// so default such cases to a string "test"
+export const accessToken = 'SCP' in window ? window.SCP.userAccessToken : 'test'
```
devonbush:
```suggestion
export const accessToken = ('SCP' in window) ? window.SCP.userAccessToken : 'test'
```
just because I never like having to remember order-of-operations for uncommon operators

devonbush: Also, if this is using `'SCP' in window` as a proxy for 'am I executing in a test environment', should it just be checking the node environment instead?

test/js/scp-api.test.js
```diff
@@ -22,4 +23,38 @@ describe('JavaScript client for SCP REST API', () => {
...
+      expect.objectContaining({
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+          'Accept': 'application/json',
+          'Authorization': 'Bearer test'
+        }
+      })
+    )
+    process.nextTick(() => {
```
devonbush: I wonder if we'd be better using isolateModules for this type of thing.  I'm not at all used to test environments that don't automatically sandbox tests from each other though, so I don't know what best practices here are

app/javascript/components/UserProvider.js
eweitz: Seems reasonable -- I've opened SCP-2237 for this and related tech debt.

test/js/scp-api.test.js
eweitz: Thanks for the pointer!  I haven't used `isolateModules`, but it seems worthwhile.  Noted as something to consider in c8670a3.

PR #451, opened 2020-03-13
fixing numeric slider bugs and adding test

devonbush:
This satisfies SCP-2201 and SCP-2203.

This also changes the UI slightly to add the facet name in front of numeric filters since otherwise they may be ambiguous.

image

### DIFF
--- a/app/javascript/components/FiltersBox.js
+++ b/app/javascript/components/FiltersBox.js
@@ -4,7 +4,6 @@ import Button from 'react-bootstrap/lib/Button'
 
 import { StudySearchContext } from 'components/search/StudySearchProvider'
 import Filters from './Filters'
-import _remove from 'lodash/remove'
 
 /**
  * Component that can be clicked to unselect filters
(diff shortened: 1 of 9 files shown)


app/javascript/components/FiltersBox.js
```diff
@@ -108,9 +88,8 @@ export default function FiltersBox(props) {
       <Filters
         facet={props.facet}
         filters={props.filters}
-        updateSelectionForFilterCheckboxes={updateSelectionForFilterCheckboxes}
-        updateSelectionForFilterSlider={updateSelectionForFilterSlider}
         selection={selection}
+        setSelection={setSelection}
```
eweitz: Much cleaner!

app/models/search_facet.rb
```diff
@@ -283,6 +283,11 @@ def is_numeric?
     self.data_type == 'number'
   end
 
+  # for now, assume it's time if it's numeric and has a base of years
```
eweitz: Is `self.unit` necessarily a year here?  If not, this might be clearer:

```suggestion
  # for now, assume it's time if it's numeric and has a known time unit
```

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "scripts": {
     "test": "jest --coverage",
+    "testDebug": "node --inspect-brk node_modules/.bin/jest --runInBand",
```
eweitz: This will be helpful!

app/models/search_facet.rb
devonbush: good catch, I updated the code, then forgot the comment

PR #454, opened 2020-03-17
Analytics for advanced search (SCP-2192, SCP-2208)

eweitz:
This enables measuring advanced search usage when it's released.

It uses both Google Analytics and Mixpanel.  Mixpanel is more promising, but will be disabled on production until questions about event quota are resolved.  Google Analytics enables us to answer basic usage questions right away.

This satisfies SCP-2192 and SCP-2208.


app/javascript/lib/metrics-api.js
```diff
@@ -0,0 +1,277 @@
...
+  log('click:other', props)
+
+  // Google Analytics fallback: remove once Bard and Mixpanel are ready for SCP
+  ga('send', 'event', 'click', 'other') // eslint-disable-line no-undef
+}
+
+/**
+ * Count terms, i.e. space-delimited strings, and consider [""] to have 0 terms
+ */
+function getNumberOfTerms(terms) {
```
devonbush: I think this function and the search-related functions below belong in one of the search js files (maybe StudySearchProvider.js ?). I'd like to keep the metrics api free of specific business logic.  The metrics api should just export generic helper methods like 'log' and 'logClick', and then it's up to the components to know how to use them

app/javascript/lib/scp-api.js
```diff
@@ -31,11 +33,36 @@ if (
   })
 }
 
+/**
+ * Populates global id-to-name map of retrieved filters, for easier analytics.
+ * See downstream use of window.SCP.filterNamesById in metrics-api.js.
+ */
```
devonbush: To the extent we need such a map, it should be maintained in the SearchFacetContext in SearchFacetProvider.js or just the individual FacetControl that grabbed them.  We shouldn't be adding globals unless we need to access them in non-React parts of the app.

app/javascript/lib/metrics-api.js
```diff
@@ -0,0 +1,277 @@
...
+  ga( // eslint-disable-line no-undef
+    'send', 'event', 'faceted-search', 'search-filter',
+    'num-terms', numTerms
+  )
+}
+
+/**
+ * Log when a download is authorized.
+ * This is our best web-client-side methodology for measuring downloads.
+ */
```
devonbush: likewise, this function should live alongside the download code. (DownloadButton.js?)

app/javascript/lib/scp-api.js
```diff
@@ -119,34 +153,55 @@ export async function fetchFacetFilters(facet, query, mock=false) {
...
+
+  return await scpApi(path, defaultInit, mock)
 }
 
+/**
+ * Constructs query string used for /search REST API endpoint
+ */
 export function buildSearchQueryString(type, terms, facets, page) {
-  return `type=${type}&terms=${terms}&facets=${buildFacetQueryString(facets)}&page=${page ? page : 1}`
+  const facetsParam = buildFacetQueryString(facets)
```
devonbush: thanks for the refactor, much more readable :)

app/javascript/lib/scp-api.js
```diff
@@ -119,34 +153,55 @@ export async function fetchFacetFilters(facet, query, mock=false) {
...
+  const pageParam = page ? page : 1
+  return `type=${type}&terms=${terms}&facets=${facetsParam}&page=${pageParam}`
 }
 
+/**
+ * Serializes "facets" URL parameter for /search API endpoint
+ */
```
devonbush: I wouldn't mind a convention where simple one-line function comments like these don't use block-comment syntax, just so that we can see more of a class at once.  But I can see both sides

app/javascript/lib/metrics-api.js
eweitz: Good point!  Per our chat, we agree that this module is too high level for such SCP-API-specific metrics functions, and that React components are too low level.

So in 483b0701, I moved them to a new module, `scp-api-metrics.js`.

app/javascript/lib/metrics-api.js
eweitz: See [comment](https://github.com/broadinstitute/single_cell_portal_core/pull/454#discussion_r393842055).

app/javascript/lib/scp-api.js
eweitz: This global variable was a bad code smell indeed.  I've refactored that to not be global, and moved it and its surrounding function as described in my [other comment](https://github.com/broadinstitute/single_cell_portal_core/pull/454#discussion_r393842055).

app/javascript/lib/scp-api.js
eweitz: Agreed, if we can get:

1. IDEs to show such docstrings in imported references, and
2. The linter to not throw warnings for non-block docstrings while still warning about functions without docstrings

eweitz: Done in f7640f5!

PR #456, opened 2020-03-17
Ensuring consistency of search logic for inferred matches (SCP-2250)

bistline:
Faceted queries in the search API use AND logic across facets, and OR logic within a given facet (if multiple filters were supplied).  This update applies the same logic to the inferred search feature by running each facet query separately to find matching studies (by converting a filter display name into keywords/phrases) and then finding the intersection of studies at the end.  Also, this fixes a small bug in computing search facet weights where all filters present were being weighted, regardless of the match.

This PR satisfies SCP-2250.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -595,8 +595,8 @@ module Api
           # in order to maintain the same behavior as normal facets, we run each facet separately and get matching accessions
           # this gives us an array of arrays of matching accessions; now find the intersection (:&)
           filters = terms.values.map {|keywords| escape_terms_for_regex(term_list: keywords)}
-          accessions_by_filter = filters.map {|filter| base_studies.any_of({name: filter}, {description: filter}).
-              where(:accession.nin => accessions).pluck(:accession)}
+          accessions_by_filter = filters.map {|filter| base_studies.any_of({name: filter}, {description: filter})
+                                                           .where(:accession.nin => accessions).pluck(:accession)}
           base_studies.where(:accession.in => accessions_by_filter.inject(:&))
         else
           # no matching query case, so perform normal text-index search
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -213,17 +213,18 @@ module Api
         # if a user ran a faceted search, attempt to infer results by converting filter display values to keywords
         if @facets.any?
           # preserve existing search terms, if present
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -677,8 +683,10 @@ def self.match_studies_by_facet(query_results, search_facets)
             facet_name = key.to_s.chomp('_val')
             matching_filter = match_results_by_filter(search_result: result, result_key: key, facets: search_facets)
             matches[accession][facet_name] ||= []
-            matches[accession][facet_name] << matching_filter unless matches[accession][facet_name].include?(matching_filter)
-            matches[accession][:facet_search_weight] += 1
+            if !matches[accession][facet_name].include?(matching_filter)
+              matches[accession][facet_name] << matching_filter
+              matches[accession][:facet_search_weight] += 1
```
devonbush: If I read this correct, we only bump the weight if the facet hadn't already been accounted for.  But if a facet occurs in both the description *and* the metadata, maybe that should that give it extra weight?  I dunno

app/controllers/api/v1/search_controller.rb
```diff
@@ -591,8 +592,12 @@ def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, que
           study_regex = escape_terms_for_regex(term_list: terms)
           base_studies.any_of({name: study_regex}, {description: study_regex}, {:accession.in => accessions})
         when :inferred
-          filter_regex = escape_terms_for_regex(term_list: terms)
-          base_studies.any_of({name: filter_regex}, {description: filter_regex}).where(:accession.nin => accessions)
+          # in order to maintain the same behavior as normal facets, we run each facet separately and get matching accessions
+          # this gives us an array of arrays of matching accessions; now find the intersection (:&)
+          filters = terms.values.map {|keywords| escape_terms_for_regex(term_list: keywords)}
+          accessions_by_filter = filters.map {|filter| base_studies.any_of({name: filter}, {description: filter}).
```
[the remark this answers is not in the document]
bistline: Yes - you are right.  I forget that sometimes.

app/controllers/api/v1/search_controller.rb
bistline: This is only matching by the actual facet filter ID value, meaning the hits are coming from BigQuery.  So we only want to count the filter value once per study.  The inferred matches happen in a different place and count all instances of the value in both name and description.

PR #461, opened 2020-03-19
Highlight keywords

knapii-developments:
This PR highlights all keyword matches (including phrases) in the study description and title. In the event where the study is more than 750 characters (about 100 words), there will opening text (first 150 characters) followed by 100 characters preceding the matched keyword then the next 750 characters. 

Screen Shot 2020-03-19 at 10 15 34 AM

This PR satisfies [SCP-2189](https://broadworkbench.atlassian.net/browse/SCP-2189)

Testing will be covered in [SCP-2175](https://broadworkbench.atlassian.net/browse/SCP-2175)


app/javascript/components/PagingControl.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
```
devonbush: nice, good move pulling this into its own file

app/javascript/components/StudyResults.js
```diff
@@ -1,30 +1,46 @@
...
-  faAngleDoubleLeft, faAngleLeft, faAngleRight, faAngleDoubleRight
-} from '@fortawesome/free-solid-svg-icons'
 import Study from './Study'
-
+import PagingControl from './PagingControl'
 /**
  * Wrapper component for studies on homepage
  */
-const StudyResults = ({ results, changePage }) => {
+export default function StudyResultsContainer(props) {
```
devonbush: The default export by convention should be named the same as the file. 
FYI: we're getting rid of the tab container for now--we'll put it back when we implement the 'files' tab

app/javascript/components/Study.js
```diff
@@ -1,46 +1,90 @@
...
+    while ((match = regex.exec(text)) != null) {
+      matchedIndicies.push(match.index)
+    }
+    if (matchedIndicies.length>0) {
+      return { styledText: text.replace(regex, `<span id='highlight'>${terms}</span>`), matchedIndicies }
+    }
+  }
+  return { styledText: text, matchedIndicies }
+}
+
```
devonbush: rename shortenDescription to shortenHighlightDescription?

app/javascript/components/Study.js
```diff
@@ -1,46 +1,90 @@
...
-                                 .join(' ')
-
-  let suffixTag = <></>
-  if (numWords > descriptionWordLimit) {
-    suffixTag = <span className="detail"> ...(continued)</span>
+function highlightText(text, terms) {
+  const matchedIndicies = []
+  if (terms) {
+    let match
+    const regex = RegExp(terms, 'gi')
```
devonbush: I don't think this will do what you expect when there are multiple terms.  Passing an array ['a','b'] to RegExp will only return matches on the string 'a,b'

app/javascript/components/Study.js
```diff
@@ -1,57 +1,103 @@
...
-  const numWords = wordArray.length
-  // take the first 100 words, or the whole thing if shorter
-  const shortenedText = wordArray.slice(0, descriptionWordLimit - 1)
-                                 .join(' ')
-
-  let suffixTag = <></>
-  if (numWords > descriptionWordLimit) {
-    suffixTag = <span className="detail"> ...(continued)</span>
+function highlightText(text, terms) {
+  const matchedIndicies = []
```
devonbush: spelling: `matchedIndices`

app/javascript/components/Study.js
```diff
@@ -1,57 +1,103 @@
...
+  if (terms) {
+    let match
+    const regex = RegExp(terms, 'gi')
+    // Find indices where match occured
+    while ((match = regex.exec(text)) != null) {
+      matchedIndicies.push(match.index)
+    }
+    if (matchedIndicies.length>0) {
```
devonbush: the code would be simpler (although slightly less performant), if you dropped this length check, and just always returned { styledText: text.replace(regex, `<span id='highlight'>${terms}</span>`), matchedIndicies }. (even if there are no terms)

app/javascript/components/Study.js
```diff
@@ -1,57 +1,103 @@
...
-      <div>
-        <span id='cell-count' className='badge badge-secondary'>
-          {study.cell_count} Cells
-        </span>
+      <div key={study.accession}>
+        <label htmlFor={study.name} id= 'result-title'>
+          <a href={study.study_url} dangerouslySetInnerHTML = {displayStudyTitle}></a>{inferredBadge}
+        </label>
+        <div>
+          <span id='cell-count' className='badge badge-secondary'>
```
devonbush: there will be many spans like this, so we shouldn't use 'id' -- put 'cell-count' as a class

app/javascript/components/Study.js
knapii-developments: This won't work due to there being multiple terms. If there are no matches there's no need to to iterate over termMatches to replace text

PR #464, opened 2020-03-19
fixing possible infinite loop in bulk_download

devonbush:


app/models/study_file.rb
```diff
@@ -664,7 +664,7 @@ def bulk_download_pathname
   # Map of StudyFile#file_type to ::BULK_DOWNLOAD_TYPES, maintaining relationship for bundled files to parent
   def bulk_download_type
     # put bundled files in a sub-directory named after the bundle parent's ID so relationship is maintained
-    if self.is_bundled?
+    if self.is_bundled? && self.bundle_parent&.id != self.id
```
bistline: There's an easier way to check this: `if self.is_bundled? && !self.is_bundle_parent?`

However, the reason for this here is that if a file is "bundled", we create a subdirectory and put all the files that are bundled together into that directory.  So if we make this change, then the "parent" files won't get put in that subdirectory.

PR #468, opened 2020-03-20
adding facet match badges to search results

devonbush:
Satisfies SCP-2264 (also some incidental linting cleanup)

image

### DIFF
--- a/app/javascript/styles/_resultsPanel.scss
+++ b/app/javascript/styles/_resultsPanel.scss
@@ -68,7 +68,15 @@
     word-wrap: break-word;
   }
   .cell-count {
-    background-color: green;
+    background-color: #74ae43;
+    font-weight: normal;
+  }
+  .facet-match {
+    background-color: #fff;
+    margin-left: 0.5em;
+    color: #74ae43;
+    border: 1px solid #74ae43;
+    font-weight: normal;
   }
   &.inferred-match {
     border-left: 4px dotted #bbb;
@@ -84,6 +92,7 @@
       color: #888;
     }
   }
+
 }
 .pagination{
   text-align: center;
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -97,7 +97,7 @@ class User
   field :feature_flags, default: {}
 
   DEFAULT_FEATURE_FLAGS = {
-    "faceted_search" => false
+    "faceted_search" => true
   }
 
   ###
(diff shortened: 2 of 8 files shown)


test/js/search/study.test.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const facetUnmatchedStudy = {
+  name: 'Tuberculosis subtypes in human male blood',
+  description: 'stuff and things'
+}
+
+describe('Facet match badges', () => {
+  it('renders no badges with no matches', async () => {
+    const wrapper = mount((
```
knapii-developments: I don't know if mount is needed here. Mount does a full rendering including child components. In this case we're not testing child behavior so I don't believe mount is necessary.  I think render or shallow would be better. It would also increase the speed of the test because a full render isn't being performed.

app/models/user.rb
```diff
@@ -97,7 +97,7 @@ def owned_by(user)
   field :feature_flags, default: {}
 
   DEFAULT_FEATURE_FLAGS = {
-    "faceted_search" => false
+    "faceted_search" => true
```
eweitz: Blocker:

```suggestion
    "faceted_search" => false
```

app/javascript/styles/_resultsPanel.scss
```diff
@@ -68,7 +68,15 @@
...
-    background-color: green;
+    background-color: #74ae43;
+    font-weight: normal;
+  }
+  .facet-match {
+    background-color: #fff;
+    margin-left: 0.5em;
+    color: #74ae43;
+    border: 1px solid #74ae43;
+    font-weight: normal;
```
eweitz: Per [UI spec](https://projects.invisionapp.com/d/main#/console/19272801/402387755/preview):

```suggestion
    font-weight: normal;
    text-transform: uppercase;
```

eweitz:
Nit: to polish the UI and better align with the [spec](https://projects.invisionapp.com/d/main#/console/19272801/402387755/preview), consider more `padding` and a slightly higher `font-weight` in (I think) `secondary-badge`.

app/javascript/styles/_resultsPanel.scss
devonbush: I tried that -- it looks substantially worse.  For the uppercase to look good, you need the padding and font size to be different, and that was surprisingly non-trivial to change while still keeping text alignment

devonbush:
concur with Eric that updating the badge styling will get us closer to the spec -- it's a finicky task that touches all our badges, so currently beyond the scope of this ticket

PR #472, opened 2020-03-23
Nightly Admin Server Statistics Email (SCP-2119)

bistline:
Consolidating the two existing nightly emails (sanity check, disk usage) into a single email, along with following daily stats:

* Total users
* Active users that day
* New studies that day
* Analysis submissions from portal in the last 2 weeks (does not include workflows run directly in Terra)
* Ingest pipeline runs for that day
* Disk usage
* Storage sanity check for missing files (does not include failed uploads)

All these features have been refactored into the `SummaryStatsUtils` class, along with integration tests for all methods.  In addition, email delivery in the development environment has been turned back on (it was disabled due to email quotas in the past, but is no longer a concern with our new Sendgrid project).

This PR satisfies SCP-2119.


app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
+##
+# PortalUtils: generic class with server stats/maintenance methods
+#
+# note: all instances of :start_date and :cutoff_date are inclusive
+##
+class SummaryStatsUtils
+
+  # get a snapshot of user counts/activity up to a given date
+  # will give count of users as of that date, and number of active users on that date
+  def self.total_and_active_user_counts(cutoff_date: Date.today)
```
devonbush: maybe rename this to daily_total_and_active_user_counts(target_day)?

app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
...
+    # make sure to make cutoff_date one day forward to include any users that were created on cutoff date
+    next_day = cutoff_date + 1.day
+    total_users = User.where(:created_at.lte => next_day).count
+    active_users = User.where(:current_sign_in_at => (cutoff_date..next_day)).count
+    {total: total_users, active: active_users}
+  end
+
+  # get a count of all submissions launch from the portal in a given 2 week period
```
devonbush:
```suggestion
  # get a count of all submissions launch from the portal in a given time period
```

app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
...
+  # defaults to a time period of the last two weeks from right now
+  def self.analysis_submission_count(start_date: DateTime.now - 2.weeks, cutoff_date: DateTime.now)
+    AnalysisSubmission.where(:submitted_on => (start_date..cutoff_date), submitted_from_portal: true).count
+  end
+
+  # get a count of all studies created on the requested day
+  def self.study_creation_count(cutoff_date: Date.today)
```
devonbush: either rename to daily_study_creation_count or make it take a range like the analysis_submission_count method

app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
...
+  def self.disk_usage
+    disk = `df -h /home/app/webapp`.split("\n").last
+    details = disk.split
+    {total_space: details[1], space_used: details[2], space_free: details[3], percent_used: details[4], mount_point: details.last}
+  end
+
+  # find out all ingest jobs run in a given time period
+  # since the "filter" parameter for list_project_operations doesn't work, check dates manually.
+  # defaults to current day
+  def self.ingest_run_count(start_date: Date.today, cutoff_date: Date.today + 1.day)
```
devonbush: 'end_date' is a more common natural pair with 'start_date'

config/environments/development.rb
```diff
@@ -49,7 +49,7 @@
 
   config.action_mailer.default_url_options = { :host => 'localhost', protocol: 'https' }
   config.action_mailer.delivery_method = :smtp
-  config.action_mailer.perform_deliveries = false
+  config.action_mailer.perform_deliveries = true
```
devonbush: did you mean to leave this turned on?  We briefly chatted about installing something that would forward all dev. emails to the coder, but wanted to confirm that's happening here

bistline: That couldn't happen - we're connected to different schemas.  However, if you boot your local instance and connect to production, you could do that... this was mainly so I could test this feature, then I realized that the whole reason I turned it off before (quota impact with Sendgrid) doesn't apply anymore.

However, point of accidental misuse taken.  I'll turn this off until we implement the feature you talked about (it's easy enough to turn on/off locally).

app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
...
+        Rails.logger.error  "Error in retrieving remotes for #{study.name}: #{e.message}"
+        missing_files << {filename: 'N/A', study: study.name, owner: study.user.email, reason: "Error retrieving remotes: #{e.message}"}
+      end
+    end
+    missing_files
+  end
+
+  # disk usage stats
+  def self.disk_usage
+    disk = `df -h /home/app/webapp`.split("\n").last
```
eweitz: I recall historic security concerns with system calls in Rails.  Is the principle for this being OK that it accepts no user input, and is read only?

app/lib/summary_stats_utils.rb
```diff
@@ -0,0 +1,76 @@
+##
+# PortalUtils: generic class with server stats/maintenance methods
+#
+# note: all instances of :start_date and :cutoff_date are inclusive
+##
+class SummaryStatsUtils
+
+  # get a snapshot of user counts/activity up to a given date
+  # will give count of users as of that date, and number of active users on that date
+  def self.total_and_active_user_counts(cutoff_date: Time.zone.today)
```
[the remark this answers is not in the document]
bistline: Yeah that makes sense.  I had originally only had `cutoff_date` for these methods but then added `start_date` to some where appropriate and didn't refactor the name.

app/lib/summary_stats_utils.rb
bistline: To be honest, I'm not psyched about this, but I don't know of another way to do it in Ruby.  In theory the concern is warranted but acceptable given the two conditions you named.

I'm going to take a quick pass through the interwebs to see if there's a gem that can safely do this when I address the other PR concerns.

bistline:
Addressing all PR comments, and using `sys-filesystem` gem for safer access to disk usage statistics.

PR #473, opened 2020-03-23
Adding API integration for study external resources (SCP-2270)

bistline:
Adding in support to the API for CRUDing `ExternalResource` objects, which are links to publications or any other "resource" that a study owner wants to include on their study page.  Following the established paradigm in the API, rather than allowing nested objects (like the normal `studies_controller` does), a separate controller and collection of endpoints has been added for managing all external resources for a given study.  This also adds seeds and integration tests like other study-related models.

This PR satisfies SCP-2270, and a related user request for this feature.

### DIFF
--- a/app/controllers/api/v1/external_resources_controller.rb
+++ b/app/controllers/api/v1/external_resources_controller.rb
@@ -64,7 +64,7 @@ module Api
           key :tags, [
               'ExternalResources'
           ]
-          key :summary, 'Find a ExternalResource'
+          key :summary, 'Find an ExternalResource'
           key :description, 'Finds a single Study'
           key :operationId, 'study_external_resource_path'
           parameter do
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/external_resources_controller.rb
```diff
@@ -0,0 +1,312 @@
...
+      def index
+        @external_resources = @study.external_resources
+      end
+
+      swagger_path '/studies/{study_id}/external_resources/{id}' do
+        operation :get do
+          key :tags, [
+              'ExternalResources'
+          ]
+          key :summary, 'Find a ExternalResource'
```
eweitz:
```suggestion
          key :summary, 'Find an ExternalResource'
```

And similarly for other instances of "a ExternalResource".

PR #474, opened 2020-03-24
refining behavior of faceted search for numeric facets

devonbush:
Jean and I spent awhile talking about this this morning, and this was our consensus for the best way to handle numeric facets for inferred search.  Otherwise, selecting a numeric range along with other facets means that studies that match the other facets, but have no numeric data at all, will be treated as 'inferred' matches, but there's nothing behind that inference.


eweitz:

PR #475, opened 2020-03-24
adding error boundaries to search page

devonbush:
See https://reactjs.org/docs/error-boundaries.html for the background.  Basically, if there's an error in component rendering (although this does NOT handle errors in event handlers), this will now show a nice error instead of the entire React UI disappearing

image

### DIFF
--- a/app/javascript/components/HomePageContent.js
+++ b/app/javascript/components/HomePageContent.js
@@ -4,19 +4,26 @@ import ResultsPanel from './ResultsPanel'
 import StudySearchProvider from 'components/search/StudySearchProvider'
 import SearchFacetProvider from 'components/search/SearchFacetProvider'
 import UserProvider from 'components/UserProvider'
+import ErrorBoundary from 'lib/ErrorBoundary'
 
 /**
  * Wrapper component for search and result panels
  */
 export default function HomePageContent() {
   return (
-    <UserProvider>
-      <SearchFacetProvider>
-        <StudySearchProvider>
-          <SearchPanel/>
-          <ResultsPanel/>
-        </StudySearchProvider>
-      </SearchFacetProvider>
-    </UserProvider>
+    <ErrorBoundary>
+      <UserProvider>
+        <SearchFacetProvider>
+          <StudySearchProvider>
+            <ErrorBoundary>
+              <SearchPanel/>
+            </ErrorBoundary>
+            <ErrorBoundary>
+              <ResultsPanel/>
+            </ErrorBoundary>
+          </StudySearchProvider>
+        </SearchFacetProvider>
+      </UserProvider>
+    </ErrorBoundary>
   )
 }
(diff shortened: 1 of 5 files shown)


app/javascript/lib/ErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // the error.stack is typically useless because EventBoundaries do NOT catch
+  // errors in event handlers, so the error.stack is just full of react internals
+  return error.message + info.componentStack.split('\n')[1]
+}
+
+/*
+ * See https://reactjs.org/docs/error-boundaries.html
+ * note that this must be a class component
+ * as hooks do not support componentDidCatch yet
```
eweitz: Helpful note!

app/javascript/components/HomePageContent.js
```diff
@@ -4,19 +4,26 @@ import ResultsPanel from './ResultsPanel'
...
   return (
-    <UserProvider>
-      <SearchFacetProvider>
-        <StudySearchProvider>
-          <SearchPanel/>
-          <ResultsPanel/>
-        </StudySearchProvider>
-      </SearchFacetProvider>
-    </UserProvider>
+    <ErrorBoundary>
```
eweitz: Needing to wrap each rendered component in `ErrorBoundary` seems tedious and error-prone -- I could see it being overlooked.

Could this boilerplate go away only if/when hooks support `componentDidCatch`, or could it be refactored beforehand?

app/javascript/lib/ErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+
+  componentDidCatch(error, info) {
+
+    logError(readableErrorMessage(error, info))
+    this.setState({ error, info })
+  }
+
+  render() {
+    if (this.state.error) {
+      // consider using node_env to decide whether or not to render the full trace
```
eweitz: Good idea.  More context might help future readers:

```suggestion
      // consider using node_env to decide whether or not to render the full trace
      // See related ticket SCP-2237
```

eweitz:

app/javascript/components/HomePageContent.js
devonbush: so this is an interesting question that I'm not sure I know the best path.  I've been operating under the assumption that it's not worth wrapping every component in an error boundary, but rather we should only wrap functional areas.  But I could see either approach being correct.  I've got a higher-order component to help with boilerplate, but you're right that a hook would be easier

app/javascript/lib/ErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+ * note that this must be a class component
+ * as hooks do not support componentDidCatch yet
+ */
+export default class ErrorBoundary extends Component {
+  constructor(props) {
+    super(props)
+    this.state = { error: null }
+  }
+
+  componentDidCatch(error, info) {
```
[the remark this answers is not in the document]
devonbush: great question, as far as I can tell, yes. The error still goes to the javascript console and I don't see any difference in what goes out to new relic in my network tab.  Definitely something I'll want to double-check in staging

PR #476, opened 2020-03-24
adding facet badges for numeric filters

devonbush:
per discussion post-standup

image

### DIFF
--- a/app/javascript/components/Study.js
+++ b/app/javascript/components/Study.js
@@ -93,7 +93,15 @@ function facetMatchBadges(study) {
           className="badge badge-secondary facet-match"
           data-toggle="tooltip"
           title={helpText}>
-          { matches[key].map(filter => filter.name).join(',') }
+          {
+            matches[key].map(filter => {
+              if (filter.min) { // numeric facet
+                return `${key} ${filter.min}-${filter.max} ${filter.unit}`
+              } else {
+                return filter.name
+              }
+            }).join(',')
+          }
         </span>)
     })}
   </>)
--- a/app/javascript/components/Study.js
+++ b/app/javascript/components/Study.js
@@ -1,6 +1,6 @@
 /* eslint-disable require-jsdoc */
 import React from 'react'
-
+import { getDisplayNameForFacet } from 'components/search/SearchFacetProvider'
 const descriptionWordLimit = 750
 const summaryWordLimit = 150
 const lengthOfHighlightTag = 21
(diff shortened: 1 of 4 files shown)


test/js/search/study.test.js
```diff
@@ -54,4 +64,12 @@ describe('Facet match badges', () => {
     expect(wrapper.find('.facet-match').first().text().trim()).toEqual('blood')
     expect(wrapper.find('.facet-match').last().text().trim()).toEqual('mouse,human')
   })
+
+  it('renders badges for numeric facets', async () => {
+    const wrapper = mount((
+      <Study study={numericFacetMatchedStudy}/>
+    ))
+    expect(wrapper.find('.facet-match').length).toEqual(1)
+    expect(wrapper.find('.facet-match').first().text().trim()).toEqual('organism_age 30-50 years')
```
eweitz:
```suggestion
    expect(wrapper.find('.facet-match').first().text().trim()).toEqual('Organism Age 30-50 years')
```

app/javascript/components/Study.js
```diff
@@ -93,7 +93,15 @@ function facetMatchBadges(study) {
           className="badge badge-secondary facet-match"
           data-toggle="tooltip"
           title={helpText}>
-          { matches[key].map(filter => filter.name).join(',') }
+          {
+            matches[key].map(filter => {
+              if (filter.min) { // numeric facet
+                return `${key} ${filter.min}-${filter.max} ${filter.unit}`
```
eweitz: The screenshot shows "organism_age: 1-180 years".  Instead of showing the ID for such matches, we should its name, e.g. "Organism Age: 1-180 years":

```suggestion
                return `${filter.name} ${filter.min}-${filter.max} ${filter.unit}`
```

devonbush: yeah, the backend doesn't return the display name in these matches, so this was me going to quick-and-dirty route.  I think you're right though that organism_age is too sloppy.  So I'll get the display name via other means

bistline: The `name` attribute is source from Mongo, so if the value you're getting back is not what we want, we should change it in the database.

bistline:
Also, the failed build is a red herring - on rare occasion I've seen that particular test fail (api/v1/study_files_controller) due to a mismatch in an expected file size.  I think this might have to do with a race condition with gzipping the file for storage in the bucket.  In any event, it's not a regression, just a slightly flaky test.

app/javascript/components/Study.js
devonbush: the backend only returns the 'name' for filters, not for facets.  And the numeric slider badge is different from the others in that we need the name of the facet, not the filter

PR #484, opened 2020-03-25
Adding Preset Searches

bistline:
Adding the `PresetSearch` model and associated functionality for running "canned" queries that include keyword/facet parameters, as well as an `accession_whitelist` which always return any declared studies at the top of the results.  This also includes an administrative UI for CRUDing searches to allow for easier and faster configuration.  Preset searches are invoked in the search API by passing a single identifier, which then loads any keywords/facets/whitelisted studies and then runs a normal search.

Whitelist-only searches (e.g. no search terms or facets provided) will return the whitelisted studies only.

This PR satisfies SCP-2264.

### DIFF
--- a/app/views/preset_searches/show.html.erb
+++ b/app/views/preset_searches/show.html.erb
@@ -1,0 +1,49 @@
+<h1><%= @preset_search.name %></h1>
+
+<dl class="dl-horizontal">
+  <dt>Identifier</dt>
+  <dd><%= @preset_search.identifier %></dd>
+  <dt>Whitelist</dt>
+  <dd><%= @preset_search.accession_whitelist.join(', ') %></dd>
+  <dt>Search Terms</dt>
+  <dd><%= @preset_search.search_terms.join(', ') %></dd>
+  <dt>Facets</dt>
+  <dd>
+    <% @preset_search.matching_facets_and_filters.each do |facet| %>
+      <%= facet_query_label(facet).html_safe %>&nbsp;
+    <% end %>
+  </dd>
+  <dt>Public?</dt>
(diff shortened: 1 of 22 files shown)


app/views/preset_searches/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+  <%= scp_link_to "<span class='fas fa-chevron-left'></span> Back".html_safe, preset_searches_path, class: 'btn btn-warning' %>
+</p>
+
+
+<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+   var baseUrl = "<%= site_url %>/study/";
+   var resultsTarget = $('#preset-search-results')
+   $(document).ready(function() {
+       $.ajax({
+           url: '<%= javascript_safe_url(api_v1_search_path(type: 'study', preset_search: @preset_search.identifier)) %>',
```
eweitz: My first thought was to [disentangle Rails and JavaScript](https://github.com/broadinstitute/single_cell_portal_core/pull/396#discussion_r380709899) here.  But given our decision to indefinitely defer migrating admin pages like this to React, I'm not sure how to handle such Rails-JS commingling in admin pages.  Downside: somewhat hardens two paradigms for development. Upside: gets stuff done quickly.

I'm inclined to consider such JS-Rails commingling in admin pages to be low-priority tech debt, and fine for the foreseeable future.

bistline: This is technically in line with what we had previously discussed before - any user-facing, high-traffic views/actions are candidates for migrating to React in the somewhat near future.  Anything that is admin-only, or would be seldom accessed by users is fine to keep in the standard Rails UI paradigm as the impact of the migration is outweighed by the effort required to do so.

PR #487, opened 2020-03-27
Clear all button for search panel

knapii-developments:
This PR adds a 'Clear All' button to the results panel.  The button is always displayed. The clear all button removes all search/ed query terms ( keyword and facets) and updates the result panel to the default state. 

Screen Shot 2020-03-27 at 2 06 30 PM

This PR satisfies [SCP-2220](https://broadworkbench.atlassian.net/browse/SCP-2220)

### DIFF
--- a/app/javascript/components/SearchQueryDisplay.js
+++ b/app/javascript/components/SearchQueryDisplay.js
@@ -1,18 +1,21 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import { faSearch } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { getDisplayNameForFacet } from 'components/search/SearchFacetProvider'
+import Button from 'react-bootstrap/lib/Button'
+import { SearchSelectionContext } from './search/SearchSelectionProvider'
+import { StudySearchContext } from 'components/search/StudySearchProvider'
 
 
 function formattedJoinedList(itemTexts, itemClass, joinText) {
   return itemTexts.map((text, index) => {
-      return (
-        <span key={index}>
-          <span className={itemClass}>{text}</span>
-          { (index != itemTexts.length - 1) &&
+    return (
+      <span key={index}>
+        <span className={itemClass}>{text}</span>
+        { (index != itemTexts.length - 1) &&
             <span className="join-text">{joinText}</span>}
-        </span>
-      )
+      </span>
+    )
   })
 }
 
@@ -20,8 +23,8 @@ function formatFacet(facet, index, numFacets) {
   let facetContent
   if (Array.isArray(facet.filters)) {
     facetContent = formattedJoinedList(facet.filters.map(filter => filter.name),
-                                       'filter-name',
-                                       ' OR ')
+      'filter-name',
+      ' OR ')
   } else { // it's a numeric facet
     facetContent = (<span className="filter-name">
       {facet.filters.min} - {facet.filters.max} {facet.filters.unit ? facet.filters.unit : '' }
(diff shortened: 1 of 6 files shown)


app/javascript/components/SearchQueryDisplay.js
```diff
@@ -36,19 +39,36 @@ function formatFacet(facet, index, numFacets) {
...
+  const selectionContext = useContext(SearchSelectionContext)
+  const searchContext = useContext(StudySearchContext)
 
-export default function SearchQueryDisplay({terms, facets}) {
+  const clearSearch = () => {
+    const defaultSearchParams = {
+      terms: '',
+      facets: {}
+    }
+    selectionContext.performSearch()
```
devonbush: this first `performSearch()` is unnecessary -- it's just executing a search with the previously selected params, which is a no-op

app/javascript/components/SearchQueryDisplay.js
```diff
@@ -36,19 +39,36 @@ function formatFacet(facet, index, numFacets) {
...
+    searchContext.updateSearch(defaultSearchParams)
+    selectionContext.updateSelection(defaultSearchParams)
```
devonbush: Sorry, I don't think I communicated this well when we were talking.  I want you to edit `SearchSelectionProvider.updateSelection` to take a `searchNow` argument, just like `updateFacet` does.  Then instead of having to have order-sensitive double-update-calls like this in each component, it can be handled in the selection provider.  Then these two lines will just become `selectionContext.updateSelection(defaultSearchParams, true)`

app/javascript/styles/_resultsPanel.scss
```diff
@@ -1,6 +1,16 @@
+@import './_colors.scss';
+
 .results-panel {
   padding-top: 14px;
 
+  button{
+    color:$action-color;
+    background-color: $default-background-color;
+    border: none;
+  }
```
devonbush:
```suggestion
  button {
    color: $action-color;
    background-color: none;
    border: none;
  }
```

app/javascript/styles/_resultsPanel.scss
```diff
@@ -1,6 +1,16 @@
...
+  button:hover{
```
devonbush:
```suggestion
  button:hover {
```

test/js/search/search-query-display.test.js
```diff
@@ -1,61 +1,79 @@
...
-    expect(wrapper.text().trim()).toEqual(': Metadata contains (species: Homo sapiens)')
+    expect(wrapper.text().trim()).toEqual(': Metadata contains (species: Homo sapiens) Clear All')
   })
 
   it('renders multiple facets', async () => {
     const wrapper = mount((
       <SearchQueryDisplay facets={twoStringFacets} terms={''}/>
     ))
-    expect(wrapper.text().trim()).toEqual(': Metadata contains (disease: disease1) AND (species: Homo sapiens)')
+    expect(wrapper.text().trim()).toEqual(': Metadata contains (disease: disease1) AND (species: Homo sapiens) Clear All')
```
devonbush: this makes a good case for either adding a prop 'showClearAll'  to SearchQueryDisplay that can be used to hide the button so that you can isolate the components for testing, or rendering the button from ResultsPanel.js instead

test/js/search/search-query-display.test.js
```diff
@@ -1,61 +1,79 @@
...
+})
+
+describe('Clearing search query', () => {
+  it('clears search params', () => {
+    const component = <PropsStudySearchProvider searchParams={{ terms: 'foo' }}>
+      <ClearAllButton/>
+      <KeywordSearch/>
+    </PropsStudySearchProvider>
+    const wrapper = mount(component)
+    expect(wrapper.find('input[name="keywordText"]').first().props().value).toEqual('foo')
```
devonbush: Do we need to test that this can also clear facets?

test/js/search/search-query-display.test.js
knapii-developments: I'm not sure how you would test this seeing that the button always appears when the SearchQuery is mounted.

app/javascript/components/SearchQueryDisplay.js
devonbush: yeah, like I said, I don't think we were understanding each other well yesterday. :-).  it wasn't a *necessary* change to get this working, but it'll make future components much better.

test/js/search/search-query-display.test.js
devonbush: Exactly, if you have showClearAll as a prop, then you can set it to false in the test, and not have to worry about exactly how the button renders when we're testing other things.  Alternatively, you can wrap the text part of the display in a separate span, so you can just get that text

PR #493, opened 2020-03-30
Refine user ID, page view tracking; COVID-19 analytics (SCP-2287, SCP-2282)

eweitz:
This fixes long-standing bugs in user ID and page view tracking, and also enhances analytics to ease measurements of COVID-19 space impact.

This satisfies SCP-2287 and SCP-2282.  See those tickets for more context.

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -35,21 +35,21 @@ if ('SCP' in window) {
 /**
  * Log page view, i.e. page load
  */
-export function logPageView(props={}) {
-  log('page:view', props)
+export function logPageView() {
+  log('page:view')
 }
 
 /** Log click on page.  Delegates to more element-specific loggers. */
-export function logClick(event, props={}) {
+export function logClick(event) {
   const target = event.target
   const tag = target.localName.toLowerCase() // local tag name
 
   if (tag === 'a') {
-    logClickLink(target, props)
+    logClickLink(target)
   } else if (tag === 'button') {
-    logClickButton(target, props)
+    logClickButton(target)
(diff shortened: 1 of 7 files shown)


app/views/layouts/application.html.erb
```diff
@@ -30,13 +33,27 @@
       window.SCP.userAccessToken = '<%= get_user_access_token(current_user) %>';
       window.SCP.environment = '<%= Rails.env %>';
 
+      window.SCP.hasLoggedPageView = false;
+
+      <% if controller_name == 'site' && action_name == 'covid19' %>
+      window.SCP.featuredSpace = 'covid19'
+      <% end %>
```
devonbush: this probably has to be here for loading timing reasons, but if possible, it would be great to relocate it to a script on `covid19.html.erb`, just so as much covid-specific stuff is kept in one place as possible

eweitz: Worth trying.  I just did, and see the current location is indeed required for timing reasons.

We manually log `page:view` to Mixpanel, and without this variable available before our React code loads, the `featuredSpace: covid19` event property is omitted in Mixpanel log posts.  So the current location stays for now.

app/javascript/lib/metrics-api.js
```diff
@@ -161,6 +161,11 @@ export function log(name, props={}) {
     env
   })
 
+  if ('SCP' in window && 'featuredSpace' in window.SCP) {
+    // For e.g. COVID-19 featured space
+    props['featureSpace'] = window.SCP.featuredSpace
```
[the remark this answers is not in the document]
eweitz: Good catch, fixed in 0b14047.

PR #508, opened 2020-04-03
Reflect server-side errors

knapii-developments:
500 errors are now caught in scp-api and can reflected in the result panel. However, this PR does not include handling of any server-side errors for facets. 
Screen Shot 2020-04-09 at 11 10 08 AM

This PR satisfies [SCP-2161](https://broadworkbench.atlassian.net/browse/SCP-2161)


test/js/scp-api.test.js
```diff
@@ -1,17 +1,20 @@
-/* eslint-disable no-undef */
+/* eslint-disable*/
```
eweitz, line 1: Could you add a code comment briefly explaining why the linter is completely disabled in this file?

I suspect this is needed to avoid VS Code from buggily auto-inserting `Promise`, as we saw while pair-programming today.  But that should be noted in code, to avoid maintainers reflexively re-adding the linter.

If that buggy auto-inserting happens in more than this one file, then we should open an SCP ticket to find a better solution than disabling the linter.

test/js/scp-api.test.js
```diff
@@ -1,17 +1,20 @@
...
 } from 'lib/scp-api'
 
 describe('JavaScript client for SCP REST API', () => {
   beforeAll(() => {
     global.fetch = fetch
   })
+  // Note: tests that mock global.fetch must be cleared after every test
+  afterEach(() => {
+    // Restores all mocks back to their original value
+    jest.restoreAllMocks()
```
eweitz, line 16: Kudos on getting this to work! :tada:

test/js/scp-api.test.js
```diff
@@ -55,8 +51,28 @@ describe('JavaScript client for SCP REST API', () => {
...
+      ok: false,
+      statusText: 'Internal Server Error'
+    }
+    jest.spyOn(global, 'fetch').mockReturnValue(Promise.resolve(mockErrorResponse))
+    const actualResponse = await scpApi('/test/path', {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json'
+      }
```
eweitz, line 73: Does this test still pass if you omit these lines?

```suggestion
```

I believe these fetch `init` properties are automatically supplied by [`defaultInit` in `scp-api.js`](https://github.com/broadinstitute/single_cell_portal_core/blob/master/app/javascript/lib/scp-api.js#L23), but I could be missing something.

app/javascript/providers/StudySearchProvider.js
```diff
@@ -104,7 +104,7 @@ export function PropsStudySearchProvider(props) {
 
     setSearchState({
       params,
-      isError: false,
+      isError: results.ok? !results.ok: false,
```
eweitz, line 107: Relatedly:

A component-level test that verifies the "Sorry, an error has occurred" error message appears in the rendered ResultsPanel UI when `isError` is `true` would prevent this bug from (re-)emerging.

I don't consider such a new test blocking, but we should at least document it as a "TODO" or "To consider" in the code.

app/javascript/lib/scp-api.js
```diff
@@ -260,10 +260,22 @@ export default async function scpApi(path, init, mock=false) {
...
   const response = await fetch(fullPath, init)
-  const json = await response.json()
-
-  // Converts API's snake_case to JS-preferrable camelCase,
-  // for easy destructuring assignment.
-  return camelcaseKeys(json)
+    .then(handleErrors)
+    .catch(error => error)
+  if (response.ok) {
+    const json = await response.json()
```
devonbush, line 275: I would also put error handling around this response.json call.  A common failure case for rails APIs is an error in a part of a controller, that due to incomplete configuration, then returns an html error page rather than json.  In which case I believe this call will fail with an 'unexpected character, failed to parse' error

app/controllers/api/v1/search_controller.rb
```diff
@@ -153,6 +153,7 @@ class SearchController < ApiBaseController
       end
 
       def index
+        render json:{error: 'blah'}, status:500
```
eweitz, line 156: Bug -- it looks like some development code slipped through.

test/js/scp-api.test.js
```diff
@@ -1,17 +1,21 @@
-/* eslint-disable no-undef */
+// Without disabling eslint code is Promises are auto inserted
```
eweitz, line 1:
```suggestion
// Without disabling eslint code, Promises are auto inserted
```

app/javascript/providers/StudySearchProvider.js
```diff
@@ -102,9 +102,10 @@ export function PropsStudySearchProvider(props) {
 
     const results = await fetchSearch('study', params)
 
+
     setSearchState({
       params,
-      isError: false,
+      isError: (results.ok || (results.ok == undefined))? false: !results.ok,
```
eweitz, line 108: At a glance I'd lean towards `!results.hasOwnProperty('ok')` over `results.ok == undefined`, but I can't actually think of a concrete case where we'd encounter this condition, and my naive check shows your updated code works as is.  So I consider the blocker resolved.

A minor non-blocking suggestion for conventional spacing around the ternary operators:
```suggestion
      isError: (results.ok || (results.ok == undefined)) ? false : !results.ok,
```

app/javascript/components/search/results/ResultsPanel.js
```diff
@@ -15,38 +14,50 @@ const ResultsPanel = props => {
...
-        Please try again.
+    panelContent = (
+      <div className="error-panel  col-md-6 col-md-offset-3">
+        <FontAwesomeIcon
+          icon={faExclamationCircle}
+          style={{ 'margin-right': '5px' }}
+        />
+        Sorry, an error has occurred. Support has been notified. Please try
+        again. If this error persists, or you require assistance, please contact
+        support at{' '}
```
devonbush, line 25: if you just need an extra space to render, use `&nbsp;` or `&#32;` depending on whether you want line breaks to be a possibility on that space

app/javascript/components/search/results/ResultsPanel.js
```diff
@@ -15,38 +14,50 @@ const ResultsPanel = props => {
...
-    panelContent =
-      <div className="error-panel">
-        <FontAwesomeIcon icon={faExclamationCircle}/>
-        Sorry, an error has occurred.  Support has been notified.
-        Please try again.
+    panelContent = (
+      <div className="error-panel  col-md-6 col-md-offset-3">
+        <FontAwesomeIcon
+          icon={faExclamationCircle}
+          style={{ 'margin-right': '5px' }}
```
devonbush, line 21: For now, we're trying to avoid inline CSS where possible.  Let's move this to a css file.  give this a className like "left-margin-icon" or something, because I suspect this will be a generally useful class.  Also, make the margin in ems rather than px so that it will scale with font resizes

app/javascript/components/search/results/ResultsPanel.js
```diff
@@ -15,38 +14,50 @@ const ResultsPanel = props => {
...
+        />
+        Sorry, an error has occurred. Support has been notified. Please try
+        again. If this error persists, or you require assistance, please contact
+        support at{' '}
+        <a href="mailto:scp-support@broadinstitute.zendesk.com">
+          scp-support@broadinstitute.zendesk.com
+        </a>
```
devonbush, line 28: nice, I think this is the first time we've included the support email in a React page.  The second time, we'll probably want to move this into a lib component

app/javascript/providers/StudySearchProvider.js
```diff
@@ -104,7 +107,7 @@ export function PropsStudySearchProvider(props) {
 
     setSearchState({
       params,
-      isError: false,
+      isError: results.ok || results.ok == undefined ? false : !results.ok,
```
devonbush, line 110: if you're trying to set isError to `false` when results.ok is either false or undefined,    you can just use `!!results.ok`.  If you're doing something else, please add a comment so it's easier to read (and probably this is too complex to put in a one-liner)

PR #509, opened 2020-04-03
UI, parser, and test bugfixes (SCP-2306, SCP-2311, SCP-2312)

bistline:
Addressing several small bugs:

- Harmonizing expression axis labels across all gene expression distribution plots (SCP-2306)
- Correctly comparing dates across timezones in `SummaryStatsUtils#ingest_run_count` (SCP-2311)
- Stripping quotes off of dense expression matrix headers during format validation (SCP-2312)
- Empty search API requests use correct order for results


app/views/site/_expression_plots_plotly.html.erb
```diff
@@ -164,11 +165,11 @@ var expressionData = [];
     <% if @expression[:all][:x].length >= ClusterGroup::MAX_THRESHOLD %>
     title += ' (subsampled to <%= ClusterGroup::MAX_THRESHOLD %> cells)';
     <% end %>
-    var data = createTracesAndLayout(formatted_array, title, jitter_var);
+    var data = createTracesAndLayout(formatted_array, title, jitter_var, expressionLabel);
```
eweitz, line 168: Required parameters before optional, per convention:

```suggestion
    var data = createTracesAndLayout(formatted_array, title, expressionLabel, jitter_var);
```

This also reveals that we should use camel case in JavaScript wherever we can; the less snake case in JS the better.  I don't think that refactoring must happen now, though.

PR #512, opened 2020-04-07
harmonizing API "timeouts" with devise User.timeout_in threshold (SCP-2304)

bistline:
Synchronizing API "session" length with user session timeouts in the UI.  Currently, API requests only validate that the OAuth token is within the standard 1 hour window.  This now adds a secondary check to validate that the user has used this access token within the configured timeout window (currently 30 minutes, as set in `config/initializers/devise.rb`, and accessible via `User.timeout_in`).  As functionality migrates from the standard Rails UI to API calls + React, this will be required in order to maintain application security guidelines of logging out "idle" users as per FISMA Moderate.

This PR satisfies SCP-2304.

### DIFF
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -27,10 +27,12 @@ module Api
                 token_url = "https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=#{api_access_token}"
                 response = RestClient.get token_url
                 credentials = JSON.parse response.body
+                now = Time.zone.now
                 token_values = {
                     'access_token' => api_access_token,
                     'expires_in' => credentials['expires_in'],
-                    'expires_at' => Time.zone.now + credentials['expires_in'].to_i
+                    'expires_at' => now + credentials['expires_in'].to_i,
+                    'last_access_at' => now
                 }
                 email = credentials['email']
                 user = User.find_by(email: email)
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -50,15 +50,19 @@ module Api
                 Rails.logger.error "Error retrieving user api credentials: #{e.class.name}: #{e.message}"
               end
             end
-            # check for token expiry and unset user if expired
-            if user.api_access_token_expired?
+            # check for token expiry and unset user if expired/timed out
+            if user.api_access_token_expired? || user.api_access_token_timed_out?
               nil
             else
+              # update last_access_at
+              user.update_api_last_access_at!
               user
             end
           elsif controller_name == 'search' && action_name == 'bulk_download'
             Rails.logger.info "Authenticating user via auth_token: #{params[:auth_code]}"
-            User.find_by(totat: params[:auth_code].to_i)
+            user = User.find_by(totat: params[:auth_code].to_i)
+            user.update_api_last_access_at!
+            user
           end
         end
 
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -55,13 +55,13 @@ module Api
               nil
             else
               # update last_access_at
-              user.update_api_last_access_at!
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -36,6 +36,9 @@ def self.big_query_client
   around_action :set_current_user
   def set_current_user
     Current.user = current_user
+    if current_user.present?
+      current_user.update_api_last_access_at!
```
devonbush, line 40: since we're updating this on any calls, not just api calls, would a better name be just `update_last_access_at!` ?

app/models/user.rb
```diff
@@ -173,14 +179,34 @@ def api_access_token_expired?
...
+
   # return an valid access token (will renew if expired) - does not apply to api access tokens, those cannot be renewed
   def valid_access_token
     self.access_token_expired? ? self.generate_access_token : self.access_token
   end
 
   # extract timezone from an access token to allow correct date math
-  def get_token_timezone(token_method)
-    self.send(token_method)[:expires_at].zone
+  def get_token_timezone(token_method, timestamp=:expires_at)
```
devonbush, line 208: I'm not sure I'm following why you have to accept that second argument -- is it possible that the two parts of the token have different timezones? Or one of them is nil?

app/controllers/application_controller.rb
bistline: I thought about that, but the `last_access_at` is something stored in the Warden/Devise session client-side, and I didn't want there to be any confusion.  But I take the point - most people probably wouldn't note the distinction.

app/models/user.rb
bistline: It's almost always `UTC`, but I noticed in some tests that sometimes it was getting set to `EDT`, so I just wanted to make sure that whatever timestamp we're looking at, we use _that_ timestamp's zone.  I wasn't sure if this is an actual usage artifact, or just me doing something strange in a test.  But I agree - it _should_ always be `UTC` since that's what we get back from OAuth.

bistline: This actually broke almost all the tests, so I'm taking this out - if that timestamp doesn't exist it errors out, but the `expires_at` will always be there.  And you're right - there shouldn't be a disagreement between the two, and taking this out will actually enforce that.

bistline: 872ac64

app/controllers/application_controller.rb
bistline: 5d2bcbc

app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -50,15 +50,19 @@ def current_api_user
                 Rails.logger.error "Error retrieving user api credentials: #{e.class.name}: #{e.message}"
               end
             end
-            # check for token expiry and unset user if expired
-            if user.api_access_token_expired?
+            # check for token expiry and unset user if expired/timed out
+            if user.api_access_token_expired? || user.api_access_token_timed_out?
```
eweitz, line 54: The difference between "expired" and "timed out" is unclear without looking up the docstring for `api_access_token_timed_out`.  

Naming the method `api_access_token_idle_timed_out` (or something to that effect) would make these easier to distinguish at a glance.

test/api/search_controller_test.rb
```diff
@@ -383,4 +383,36 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
+    api_test_study = Study.find_by(name: /API Test Study/)
+    api_test_study.update(public: false)
+    search_terms = "\"API Test Study\""
+    execute_http_request(:get, api_v1_search_path(type: 'study', terms: search_terms))
+    assert_response :success
+    expected_results = [api_test_study.accession]
+    assert_equal expected_results, json['matching_accessions'],
+                 "Found wrong study with keywords search; expected #{expected_results} but found #{json['matching_accessions']}"
+    # now simulate a user "timing out" by back-dating last_access_at timestamp by double the timeout threshold
+    # and then re-run search
```
eweitz, line 403: Clever -- and helpful insight.

PR #513, opened 2020-04-07
adding support to not show ontology ID-based annotations

bistline:
Cell Metadata annotations that consist of ontology IDs (e.g. NCBITaxon_9606, MONDO_0000001) are no longer candidates for visualization if there is a corresponding "labels" annotation.  This is determined by the presence another group-based annotation with the same name, but suffixed with `__ontology_label`, as per the Alexandria metadata convention.  This also adds information to the "Study Details" page that will let the user know which annotations are "viewable", as well as a short tooltip to explain the criteria.

This also fixes a small regression with an alert incorrectly showing when a user selects a gene list to view from the study visualize tab.

This PR satisfies SCP-2227.

### DIFF
--- a/app/models/cell_metadatum.rb
+++ b/app/models/cell_metadatum.rb
@@ -71,10 +71,16 @@ class CellMetadatum
 
   def can_visualize?
     if self.annotation_type == 'group'
-      GROUP_VIZ_THRESHOLD === self.values.count
+      GROUP_VIZ_THRESHOLD === self.values.count && !self.is_ontology_ids?
     else
       true
     end
+  end
+
+  # determine if there is another metadatum that represents labels that map to these IDs
+  # e.g. disease vs. disease__ontology_label
+  def is_ontology_ids?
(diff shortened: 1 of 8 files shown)


app/models/cell_metadatum.rb
```diff
@@ -71,12 +71,18 @@ def annotation_select_option
 
   def can_visualize?
```
devonbush, line 72: looking through the codebase, this method looks like it's is always accessed in a keep_if? loop.  So if performance matters, it might be worth making a study-level 'visualizable_metadata' method that grabs all the visualizable fields in one go without requiring a DB query per column.  But that's a non-trivial query to write, so it may not be worth it.

bistline: That's a really good point - I think this method still has a place (specifically the "study info" page), but in the context of getting all visualizable annotations for a study, we want to optimize that.  I'll work on that refactoring.

bistline: 40a51e6

PR #515, opened 2020-04-10
refactoring API search to use json render instead of partials

devonbush:
Ok, this gives me a 10x (!!!) view render performance boost.  And all it does is take exactly the same code Jon wrote in partials, and move it to inside the controller.  Reading some Ruby threads, it seems like this is only an issue in development environments (rails has to re-interpret every partial on every call), but it's slow enough in my dev environment (5+ seconds to render a search) that I think this may be worth just switching to just for development speed.  The difference between a 5sec lag to reload the page and 0.3sec lag when developing is big!

Let me know your thoughts about whether it's worth switching, and if we think it's worth switching, I can clean up this code and move some of the methods (e.g. rendering the json for a single study) to shared libs somewhere.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -272,6 +272,7 @@ module Api
         @matching_accessions = @studies.map(&:accession)
         Rails.logger.info "Final list of matching studies: #{@matching_accessions}"
         @results = @studies.paginate(page: params[:page], per_page: Study.per_page)
+        render json: assemble_json_repsonse_obj, status: 200
       end
 
       swagger_path '/search/facets' do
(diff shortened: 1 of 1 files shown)


eweitz:
Wow!  Big difference indeed -- now an engineer can maintain mental context while iterating.

Could you link those Rails discussion threads that led you here, for reference?

Assuming the [build issue](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/159725983) is unrelated to this change, I see no issues.  Jon will have a more qualified review given his Rails expertise.

bistline:
I think this build failure is actually manifestation of the issue Tim has reported in the difference between the compressed/uncompressed file sizes and a race condition.  But not a blocking regression.

devonbush:
closing as it's included and further refined in the db-gene-search-links branch

PR #516, opened 2020-04-10
Encoding request URLs for analytics, allowing comma-delimited gene lists (SCP-1238, SCP-2272)

bistline:
Addressing two small issues:

1. Fixing URL encoding when reporting some requests to Google Analytics - lists of genes sometimes included control characters (^R, ^M) that would break when trying to convert
1. Allowing gene searches to delimit by both space and comma

This PR satisfies both SCP-1238 and SCP-2272.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -216,7 +216,7 @@ $(document).on('change', '#panel-genes-search input, #panel-genes-search select'
   $('#perform-gene-search').click();
 });
 
-// split a string on spaces, used for extractLast()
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/application.js
```diff
@@ -218,12 +218,12 @@ $(document).on('change', '#panel-genes-search input, #panel-genes-search select'
 
 // split a string on spaces, used for extractLast()
```
devonbush, line 219: update the comment to reflect that this also does commas now

bistline: d9d00de

PR #518, opened 2020-04-13
Continuous integration for React UI code

knapii-developments:
This PR adds CI for our React UI code by using```yarn ui-test```. It should be noted that tests that make an api call have been moved to a [separate folder](https://github.com/broadinstitute/single_cell_portal_core/pull/518/commits/53fbee9152f1837f340d1c5bccb9b2ea29ec7852#diff-a591b532f7318f248fbfe8ec5d3ea5ac) and are not included in the ui-test run. However, all tests are still included when running ```jest test``` and are utilized in the pre-push git hooks.

Commit d04d8be shows the [build](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/161247984) succeeding. 

This satisfies [SCP-2216](https://broadworkbench.atlassian.net/browse/SCP-2216)


package.json
```diff
@@ -1,10 +1,16 @@
...
   "scripts": {
     "test": "jest --coverage",
+    "ui-test": "jest test/js --testPathIgnorePatterns=scp-api.test.js",
     "testDebug": "node --inspect-brk node_modules/.bin/jest --runInBand",
     "lint": "eslint app/javascript",
     "fix-styles": "prettier-stylelint --write 'src/**/*.{css,scss}' "
   },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm test"
```
eweitz, line 11: Is there a reason to use `npm` locally and `yarn` in CI?  I personally prefer npm over yarn, but I prefer consistency even more.  

So, unless there's a reason I'm missing, for consistency I'd propose we either A) switch to `yarn test` here, or B) use `npm ui-test` in CI (and use npm instead of yarn everywhere else).

```suggestion
      "pre-push": "yarn test"
```

devonbush, line 11: curious why we're using 'npm test' here instead of 'yarn test'.  does it matter?

package.json
```diff
@@ -1,10 +1,16 @@
 {
   "scripts": {
     "test": "jest --coverage",
+    "ui-test": "jest test/js --testPathIgnorePatterns=scp-api.test.js",
```
eweitz, line 4: Blocker: my understanding is the intent here is to ignore `test/js/fetch/scp-api.test.js`, but it also ignores `test/js/scp-api.test.js` -- which seems like a functional bug insofar as CI is concerned.

Something like the following would likely test and skip as desired.

```suggestion
    "ui-test": "jest test/js --testPathIgnorePatterns=fetch/**",
```

package.json
eweitz, line 11: Separately, kudos on this elegant solution to the CI testing issue with fetch!

package.json
knapii-developments: Made updates in 0e230023. 
Results after update to use ui-test:
```Test Suites: 14 passed, 14 total
Tests:       40 passed, 40 total
Snapshots:   0 total
Time:        4.678s, estimated 6s
 ```
Results using yarn test: 
```
Test Suites: 15 passed, 15 total
Tests:       42 passed, 42 total
Snapshots:   0 total
Time:        5.55s
Ran all test suites.
✨  Done in 7.00s.
```
UI-test excludes 2 test so these results are expected

package.json
knapii-developments: I've just been using npm. I've updated it to yarn 0e23002

PR #519, opened 2020-04-13
API endpoints for metadata convention schemas (SCP-2322)

bistline:
Adding API endpoint to serve up metadata convention schema files, in JSON and TSV format.  Will include all previous "snapshots", as well as a stable URL that will always point to the latest version of any convention.

This PR satisfies SCP-2322.

### DIFF
--- a/app/controllers/api/v1/metadata_schemas_controller.rb
+++ b/app/controllers/api/v1/metadata_schemas_controller.rb
@@ -2,12 +2,7 @@ module Api
   module V1
     class MetadataSchemasController < ApiBaseController
       include Swagger::Blocks
-      before_action :set_available_schemas, only: :index
-      before_action :set_schema_filename, only: :load_schema
-      before_action :set_schema_filepath, only: :load_schema
-      before_action :set_response_type, only: :load_schema
-
-      SCHEMAS_BASE_DIR = Rails.root.join('lib', 'assets', 'metadata_schemas')
+      include Concerns::ConventionSchemas
 
       swagger_path '/metadata_schemas' do
         operation :get do
@@ -24,7 +19,8 @@ module Api
       end
 
       def index
-        render json: @schemas
+        schemas = set_available_schemas
+        render json: schemas
       end
 
       swagger_path '/metadata_schemas/{project_name}/{version}/{schema_format}' do
@@ -33,14 +29,13 @@ module Api
               'MetadataSchemas'
           ]
           key :summary, 'Load a metadata convention schema file'
-          key :description, 'Returns a schema definition file for the requested metadata convention, in JSON or TSV format'
+          key :description, "Returns a schema definition file for the requested metadata convention, in JSON or TSV format. Refer above for available schemas."
           key :operationId, 'metadata_schemas_load_schema_path'
           parameter do
             key :name, :project_name
             key :in, :path
             key :description, 'Project name of metadata convention'
             key :required, true
-            key :enum, %w(alexandria_convention)
             key :type, :string
           end
           parameter do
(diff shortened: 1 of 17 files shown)


app/controllers/api/v1/metadata_schemas_controller.rb
```diff
@@ -0,0 +1,125 @@
+module Api
+  module V1
+    class MetadataSchemasController < ApiBaseController
+      include Swagger::Blocks
```
devonbush, line 4: since these actions all only apply to a single route, I could see either moving each before_action declaration to right before each route function, or not using before_action and just calling the methods explicitly as the first statement of each route.  Totally a matter of taste, so feel free to ignore :)

app/controllers/api/v1/metadata_schemas_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+      def load_schema
+        if File.exists?(@schema_pathname)
+          send_file @schema_pathname, type: @response_type, filename: @schema_filename
+        else
+          head 404 and return
+        end
+      end
+
+      private
+
```
devonbush, line 83: this seems like it belongs in a helper lib somewhere as get_available_schema, since there's nothing request-specific about it.  But we can wait until we have a second consumer for it to move it

app/controllers/api/v1/metadata_schemas_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+            key :required, true
+            key :enum, %w(alexandria_convention)
+            key :type, :string
+          end
+          parameter do
+            key :name, :version
+            key :in, :path
+            key :description, 'Version of requested convention'
+            key :required, true
+            key :enum, %w(latest 2.0.0 1.1.5 1.1.4 1.1.3)
```
eweitz, line 51: This validation using `enum` seems like it would not suffice for multiple projects.

Let's say we release a new convention for a second project, "Foo Metadata Convention 1.0.0", as is foreseeable given we have a URL path segment for `{project_name}` .  Line 43 becomes `key :enum, %w(alexandria_convention foo)` and line 51 becomes `key :enum, %w(latest 2.0.0 1.1.5 1.1.4 1.1.3 1.0.0)`.  Then `/metadata_schemas/foo/1.1.3/json` would pass URL validation -- but it should fail.  Same for `/metadata_schemas/alexandria_convention/1.0.0/json`. 

Perhaps something like a namespaced `enum` in line 51 would address that.  I don't know if that's possible.

I don't think this lurking issue needs to be fixed now, but it seems foreseeable if we add a `project` beyond `alexandria_convention`.

app/controllers/api/v1/metadata_schemas_controller.rb
bistline: Yeah I went back and forth about where to put these methods, and whether they should be static or not.  I'll ponder it some more and see if something else makes more sense.

app/controllers/api/v1/metadata_schemas_controller.rb
bistline: So the `enums` are just for Swagger, and not part of any validation server-side.  And since we only have the one project, it works fine for now.  But you are correct in that if we do get another project that has different version numbers, something different will need to be done in order to make those dropdowns consistent.  I'm not aware of anything in Swagger that would support this, but I can check.

bistline: 03f64f7

app/controllers/api/v1/metadata_schemas_controller.rb
bistline: 03f64f7

app/controllers/api/v1/metadata_schemas_controller.rb
bistline: 03f64f7

PR #521, opened 2020-04-14
Skipping nil/optional outputs when syncing workflow submissions (SCP-2339)

bistline:
Syncing workflow submissions currently breaks if a submission had "optional" outputs the were not generated or defined.  To avoid throwing errors, the portal now skips any optional/nil outputs in workflows if they are not present.  This also addresses a similar issue when attempting to generate an `analysis.json` file with submission metadata.

This PR satisfies SCP-2339.

### DIFF
--- a/app/models/analysis_metadatum.rb
+++ b/app/models/analysis_metadatum.rb
@@ -268,12 +268,14 @@ class AnalysisMetadatum
           outputs = []
           workflows.each do |workflow|
             outs = workflow['outputs'].values
-            outs.each do |o|
-              outputs << {
-                  'name' => o.split('/').last,
-                  'file_path' => o,
-                  'format' => o.split('.').last
-              }
+            if outs.present?
+              outs.each do |o|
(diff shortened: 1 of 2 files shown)


app/models/analysis_metadatum.rb
```diff
@@ -268,12 +268,14 @@ def create_payload
           outputs = []
           workflows.each do |workflow|
             outs = workflow['outputs'].values
-            outs.each do |o|
-              outputs << {
-                  'name' => o.split('/').last,
-                  'file_path' => o,
-                  'format' => o.split('.').last
-              }
+            if outs.present?
```
devonbush, line 271: You could skip the if check and do `outs&.each do |o|` instead --  dealer's choice :)

bistline: Ooooh snazzy.  Didn't know that particular shorthand.  It's not as readable but I do like the concision.

PR #525, opened 2020-04-15
Fix confusing Download tooltip when not signed in

eweitz:
This refines tooltip messaging when the user is not signed in.

Screen Shot 2020-04-15 at 11 24 14 AM

It satisfies SCP-2303.


app/models/user.rb
```diff
@@ -104,9 +104,9 @@ def owned_by(user)
 
   DEFAULT_FEATURE_FLAGS = {
     # whether the home page uses React and the new search API
-    "advanced_search" => false,
+    "advanced_search" => true,
     # whether the facet search controls are shown
-    "faceted_search" => false,
+    "faceted_search" => true,
```
devonbush, line 109: blocker: i think this was left in accidentally

test/js/download.test.js
```diff
@@ -59,14 +61,39 @@ describe('Download components for faceted search', () => {
...
+    const wrapper = mount((< DownloadButton />))
+    wrapper.find('#download-button > span').simulate('mouseenter')
+
+    const tooltipHint =
+      wrapper.find('OverlayTrigger').prop('overlay').props['children']
+
+    expect(tooltipHint).toBe('To download, please sign in')
   })
+
+  // TODO (SCP-2333): Restore test for showing modal upon clicking Download button
```
devonbush, line 81: thanks for making the ticket

app/models/user.rb
eweitz: Good catch!  Fixed.

PR #529, opened 2020-04-15
adding visualizable check to gene search

devonbush:
image

### DIFF
--- a/app/controllers/api/v1/study_search_results_objects.rb
+++ b/app/controllers/api/v1/study_search_results_objects.rb
@@ -58,6 +58,7 @@ module Api
         end
         if @gene_results.present?
           study_obj[:gene_matches] = @gene_results[:genes_by_study][study.id]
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/study_search_results_objects.rb
```diff
@@ -58,6 +58,7 @@ def study_response_obj(study)
         end
         if @gene_results.present?
           study_obj[:gene_matches] = @gene_results[:genes_by_study][study.id]
+          study_obj[:can_visualize_clusters] = study.can_visualize_clusters?
```
bistline, line 61: If we're adding this, you should also add `study_obj[:has_expression_data] = study. has_expression_data?` since that is also required for viewing violin/heatmaps

bistline: After discussion, I realized that you won't even get a response for this study unless it has expression data, so this is superfluous.

PR #531, opened 2020-04-15
Ingest Pipeline v1.3.1 (SCP-2280, SCP-2285)

bistline:
Updates to the latest release of `scp-ingest-pipeline`, which includes support for Alexandria metadata convention v2.0.0.  This also adds information to metadata parse emails about what version of the convention was used, if any.

This PR satisfies in part SCP-2280 and SCP-2285.


app/controllers/api/v1/concerns/convention_schemas.rb
```diff
@@ -18,6 +18,15 @@ def set_available_schemas
           end
           schemas
         end
+
+        # get the latest version number of a given project/schema
+        def get_latest_schema_version(project_name)
+          schemas = set_available_schemas
```
devonbush, line 24: should we rename this method now to 'get_available_schemas'?

PR #544, opened 2020-04-17
Fallback to first available annotation on invalid requests (SCP-2342)

bistline:
Race conditions in the UI can sometimes lead to invalid requests where a given annotation may not exist for the requested cluster.  In addition to previous client-side fixes, this server-side patch will fall back to the first available annotation from a study's cell metadata file (which has to be present in order for visualization to work in the first place).

This PR satisfies SCP-2342.


app/controllers/site_controller.rb
```diff
@@ -1410,24 +1410,37 @@ def set_selected_annotation
     # determine which URL param to use for selection and construct base object
     selector = params[:annotation].nil? ? params[:gene_set_annotation] : params[:annotation]
     annot_name, annot_type, annot_scope = selector.nil? ? @study.default_annotation.split('--') : selector.split('--')
+
     # construct object based on name, type & scope
```
devonbush, line 1414: Is there a way to refactor this so that this calls  the method in ApiParamsUtils?  Or is the logic different.
Also, maybe we need a better name for ApiParamsUtils since the utility methods may not be restricted to API requests.

bistline: Possibly - let me look it over.  I agree that the duplication isn't ideal, but the difference in the controllers made me think that it wasn't a good idea.  But I'll take a closer look.

bistline: Don't know why I thought that wouldn't be a good idea - much cleaner now.  0086954.

bistline:
Failed test is another race condition surrounding gzipping files and pushing to the bucket (updated_at timestamps don't match).  This is not a regression, just further proof we need to look at that logic regarding file compression/uploads.

PR #545, opened 2020-04-21
Db fix accession search

devonbush:
The production query that triggered this was a keyword search for 'SCP-502', which gets treated differently by the accession matcher and the term splitter.  We may want a separate ticket to handle the underlying treatment of '-' in terms, but this at least makes the method not error.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -250,7 +250,7 @@ module Api
             accession_index = possible_accessions.index(study.accession)
             if accession_index.nil?
               # study was not a true accession match, it matches the accession term in its description
-              # make this appear after the proper accession matches, in order of wieght match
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -246,7 +246,15 @@ def index
         when :keyword
           @studies = @studies.sort_by {|study| -study.search_weight(@term_list)[:total] }
         when :accession
-          @studies = @studies.sort_by {|study| possible_accessions.index(study.accession) }
+          @studies = @studies.sort_by do |study|
+            accession_index = possible_accessions.index(study.accession)
+            if accession_index.nil?
+              # study was not a true accession match, it matches the accession term in its description
+              # make this appear after the proper accession matches, in order of wieght match
```
eweitz, line 253:
```suggestion
              # make this appear after the proper accession matches, in order of weight match
```

PR #546, opened 2020-04-22
Prevent checking permissions/groups in FireCloud API during outages (SCP-2325)

bistline:
During a FireCloud service outage (defined as any service that does not report "ok" in the FireCloud API [status](https://api.firecloud.org/#!/Status/status) endpoint), many of the permission or group membership checks the portal makes when loading the study overview page will fail with a 502 or 503.  Since these error codes are marked as being able to "retry", the portal will employ a staggered backoff approach to making API calls.  The end result is that a study can take up to 3 minutes to load while all retries are exhausted.

As a result, this fix will now skip making any such calls if any of their dependent services are not functioning normally, and the associated portal functionality will be disabled for that request.  For example: if `Rawls` or `Sam` are not reporting `"ok: true"`, then a user that owns the project/workspace associated with a portal study will not see either the analysis or study settings tab load when viewing their study while signed in.  The permission check itself has also been refactored into a single status call, and each endpoint is evaluated from the single call (previously it had be 3 separate calls).  Some checks are now skipped based on the results of previous checks.

Also, this fixes a lingering issue with degraded performance when loading the study download tab.  Each file download link was making a permission check if the signed-in user had view permission for a given study, which in some circumstances involved an API call.  This has been changed to a single call and the result is shared across all files.

This PR satisfies SCP-2325.


app/controllers/site_controller.rb
```diff
@@ -1444,10 +1445,29 @@ def set_workspace_samples
...
+          buckets_ok = system_status.dig('GoogleBuckets', 'ok') == true
+          @allow_downloads = buckets_ok
+          @allow_edits = sam_ok && rawls_ok
+        end
+      rescue => e
+        logger.error "Error checking FireCloud API status: #{e.class.name} -- #{e.message}"
+        error_context = ErrorTracker.format_extra_context(@study, {firecloud_status: api_status})
+        ErrorTracker.report_exception(e, current_user, error_context)
+      end
+    end
```
eweitz, line 1470: An early `return` would decrease cyclomatic complexity and line length.  Opinions can differ, but I've been finding early returns to help readability, too.

```suggestion
    return if study_detached
    
    begin
      @allow_firecloud_access = AdminConfiguration.firecloud_access_enabled?
      api_status = Study.firecloud_client.api_status
      # reuse status object because firecloud_client.services_available? each makes a separate status call
      # calling Hash#dig will gracefully handle any key lookup errors in case of a larger outage
      if api_status.is_a?(Hash)
        system_status = api_status['systems']
        sam_ok = system_status.dig('Sam', 'ok') == true # do equality check in case 'ok' node isn't present
        rawls_ok = system_status.dig('Rawls', 'ok') == true
        buckets_ok = system_status.dig('GoogleBuckets', 'ok') == true
        @allow_downloads = buckets_ok
        @allow_edits = sam_ok && rawls_ok
      end
    rescue => e
      logger.error "Error checking FireCloud API status: #{e.class.name} -- #{e.message}"
      error_context = ErrorTracker.format_extra_context(@study, {firecloud_status: api_status})
      ErrorTracker.report_exception(e, current_user, error_context)
    end
```

app/views/site/study.html.erb
```diff
@@ -17,7 +17,7 @@
         <%= render partial: 'study_download_data' %>
       </div>
     <% end %>
-    <% if @allow_firecloud_access && @allow_computes && @user_can_compute %>
+    <% if @allow_firecloud_access && @user_can_compute %>
```
eweitz, line 20: Nice simplification!

app/controllers/site_controller.rb
bistline: Interesting - I like that.  Less cyclomatic complexity, and still very readable.  I'll give that a spin.

app/controllers/site_controller.rb
```diff
@@ -1444,10 +1445,31 @@ def set_workspace_samples
...
   def set_firecloud_permissions(study_detached)
-    @allow_firecloud_access = study_detached ? false : AdminConfiguration.firecloud_access_enabled?
-    @allow_downloads = study_detached ? false : Study.firecloud_client.services_available?('GoogleBuckets')
-    @allow_computes = study_detached ? false : Study.firecloud_client.services_available?('Sam', 'Agora', 'Rawls')
-    @allow_edits = study_detached ? false : Study.firecloud_client.services_available?('Sam', 'Rawls')
+    @allow_firecloud_access = false
+    @allow_downloads = false
+    @allow_edits = false
+    @allow_computes = false
+    return if study_detached
```
devonbush, line 1446: this was a nice refactor -- much easier to read

app/controllers/studies_controller.rb
```diff
@@ -23,8 +23,8 @@ class StudiesController < ApplicationController
     check_access_settings
   end
   # special before_action to make sure FireCloud is available and pre-empt any calls when down
-  #before_action :check_firecloud_status, except: [:index, :do_upload, :resume_upload, :update_status,
-  #                                                :retrieve_wizard_upload, :parse]
+  before_action :check_firecloud_status, except: [:index, :do_upload, :resume_upload, :update_status,
+                                                  :retrieve_wizard_upload, :parse]
```
devonbush, line 27: were these commented out before for performance reasons?

app/models/study.rb
```diff
@@ -764,28 +764,34 @@ def can_compute?(user)
...
-        if workspace_acl['acl'][user.email].nil?
-          # check if user has project-level permissions
-          user_client = FireCloudClient.new(user, self.firecloud_project)
-          projects = user_client.get_billing_projects
-          # billing project users can only create workspaces, so unless user is an owner, user cannot compute
-          projects.detect {|project| project['projectName'] == self.firecloud_project && project['role'] == 'Owner'}.present?
-        else
-          workspace_acl['acl'][user.email]['canCompute']
+      # don't check permissions if API is not 'ok'
+      if Study.firecloud_client.services_available?('Sam', 'Rawls', 'Agora')
```
devonbush, line 796: do you think we should make these strings constants somewhere?  e.g.  FireCloudClient::SAM_SERVICE = 'Sam'.

test/models/study_test.rb
```diff
@@ -74,4 +78,40 @@ def setup
     puts "#{File.basename(__FILE__)}: '#{self.method_name}' successful!"
 
   end
+
+  test 'should skip permission and group check during firecloud service outage' do
```
devonbush, line 82: nice

app/controllers/studies_controller.rb
bistline: No - I think this was actually an error where we were testing something and forgot to uncomment it.  It most definitely _should_ be there as API outage will cause many of these methods to barf if we don't check.

app/models/study.rb
bistline: That's actually a great idea.  It should be there, and comments to the effect of what service covers what.  That data only lives in my brain and our playbook... it would be great to have them in code.

app/controllers/site_controller.rb
bistline: c984b19

app/views/site/study.html.erb
bistline: Unfortunately I had to take it out because I realized that it was there specifically for thenav-tabs.  But I did make it so that it can source existing calls rather than make a new one.

app/models/study.rb
bistline: 555b181

PR #548, opened 2020-04-23
adding validations to analysis_parameter_filters to prevent breaking UI (SCP-2327)

bistline:
Adding filters to analysis parameters helps refine the UI when a user is trying to select inputs for a given analysis.  However, we had no validations attached to the `AnalysisParameterFilter` model, and therefore it was possible to save an "invalid" filter value that would then throw and error when trying to render the UI that was cryptic, at best.

This refinement adds standard presence/content validations to `AnalysisParameterFilter` and leverages the standard errors block in Rails forms to communicate the issue back the the pipeline curator.

This PR satisfies SCP-2327.


bistline:
Failed build is the same flaky test, not a regression.

PR #552, opened 2020-04-28
Ideogram for exploring related genes (SCP-2363)

eweitz:
This is a draft enhancement of single-gene search in the Study Overview "Explore" tab.  In previously unused space at top right on the page, it shows an ideogram with the searched gene and related genes plotted across the genome.  Users can click on the related genes -- evolutionarily similar genes in the same organism, i.e. paralogs -- to trigger a new gene search.

Without something like this, users must know _a priori_ which genes are related to a gene they search.  This genome visualization gives users a hint to discover related genes, and a graphical way to explore them.

This draft feature was developed during the first "SCP Developer's Choice Days".  If promising, it could be feature-flagged and further developed to test with more representative datasets, adapt for narrow viewports, refine code, etc.

This brief animation shows the gist:
scp_related_genes_via_ideogram_2020-04-28

This relates to SCP-2363.


app/controllers/site_controller.rb
```diff
@@ -417,6 +417,9 @@ def view_gene_expression
     @cluster_annotations = load_cluster_group_annotations
     @top_plot_partial = @selected_annotation[:type] == 'group' ? 'expression_plots_view' : 'expression_annotation_plots_view'
     @y_axis_title = load_expression_axis_title
+    matrix = @study.expression_matrix_files.first
```
bistline, line 420: Non-blocking refinement: while this will give you a matrix from which you can extract the `taxon`, a nice enhancement might be to extract that from each matrix (rather than just the first) and store in an object that you could use in the front end in a select (in case the study has multiple species).  For example:

```
@taxons = []
@study.expression_matrix_files.each do |matrix|
    @taxons << {taxid: matrix.taxon.ncbi_taxid, organism: matrix.taxon.scientific_name}
end
```

config/environments/development.rb
```diff
@@ -76,4 +76,6 @@
     SwaggerUiEngine::SwaggerDocsController.send(:include, Api::V1::Concerns::CspHeaderBypass)
     SwaggerUiEngine::SwaggerDocsController.send(:layout, 'swagger_ui_engine/layouts/swagger')
   end
+
+  config.force_ssl = true
```
bistline, line 80: We will need to test this to make sure that it does not conflict with Dockerized development via nginx/Passenger.

devonbush: > Looks good - I'd just like a quick check that adding `config.force_ssl` in `config/environments/development.rb` will not impact Dockerized development.

Yes, I think all the local https changes (thin in the gemfile, requiring ssl, etc...) should not be part of this branch

app/controllers/site_controller.rb
eweitz: Good catch here.  I've stashed [some scratch code to enable this](https://github.com/broadinstitute/single_cell_portal_core/pull/552/commits/990b6abd944b62a1bce121558c3592328ec74de6), but it needs more work.

Tandem changes will be needed for gene autocomplete.  These would preserve current gene autocomplete behavior in Study Overview, while ensuring the correct organism's genome is shown for the selected gene.

config/environments/development.rb
eweitz: I've incorporated recent updates here that do away with `thin` (and without which this doesn't work locally).

app/views/site/_view_gene_expression_title_bar.html.erb
```diff
@@ -5,4 +5,14 @@
       <h3 class="help-block lead">Expression values are shown as the <strong id="selected-consensus"><%= params[:consensus].nil? ? 'mean' : params[:consensus] %></strong> of expression values across all genes for each cell.</h3>
     <% end %>
   </div>
-</div>
\ No newline at end of file
+</div>
+
+<script>
+  <% if User.feature_flags_for_user(current_user).key?('ideogram_related_genes') && User.feature_flags_for_user(current_user)['ideogram_related_genes'] == true %>
```
devonbush, line 11:
```suggestion
  <% if User.feature_flag_for_user(current_user, 'ideogram_related_genes') %>
```
If that doesn't work, let me know, because the intent of feature_flag_for_user is that it *should* work that way :)

app/controllers/site_controller.rb
devonbush: My impression is that there are relatively painless and risk-free calls (just a couple of DB lookups), but I think it's worth wrapping the additions to this method in a feature flag block just so that we know what they belong to

PR #553, opened 2020-04-28
Ensure categorical labels in violin plot x-axis labels (SCP-2213)

eweitz:
This fixes Plotly violin plots so "group" type annotations with numeric values are honored as "group".  Previously, if x-axis labels were integer strings ("1", "2", ..., "10"), some labels would be omitted.  Now, all labels are shown as intended.

I tested this manually.

This satisfies SCP-2213.

### DIFF
--- a/app/assets/javascripts/kernel-functions.js
+++ b/app/assets/javascripts/kernel-functions.js
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
 /* Advanced Rule of thumb bandwidth selector from :
     https://en.wikipedia.org/wiki/Kernel_density_estimation#Bandwidth_selection
     and https://stat.ethz.ch/R-manual/R-devel/library/stats/html/bandwidth.html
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/kernel-functions.js
```diff
@@ -111,6 +113,7 @@ function createTracesAndLayout(arr, title, jitter='all', expressionLabel){
     }
     var layout = {
         title: title,
+        xaxis: 'category',
```
bistline, line 116: I guess I don't understand how this fixes it... explanation?

eweitz: I just added a note in 518509c.  Unfurling the URL there for convenience:

> Forcing an axis to be categorical
If you pass string values for the x or y parameter, plotly will automatically set the corresponding axis type to category, with the exception of string of numbers, in which case the axis is linear. It is however possible to force the axis type by setting explicitely xaxis_type to be category.

There's oddly no explanation of this in the Plotly JavaScript docs, but a similar generic API applies for all languages they support.

bistline: Great - thanks!

PR #563, opened 2020-05-04
adding doUpload duplicate check logging

devonbush:


app/controllers/studies_controller.rb
```diff
@@ -452,32 +452,38 @@ def do_upload
...
-        render json: { file: { name: study_file.upload_file_name,size: upload.size } } and return
-      else
+      begin
+        study_file.update!(study_file_params)
+      rescue => e
         logger.error "#{Time.zone.now} #{study_file.errors.full_messages.join(", ")}"
+        existing_file = StudyFile.find(study_file.id)
+        if existing_file
+          ErrorTracker.report_exception(e, current_user, params)
+          logger.error("DoUpload Failed: Existing file for #{study_file.id} -- type:#{existing_file.type} name:#{existing_file.name}")
```
eweitz, line 462:
```suggestion
          logger.error("do_upload Failed: Existing file for #{study_file.id} -- type:#{existing_file.type} name:#{existing_file.name}")
```

app/controllers/studies_controller.rb
```diff
@@ -452,32 +452,38 @@ def do_upload
...
-        end
-        render json: { file: { name: study_file.upload_file_name,size: upload.size } } and return
-      else
+      begin
+        study_file.update!(study_file_params)
+      rescue => e
         logger.error "#{Time.zone.now} #{study_file.errors.full_messages.join(", ")}"
+        existing_file = StudyFile.find(study_file.id)
+        if existing_file
+          ErrorTracker.report_exception(e, current_user, params)
```
bistline, line 461: So we're pivoting to logging the error rather than trying to capture and recover from it?  I seem to remember this was the conversation.  I agree it's probably the best we can do since reproducing this is so difficult.

devonbush: yeah, I'm not sure what the right recovery path is, so we'll log it and see.

PR #564, opened 2020-05-05
Roll back deployments (SCP-2385)

bistline:
If a deployment of SCP to a remote host fails for some reason (portal fails to boot properly due to an error, or a showstopper bug is found during release verification), currently rollbacks have to be performed manually on the host itself using the SCP playbook.  Now, using our Jenkins server, we can automate a rollback by one (1) release using the Github [releases API endpoint](https://api.github.com/repos/broadinstitute/single_cell_portal_core/releases) for this repository and checking out the previous release tag on the remote host.  Rollbacks will be completely automated, not requiring any action from a portal administrator other than invoking the Jenkins job.  This will allow safe recovery of a broken instance while development staff investigate and prioritize a hotfix, without having to worry about users being blocked in the meantime.

The rollback can be customized by passing the `-t [OFFSET]` flag to `bin/deploy.sh` in the Jenkins configuration to increase the number of tags to roll back (default is one release, but can go back as many as are available, however due to ongoing schema/infrastructure changes, this is not advised unless absolutely necessary).

Also, this update increases the final portal health check timeout from 3 to 6 minutes to help account for increased JS asset compilation time, and eliminate any "false negative" failed builds when the deployment was a success, just not in the time allotted.

This PR satisfies SCP-2385.

### DIFF
--- a/bin/github_releases.sh
+++ b/bin/github_releases.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# functions for retrieving Github releases for SCP and extracting requested release attributes
+# by walking back through release history
+#
+# Requirements: curl, jq
+
+REPO_URL='https://api.github.com/repos/broadinstitute/single_cell_portal_core/releases'
+
+# Get all Github published releases for $REPO_URL
+function get_all_releases {
+  echo $(curl -X GET --header 'application/json' --silent $REPO_URL)
+}
+
+# extract an attribute from a Github release object JSON
+# will default to using "tag_name", and targeting the previous release from the most recent
+function extract_release_attribute {
+  OFFSET="$1"
+  RELEASE_ATTR="$2"
+  if [[ -z "$OFFSET" ]]; then
+    OFFSET="1"
+  fi
+  if [[ -z "$RELEASE_ATTR" ]]; then
+    RELEASE_ATTR="tag_name"
+  fi
+  echo $(get_all_releases | jq ".[$OFFSET].$RELEASE_ATTR" | sed 's/"//g')
+}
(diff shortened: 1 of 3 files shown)


bin/github_releases.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+# functions for retrieving Github releases for SCP and extracting requested release attributes
```
eweitz, line 3:
```suggestion
# functions for retrieving GitHub releases for SCP and extracting requested release attributes
```

And similarly throughout this file.

bin/github_releases.sh
```diff
@@ -0,0 +1,27 @@
...
+function extract_release_attribute {
+  OFFSET="$1"
+  RELEASE_ATTR="$2"
+  if [[ -z "$OFFSET" ]]; then
+    OFFSET="1"
+  fi
+  if [[ -z "$RELEASE_ATTR" ]]; then
+    RELEASE_ATTR="tag_name"
+  fi
+  echo $(get_all_releases | jq ".[$OFFSET].$RELEASE_ATTR" | sed 's/"//g')
```
eweitz, line 26: This second parameter strikes me as unnecessary -- I can't think of another `RELEASE_ATTR` that'd make sense here, and the parameter isn't exposed upstream.

So perhaps consider this simplification:

```suggestion
  echo $(get_all_releases | jq ".[$OFFSET].tag_name" | sed 's/"//g')
```

bistline: Yeah, I debated that - there definitely are things you could think that might be useful to grab from that endpoint, but in the context of rolling back a release, `tag_name` is the only thing that we actually need.  I'll change that and rename the method to `extract_release_tag`.

bistline: 52d13a5

bin/github_releases.sh
bistline: 52d13a5

PR #567, opened 2020-05-07
Fixing bug with NaN in scatter/violin plots (SCP-2079)

bistline:
Arrays of numeric data that contain NaN values (which are technically `Float::NAN` instances in Ruby) will throw an error when trying to compute the min/max values, which are needed when trying to determine the bounds of Plotly axes.  This patch captures the error, and then sanitizes the array by rejecting any NaN values before calling minmax.  This is approach is taken as most of our arrays of numeric data do not contain NaN values, so calling `minmax` first rather than always sanitizing the array will reduce load times in most cases.

This PR does not address the issues raised regarding the UX and scientific validity of showing NaN values in plots - further refinement is needed on that front.  This simply handles the error to allow plots to render.  This PR satisfies the main goal of SCP-2079.

### DIFF
--- a/app/lib/request_utils.rb
+++ b/app/lib/request_utils.rb
@@ -55,4 +55,13 @@ class RequestUtils
     end
     page_num
   end
+
+  # safely determine min/max bounds of an array, accounting for NaN value
+  def self.get_minmax(values_array)
+    begin
+      values_array.minmax
+    rescue TypeError, ArgumentError
+      values_array.dup.reject!(&:nan?).minmax
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


app/lib/request_utils.rb
```diff
@@ -55,4 +55,13 @@ def self.sanitize_page_param(page_param)
     end
     page_num
   end
+
+  # safely determine min/max bounds of an array, accounting for NaN value
+  def self.get_minmax(values_array)
```
devonbush, line 60: really like the approach here

PR #570, opened 2020-05-08
Prevent loading missing subsampled clusters (SCP-2409)

bistline:
Since subsampling happens after a cluster has finished parsing, it is currently possible for users to request subsampled clustering data before it has finished computing.  This leads to both empty cluster plots, and errors when trying to view gene expression data.  This patch removes all subsampling options from the Subsampling threshold dropdown menu, and provides a small warning telling users that the cluster has not completed subsampling yet.  Also, if a request is somehow made for subsampled data that does not exist, the cluster will fall back to full-resolution data (i.e. "All Cells") instead of returning an empty array.

This PR satisfies SCP-2409.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -222,9 +222,10 @@ module ApplicationHelper
   end
 
   # return an array of values to use for subsampling dropdown scaled to number of cells in study
-  # only options allowed are 1000, 10000, 20000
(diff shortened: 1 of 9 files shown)


bistline:
I checked and the build passed - this looks like GitHub never picked up the event.

https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/164791683

PR #571, opened 2020-05-09
fix advanced global gene search

devonbush:
Fixes the bug noted by Eric where keyup events in gene search were causing infinite repaints.  We suspect this was a latent problem revealed by the recent yarn upgrade.


eweitz:

PR #573, opened 2020-05-10
enabling admin accounts to download study files (SCP

devonbush:
Fixes SCP-2401 (bulk download study modal sometimes empty) 
 
The issue here is that the check for showing modal contents for non-public studies (Study.has_bucket_access?) doesn't line up with the check for visibility of the study download button (@allow_downloads).  Specifically, allow_downloads allows anyone who can edit the study to download as long as Firecloud is up, and the can_edit? check includes an allowance for all admins.  The has_bucket_access? has no such admin allowance.  So the simplest fix is just to add the admin allowance in, but I'm not sure if there's not a more elegant refactoring to be had.  

Jon, you were the one that told me about this issue -- did it impact anyone that didn't have an admin account?  (i.e. do we need to look for other gaps in the has_bucket_access? method)

### DIFF
--- a/app/views/site/_study_download_data.html.erb
+++ b/app/views/site/_study_download_data.html.erb
@@ -29,28 +29,31 @@
                   </thead>
                 </table>
               <% end %>
-            <% elsif @study.public? && !@user_embargoed %>
-                <p class="lead">To download all files using <code>curl</code>, click the button below to get the download command:</p>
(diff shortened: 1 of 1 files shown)


app/models/study.rb
```diff
@@ -763,6 +763,8 @@ def has_bucket_access?(user)
     else
       if self.user == user
         return true
+      elsif user.admin? # need a separate case for this since admin users won't be in user_in_group_share?
+        return true
```
bistline, line 767: The reason that this method had no special case for admins is that "admin" accounts only have special access in the portal, not in Terra or the bucket.  So we wouldn't want to return true here - it will almost never be the case.

PR #572, opened 2020-05-10
Nondockerized ssl

devonbush:
enables non-dockerized development to use https over port 3000.  (I could see an argument for moving it to port 8443, but that would require changing a number of paths in the code)

PR #575, opened 2020-05-11
CI-runnable tests for violin plots (SCP-2213, SCP-2391, SCP-2403)

eweitz:
This adds CI-runnable tests for violin plots.  It introduces a general model for fast, reliable, and higher-coverage tests of React components that use Plotly.

Previously, it was impractical to have automated tests for components that use Plotly.  Existing end-to-end tests in `ui_test_suite.rb` have  high false-positive rates -- e.g. throwing an error when no bug exists -- and are not runnable in Travis CI due to authentication issues.  Also, code in `useEffect` and downstream was not testable at a component level.

Now, tests for such components can run in CI by mocking `fetch` with a file-system read (rather than an HTTP request) using high-fidelity mock data.  Using React Testing Library (originally researched by Eno), we can now test code in `useEffect` -- and thus have reliable regression tests for issues like #560.  These tests are also fast; this one typically takes < 100 ms.

The main limitation here is that it does not actually test rendered SVG, due to jsdom/jsdom#1330.  Something like [`jest-puppeteer`](https://github.com/smooth-code/jest-puppeteer) could enable that in a similarly fast, reliable, and Travis CI-friendly manner.

This satisfies SCP-2213, SCP-2391, and SCP-2403.


eweitz:
This PR includes changes from #572, and will get a simpler to review once that is merged.

rails_local.bash
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+source config/secrets/.source_env.bash
+
+thin start --ssl --ssl-key-file config/local_ssl/localhost.key --ssl-cert-file config/local_ssl/localhost.crt  --ssl-disable-verify
```
bistline, line 5: Since we're using `puma` now, is this still needed?

eweitz: Nope, good catch, removed in 2657056.

config/local_ssl/localhost.crt
```diff
@@ -0,0 +1,28 @@
+-----BEGIN CERTIFICATE-----
```
devonbush, line 1: this file should be removed -- it got moved to config/certs

config/local_ssl/localhost.key
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
devonbush, line 1: same with this file -- moved to config/certs

config/local_ssl/localhost.crt
eweitz: Done in 4ee6b23.

config/local_ssl/localhost.key
eweitz: Done in 4ee6b23.

PR #582, opened 2020-05-13
Adding remote asset backup for user-uploaded assets (SCP-2374)

bistline:
Adds the `UserAssetService` to back up any user-provided assets (i.e. images) that are used either for branded spaces, or legacy support of previously uploaded images for study descriptions.  Assets will be backed up to a Google bucket after new branded spaces are created, and every time the portal reboots, the asset service will localize any files that are missing on disk.  This is mainly useful for when new host VMs are provisioned (which would have new persistent disks).

Each portal GCP project will have a bucket per Rails environment used.  Buckets are created on-demand, so no prior setup is required.  Also includes a migration to perform the initial backup of all assets present for each portal instance.

This PR satisfies SCP-2374.

### DIFF
--- a/app/lib/user_asset_service.rb
+++ b/app/lib/user_asset_service.rb
@@ -1,0 +1,203 @@
+##
+# UserAssetService: manages uploading/localization of "user-provided" static asset files
+# Used primarily for BrandingGroup images, but also supports legacy CKEditor uploads
+##
+
+class UserAssetService
+
+  # GCP Compute project to run reads/writes in
+  COMPUTE_PROJECT = ENV['GOOGLE_CLOUD_PROJECT'].blank? ? '' : ENV['GOOGLE_CLOUD_PROJECT']
+  # Service account JSON credentials
+  SERVICE_ACCOUNT_KEY = !ENV['SERVICE_ACCOUNT_KEY'].blank? ? File.absolute_path(ENV['SERVICE_ACCOUNT_KEY']) : ''
+
+  # Asset Path helpers
+  RAILS_PUBLIC_PATH = Rails.root.join('public', 'single_cell')
+  ASSET_PATHS_BY_TYPE = {
+      ckeditor_pictures: RAILS_PUBLIC_PATH.join('ckeditor_assets', 'pictures'),
+      ckeditor_attachments: RAILS_PUBLIC_PATH.join('ckeditor_assets', 'attachments'),
+      branding_images: RAILS_PUBLIC_PATH.join('branding_groups')
+  }.with_indifferent_access
+  ASSET_TYPES = ASSET_PATHS_BY_TYPE.keys.freeze
+
+  # Bucket info
+  STORAGE_BUCKET_NAME = (COMPUTE_PROJECT.dup + '-asset-storage').freeze
+
+  # initialize GCS driver with same credentials as FireCloudClient
+  #
+  # * *yields*
+  #   - +Google::Cloud::Storage+
+  def self.initialize_storage_service
+    Google::Cloud::Storage.new(keyfile: SERVICE_ACCOUNT_KEY)
+  end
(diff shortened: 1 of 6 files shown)


bistline:
> Looks good, except for a seemingly relevant build failure:
> 
> https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/165806123#L2603

Ah, I think the problem here is how the test is written.  Because I'm using `ObjectIds` to mimic what the paths would be like in normal usage, that means every time we run CI it will create new entries in the bucket, so that number will always increase.  I'll change it so that the paths are idempotent.

app/lib/user_asset_service.rb
```diff
@@ -0,0 +1,229 @@
...
+  RAILS_PUBLIC_PATH = Rails.root.join('public', 'single_cell')
+  ASSET_PATHS_BY_TYPE = {
+      ckeditor_pictures: RAILS_PUBLIC_PATH.join('ckeditor_assets', 'pictures'),
+      ckeditor_attachments: RAILS_PUBLIC_PATH.join('ckeditor_assets', 'attachments'),
+      branding_images: RAILS_PUBLIC_PATH.join('branding_groups')
+  }.with_indifferent_access
+  ASSET_TYPES = ASSET_PATHS_BY_TYPE.keys.freeze
+
+  # Bucket info; each Rails environment per project has a bucket
+  STORAGE_BUCKET_NAME = (COMPUTE_PROJECT.dup + "-#{Rails.env}-asset-storage").freeze
```
devonbush, line 23:
```suggestion
  STORAGE_BUCKET_NAME = "#{COMPUTE_PROJECT.dup}-#{Rails.env}-asset-storage").freeze
```

app/lib/user_asset_service.rb
```diff
@@ -0,0 +1,229 @@
...
+
+  # initialize GCS driver with same credentials as FireCloudClient
+  #
+  # * *yields*
+  #   - +Google::Cloud::Storage+
+  def self.initialize_storage_service
```
devonbush, line 29: it seems odd to call this 'initialize' since it's not the one that sets the class variable.  I'd rename this `create_storage_service_instance` or somesuch. (and maybe make it private since the goal is to have access through the storage_service getter)

db/migrate/20200513190742_backup_user_assets.rb
```diff
@@ -0,0 +1,10 @@
+class BackupUserAssets < Mongoid::Migration
+  def self.up
+    UserAssetService.push_assets_to_remote
+  end
+
+  def self.down
+    bucket = UserAssetService.get_storage_bucket
+    bucket.delete
```
devonbush, line 8: hmm...  I think this might be a rare case where we don't want a true 'undo' migration.  I'm imagining a scenario where we're rolling back production because of massive errors, and having the bucket as a backup source of user data could be a lifesaver.   Essentially, if we're rolling this back, I'd want to make sure the decision to delete the bucket doesn't happen automatically

app/lib/user_asset_service.rb
bistline: Ah - I see what you mean.  This method is in fact superfluous - `self.storage_service` can call `Google::Cloud::Storage.new` directly.  Good catch.

db/migrate/20200513190742_backup_user_assets.rb
bistline: Fair point.

app/lib/user_asset_service.rb
bistline: 87754ed

db/migrate/20200513190742_backup_user_assets.rb
bistline: f475a86

PR #585, opened 2020-05-14
Merge user identities in Mixpanel via Bard (SCP-2321)

eweitz:
This restores logging distinct users to Mixpanel, by merging unauth’d and auth’d user identities in Mixpanel via Bard.
 
Bard is a DSP service that mediates writes to Mixpanel. To enable tracking users when they are signed in and not, Bard provides an endpoint [`/api/identify`](https://terra-bard-dev.appspot.com/docs/) to merge identities.  In SCP's case, the anonynmous ID (`anonId`) is a random UUIDv4 string set as a cookie for all users -- auth'd or not -- upon visiting SCP.

This call links that `anonId` to the user's bearer token used by DSP's Sam  service.  That bearer token is in turn linked to a de-identified "distinct ID" used to track users across auth states in Mixpanel.

This satisfies SCP-2321.


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -6,23 +6,76 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+
+    begin
+      response = RestClient::Request.execute(
+        method: 'POST',
+        url: bard_path,
+        headers: headers,
+        payload: post_body
+      )
+    rescue RestClient::ExceptionWithResponse => e
+      Rails.logger.error "Bard error: #{e.message}"
```
bistline, line 55: Non-blocking:  It would be nice if we logged these errors to Sentry via the [ErrorTracker](https://github.com/broadinstitute/single_cell_portal_core/blob/master/lib/error_tracker.rb#L4) module.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -6,23 +6,76 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+      Rails.logger.error e
```
bistline, line 56: I'm not sure if this will parse by just logging the error object...

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -6,23 +6,76 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+      return nil
+    end
+
+    Rails.logger.info "Merging user identity in Mixpanel via Bard"
+
+    bard_domains_by_env = {
+      'development': 'https://terra-bard-dev.appspot.com',
+      'staging': 'https://terra-bard-alpha.appspot.com',
+      'production': 'https://terra-bard-prod.appspot.com'
+    }
```
devonbush, line 35: these should be moved to the appropriate config/environments/{devolopment|production|staging}.rb files as 
`config.bard_metrics_url = '<<url>>'`

then you can access them here as 
`Rails.application.config.bard_metrics_url`

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -6,23 +6,76 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+  # Bard is a DSP service that mediates writes to Mixpanel. To enable tracking
+  # users when they are signed in and not, Bard provides an endpoint
+  # `/api/identify` to merge identities.  In SCP's case, the anonynmous ID
+  # (`anonId`) is a random UUIDv4 string set as a cookie for all users --
+  # auth'd or not -- upon visiting SCP.
+
+  # This call links that anonId to the user's bearer token used by DSP's Sam
+  # service.  That bearer token is in turn linked to a deidentified
+  # "distinct ID" used to track users across auth states in Mixpanel.
+  def merge_identities
```
devonbush, line 22: I would make this method static and accept the user as an argument.  That will help if we need to write, e.g. a migration that merges a bunch of old users

app/controllers/users/omniauth_callbacks_controller.rb
eweitz: Done in c9f4d11.

app/controllers/users/omniauth_callbacks_controller.rb
eweitz: Fixed in 623918d.

app/controllers/users/omniauth_callbacks_controller.rb
eweitz: Done in 623918d.

PR #588, opened 2020-05-15
Updating ingest email messages, auth error logging (SCP-2420, SCP-2423)

bistline:
Ingest success emails will now report all annotation values parsed from a file, unless they are over 100 entries long, in which case the default `list too long for email` message will still appear.  This patch also updates error logging for API auth failures to include the auth token to allow for better error debugging.

This PR satisfies SCP-2420, SCP-2423.

### DIFF
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -3,6 +3,7 @@ module Api
     module Concerns
       module Authenticator
         extend ActiveSupport::Concern
+        OAUTH_V3_TOKEN_INFO = 'https://www.googleapis.com/oauth2/v3/tokeninfo'
 
         def authenticate_api_user!
           head 401 unless api_user_signed_in?
@@ -24,8 +25,7 @@ module Api
             if user.nil?
               # extract user info from access_token
               begin
-                token_url = "https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=#{api_access_token}"
-                response = RestClient.get token_url
+                response = RestClient.get OAUTH_V3_TOKEN_INFO + "?access_token=#{api_access_token}"
                 credentials = JSON.parse response.body
                 now = Time.zone.now
                 token_values = {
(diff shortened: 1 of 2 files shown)


app/models/ingest_job.rb
```diff
@@ -484,22 +484,20 @@ def get_annotation_message(annotation_source:, cell_annotation: nil)
     case annotation_source.class
     when CellMetadatum
       message = "#{annotation_source.name}: #{annotation_source.annotation_type}"
-      if annotation_source.can_visualize? || annotation_source.values.size == 1
+      if annotation_source.values.size < CellMetadatum::GROUP_VIZ_THRESHOLD.max || annotation_source.annotation_type == 'numeric'
         values = annotation_source.values.any? ? ' (' + annotation_source.values.join(', ') + ')' : ''
       else
         values = ' (List too large for email)'
```
eweitz, line 490: Non-blocking refinement to consider:

It'd be nice if this message had slightly more detail, e.g. how large the list was (`annotation_source.values.size`), and how large is too large for email (`CellMetadatum::GROUP_VIZ_THRESHOLD.max`).

app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -46,7 +46,8 @@ def current_api_user
                 error_context = {
                     auth_response_body: response.present? ? response.body : nil,
                     auth_response_code: response.present? ? response.code : nil,
-                    auth_response_headers: response.present? ? response.headers : nil
+                    auth_response_headers: response.present? ? response.headers : nil,
+                    api_access_token: api_access_token
                 }
```
devonbush, line 51: do we really want to log the token -- isn't that a security issue?  Should we maybe log some subset of the token or metainfo e.g. whether it's nil/malformed or somesuch?

app/models/ingest_job.rb
bistline: That's a nice refinement - I'll incorporate that.

app/controllers/api/v1/concerns/authenticator.rb
bistline: If it were a refresh token, I would agree - but OAuth tokens only last for an hour, so I think the issue here is negligible.  The main issue I'm trying to solve here is that when there's an error without a response, we don't know anything else about the request.  The token would be the only other piece of information we have.

Having said that, I think your approach of just logging that there _was_ a token would help - if we see errors saying there was no token.  Also, after some reading and testing, it does appear that this API endpoint will respond `400` if the token has expired, rather than `401`.  It looks like this updated endpoint behaves more normally: https://www.googleapis.com/oauth2/v3/userinfo.

I'm going to make some more substantive updates so that we have better behavior on this method for expired auth tokens.

devonbush: ok, then lets do that of just logging presence/absence (and maybe last 5 chars if it's present or somesuch).  I can see FISMA auditors not liking seeing tokens in our logs, even if we tell them that they were probably expired

bistline: After some more reading/testing, the above endpoint works, but doesn't give us any info about the _token_, just the _user_.  More reading has shown that the `400` on expired tokens for the token_info endpoints is expected behavior.  So we may just want to catch the `400` explicitly and treat it like a 401.

bistline: 346dbd9

app/models/ingest_job.rb
bistline: 3975cd3

PR #593, opened 2020-05-18
Ignoring study_name, default to accession (SCP-2422)

bistline:
Always default to using the study accession when loading a study, regardless of what is provided for the study name.  Since this update affects only `download_private_file` (all other download/study viewing methods use only accession), no further refactoring is required.

This PR satisfies SCP-2422.

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -580,8 +580,11 @@ class StudiesController < ApplicationController
   # method to download files if study is private, will create temporary signed_url after checking user quota
   def download_private_file
     @study = Study.find_by(accession: params[:accession])
-    # make sure user is signed in
-    if !user_signed_in? || !@study.can_view?(current_user)
+    # make study exists, then ensure user is signed in
+    if @study.nil?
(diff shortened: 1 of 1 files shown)


app/controllers/studies_controller.rb
```diff
@@ -579,7 +579,7 @@ def parse
 
   # method to download files if study is private, will create temporary signed_url after checking user quota
   def download_private_file
-    @study = Study.find_by(accession: params[:accession], url_safe_name: params[:study_name])
+    @study = Study.find_by(accession: params[:accession])
```
devonbush, line 582:
```suggestion
    @study = Study.find_by!(accession: params[:accession])
```
might not be necessary if there are other checks beforehand, but probably can't hurt

bistline: No, good point.  There aren't other checks, so this would fail downstream.  At least this way it will fail with a predictable error or `Record Not Found` from Mongo.

bistline: I went a different route - we handle missing studies and redirect for "public" download requests, so I'm applying similar logic here.

devonbush:
nice

PR #594, opened 2020-05-19
Adding service account workspace owner user group (SCP-2399)

bistline:
All Google accounts (including service accounts) have a quota on memberships in Google groups (2000 direct memberships, 5000 indirect memberships), and being added to the ACL on a Terra workspace adds that account to a user group that is specific to that workspace that grants the requested permission.  This means that we would have in effect a hard limit of 2000 studies in SCP (per instance) unless we begin pooling service accounts.  As part of ameliorating the impact on that quota, this update will add a single user group that will be an owner of all workspaces registered in the portal that are _not_ part of the "default" free Terra billing project.

Future work will be needed to redact all direct service account permissions on portal workspaces, but additional upstream work in the Terra Orchestration API is needed before that work can commence.

This PR satisfies SCP-2399.

### DIFF
--- a/app/models/admin_configuration.rb
+++ b/app/models/admin_configuration.rb
@@ -226,10 +226,13 @@ class AdminConfiguration
     ws_owner_group = groups.detect {|group| group['groupName'] == FireCloudClient::WS_OWNER_GROUP_NAME &&
         group['role'] == 'Admin'}
     # create group if not found
-    unless ws_owner_group.present?
-      ws_owner_group = Study.firecloud_client.create_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
+    if ws_owner_group.present?
+      ws_owner_group
+    else
+      # create and return group
+      Study.firecloud_client.create_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
+      Study.firecloud_client.get_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
     end
-    ws_owner_group
   end
 
   # getter to return all configuration options as a hash
--- a/app/models/admin_configuration.rb
+++ b/app/models/admin_configuration.rb
@@ -224,7 +224,7 @@ class AdminConfiguration
   def self.find_or_create_ws_user_group!
     groups = Study.firecloud_client.get_user_groups
     ws_owner_group = groups.detect {|group| group['groupName'] == FireCloudClient::WS_OWNER_GROUP_NAME &&
-        group['role'].downcase == 'owner'}
+        group['role'] == 'Admin'}
     # create group if not found
     unless ws_owner_group.present?
       ws_owner_group = Study.firecloud_client.create_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
--- a/app/models/admin_configuration.rb
+++ b/app/models/admin_configuration.rb
@@ -221,6 +221,17 @@ class AdminConfiguration
     end
   end
 
+  def self.find_or_create_ws_user_group!
+    groups = Study.firecloud_client.get_user_groups
+    ws_owner_group = groups.detect {|group| group['groupName'] == FireCloudClient::WS_OWNER_GROUP_NAME &&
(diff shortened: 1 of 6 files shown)


db/migrate/20200518190036_set_sa_group_owner_on_workspaces.rb
```diff
@@ -0,0 +1,27 @@
+# Backfilling persmisions on existing workspaces not in the 'default' portal project by adding a
```
eweitz, line 1:
```suggestion
# Backfilling permissions on existing workspaces not in the 'default' portal project by adding a
```

app/models/admin_configuration.rb
```diff
@@ -221,6 +221,20 @@ def self.set_readonly_service_account_permissions(grant_access)
...
+    groups = Study.firecloud_client.get_user_groups
+    ws_owner_group = groups.detect {|group| group['groupName'] == FireCloudClient::WS_OWNER_GROUP_NAME &&
+        group['role'] == 'Admin'}
+    # create group if not found
+    if ws_owner_group.present?
+      ws_owner_group
+    else
+      # create and return group
+      Study.firecloud_client.create_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
+      Study.firecloud_client.get_user_group(FireCloudClient::WS_OWNER_GROUP_NAME)
```
devonbush, line 234: huh, does the create not return the group itself?  That seem un-RESTful of Terra.

app/models/study.rb
```diff
@@ -3313,12 +3313,13 @@ def set_service_account_permissions
     # only perform check if this is not the default portal project
     if self.firecloud_project != FireCloudClient::PORTAL_NAMESPACE
       begin
+        sa_owner_group = AdminConfiguration.find_or_create_ws_user_group!
         client = FireCloudClient.new(self.user, self.firecloud_project)
-        service_account = Study.firecloud_client.issuer
-        acl = client.create_workspace_acl(service_account, 'OWNER', true, true)
+        group_email = sa_owner_group['groupEmail']
+        acl = client.create_workspace_acl(group_email, 'OWNER', true, false)
```
devonbush, line 3319: can you explain why we used to set compute_permission to true and now you're moving it to false?

db/migrate/20200518190036_set_sa_group_owner_on_workspaces.rb
```diff
@@ -0,0 +1,27 @@
+# Backfilling permissions on existing workspaces not in the 'default' portal project by adding a
+# Service Account owned user group to all workspaces
+# This is in preparation of revoking all direct SA ownerships on workspaces
```
devonbush, line 3: I REALLY like how you're not revoking the direct SA ownerships until this change has gone out and we've had time to test.

app/models/admin_configuration.rb
bistline: Actually, in looking at this, I think that I have a bug still.  The return object for those two endpoints are the same, but it's not the same as the top-level "show me all groups" response.  I need to make an update to account for that structure.  The information is all there, it's just formatted differently.

app/models/study.rb
bistline: In truth, setting it to `true` would never allow the service account to submit workflows.  Since SAs can't get refresh tokens, they can't submit to Cromwell.  Also we never used that permission ever, so I'm just accounting for what has been the de facto state all along.

app/models/admin_configuration.rb
bistline: So I was wrong - while the structure is different, the `groupEmail` value is still in the same place, and that's all we need from the response.

db/migrate/20200518190036_set_sa_group_owner_on_workspaces.rb
bistline: In truth, we don't have a way to do that yet (at least not programmatically), so even if I wanted to I can't.  But I agree that doing this in a multi-stage process is still better than all at once.

PR #598, opened 2020-05-20
putting focus on keyword input after clear

devonbush:


app/javascript/styles/_keywordSearch.scss
```diff
@@ -50,6 +50,9 @@
     svg {
       color: #aaa;
     }
+    &:focus {
+      outline: none;
+    }
```
eweitz, line 55: A few weeks ago I'd have supported this without reservation, but then I read https://a11yproject.com/posts/never-remove-css-outlines/.

For accessibility, some visible indication of focus is needed keyboard users.  The simplest solution is to preserve the browser default, per below; the article linked above gives a few other options too.  

```suggestion
```

devonbush: makes sense, thanks

PR #600, opened 2020-05-21
Refactoring usage of Mongoid associations for DataArrays, Genes (SCP-2439)

bistline:
Rather than using the auto-generated Mongoid association, this change directly leverages the DataArray and Gene models and their declared indices for better performance when creating, reading, and updating individual records.  This also changes the behavior of both getting unique genes for a study, and setting the unique gene count.  These updates are intended to ameliorate any potential runaway queries or inserts that can cause stability issues with the portal as a whole.

In addition, this update removes unnecessary updates to DataArrays when clusters are renamed, as the `cluster_name` value does not need to be used in the query to leverage the index (`linear_data_id` resolve to the same object).

This PR satisfies SCP-2439.

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -684,9 +684,11 @@ class StudiesController < ApplicationController
           @study.default_options[:cluster] = @study_file.name
           @study.save
         end
+        old_name = @cluster.name.dup
         @cluster.update(name: @study_file.name)
-        # also update data_arrays
-        @cluster.data_arrays.update_all(cluster_name: @study_file.name)
+        # also update data_arrays, using declared indices rather than Mongoid association for better performance
+        DataArray.where(study_id: @study.id, study_file_id: @cluster.study_file.id, cluster_name: old_name,
+                        linear_data_type: 'ClusterGroup', linear_data_id: @cluster.id).update_all(cluster_name: @study_file.name)
       elsif ['Expression Matrix', 'MM Coordinate Matrix'].include?(study_file_params[:file_type]) && !study_file_params[:y_axis_label].blank?
         # if user is supplying an expression axis label, update default options hash
         @study.update(default_options: @study.default_options.merge(expression_label: study_file_params[:y_axis_label]))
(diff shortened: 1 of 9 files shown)


app/controllers/studies_controller.rb
```diff
@@ -741,7 +743,11 @@ def update_study_file_from_sync
           @study.default_options[:cluster] = @study_file.name
           @study.save
         end
+        old_name = @cluster.name.dup
         @cluster.update(name: @study_file.name)
+        # also update data_arrays, using declared indices rather than Mongoid association for better performance
+        DataArray.where(study_id: @study.id, study_file_id: @cluster.study_file.id, cluster_name: old_name,
+                        linear_data_type: 'ClusterGroup', linear_data_id: @cluster.id).update_all(cluster_name: @study_file.name)
```
devonbush, line 750: maybe pull the query part of this into a class method on DataArray.  Something like
 ```
def self.find_for_cluster(cluster)
   return DataArray.where(study_id: @study.id, study_file_id: @cluster.study_file.id, cluster_name: old_name,
                        linear_data_type: 'ClusterGroup', linear_data_id: @cluster.id)
end
```

File: app/models/study.rb [hunk 2]
```diff
@@ -1095,15 +1091,26 @@ def default_cluster_point_alpha
...
     Rails.logger.info "Setting gene count in #{self.name} to #{gene_count}"
     self.update(gene_count: gene_count)
+    Rails.logger.info "Gene count set for #{self.name}"
+  end
+
+  # get all unique gene names for a study; leverage index on Gene model to improve performance
+  def unique_genes
+    genes = []
+    self.expression_matrix_files.each do |file|
+      genes += Gene.where(study_id: self.id, study_file_id: file.id).pluck(:name)
```
devonbush, line 1111: I might rewrite this as `genes = Gene.where(study_id: self.id, :study_file_id.in self.expression_matrix_files.map(&:id)).pluck(:name)`  (I'm not sure I'm remembering the mongoid 'in' syntax appropriately, but you get the idea).  Just to avoid the for loop

bistline: We actually do that in other places... good call.

bistline: 5d68378

app/controllers/studies_controller.rb
bistline: 73c287e

PR #604, opened 2020-05-22
Prevent users from deleting cluster/metadata files that are subsampling (SCP-2445)

bistline:
Currently a use can delete either a cluster or metadata file that is in the process of being subsampled.  This leads to various unexpected behavior modes, including a scenario where a cluster is "locked" in a mode where it stays labeled as in the process of subsampling, and cannot be unlocked without direct DB manipulation.

This update will block a user from deleting any file that is "locked", i.e being used in an active parse/subsampling job.  API requests to this effect will respond with an error and a status of `423 Locked`.  UI requests will show an alert notifying the user that they cannot delete this file until active jobs are done.  Emails notifying users of cluster file parse completions now also state that the cluster file will be subsampled, and that file (along with the metadata file) cannot be deleted while that job is in process.

This PR satisfies SCP-2445.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -345,6 +345,7 @@ class IngestJob
     when :ingest_subsample
       cluster_group = ClusterGroup.find_by(study_id: self.study.id, study_file_id: self.study_file.id)
       if cluster_group.is_subsampling? && cluster_group.find_subsampled_data_arrays.any?
+        Rails.logger.info "Setting subsampled flags for #{self.study_file.upload_file_name}:#{self.study_file.id} (#{cluster_group.name}) for visualization"
         cluster_group.update(subsampled: true, is_subsampling: false)
       end
     end
@@ -418,6 +419,12 @@ class IngestJob
           end
         end
         message << "Total points in cluster: #{cluster.points}"
+        # notify user that subsampling is about to run and inform them they can't delete cluster/metadata files
+        if cluster.can_subsample? && self.study.metadata_file.present?
+          message << "This cluster file will now be processed to compute representative subsamples for visualization."
+          message << "You will receive an additional email once this has completed."
+          message << "While subsamples are being computed, you will not be able to remove this cluster file or your metadata file."
+        end
       else
         message << "Subsampling has completed for #{cluster.name}"
         message << "Subsamples generated: #{cluster.subsample_thresholds_required.join(', ')}"
--- a/app/models/study_file.rb
+++ b/app/models/study_file.rb
@@ -906,7 +906,8 @@ class StudyFile
       when 'Metadata'
         !self.study.cluster_groups.where(is_subsampling: true).any?
       when 'Cluster'
-        !ClusterGroup.where(study_file_id: self.id).is_subsampling?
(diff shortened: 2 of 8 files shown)


app/models/ingest_job.rb
```diff
@@ -345,6 +345,7 @@ def set_subsampling_flags
     when :ingest_subsample
       cluster_group = ClusterGroup.find_by(study_id: self.study.id, study_file_id: self.study_file.id)
       if cluster_group.is_subsampling? && cluster_group.find_subsampled_data_arrays.any?
+        Rails.logger.info "Setting subsampled flags for #{self.study_file.upload_file_name}:#{self.study_file.id} (#{cluster_group.name}) for visualization"
```
devonbush, line 348: I like this log statement, nice

app/models/ingest_job.rb
```diff
@@ -418,6 +419,12 @@ def generate_success_email_array
           end
         end
         message << "Total points in cluster: #{cluster.points}"
+        # notify user that subsampling is about to run and inform them they can't delete cluster/metadata files
+        if cluster.can_subsample? && self.study.metadata_file.present?
+          message << "This cluster file will now be processed to compute representative subsamples for visualization."
+          message << "You will receive an additional email once this has completed."
+          message << "While subsamples are being computed, you will not be able to remove this cluster file or your metadata file."
+        end
```
devonbush, line 427: Possible other language below -- feel free to pick and choose sentences/phrases you like.   

This cluster file is now being subsampled for visualization. Please do not delete your cluster or metadata file while this process is occurring

You will receive an email once visualization processing is complete, letting you know this cluster is ready for viewing.

app/models/study_file.rb
```diff
@@ -896,6 +896,23 @@ def remove_local_copy
     Rails.logger.info "Removal of local copy of #{self.upload_file_name} complete"
   end
 
+  # check if this file can be deleted "safely"; e.g. not being used in any running parse jobs
+  # most files just need to check if they are still parsing; cluster/metadata files need to check for subsampling
+  def can_delete_safely?
```
devonbush, line 901: this logic is juuuust complex enough that it might be worth a model-level test.  E.g. just create a study file object with the file type and is_parsing  set/unset and manually create a clustergroup object for it with is_subsampling set/unset.  (this may not be possible depending on what kind of hooks/model validations have to pass, but if you can do it in <10mins, I think it's worth it)

bistline: I initially wasn't going to because I thought that it would involve another long-running integration test that would add several minutes to our CI runs, but on reflection your suggestion will be more than sufficient.  I'll put something together.

app/models/ingest_job.rb
bistline: My only issue with "please do not delete your file" is that it makes it seem as though they have the option, when in fact they do not.  Also, with the second option - they can view the cluster before it is subsampled, just only at full resolution.  Therefore I feel like both options are misleading.

Since I need to make further updates anyway, I'll ruminate over the weekend and see if I come up with something better.

app/models/study_file.rb
bistline: 5e47004

PR #622, opened 2020-06-08
simplifying detached logic, adding fallback workspace check (SCP-1924)

bistline:
This simplifies the logic for marking a study as "detached", which is how SCP labels studies that have had their associated Terra workspace and/or GCS bucket deleted.  Now instead of matching on error class, the handler will first look for common error messages associated with workspaces/buckets not being present, and then fall back to attempting to load the Terra workspace via the Orchestration API.  If any of these evaluate to being true, then the study is marked detached.  SCP Admins will still be notified the first time a study is found to be "detached", and subsequent sanity checks will then ignore this study.

This PR satisfies SCP-1924.

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -2978,14 +2978,19 @@ class Study
 
   # check whether a study is "detached" (bucket/workspace missing)
   def set_study_detached_state(error)
-    case error.class.name
-    when 'NoMethodError'
-      if error.message == "undefined method `files' for nil:NilClass"
-        Rails.logger.error "Marking #{self.name} as 'detached' due to missing bucket: #{self.bucket_id}"
-        self.update(detached: true)
-      end
-    when 'RuntimeError'
-      if error.message == "#{self.firecloud_project}/#{self.firecloud_workspace} does not exist"
+    # missing bucket errors should have one of three messages
+    #
+    # nil:NilClass => returned from a NoMethodError when calling bucket.files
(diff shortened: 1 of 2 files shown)


app/models/study.rb
```diff
@@ -2978,14 +2978,19 @@ def set_readonly_access(grant_access=true, manual_set=false)
...
-        self.update(detached: true)
-      end
-    when 'RuntimeError'
-      if error.message == "#{self.firecloud_project}/#{self.firecloud_workspace} does not exist"
+    # missing bucket errors should have one of three messages
+    #
+    # nil:NilClass => returned from a NoMethodError when calling bucket.files
+    # forbidden, does not have storage.buckets.get access => resulting from 403 when accessing bucket as ACLs
+    # have been revoked pending delete
+    if /(nil\:NilClass|does not have storage.buckets.get access|forbidden)/.match(error.message)
```
devonbush, line 2986: I might pull this regex out into a constant, but don't feel strongly about it

bistline:
> I'm assuming this is something that would be relatively difficult to unit/integration test, and is a rare enough occurrence that it's not worth that effort and the corresponding time increase in our test suite runs. But if by chance it is easy, then please add one.

It is a rare occurrence, and in order to recreate it, we would need to create studies through the portal, then remove the workspaces in Terra, and then wait up to 24 hours for the bucket to be removed.  Since the bucket is the entrypoint for the storage sanity check, and not the workspace, it didn't seem worth it to try to simulate this in a test.

PR #626, opened 2020-06-09
Support Macaca fascicularis genome in igv.js (SCP-2448)

eweitz:
This adds a special case of rudimentary support for _Macaca fascicularis_ (crab-eating macaque) in the igv.js genome browser embedded in Single Cell Portal.  It enables viewing BAMs for the genome assembly Macaca_fascicularis_5.0.

I generated the files referenced here manually.  A more generic and reproducible approach is under development to generate these files.  That is needed for other current SCP organisms that do not have built-in support in igv.js, and to address tech debt in this quick and dirty solution.  But this fixes a known user issue for now.

This satisfies SCP-2448.

### DIFF
--- a/app/assets/javascripts/scp-igv.js
+++ b/app/assets/javascripts/scp-igv.js
@@ -6,8 +6,6 @@
  * Provides a way to view nucleotide sequencing reads in genomic context.
  */
 
-console.log('start scp-igv.js')
-
 // Persists 'Genome' tab and IGV embed across view states.
 // Ensures that IGV doesn't disappear when user clicks 'Browse in genome' in
 // default view, then searches a gene.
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/scp-igv.js
```diff
@@ -140,15 +140,38 @@ function initializeIgv() {
...
 
   const genes = $('.queried-gene')
-  const locus = (genes.length === 0) ? ['myc'] : [genes.first().text()]
 
-  const genome = bamsToViewInIgv[0].genomeAssembly
+  const genomeId = bamsToViewInIgv[0].genomeAssembly
+
+  let reference
+  let locus
+  if (genomeId === 'Macaca_fascicularis_5.0') {
```
devonbush, line 148: this is nice -- I like how clearly this is hardcoded, rather than half-abstracted, given that this is the only kind of this assembly we have now.  This makes both this easier to read now, and hopefully will make subsequent refactoring to support more of these kinds of assemblies easier.

PR #634, opened 2020-06-15
Adding combined facet UI for cell type and organ (SCP-2510)

devonbush:
adds the new combined facet UI for organ and cell type
image
this involved a lot of refactoring the facet controls, including pulling the closeable modal logic into a custom hook

Also polishes the facet UI generally to smooth over rough edges from Q1's release and make it closer to the original design

image

Lots of linting fixes.


app/javascript/components/search/controls/CombinedFacetControl.js
```diff
@@ -0,0 +1,123 @@
...
+import _kebabCase from 'lodash/kebabCase'
+import _compact from 'lodash/compact'
+import _flatten from 'lodash/flatten'
+import _clone from 'lodash/clone'
+
+/**
+ * Component for combining two facets into a single panel control.
+ * They will be rendered more or less independently, but with a single 'apply' button
+ *
+ * Note that this does NOT support numeric facets -- to support those the display CSS would have to be
```
eweitz, line 19: This line is 101 characters long, which I believe throws a linting error.

If so, to keep linting high-signal, please either shorten long lines like this (preferable) or adjust the line length (`max-len`) style rule to something more lenient.  I prefer 80 characters, but not strongly.

Current style rule:
```
./node_modules/.bin/eslint --print-config app/javascript/components/search/controls/KeywordSearch.js  | grep -A 8 max-len
    "max-len": [
      2,
      {
        "code": 80,
        "tabWidth": 2,
        "ignoreUrls": true,
        "ignorePattern": "goog.(module|require)"
      }
    ],
```

app/javascript/components/search/controls/CombinedFacetControl.js
```diff
@@ -0,0 +1,123 @@
...
+                  { facetContents.length > 1 && <h4>{facetContent.facet.name}</h4> }
+                  <FiltersBoxSearchable
+                    show={showFilters}
+                    facet={facetContent.facet}
+                    setShow={ () => {} }
+                    selection={facetContent.selection}
+                    setSelection={selection =>
+                      selectionContext.updateFacet(facetContent.facet.id, selection)
+                    }
+                    hideControls={ true }/>
```
eweitz, line 101: Consistent padding (`{foo}` or `{ foo }`, but not both) would help readability.

app/javascript/components/search/controls/CombinedFacetControl.js
```diff
@@ -0,0 +1,123 @@
...
+                </div>
+            })
+            }
+          </div>
+          <div className='filters-box-footer'>
+            {canClear &&
+                <ClearFilters
+                  facetId={_kebabCase(controlName)}
```
eweitz, line 109: Neat!

app/javascript/components/search/controls/FacetsPanel.js
```diff
@@ -1,22 +1,32 @@
...
  * Container for horizontal list of facet buttons, and "More Facets" button
  */
 export default function FacetsPanel() {
   const searchFacetContext = useContext(SearchFacetContext)
-
+  const defaultFacets = searchFacetContext.facets.filter(facet => defaultFacetIds.includes(facet.id))
+  const moreFacets = searchFacetContext.facets.filter(facet => moreFacetIds.includes(facet.id))
   return (
     <>
+      <CombinedFacetControl controlName="cell type" facetIds={ ['cell_type', 'cell_type__custom'] }/>
```
eweitz, line 22: Delimiting with a space strikes me as slightly unusual for non-user-facing strings (underscore or hyphen being more typical).

This isn't necessarily an issue, but it seems worth verifying -- is space-delimiting intended here?

app/javascript/hooks/closeableModal.js
```diff
@@ -0,0 +1,58 @@
...
+  Example usage might be:
+
+  const [show, setShow] = useState(false)
+  const { node, clearNode, handleButtonClick } = useCloseableModal(show, setShow)
+  <span ref={node}>
+    <a onClick={handleButtonClick}>Toggle thingy</a>
+    <a ref={clearNode}>Close thingy</a>
+    { show && <thingy/> }
+  </span>
+ */
```
eweitz, line 22: Very helpful!

app/javascript/hooks/closeableModal.js
```diff
@@ -0,0 +1,58 @@
...
+export default function useCloseableModal(show, setShow) {
+  const clearNode = useRef()
+  /**
+    * Click on the facet control itself
+    */
```
eweitz, line 27: A few months ago you suggested single-line IDE-hookable docstrings (e.g. `/** Click on the facet control itself */`) for such brief comments.  I've found that useful and would do so here and in `clearFacet`.

app/javascript/providers/SearchSelectionProvider.js
```diff
@@ -10,6 +11,31 @@ export const SearchSelectionContext = React.createContext({
...
+  }
+  return selection
+}
+
+/** returns whether or not the current selection for the given facet object is different
+  * than what the most recently performed search was */
+export function isFacetApplicable(facet, selectionContext, searchContext) {
+  const selection = getSelectionForFacet(facet, selectionContext)
+  const appliedSelection = getAppliedParamsForFacet(facet, searchContext)
+  const isSelectionValid = facet.type != 'number' ||
```
eweitz, line 28: Strict equality, for consistency with comparison for positive case (`===` and `!==`).

```suggestion
  const isSelectionValid = facet.type !== 'number' ||
```

app/javascript/components/search/controls/CombinedFacetControl.js
```diff
@@ -0,0 +1,123 @@
...
+          <div className="multi-facet-container">
+            { facetContents.map((facetContent, index) => {
+              return facetContent.facet &&
+                <div className="single-facet" key={facetContent.facet.id}>
+                  { /* only show the facet names if there are more than one */ }
+                  { facetContents.length > 1 && <h4>{facetContent.facet.name}</h4> }
+                  <FiltersBoxSearchable
+                    show={showFilters}
+                    facet={facetContent.facet}
+                    setShow={ () => {} }
```
eweitz, line 96: This looks extraneous, at a glance.

devonbush:
> Looks good! I noted some code style nits, nothing blocking.
> 
> The screenshot helps convey what this does -- thanks for that. Some UI nits that aren't necessarily in scope here, but timely to note given the screenshot:
> 
> * 15 items is perhaps too many for a list called "Top filters".  The screenshot seems to be just testing data (stuff1 - stuff15), but if we'll likely encounter this many top filters in production data then it's probably worth somehow automatically keeping the "Top filters" shorter.
agree -- do you remember where 15 came from?  Did one of us just make up that number?
> * "Top filters" for Cell Type includes an item labeled with an empty-string.
yup, that's just my DB having some wonky data
> * The facet boxes are quite wide -- the amount of negative space at right in the Species box seems like too much.  I vaguely recall having this due to long ontology names linked beside "Top Filters".  If so, it might be worth shortening the display name of rare very-long ontology names and slightly narrowing the facet boxes.  (Long filter names could presumably be wrapped.)
> 
That's actually how much space is in the original design https://projects.invisionapp.com/d/main#/console/19272801/402387753/preview. I think it looks a bit wider in the screenshot since I didn't grab the entire screen
> None of these minor UI concerns are blocking.

app/javascript/components/search/controls/FacetsPanel.js
devonbush: I definitely didn't intend to use space delimiting -- where do you see it?

app/javascript/hooks/closeableModal.js
devonbush: agree

app/javascript/providers/SearchSelectionProvider.js
devonbush: good catch

app/javascript/components/search/controls/CombinedFacetControl.js
devonbush: yup

app/javascript/components/search/controls/CombinedFacetControl.js
devonbush: the idea is that filtersBoxSearchable expects a function for this argument, and so I'm passing it a no-op.  The alternative is to add checks before any calls in filtersBoxSearchable to setShow that setShow exists. I could see either one being correct

app/javascript/components/search/controls/CombinedFacetControl.js
devonbush: yeah, I remember the discussion when we agreed on 80 -- our code is unfortunately chock-full of longer lines (I'm definitely a guilty party, though not the only one).  Let's have a brief chat with Eno tomorrow about whether it's worth cleaning them all up, or increasing the length.  (I lean towards longer lines, but not strongly, so she can break the tie :)   Once we settle that, then we can add linting to the server-side PR checks

app/javascript/components/search/controls/CombinedFacetControl.js
devonbush: added an eslint rule for it

app/javascript/components/search/controls/FacetsPanel.js
eweitz: Sorry, I should've been more specific -- I was referring to `cell name`.  I mistakenly thought it was not user-facing.  No issues here.

> Maybe rename 'controlName' to 'controlDisplayName'?

A helpful measure for easily-confused readers like me :).

app/javascript/components/search/controls/CombinedFacetControl.js
eweitz: Agreed, either works.  Thanks for helping me understand here.

app/javascript/components/search/controls/CombinedFacetControl.js
eweitz: Sounds good!

eweitz:

app/models/search_facet.rb
```diff
@@ -358,12 +358,14 @@ def generate_bq_query_string
   def generate_array_query
     "SELECT DISTINCT id, name FROM(SELECT id_col AS id, name_col as name " + \
     "FROM #{CellMetadatum::BIGQUERY_TABLE}, UNNEST(#{self.big_query_id_column}) AS id_col WITH OFFSET id_pos, " + \
-    "UNNEST(#{self.big_query_name_column}) as name_col WITH OFFSET name_pos WHERE id_pos = name_pos)"
+    "UNNEST(#{self.big_query_name_column}) as name_col WITH OFFSET name_pos WHERE id_pos = name_pos) WHERE id IS NOT NULL " + \
```
bistline, line 361: I'm unclear how we could have a situation where `id` is `NULL` - is this because these new "custom" facets aren't pairs, and are not required?

test/api/search_controller_test.rb
```diff
@@ -430,4 +430,101 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
+
+  sample_disease_facet = SearchFacet.new(identifier: 'disease',
```
bistline, line 434: Is there a reason why you are declaring these facets outside of the test or setup block?  Since we have facets already being created in `db/seeds.rb`, and real data going into BigQuery from upstream integration tests, shouldn't we leverage those instead?  I'm all for a self-contained test, but this seems strange to me.

app/controllers/api/v1/search_controller.rb
```diff
@@ -683,45 +683,32 @@ def self.generate_bq_query_string(facets)
...
-            max_value = facet_obj[:filters][:max]
-            unit = facet_obj[:filters][:unit]
-            if search_facet.must_convert?
-              query_on = search_facet.big_query_conversion_column
-              min_value = search_facet.calculate_time_in_seconds(base_value: min_value, unit_label: unit)
-              max_value = search_facet.calculate_time_in_seconds(base_value: max_value, unit_label: unit)
-            end
-            where_clauses << "#{query_on} BETWEEN #{min_value} AND #{max_value}"
+        or_facets = [['cell_type', 'cell_type__custom'], ['organ', 'organ_region']]
+        leading_or_facets = or_facets.map{|g| g[0]}
```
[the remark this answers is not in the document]
devonbush: good point, concur

app/controllers/api/v1/search_controller.rb
```diff
@@ -683,45 +683,32 @@ def self.generate_bq_query_string(facets)
...
+        trailing_or_facets = or_facets.map{|g| g[1]}
```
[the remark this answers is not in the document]
devonbush: yup

app/controllers/api/v1/search_controller.rb
```diff
@@ -683,45 +683,32 @@ def self.generate_bq_query_string(facets)
...
+        or_grouped_where_clause = nil
+        # sort the facets so that OR'ed facets will be next to each other
+        sorted_facets = facets.sort_by {|facet| or_facets.flatten.find_index(facet[:id]) || 99}
```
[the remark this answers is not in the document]
devonbush: yup, I'll add a comment

app/controllers/api/v1/search_controller.rb
```diff
@@ -683,45 +683,32 @@ def self.generate_bq_query_string(facets)
...
+
+        sorted_facets.each_with_index do |facet_obj, index|
+          query_elements = get_query_elements_for_facet(facet_obj)
+          from_clause += ", #{query_elements[:from]}" if query_elements[:from]
+          base_query += ", #{query_elements[:select]}" if query_elements[:select]
+          with_clauses << query_elements[:with] if query_elements[:with]
+
+          # check if we're at the start of a pair of facets that should be grouped by OR
+          or_group_index = leading_or_facets.find_index(facet_obj[:id])
```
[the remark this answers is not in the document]
devonbush: yeah, agree this needs some rework, I like pulling some of the complexity out of the if statement.  Not sure about the 'next', let me know your thoughts on how it looks with much better commenting and your logic grouping improvements

app/models/search_facet.rb
devonbush: yeah, it's possible that it's just relics from wonky populates in my development db, but I had null values that were throwing off the results.  It might not be a real risk for production, but I figure it can't hurt

test/api/search_controller_test.rb
devonbush: yeah, I see how weird this looks.  the goal was self-contained unit tests, (which is why I took the time to refactor generate_bq_query_string so that it took the actual db facet rather than an id as part of its parameters).  Then I realized all four of the unit tests wanted the same facets, so I pulled them out of the indvidiual tests.  It doesn't feel good to me to have tests this low-level rely on db/seed, since these are just validating a single method with exact string comparison asserts.  Maybe group the tests in a block/module?  I'm open to suggestions

bistline: Yeah, maybe take these tests out and make them their own module/file, and then have those facets declared in the `setup` block.  That would alleviate my concerns over the construction of the tests.

devonbush: cool, I pulled them into a separate file so they can have a dedicated setup method -- agree, it looks better this way

PR #636, opened 2020-06-17
remove MONDO_0000001 from synthetic data (SCP-2504)

jlchang:
In SCP-2481, we manually removed MONDO_0000001 from the metadata backend for production and staging. Rationale: having "disease or disorder" is confusing as a proxy for "normal". Replace instances of "MONDO_0000001/disease or disorder" in the backend with "PATO_0000461/normal" to reduce potential for confusion. Removing MONDO_0000001 from our synthetic files will prevent re-introduction of this source of confusion.

This satisfies SCP-2504.


eweitz:

PR #637, opened 2020-06-18
API Bulk/Single file download updates (SCP-2512, SCP-2513, SCP-2514)

bistline:
Refactoring authentication and download logic for file download requests to the API:

* SCP-2514: Bulk download requests correctly ensure user is authenticated via `auth_code`, and responds `401 Unauthorized` if a bad/missing `auth_code` is provided.  Additionally, user permissions are now checked per study accession to ensure user requesting bulk download has permission to download from the requested studies.  This was enforced previously at the search level, but not at the download level. 
* SCP-2513: Externally stored sequence data (Fastq, BAM, etc) now returns correct URL for downloads, and is excluded from bulk download requests since data is not in study bucket, and may require additional authentication to retrieve
* SCP-2512: Adding documentation detailing that file download requests do not function from the Swagger documentation due to CORS issues (not resolvable, but `curl` commands with `-L` flags will function as expected).  Also updating the timeout on `signed_urls` in single-file download to 1 minute to allow users extra time to issue the `curl` command.

Also, this PR addresses issues with GCP firewall rules for new staging assets provisioned by Terraform by making deployed hosts (`production`, `staging`, `pentest`) access MongoDB via the internal IP address, and `development` and `test` environments via the external IP address.


app/controllers/api/v1/site_controller.rb
```diff
@@ -94,7 +97,9 @@ def view_study
               'Site'
           ]
           key :summary, 'Download a StudyFile'
-          key :description, 'Download a single StudyFile (via signed URL)'
+          key :description, "Download a single StudyFile (via signed URL)<br/><br/><strong>NOTE</strong>: Due to CORS issues, files cannot be " + \
+                            "downloaded via Swagger.  To download a file, either use a client such as Postman, or copy/paste " + \
+                            "the CURL command into a terminal and add the '-L' flag.".html_safe
```
eweitz, line 102: This works, assuming we can't hide the "Request URL" nor add the curl `-L` flag.  But a refinement for the note would help:

```suggestion
                            "the CURL command into a terminal and add the '-L' flag immediately before the URL.".html_safe
```

bistline: I tried to figure out how to add in the `-L` flag, but it seems that the curl code is auto-generated in the Swagger JS, so we'd have to monkey-patch it, and only for this method.  That didn't seem like it would be very feasible.

eweitz: I see -- agreed!

app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -71,9 +71,14 @@ def current_api_user
...
+            return nil if params[:auth_code].blank?
             user = User.find_by(totat: params[:auth_code].to_i)
-            user.update_last_access_at!
-            user
+            if user.present?
+              user.update_last_access_at!
+              user
+            else
+              nil
+            end
```
devonbush, line 81:
```suggestion
           user.try(:update_last_access_at!)
           user
```

app/views/api/v1/site/_study_file.json.jbuilder
```diff
@@ -3,5 +3,9 @@ json.set! :file_type, study_file.file_type
 json.set! :description, study_file.description
 json.set! :bucket_location, study_file.bucket_location
 json.set! :upload_file_size, study_file.upload_file_size
-json.set! :download_url, api_v1_site_study_download_data_url(accession: study.accession, filename: study_file.bucket_location)
-json.set! :media_url, api_v1_site_study_stream_data_url(accession: study.accession, filename: study_file.bucket_location)
\ No newline at end of file
+if study_file.human_fastq_url.present?
+  json.set! :download_url, study_file.human_fastq_url
```
devonbush, line 7: nice

test/api/site_controller_test.rb
```diff
@@ -77,4 +77,65 @@ class SiteControllerTest < ActionDispatch::IntegrationTest
     @study.update(detached: false)
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
+
+  test 'should download file' do
```
devonbush, line 81: nice, these tests are good additions

PR #638, opened 2020-06-22
changing normal keyword search to POST from GET (SCP-2444)

bistline:
This addresses an issue with serializing query parameters back into the DOM when a user has supplied non-ASCII characters in their search request.  This is only affecting the normal, non-React search front end.  On a `GET` with non-ASCII query string parameters that are used somewhere in the DOM, an `ActionView::Template::Error: incompatible character encodings: ASCII-8BIT and UTF-8` error is thrown.  A `POST` with the same parameters does not throw this error.  

This PR satisfies SCP-2444.

### DIFF
--- a/app/lib/request_utils.rb
+++ b/app/lib/request_utils.rb
@@ -73,7 +73,7 @@ class RequestUtils
   # safely strip unsafe characters and encode search parameters for query/rendering
   # strips out unsafe characters that break rendering notices/modals
   def self.sanitize_search_terms(terms)
-    inputs = terms.is_a?(Array) ? terms.join(',') : terms
+    inputs = terms.is_a?(Array) ? terms.join(',') : terms.to_s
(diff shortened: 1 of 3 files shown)


bistline:
Note: There is no added test for this as it is covered by an existing UI test:

```ui_test_suite.rb: 'test: front-end: search'
ui_test_suite.rb: 'test: front-end: search' successful!
.
Finished in 139.447991 seconds.
-------------------------------------------------------------------------------------
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------
0.01 tests/s, 0.01 assertions/s
```

bistline:
> Normally I would object to changing a search from POST to GET, because it degrades UX by making searching not linkable. But given our upcoming switch to React -- which searches via GET -- I think this is a reasonable short-term solution. (I assume this bug isn't reproducible in the React search front-end.)

This is correct - the bug only affects the non-React keyword search.

PR #646, opened 2020-06-24
Improve igv.js embed: Gene search and better default coordinates for custom genomes, and multiple BAMs (SCP-2505)

eweitz:
This enhances the igv.js embed in Single Cell Portal.  

It extends support for custom genome references (introduced in #626) with support for gene search, and better default coordinates.  Previously, searching a gene like "BRCA1" in igv.js was not possible in a custom genomes (i.e. those without built-in support in igv.js), e.g. Macaca_fascicularis_5.0.  This is now possible via some patches to igv.js to support the [Ensembl REST API](https://rest.ensembl.org/documentation/info/symbol_lookup) as a custom igv.js search backend.  The patches exist in [a new SCP-scoped igv.js package](https://www.npmjs.com/package/@single-cell-portal/igv) ([source](https://github.com/igvteam/igv.js/compare/master...broadinstitute:ensembl-search-support)) -- a short-term fork before merging upstream.  I've run that new package by Devon and the IGV team, but welcome more review here too.

Additionally, an issue with showing multiple BAM tracks has been fixed.  All BAM tracks now display by default; the user can hide them as desired through the igv.js UI.

This satisfies SCP-2505.


app/assets/javascripts/scp-igv.js
```diff
@@ -139,42 +139,74 @@ function initializeIgv() {
...
     //  - Update genomes pipeline to make such files automatically reproducible
     const genomeAnnotationObj = bamsToViewInIgv[0].genomeAnnotation
-    const genomePath = `${genomeAnnotationObj.link.split('/').slice(0, -2).join('%2F')}%2F`
+    const genomePath =
+      `${genomeAnnotationObj.link.split('/').slice(0, -2).join('%2F')}%2F`
     const bucket = genomeAnnotationObj.bucket_id
     const gcsBase = 'https://www.googleapis.com/storage/v1/b/'
     const macacaFascicularisBase = `${gcsBase + bucket}/o/${genomePath}`
 
+    const fasta = 'Macaca_fascicularis.Macaca_fascicularis_5.0.dna.toplevel.fa'
```
devonbush, line 159: would this section read easier if all the macaca constants were defined in a single object?  e.g. 
```
const macaca_genome_info = {
 genomeId: 'Macaca_fascicularis_5.0', 
fasta: '...',
  url: '...',
 cytoband: '...'
}
```

package.json
```diff
@@ -23,7 +23,7 @@
     "camelcase-keys": "^6.1.2",
     "core-js": "^3.6.4",
     "ideogram": "1.19.0",
-    "igv": "2.1.0",
+    "@single-cell-portal/igv": "2.5.6-beta.2",
```
devonbush, line 26: nice, yeah, I'm definitely glad you took the effort to make the repo rather than manually including the file

app/assets/javascripts/scp-igv.js
```diff
@@ -139,42 +139,74 @@ function initializeIgv() {
...
 
-  const genes = $('.queried-gene')
-
   const genomeId = bamsToViewInIgv[0].genomeAssembly
 
   let reference
-  let locus
+  let searchOptions
+  let fallbackLocus
   if (genomeId === 'Macaca_fascicularis_5.0') {
```
devonbush, line 147: is this something that will get picked up in our analytics?  e.g. will we be able to answer the question "How many people are using IGV to view macaca genomes?"   Will we be able to answer the more general question of how often/what do people use IGV in SCP for?

app/assets/javascripts/scp-igv.js
eweitz: I started this refactor, but found it's basically redundant with the `reference` object a few lines below, without quite the latter's succinctness.

app/assets/javascripts/scp-igv.js
eweitz: > will we be able to answer the question "How many people are using IGV to view macaca genomes?"

No, not yet.  SCP-2506 is a prerequisite.  That ticket has value beyond igv.js tracking.  This is the second time this question has been asked recently, so I've nominated the ticket for next sprint.

> Will we be able to answer the more general question of how often/what do people use IGV in SCP for?

_How often_ -- yup!

_What do people use it for_ -- not with analytics.  Qualitative interviews and value statements from users (e.g. original user report [here](https://broadinstitute.zendesk.com/agent/tickets/5718)) can answer this question more efficiently than quantitative analytics.  I'd love to set up some user interviews!

PR #650, opened 2020-06-26
Allowing for filenames with spaces in upload wizard (SCP-1953)

bistline:
This bug fix allows for user-uploaded files to have spaces in their names (this is converted to `_` server-side to avoid path issues).  Previously, if a file was larger that 10MB and contained a space in the name, the upload process would incorrectly view each chunk as a new file due to the mismatch in updated and original filenames, which then throws a `422 Unprocessable Entity` error due to "duplicate" filenames.

In addition, this update also adds an integration test for chunked uploading as well as spaces in filenames.

This PR satisfies SCP-1953.

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -445,8 +445,9 @@ class StudiesController < ApplicationController
   # method to perform chunked uploading of data
   def do_upload
     upload = get_upload
-    # remove spaces from filename since this converted client-side, but original_filename is unchanged in headers
-    filename = upload.original_filename.gsub(/\s/, '_')
+    # remove spaces and + (encoded whitespace character) from filename since this converted client-side,
+    # but original_filename is unchanged in headers
+    filename = upload.original_filename.gsub(/[\s\+]/, '_')
     study_file = @study.study_files.detect {|sf| sf.upload_file_name == filename}
     # If no file has been uploaded or the uploaded file has a different filename,
     # do a new upload from scratch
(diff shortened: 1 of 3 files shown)


app/controllers/studies_controller.rb
```diff
@@ -445,7 +445,9 @@ def new_study_file
   # method to perform chunked uploading of data
   def do_upload
     upload = get_upload
-    filename = upload.original_filename
+    # remove spaces and + (encoded whitespace character) from filename since this converted client-side,
+    # but original_filename is unchanged in headers
+    filename = upload.original_filename.gsub(/[\s\+]/, '_')
```
eweitz, line 450: Does this mean that the names of downloaded files would have a "_" if the original file name had a space?

If so, names of downloaded files would not match those of uploaded files in that rare case.  But I can't think of any realistic concrete cases where that would be a issue -- so this seems OK.

bistline: This really is only affecting the query when trying to figure out if this is a "new" upload, or the next chunk in an "existing" upload.  The filename conversion already happens client-side in the jQuery upload code, but since the `original_filename` header is unchanged and is being used for the query, that mismatch caused the query to return `nil`, which then throws an error when it tries to insert another `StudyFile` with the converted filename.

The end result is that _all_ uploaded files have spaces removed from their filenames, and that's what shows up in the download tab.  This has been the behavior for uploads in SCP since the beginning.

PR #657, opened 2020-07-08
Log branding group to Mixpanel (SCP-2568)

eweitz:
This logs the SCP branding group for all events to Mixpanel, by introducing the `brand` event property.  It will enable analysts to compare user behavior between default SCP and branded spaces / branding groups, like the Alexandria Project.

This satisfies SCP-2568.

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -6,6 +6,7 @@
  */
 
 import { accessToken } from 'providers/UserProvider'
+import { getBrandingGroup } from 'lib/scp-api'
 
 let metricsApiMock = false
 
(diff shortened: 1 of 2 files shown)


app/javascript/lib/metrics-api.js
```diff
@@ -200,6 +201,12 @@ export function log(name, props={}) {
     props['featuredSpace'] = window.SCP.featuredSpace
   }
 
+  if ('SCP' in window && window.location.search.includes('scpbr=')) {
```
devonbush, line 204: Why do you need the check for window.location.search.includes('scpbr=')?  (or the check for SCP in window?)

 Could you do 
```
const brandingGroup = getBrandingGroup() 
props['brand'] = brandingGroup ? brandingGroup : ''
```
instead?

eweitz: Good point, done in a08cd64.

PR #661, opened 2020-07-14
Changing SingleCellMailer.users_email behavior (SCP-2586)

bistline:
The "All Users" email for admins was recently changed to deliver emails in batches of 500.  It was then discovered that blocks in ActionMailer methods that call `mail` only perform deliveries for the last entry in the block.  This update changes that behavior to iterate over a block in a separate method called `batch_users_email` that then calls the `users_email` method to deliver an email to a batch of 500 users.

This was tested manually in a local instance using a batch size of `2`, and it was confirmed that each email address in the total list of users received the test email.

This PR satisfies SCP-2586.

### DIFF
--- a/app/mailers/single_cell_mailer.rb
+++ b/app/mailers/single_cell_mailer.rb
@@ -7,18 +7,22 @@ class SingleCellMailer < ApplicationMailer
   ###
 
   default from: 'no-reply@broadinstitute.org'
+  BATCH_DELIVERY_SIZE = 500
+
+  # deliver an email to a batch users (from SingleCellMailer.batch_users_email)
+  def users_email(users, from, subject, message)
+    @message = message
+    mail(to: from, bcc: users, subject: "[Single Cell Portal Message] #{subject}", from: from) do |format|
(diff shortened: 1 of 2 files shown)


app/mailers/single_cell_mailer.rb
```diff
@@ -7,18 +7,22 @@ class SingleCellMailer < ApplicationMailer
...
-    @from = email_params[:from]
-    @message = email_params[:contents]
-    # send email in batches to reduce likelihood of errors
-    @users.each_slice(500) do |email_batch|
-      mail(to: @from, bcc: email_batch, subject: "[Single Cell Portal Message] #{@subject}", from: @from) do |format|
-        format.html {@message.html_safe}
-      end
+  def batch_users_email(email_params, user)
+    users = email_params[:preview] == "1" ? [user.email] : User.where(admin_email_delivery: true).map(&:email)
+    # send email in batches to sidestep Sendgrid API quotas
```
eweitz, line 23:
```suggestion
    # send email in batches to sidestep SendGrid API quotas
```

PR #662, opened 2020-07-15
Refactoring feature flags functionality into FeatureFlaggable concern (SCP-2578)

bistline:
Functionality relating to feature flags (formerly only associated with the `User` model) is now abstracted into a Rails concern that can be included into any persisted model, which standardizes the interface and reduces potentially duplicated code.  In addition, the `BrandingGroup` model now implements feature flags.

This PR satisfies SCP-2578.

### DIFF
--- a/app/models/branding_group.rb
+++ b/app/models/branding_group.rb
@@ -2,7 +2,7 @@ class BrandingGroup
   include Mongoid::Document
   include Mongoid::Timestamps
   include Mongoid::Paperclip
-  include Featureable
+  include FeatureFlaggable
 
   field :name, type: String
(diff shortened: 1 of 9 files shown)


bistline:
> I haven't fully digested the best way to indicate priority across multiple instances (e.g. get current applicable feature flags given a user and a branded space, with user flags taking priority). I wonder if we should have a template helper method like 'get currently applicable feature flags' that takes into account both the user and branded space
> (which may be beyond the scope of this PR)

That's an interesting point, but I agree - outside the scope of this PR.  But I could see us adding something akin to `priority` to FFs.

PR #664, opened 2020-07-16
Adding visible to SearchFacet, facet_list to BrandingGroup (SCP-2575)

bistline:
Search facets now have a `visible` attribute that will govern their default visibility.  Facets with `visibility: false` can still be used for queries, but do not show up in the UI, or in API calls for available facets.  In addition, branding groups now have an attribute called `facet_list`, which is a whitelist of search facet identifiers that will control which facets are shown when viewing a branding group on the homepage.  The default behavior is now all "visible" facets will show if no `facet_list` has been configured.

This PR satisfies SCP-2575.

### DIFF
--- a/app/models/branding_group.rb
+++ b/app/models/branding_group.rb
@@ -12,6 +12,9 @@ class BrandingGroup
   field :font_color, type: String, default: '#333333'
   field :feature_flags, type: Hash, default: {}
 
+  # list of facets to show for this branding group (will restrict to only provided identifiers, if present)
+  field :facet_list, type: Array, default: []
+
   has_many :studies
   belongs_to :user
 
@@ -54,6 +57,11 @@ class BrandingGroup
 
   before_validation :set_name_as_id
   before_destroy :remove_branding_association
+
+  # helper to return list of associated search facets
+  def facets
+    self.facet_list.any? ? SearchFacet.where(:identifier.in => self.facet_list) : SearchFacet.visible
+  end
 
   private
 
--- a/test/api/search_controller_test.rb
+++ b/test/api/search_controller_test.rb
@@ -20,10 +20,41 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
   test 'should get all search facets' do
     puts "#{File.basename(__FILE__)}: #{self.method_name}"
 
-    facet_count = SearchFacet.count
+    facet_count = SearchFacet.visible.count
     execute_http_request(:get, api_v1_search_facets_path)
     assert_response :success
     assert json.size == facet_count, "Did not find correct number of search facets, expected #{facet_count} but found #{json.size}"
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
+  end
+
+  test 'should get visible search facets' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    # make one random facet not visible
+    invisible_facet = SearchFacet.all.sample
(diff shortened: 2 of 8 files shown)


app/models/branding_group.rb
```diff
@@ -12,6 +12,9 @@ class BrandingGroup
   field :font_color, type: String, default: '#333333'
   field :feature_flags, type: Hash, default: {}
 
+  # list of facets to show for this branding group (will restrict to only provided identifiers, if present)
+  field :facet_list, type: Array, default: []
```
devonbush, line 16: this is good.  I can't think of a case where we'd want to have a branding group with no facets (but still have advanced search), so I like how you're using the empty array as the default so we don't have to have nil checks everywhere

test/api/search_controller_test.rb
```diff
@@ -20,14 +20,45 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
 
+  test 'should get visible search facets' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    # make one random facet not visible
+    invisible_facet = SearchFacet.all.sample
+    invisible_facet.update(visible: false)
```
devonbush, line 36:
```suggestion
    invisible_facet.update!(visible: false)
```
To avoid case where the update silently fails, and then the test isn't testing what we want

devonbush: For belt-and-suspenders, I would also add an `assert SearchFacet.visible.count == SearchFacet.count - 1`

test/api/search_controller_test.rb
```diff
@@ -20,14 +20,45 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
+  end
+
+  test 'should get search facets for branding group' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    branding_group = BrandingGroup.first
+    facet_list = SearchFacet.pluck(:identifier).take(2).sort
+    branding_group.update(facet_list: facet_list)
```
devonbush, line 51:
```suggestion
    branding_group.update!(facet_list: facet_list)
```

test/api/search_controller_test.rb
```diff
@@ -20,14 +20,45 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
+    execute_http_request(:get, api_v1_search_facets_path(scpbr: branding_group.name_as_id))
+    assert_response :success
+    response_facets = json.map {|entry| entry['id']}.sort
+    assert response_facets == facet_list,
+           "Did not find correct facets for #{branding_group.name_as_id}, expected #{facet_list} but found #{response_facets}"
+    branding_group.update(facet_list: [])
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
+  end
+
```
devonbush, line 61: this is a really nice test -- concise, and tests exactly what we want to test

test/api/search_controller_test.rb
bistline: Great point.

bistline: a30a464

test/api/search_controller_test.rb
bistline: a30a464

PR #665, opened 2020-07-17
disabling common edit/update functions

devonbush:
I've created a new branch 'old-production' that is built to track what is currently on the (soon-to-be) old production server.  This PR will merge in code that disables/hides most common editing operations. Then we can do a release to production from the old-production branch this afternoon immediately before the migration


eweitz:
Linking this to a Jira ticket -- whether existing or new -- would help traceability.

PR #666, opened 2020-07-17
Log performance and other Ingest Pipeline metrics from Rails to Mixpanel (SCP-2433, SCP-2553)

eweitz:
This logs Ingest Pipeline metrics like latency (i.e. duration or time taken, as `perfTime`), file type, and file size to Mixpanel.  This lets us analyze how ingest performance correlates with file size, file type, and user interactions in SCP.

The newly-tracked event is named "ingest" in Mixpanel.  See the ["Live View" in Mixpanel development](https://mixpanel.com/report/2085496/view/19055/live#chosen_columns:!('$browser','$city',mp_country_code,distinct_id,'$referring_domain'),column_widths:!(200,100,219,149,218,250,180),search:ingest).  Click on an event there to see the various properties logged for it.

The Ruby library `metrics_service.rb` essentially ports the JavaScript library [`metrics-api.js`](https://github.com/broadinstitute/single_cell_portal_core/blob/3c133bab4b11a3ef224b8645dee0f3b63198813f/app/javascript/lib/metrics-api.js).  Both write logs to Mixpanel via [Bard](https://terra-bard-prod.appspot.com/docs/), the DSP service that mediates writes to Mixpanel.

This satisfies SCP-2433 and SCP-2553.

### DIFF
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -77,13 +77,10 @@ class MetricsService
       env: Rails.env
     })
 
-    puts 'user'
-    puts user
-
-    access_token = user.access_token['access_token'] || ''
-    user_id = user.id
-
     headers = get_default_headers(user)
+
+    access_token = user.access_token['access_token']
+    user_id = user.id
 
     if access_token === ''
       # User is unauthenticated / unregistered / anonynmous
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -22,7 +22,7 @@ class MetricsService
       # Uncomment line below to get known-good test data
       # puts "Posting to Mixpanel.  Params: #{params}"
       RestClient::Request.execute(params)
-    rescue RestClient::ExceptionWithResponse => e
+    rescue RestClient::Exception => e
       Rails.logger.error "#{Time.zone.now}: Bard error in call to #{params[:url]}: #{e.message}"
       # Rails.logger.error e.to_yaml
       ErrorTracker.report_exception(e, user, params)
@@ -30,8 +30,8 @@ class MetricsService
   end
 
   def self.get_default_headers(user)
-    return {
-      'Authorization': "Bearer #{user.access_token['access_token']}",
+    {
+      'Authorization': "Bearer #{user.valid_access_token['access_token']}",
       'Content-Type': 'application/json'
     }
   end
@@ -82,7 +82,7 @@ class MetricsService
 
     headers = get_default_headers(user)
 
-    access_token = user.access_token['access_token']
+    access_token = user.valid_access_token['access_token']
     user_id = user.id
 
     if access_token === ''
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -71,6 +71,7 @@ class MetricsService
   #
   # @param {String} name Name of the event
   # @param {Hash} props Properties associated with the event
+  # @param {User} user User model object
   def self.log(name, props={}, user)
     Rails.logger.info "Logging analytics to Mixpanel for event name: #{name}"
 
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -13,8 +13,6 @@ class MetricsService
 
   BARD_ROOT = Rails.application.config.bard_host_url
 
-  EVENT_PATH = BARD_ROOT + '/api/event'
-
   def self.post_to_bard(params, user)
     params.merge!({
       method: 'POST'
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -72,9 +72,9 @@ class MetricsService
   # @param {Hash} props Properties associated with the event
   def log(name, props = {}, user, cookies)
     props.merge!({
-      appId: 'single-cell-portal',
-      timestamp: Time.now.in_milliseconds,
-      env: Rails.env
+      :appId => 'single-cell-portal',
+      :timestamp => Time.now.in_milliseconds,
+      :env => Rails.env
     })
 
     access_token = user.access_token['access_token']
@@ -82,14 +82,14 @@ class MetricsService
 
     headers = get_default_headers(user)
 
-    if (access_token === '') {
+    if access_token === ''
       # User is unauthenticated / unregistered / anonynmous
       props['distinct_id'] = userId
       headers.delete('Authorization')
       props['authenticated'] = false
-    } else {
+    else
       props['authenticated'] = true
-    }
+    end
 
     post_body = {'event': name, 'props': props}.to_json
 
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -46,7 +46,7 @@ class MetricsService
   # This call links that anonId to the user's bearer token used by DSP's Sam
   # service.  That bearer token is in turn linked to a deidentified
   # "distinct ID" used to track users across auth states in Mixpanel.
-  def self.identify(user)
+  def self.merge_identities(user)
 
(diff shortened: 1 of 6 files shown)


app/lib/metrics_service.rb
```diff
@@ -0,0 +1,108 @@
...
+  def self.post_to_bard(params, user)
+    params.merge!({
+      method: 'POST'
+    })
+    begin
+      Rails.logger.info "#{Time.zone.now}: Posting to Mixpanel.  Params: #{params}"
+      # Uncomment line below to get known-good test data
+      # puts "Posting to Mixpanel.  Params: #{params}"
+      RestClient::Request.execute(params)
+    rescue RestClient::ExceptionWithResponse => e
```
bistline, line 25: We might actually want to use `RestClient::Exception` here, as there could be a scenario where Bard responds with simply an error code (503, 400) and doesn't have a response body.  In that case, I'm don't think that it would be caught here.

app/lib/metrics_service.rb
```diff
@@ -0,0 +1,108 @@
...
+      Rails.logger.error "#{Time.zone.now}: Bard error in call to #{params[:url]}: #{e.message}"
+      # Rails.logger.error e.to_yaml
+      ErrorTracker.report_exception(e, user, params)
+    end
+  end
+
+  def self.get_default_headers(user)
```
bistline, line 32: I would rewrite this as:

```
{
  'Authorization': "Bearer #{user.valid_access_token['access_token']}",
  'Content-Type': 'application/json'
}
```

We don't need to call `return` as the default Ruby behavior is to return the last object in memory.  Also, calling `valid_access_token` will ensure that we have a good OAuth token, in case it has expired (these tokens are not renewed automatically with a user's continued session, so we have to manually refresh them).

app/lib/metrics_service.rb
```diff
@@ -0,0 +1,108 @@
...
+    Rails.logger.info "#{Time.zone.now}: Logging analytics to Mixpanel for event name: #{name}"
+
+    props.merge!({
+      appId: 'single-cell-portal',
+      env: Rails.env
+    })
+
+    headers = get_default_headers(user)
+
+    access_token = user.access_token['access_token']
```
bistline, line 85: Same thing - `valid_access_token`.

app/models/ingest_job.rb
```diff
@@ -383,23 +394,44 @@ def read_parse_logfile(filepath)
...
   #   - (Array) => List of message strings to print in a completion email
-  def generate_success_email_array
+  def get_email_and_log_to_mixpanel
+    file_type = self.study_file.file_type
+
     message = ["Total parse time: #{self.get_total_runtime}"]
-    case self.study_file.file_type
+
+    # Event properties to log to Mixpanel.  Mixpanel uses camelCase for props.
+    mixpanel_log_props = {
```
bistline, line 407: Just a style comment - I know this is getting logged externally to Mixpanel, but the Ruby convention for variable names is `snake_case`, as opposed to `camelCase`.  Not a blocking concern.

test/integration/lib/metrics_service_test.rb
```diff
@@ -0,0 +1,56 @@
...
+      :payload=>{event: event, properties: expected_output_props}.to_json,
+      :method=>"POST"
+    }
+
+    # A high-fidelity test double
+    user = User.new(access_token: {access_token: 'foo'})
+
+    # Mock network traffic to/from Bard, the DSP service proxying Mixpanel
+    mock = Minitest::Mock.new
+    mock.expect :call, mock, [expected_args] # Mock request
```
bistline, line 45: What the explanation for this `expect`?  When is the method `call` invoked?

app/lib/metrics_service.rb
eweitz: Done in 6265f13.

app/lib/metrics_service.rb
eweitz: Good points, done in 6265f13.

app/lib/metrics_service.rb
eweitz: Done in 6265f13.

app/models/ingest_job.rb
eweitz: Yup, by "Mixpanel uses camelCase for props", I mean using Ruby-conventional casing here would cause inconsistent UX in Mixpanel.  (The Mixpanel analyst would need to switch between cases, depending on the language of the logging agent.)  I've refined the comment in 6265f13.

bistline: Ah, understood.  I missed that comment.

test/integration/lib/metrics_service_test.rb
bistline: Got it - thanks for the explanation.

test/integration/lib/metrics_service_test.rb
```diff
@@ -0,0 +1,59 @@
...
+    })
+
+    # Mock network traffic to/from Bard, the DSP service proxying Mixpanel
+    mock = Minitest::Mock.new
+    mock.expect :call, mock, [expected_args] # Mock `execute` call (request)
+    mock.expect :code, 200 # Mock response
+
+    RestClient::Request.stub :execute, mock do
+      response = MetricsService.log(event, input_props, user)
+      assert response.code, 200
```
devonbush, line 53: I would remove this assert--it's only testing the configuration of the mock object, not actual code paths

app/lib/metrics_service.rb
```diff
@@ -0,0 +1,108 @@
...
+  #
+  # To enable tracking users when they are signed in and not, Bard provides an
+  # endpoint `/api/identify` to merge identities.  In SCP's case, the
+  # anonynmous ID (`anonId`) is a random UUIDv4 string set as a cookie for all
+  # users -- auth'd or not -- upon visiting SCP.
+
+  # This call links that anonId to the user's bearer token used by DSP's Sam
+  # service.  That bearer token is in turn linked to a deidentified
+  # "distinct ID" used to track users across auth states in Mixpanel.
+  def self.identify(user)
```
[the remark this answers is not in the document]
eweitz: Updated to `merge_identities_in_mixpanel` in 918339a.  This extends the previously approved name (`merge_identifies`) with a bit more indication of what's going on.

app/models/ingest_job.rb
```diff
@@ -209,7 +220,7 @@ def poll_for_completion(run_at: 1.minute.from_now)
       self.study.reload # refresh cached instance of study
       self.study_file.reload # refresh cached instance of study_file
       subject = "#{self.study_file.file_type} file: '#{self.study_file.upload_file_name}' has completed parsing"
-      message = self.generate_success_email_array
+      message = self.get_email_and_log_to_mixpanel
```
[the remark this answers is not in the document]
eweitz: Good points.  I restored the original name; see 78a4f07.

test/integration/lib/metrics_service_test.rb
eweitz: Nice catch!  Done in 78a4f07.

app/models/ingest_job.rb
```diff
@@ -195,6 +195,17 @@ def get_total_runtime
...
 
+   # Get the total runtime of parsing from event timestamps, in milliseconds
+  #
+  # * *returns*
+  #   - (Integer) => Total elapsed time in milliseconds
+  def get_total_runtime_ms
+    events = self.events
+    start_time = DateTime.parse(events.first['timestamp'])
+    completion_time = DateTime.parse(events.last['timestamp'])
+    (TimeDifference.between(start_time, completion_time).in_seconds * 1000).to_i
```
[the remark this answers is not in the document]
eweitz: Same, I was surprised [`TimeDifference`](https://github.com/tmlee/time_difference) doesn't include `in_milliseconds`.

PR #671, opened 2020-07-21
Adding SyntheticBrandingGroupPopulator (SCP-2576)

bistline:
The `SyntheticBrandingGroupPopulator` class will create synthetic branding group instances using the synthetic studies created via `SyntheticStudyPopulator`.  This can be used to test various behaviors/features of branding groups in a non-production setting.  The associated database migration will trigger in all non-production environments.

Note: tests are not included for this functionality, which is in keeping with the `SyntheticStudyPopulator`, as this is not user-facing functionality, and is only intended for development purposes.

This PR satisfies SCP-2576.

### DIFF
--- a/app/lib/synthetic_branding_group_populator.rb
+++ b/app/lib/synthetic_branding_group_populator.rb
@@ -14,7 +14,7 @@ class SyntheticBrandingGroupPopulator
     SyntheticStudyPopulator.populate_all if create_studies
     configurations = self.load_all_config_files
     configurations.each do |configuration|
-      self.populate(configuration, user: user, study_list: study_list)
+      self.populate(configuration, user: user)
     end
   end
 
--- a/app/lib/synthetic_branding_group_populator.rb
+++ b/app/lib/synthetic_branding_group_populator.rb
@@ -8,33 +8,24 @@ class SyntheticBrandingGroupPopulator
   # populate all branding groups using synthetic studies
   #
   # @param user (User)                 => User account to own the branding groups
-  # @param overwrite_groups (Boolean)  => T/F for destroying/recreating any existing groups
-  # @param overwrite_studies (Boolean) => T/F for destroying/recreating any existing studies
-  def self.populate_all(user: User.first, overwrite_groups: true, overwrite_studies: false)
-    # populate all studies first to ensure they exist
-    SyntheticStudyPopulator.populate_all(overwrite: overwrite_studies)
-    synthetic_studies_ids = self.collect_synthetic_studies.pluck(:id)
-    # divide into two groups, one for human data, another for mouse
-    human_studies = Study.where(name: /human/i, :id.in => synthetic_studies_ids)
-    mouse_studies = Study.where(name: /(mouse|murine)/i, :id.in => synthetic_studies_ids)
-    if human_studies.empty? || mouse_studies.empty?
-      raise ArgumentError.new("Required synthetic studies not present, please manually run SyntheticStudyPopulator.populate_all")
-    end
-    group_dirs = Dir.glob(BASE_BRANDING_GROUPS_CONFIG_PATH.join('*')).select {|f| File.directory? f}
-    group_dirs.each do |group_dir|
-      # detect if this is the human or mouse group
-      group_type = group_dir.split('/').last.split('_').first
-      study_list = group_type == 'human' ? human_studies : mouse_studies
-      self.populate(group_dir, user: user, study_list: study_list, overwrite: overwrite_groups)
+  # @param create_studies (Boolean)    => Set to true to run SytheticStudyPopulator to create/overwrite studies
+  def self.populate_all(user: User.first, create_studies: false)
+    # populate all studies first to ensure they exist, if specified
+    SyntheticStudyPopulator.populate_all if create_studies
+    synthetic_studies_ids = SyntheticStudyPopulator.collect_synthetic_studies.pluck(:id)
+    configurations = self.load_all_config_files
+    configurations.each do |configuration|
+      study_regex = /#{configuration.dig('study_title_regex')}/i
+      study_list = Study.where(name: study_regex, :id.in => synthetic_studies_ids)
+      self.populate(configuration, user: user, study_list: study_list)
     end
   end
 
   # remove all synthetic branding groups, leaving studies in place
   def self.remove_all
-    config_dirs = Dir.glob(BASE_BRANDING_GROUPS_CONFIG_PATH.join('*')).select {|f| File.directory? f}
-    config_dirs.each do |config_dir|
-      branding_group_config = self.parse_configuration_file(config_dir)
-      group_name = branding_group_config['branding_group']['name']
+    configurations = self.load_all_config_files
+    configurations.each do |configuration|
+      group_name = configuration.dig('branding_group', 'name')
       group = BrandingGroup.find_by(name: group_name)
       if group.present?
         puts "Removing branding group #{group_name}"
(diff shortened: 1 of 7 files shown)


app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+  # @param overwrite (Boolean)            => T/F for destroying/recreating an existing group
+  def self.populate(config_dir, user: User.first, study_list:, overwrite: true)
+    branding_group_config = self.parse_configuration_file(config_dir)
+
+    puts("Populating synthetic branding group from #{config_dir}")
+    branding_group = create_branding_group(branding_group_config, user, overwrite: overwrite)
+    # assign new branding group to all studies
+    study_list.update_all(branding_group_id: branding_group.id)
+  end
+
```
devonbush, line 60: This method isn't doesn't have any branding group-specific functionality, and might be useful elsewhere --move it to SyntheticStudyPopulator?

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+  # populate all branding groups using synthetic studies
+  #
+  # @param user (User)                 => User account to own the branding groups
+  # @param overwrite_groups (Boolean)  => T/F for destroying/recreating any existing groups
+  # @param overwrite_studies (Boolean) => T/F for destroying/recreating any existing studies
+  def self.populate_all(user: User.first, overwrite_groups: true, overwrite_studies: false)
+    # populate all studies first to ensure they exist
+    SyntheticStudyPopulator.populate_all(overwrite: overwrite_studies)
+    synthetic_studies_ids = self.collect_synthetic_studies.pluck(:id)
+    # divide into two groups, one for human data, another for mouse
```
devonbush, line 17: There should be a field in the json seed files 'study_title_regex' that contains the filter for which studies are in the group, rather than hardcoded in the populator—I think that will make this code cleaner.  

That way this method can handle finding the config directories and parsing the config files, and then populate can take a user, the parsed config object, and the options, and populate will handle finding the study lists

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+
+class SyntheticBrandingGroupPopulator
+  BASE_BRANDING_GROUPS_CONFIG_PATH = Rails.root.join('db', 'seed', 'synthetic_branding_groups')
+
+  # populate all branding groups using synthetic studies
+  #
+  # @param user (User)                 => User account to own the branding groups
+  # @param overwrite_groups (Boolean)  => T/F for destroying/recreating any existing groups
+  # @param overwrite_studies (Boolean) => T/F for destroying/recreating any existing studies
+  def self.populate_all(user: User.first, overwrite_groups: true, overwrite_studies: false)
```
devonbush, line 13: It would be nice to have an option to not populate any new studies on creation--basically 'just give me the branding groups, and I don't care if they are empty (or maybe I want them to be empty to test that particular case). then If that option is specified, we shouldn't error if a branding group has no study ids found.

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+  # @param config_dir (Pathname, String)  => Directory pathname containing configuration JSON for branding group
+  # @param user (User)                    => User account to own branding group
+  # @param study_list (Mongoid::Criteria) => List of studies to associate with new branding group
+  # @param overwrite (Boolean)            => T/F for destroying/recreating an existing group
+  def self.populate(config_dir, user: User.first, study_list:, overwrite: true)
+    branding_group_config = self.parse_configuration_file(config_dir)
+
+    puts("Populating synthetic branding group from #{config_dir}")
+    branding_group = create_branding_group(branding_group_config, user, overwrite: overwrite)
+    # assign new branding group to all studies
```
devonbush, line 57: This update_all shouldn't happen if overwrite is false, and the study already existed, right?

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+##
+
+class SyntheticBrandingGroupPopulator
+  BASE_BRANDING_GROUPS_CONFIG_PATH = Rails.root.join('db', 'seed', 'synthetic_branding_groups')
+
+  # populate all branding groups using synthetic studies
+  #
+  # @param user (User)                 => User account to own the branding groups
+  # @param overwrite_groups (Boolean)  => T/F for destroying/recreating any existing groups
+  # @param overwrite_studies (Boolean) => T/F for destroying/recreating any existing studies
```
devonbush, line 12: When I read these comments, I thought that if overwrite was false, the groups wouldn't be destroyed&recreated, but would still be updated-in-place.  I think it's relatively uncommon to run a populator and *not* have it update existing elements (remember when we kept getting screwed by SearchFacetPopulator not updating existing facets?).  So if you want to keep this behavior, I would suggest renaming the option to update_existing (and then inverting the logic) so it's really clear what you mean

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+      group_type = group_dir.split('/').last.split('_').first
+      study_list = group_type == 'human' ? human_studies : mouse_studies
+      self.populate(group_dir, user: user, study_list: study_list, overwrite: overwrite_groups)
+    end
+  end
+
+  # remove all synthetic branding groups, leaving studies in place
+  def self.remove_all
+    config_dirs = Dir.glob(BASE_BRANDING_GROUPS_CONFIG_PATH.join('*')).select {|f| File.directory? f}
+    config_dirs.each do |config_dir|
```
devonbush, line 35: Lets factor out a common method "get_all_branding_group_configs", that handles going through the directories, getting the files, and parsing the configuration files

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,107 @@
...
+    branding_group_config = self.parse_configuration_file(config_dir)
+
+    puts("Populating synthetic branding group from #{config_dir}")
+    branding_group = create_branding_group(branding_group_config, user, overwrite: overwrite)
+    # assign new branding group to all studies
+    study_list.update_all(branding_group_id: branding_group.id)
+  end
+
+  # find all matching instances of synthetic studies
+  def self.collect_synthetic_studies
```
devonbush, line 62: This method isn't doesn't have any branding group-specific functionality, and might be useful elsewhere --move it to SyntheticStudyPopulator?

app/lib/synthetic_branding_group_populator.rb
bistline: Makes sense.

app/lib/synthetic_branding_group_populator.rb
bistline: This was required because of the corner case where you had _some_ synthetic studies, but not all.  I needed to be sure that you would assign everything.  Even if you aren't overwriting anything, there was still the possibility that a study was created.  If that logic changes, then yes - this would be true here.

app/lib/synthetic_branding_group_populator.rb
bistline: It was needed simply for mass assignment, but I take the point.

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -0,0 +1,86 @@
...
+  # @param user (User)                 => User account to own the branding groups
+  # @param create_studies (Boolean)    => Set to true to run SytheticStudyPopulator to create/overwrite studies
+  def self.populate_all(user: User.first, create_studies: false)
+    # populate all studies first to ensure they exist, if specified
+    SyntheticStudyPopulator.populate_all if create_studies
+    synthetic_studies_ids = SyntheticStudyPopulator.collect_synthetic_studies.pluck(:id)
+    configurations = self.load_all_config_files
+    configurations.each do |configuration|
+      study_regex = /#{configuration.dig('study_title_regex')}/i
+      study_list = Study.where(name: study_regex, :id.in => synthetic_studies_ids)
```
devonbush, line 19: let's move these two lines into the populate method.  So the division of concerns is populate_all understands how to find all the configuration objects, and populate understands how to take a config object and do everything needed to populate it

bistline: Gotcha

bistline: 6ba85a7

app/lib/synthetic_branding_group_populator.rb
bistline: 568c460, 7290786

app/lib/synthetic_branding_group_populator.rb
bistline: 568c460, 7290786

app/lib/synthetic_branding_group_populator.rb
bistline: 568c460, 7290786

app/lib/synthetic_branding_group_populator.rb
bistline: 568c460, 7290786

app/lib/synthetic_branding_group_populator.rb
bistline: 568c460, 7290786

PR #672, opened 2020-07-22
supporting ontology browsing urls in search facets  (SCP-2562)

devonbush:
Also updates the facet populator to no longer use the outdated convention version that was stored in our public bucket.

### DIFF
--- a/app/lib/search_facet_populator.rb
+++ b/app/lib/search_facet_populator.rb
@@ -5,7 +5,7 @@ class SearchFacetPopulator
   EXCLUDED_BQ_COLUMNS = %w(CellID donor_id biosample_id)
   # loads the alexandria convention schema and populates search facets from it
   def self.populate_from_schema
-    schema_object = fetch_json_from_url(alexandria_convention_config[:url])
+    schema_object = alexandria_convention_config
     required_fields = schema_object['required']
     required_fields.each do |field_name|
       if !EXCLUDED_BQ_COLUMNS.include?(field_name) && !field_name.include?('__ontology_label')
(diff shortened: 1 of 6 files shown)


app/lib/search_facet_populator.rb
```diff
@@ -37,22 +37,36 @@ def self.populate_facet_by_name(facet_name, schema_object=nil)
...
     updated_facet.save!
     updated_facet
   end
 
+
   def self.alexandria_convention_config
     {
-      url: 'https://storage.googleapis.com/broad-singlecellportal-public/AMC_v1.1.3.json',
-      version: '1.1.3' # hardcoded here since the version is not part of the schema file
+      url: 'https://github.com/broadinstitute/scp-ingest-pipeline/raw/master/schema/alexandria_convention/alexandria_convention_schema.json',
```
bistline, line 69: We don't need to use the Github URL for this.  The latest JSON already lives in this repo:

https://raw.githubusercontent.com/broadinstitute/single_cell_portal_core/master/lib/assets/metadata_schemas/alexandria_convention/alexandria_convention_schema.json

The only change we would need to make is that in the `$id` field for the JSON that gets returned, we currently show "latest" to match the URL.  We could just change that to in fact be the actual version number of the release.  Specifically, change the version of this file to be what the latest snapshot is, which currently is `2.1.4`.  This way, the code in this repository is self-referential, and updates in `scp-ingest-pipeline` won't break the populator or tests, as it will be internally consistent.

eweitz, line 69: Could someone refresh my memory on how the use case for the raw GitHub URL used here differs from the use case for the API URL, i.e. the following?

https://singlecell.broadinstitute.org/single_cell/api/v1/metadata_schemas/alexandria_convention/latest/json

Assuming you do indeed need a URL to the JSON file -- and the downloaded JSON file that that API URL responds with -- would parameter to that API endpoint (e.g. `download=no`, but defaulting to `yes`) be a viable way to get this URL?

The current approach feels a bit duplicative, but I might be missing a nuance.

PR #673, opened 2020-07-22
Fix merge of user identities in Mixpanel via Bard (SCP-2605)

eweitz:
This fixes a bug in #666 that broke merging of user identities in Mixpanel via Bard.  In brief, the _anonymous_ ID `anonId` should be a [random UUIDv4 string](https://github.com/broadinstitute/single_cell_portal_core/blob/14a9566d30800d5c0a73977021263fe75156b9c2/app/views/layouts/application.html.erb#L45), and _not a signed-in user's_ `user.id` (nor `user.id.to_s`).  See #585 for background.

An automated test verifies the fix.  You can also manually test it like so:

* Go to SCP staging
* Sign in
* Go to [Live View in Terra Analytics Dev in Mixpanel](https://mixpanel.com/report/2085496/view/19055/live)
* Search "identify"
* You should see a recent matching event

Analogous flows can manually verify the fix in development and production environments.

This satisfies SCP-2605.

### DIFF
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -46,13 +46,15 @@ class MetricsService
   # This call links that anonId to the user's bearer token used by DSP's Sam
   # service.  That bearer token is in turn linked to a deidentified
   # "distinct ID" used to track users across auth states in Mixpanel.
-  def self.merge_identities_in_mixpanel(user)
+  def self.merge_identities_in_mixpanel(user, cookies)
 
     Rails.logger.info "#{Time.zone.now}: Merging user identity in Mixpanel via Bard"
 
     headers = get_default_headers(user)
 
-    post_body = {'anonId': user.id}.to_json
+    post_body = {
+      'anonId': cookies['user_id'] # Random UUIDv4 string
+    }.to_json
 
     params = {
       url: BARD_ROOT + '/api/identify',
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -17,7 +17,7 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       @user.delay.update_firecloud_status
       sign_in(@user)
       if TosAcceptance.accepted?(@user)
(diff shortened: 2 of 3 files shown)


eweitz:
The [failed build](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/176777273#L2678) seems unrelated to these changes.  Details:

```
EE
Finished in 2.428508s, 0.8236 runs/s, 0.0000 assertions/s.
  1) Error:
SearchFacetPopulatorTest#test_populate_facets_from_alexandria_convention_data:
NoMethodError: undefined method `[]' for true:TrueClass
    app/lib/search_facet_populator.rb:9:in `populate_from_schema'
    test/integration/lib/search_facet_populator_test.rb:8:in `block in <class:SearchFacetPopulatorTest>'
  2) Error:
SearchFacetPopulatorTest#test_updates_existing_facets:
NoMethodError: undefined method `[]' for true:TrueClass
    app/lib/search_facet_populator.rb:9:in `populate_from_schema'
    test/integration/lib/search_facet_populator_test.rb:36:in `block in <class:SearchFacetPopulatorTest>'
2 runs, 0 assertions, 0 failures, 2 errors, 0 skips
```

bistline:
Base was against `master` - that may have something to do with the test breaking perhaps.

app/lib/metrics_service.rb
```diff
@@ -46,13 +46,15 @@ def self.get_default_headers(user)
   # This call links that anonId to the user's bearer token used by DSP's Sam
   # service.  That bearer token is in turn linked to a deidentified
   # "distinct ID" used to track users across auth states in Mixpanel.
-  def self.merge_identities_in_mixpanel(user)
+  def self.merge_identities_in_mixpanel(user, cookies)
```
devonbush, line 49: don't you need to update the corresponding call to this function in OmniauthCallbacksController with the new argument?  Or is there some rails auto-magic that makes that happen?

eweitz: Good catch.  I had it locally, but overlooked `add`ing the changed file.  Done in f9f3429.

PR #676, opened 2020-07-24
cleanup facet ontology link display (SCP-2619)

devonbush:
Used to look like this:
image

_______________

now looks like this:
image

Also cleans up some issues in the facet populator form schema -- it would error if you tried to use it to populate a facet whose identifier was not equal to its name (e.g. cell_type)

### DIFF
--- a/app/lib/search_facet_populator.rb
+++ b/app/lib/search_facet_populator.rb
@@ -3,11 +3,13 @@
 class SearchFacetPopulator
 
   EXCLUDED_BQ_COLUMNS = %w(CellID donor_id biosample_id)
+  INCLUDED_OPTIONAL_COLUMNS = %w(cell_type organ_region)
   # loads the alexandria convention schema and populates search facets from it
   def self.populate_from_schema
     schema_object = alexandria_convention_config
     required_fields = schema_object['required']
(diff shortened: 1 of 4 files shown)


app/lib/search_facet_populator.rb
```diff
@@ -17,25 +19,25 @@ def self.populate_from_schema
   # creates/updates a facet from a name, and returns the new SearchFacet.
   # To manually populate a new Alexandria convention facet from the rails console, run e.g.
   # SearchFacetPopulator.populate_facet_by_name('vaccination__route')
-  def self.populate_facet_by_name(facet_name, schema_object=nil)
+  def self.populate_facet_by_name(facet_identifier, schema_object=nil)
```
eweitz, line 22: Could `facet_id` work here, for less verbosity?

devonbush: since `identifier` is the name of the DB column it's stored in, I like using identifier here so it's not confused with the DB object ID.  Although, looking at this, it now doesn't make sense to have this method be called 'populate_facet_by_name'. :).  I'll update

eweitz: Makes sense!

PR #677, opened 2020-07-24
Automatically delete failed uploads after 24 hours (SCP-2612)

bistline:
If a user uploads a file, and that upload process is interrupted for some reason, that file is stuck in a state where it cannot be removed by the user due to constraints placed both client- and server-side to prevent users from accidentally deleting a file that is still being uploaded or processed.  This change will now automatically clean up any "failed" uploads, which are defined as: any file that was created more than 24 hours ago, with a `status: uploading`, does not have a `generation` tag (meaning the file never made it to the bucket), and is not currently being parsed.

Before being removed, a final sanity check is run to see if there is a matching file already in the bucket, and the database state of the corresponding file was simply never updated.  If all checks pass, then the file is queued for deletion, and the study owner is notified of the failed uploaded and encouraged to retry.

This PR satisfies SCP-2612.

### DIFF
--- a/app/mailers/single_cell_mailer.rb
+++ b/app/mailers/single_cell_mailer.rb
@@ -36,6 +36,15 @@ class SingleCellMailer < ApplicationMailer
     end
   end
 
+  def notify_user_upload_fail(study_file, study, user)
+    @study_file = study_file
+    @study = study
+    dev_email_config = AdminConfiguration.find_by(config_type: 'QA Dev Email')
+    dev_email = dev_email_config.present? ? dev_email_config.value : nil
+    title = "#{study_file.upload_file_name} did not finish uploading"
+    mail(to: user.email, bcc: dev_email, subject: '[Single Cell Portal Notifier] ' + title)
+  end
+
   def notify_user_parse_complete(email, title, message, study)
     @message = message
     @study = study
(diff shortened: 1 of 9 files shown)


app/models/delete_queue_job.rb
```diff
@@ -133,6 +133,36 @@ def perform
     end
   end
```
devonbush, line 135: I'd rather see this in its own library called, e.g. 'UploadCleanupUtils' or 'ServerCleanupTasks' or somesuch, so that the job classes stay as single-purpose 'classes that perform the job in their name'

app/mailers/single_cell_mailer.rb
```diff
@@ -36,6 +36,15 @@ def notify_admin_upload_fail(study_file, error)
     end
   end
 
+  def notify_user_upload_fail(study_file, study, user)
```
devonbush, line 39: add a method comment 'notifies the user and cc'es dev team if appropriate for the environment' or somesuch

app/models/delete_queue_job.rb
bistline: Fair enough.

app/mailers/single_cell_mailer.rb
bistline: 4fb8abb

app/models/delete_queue_job.rb
bistline: 4fb8abb

PR #680, opened 2020-07-31
Allowing image uploads via CKEditor, expanded toolbar (SCP-2499)

bistline:
Users previously were able to upload images to study descriptions when SCP utilized CKEditor 4, but that feature was lost during the upgrade to CKEditor 5 due to CSP issues and the lack of an out-of-the-box Ruby integration for handling image file uploads.  This update restores that functionality via the [`Base64UploadAdapter`](https://ckeditor.com/docs/ckeditor5/latest/features/image-upload/base64-upload-adapter.html) that converts uploaded images to base64 data strings.  This is advantageous as it does not require any server-side integration, and eliminates the need to manage/backup user-uploaded image files.  Also, the editor toolbar has been expanded to include some more common tools that users might expect in a WYSIWYG editor, such as text/image alignment, inserting/editing tables, and format clearing.

CKEditor toolbar before the update:
ckeditor_toolbar_small

CKEditor toolbar after the update:
ckeditor_toolbar_expanded

A side effect of this update is that CKEditor is no longer a package dependency, but instead is a compiled library included via the `Sprockets` asset pipeline.  Also, as base64 encoded data strings can be very long for larger images, this can increase response times for any API or server call that requires loading a study description.  To help mitigate this, study descriptions will now have all HTML tags stripped from all API/server responses, except when loading an individual study in the "Study Overview" tab.

This PR satisfies SCP-2499.


app/assets/javascripts/application.js
```diff
@@ -129,6 +130,25 @@ var paginationOpts = {
     position: 'relative' // Element positioning
 };
 
+// global config for CKEditor instances
```
devonbush, line 133: I know this file is inconsistent throughout, but at least for new stuff, let's settle on 2-spaces indent for JS

app/javascript/styles/_brand.scss
```diff
@@ -88,6 +88,10 @@
   font-size: 24px;
 }
 
+figcaption {
```
devonbush, line 91: can you add a comment that this is related to CKeditor?   and maybe move it to _global with the other ckeditor specific stuff?

app/views/admin_configurations/compose_users_email.html.erb
```diff
@@ -64,12 +64,10 @@
 
     var ckEditor = ClassicEditor
         .create( document.querySelector( '#email_contents' ),
-            {
-                toolbar: ["heading", '|',  "bold", "italic", "link", "bulletedList", "numberedList", "blockQuote", '|', "undo", "redo"]
-            }
+            fullCKEditorConfig
```
devonbush, line 67: nice, I like how this moved the actual configuration to the JS file

bistline:
> So, at a minimum, I think we should update api/search_controller so that it doesn't pull the description fields of the studies until after the array is paged (so it only pulls the descriptions of at most 5 studies). Then we can push off the more robust change -- moving the full study description into a separate 'study_details' object, and only storing the text part of the description in study.description -- for later

I'm not entirely sure how we prevent the loading of descriptions.  I understand (and frankly share) the concern, but I'm not aware of any method for saying "give me all of these objects and attributes but leave out the descriptions".  MongoDB supports what they call `projections`, which could do this, but Mongoid does not.  To do this would require completely rewriting our query logic from the ground up, which I feel is both unwarranted and out of scope.  Also, moving study descriptions into another collection will have its own impact on query performance in terms of how we weight/order results during "inferred" searches.

I'm open to suggestions if you have any.

bistline:
It appears that Mongoid _does_ in fact support projections, in a roundabout way.  Line 244 of api/v1/search_controller.rb could be rewritten as such:

```
@studies = @studies.collection.find(studies.selector).projection({description: 0}).to_a
```

This does exactly what we want in trimming the description out of the results.  The downside of this is that we now have an array of `Hash` objects instead of instances of the `Study` class.  Therefore a lot of downstream code needs to be updated to support that.  It also introduces an issue with computing weights for inferred search, which would require loading the full study object again (one at a time, this time around), which will then lead to the same issue we are trying to avoid.

TL;DR - I don't think projections are worth it.  Moving the study description into a separate collection seems like the only real, viable option, but that's a significant amount of work, and changes the scope of this PR.

app/assets/javascripts/application.js
eweitz: (Seconded.)  If not done already, I think this might be best done by integrating [this repo's `.eslintrc.js`](https://github.com/broadinstitute/single_cell_portal_core/blob/master/.eslintrc.js#L58) into your IDE's settings.  That ought to fix all JS style issues.  If the `eslintrc.js` file doesn't catch this spacing issue, we should amend it.

app/views/studies/_form.html.erb
```diff
@@ -119,16 +119,14 @@
...
       .create( document.querySelector( '#study_description' ),
-          {
-              toolbar: ["heading", '|',  "bold", "italic", "link", "bulletedList", "numberedList", "blockQuote", '|', "undo", "redo"]
-          }
+          fullCKEditorConfig
       )
       .then( function(editor) {
-          console.log(Array.from( editor.ui.componentFactory.names() ));
+          console.log('Editor inserted');
       } )
```
eweitz, line 126: Like we remove debug logging in Ingest Pipeline, it's also worth removing debug logging here.  If needed, this can be debugged in production via DevTools.

app/views/site/search/_studies_list.html.erb
```diff
@@ -1,5 +1,5 @@
 <% @studies.each do |study| %>
-  <% study_description = strip_tags(study.description).truncate_words(100, omission: '... (continued)') %>
+  <% study_description = study.description.truncate_words(25, omission: '... (continued)') %>
```
devonbush, line 2: is there a reason this is getting switched from 100 to 25?

test/api/studies_controller_test.rb
```diff
@@ -54,16 +54,19 @@ class StudiesControllerTest < ActionDispatch::IntegrationTest
...
+            study_detail_attributes: {
+                full_description: "<p>Test description #{SecureRandom.uuid}</p>"
+            }
         }
     }
     execute_http_request(:patch, api_v1_study_path(id: study_id), update_attributes)
     assert_response :success
-    assert json['description'] == update_attributes[:study][:description], "Did not set name correctly, expected #{update_attributes[:study][:description]} but found #{json['description']}"
+    plain_text_description = ActionController::Base.helpers.strip_tags update_attributes[:study][:description]
+    assert json['description'] == plain_text_description, "Did not set description correctly, expected #{plain_text_description} but found #{json['description']}"
```
devonbush, line 69: nice

db/migrate/20200731180439_create_study_detail_from_descriptions.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.down
+    StudyDetail.all.each do |study_detail|
+      study = study_detail.study
+      puts "Resetting study description for #{study.accession}"
+      study.update!(description: study_detail.full_description)
+      puts "Reset for #{study.accession} complete"
+    end
+    StudyDetail.destroy_all
+  end
+end
```
devonbush, line 21: nice

app/views/site/search/_studies_list.html.erb
bistline: That's actually a typo - good catch.

PR #681, opened 2020-07-31
changing tracking event name for page views (SCP-2518)

devonbush:
This changes the page view events to have an event name of, e.g. `page:view:/single_cell/study/SCP226` instead of just `page:view`.  @eweitz  You'd marked this issue as a '3' for effort, so I suspect there's some complexity in this that I'm missing--let me know!


app/helpers/application_helper.rb
```diff
@@ -364,4 +364,17 @@ def get_user_access_token(user)
   def pluralize_without_count(count, noun, text=nil)
     count.to_i == 1 ? "#{noun}#{text}" : "#{noun.pluralize}#{text}"
   end
+
+  # get the rails route specification in a format suitable
+  # for use in analytics or logging
+  # e.g. /single_cell/study/:accession/:study_name
+  def get_route_spec_string
+    spec_name = 'undefined'
```
devonbush, line 372: @bistline this is waaaay grosser than I thought it would have to be to get the string from routes.rb.  I spent about half an hour digging through SO and apiDock before settling on this--I'm hoping you know an easier way.  I had assumed there'd be an easy method for this named something like  `current_route.match_spec`

app/javascript/lib/metrics-api.js
```diff
@@ -208,11 +225,10 @@ function trimStudyName(appPath) {
...
  */
 export function log(name, props={}) {
-  const appPath = trimStudyName(window.location.pathname)
-
-  props = Object.assign(props, {
+   props = Object.assign(props, {
     appId: 'single-cell-portal',
-    appPath,
+    appPath: getAnalyticsPageName(),
+    appFullPath: getAppFullPath(),
```
eweitz, line 231: @vickylihorst, this change might affect some existing analytics.

Changing `appPath` to report a page name instead of a path is consistent with Terra (so Devon's change here is worthwhile), but it would affect the data used by any Flows, Funnels, and other analyses that depend on `appPath`.  I think most such Mixpanel queries use constructs like "contains", and would be insulated from this change.

If any of your reports involve e.g. `appPath ends with <a>`, it might be worth updating those to use `appPath contains <a>` to ensure they continue working after this update.

app/helpers/application_helper.rb
```diff
@@ -364,4 +364,17 @@ def get_user_access_token(user)
...
+    Rails.application.routes.router.recognize(request) do |route, match|
+      spec_name = route.path.spec.to_s
+      # remove trailing term that gets auto-added by actionDispatch
+      spec_name = spec_name.sub('(.:format)', '')
```
eweitz, line 376: For each declaration in `routes.rb`, e.g. in:

```
get 'studies/:accession/submissions/:submission_id', to: 'site#get_study_submission', as: :site_get_study_submission
```

`spec` would be the part at the end, e.g. `site_get_study_submission`, right?  If so, could you replace underscores with hyphens (`site-get-study-submission`), to help consistent Mixpanel UX for Terra and SCP?

Also along those lines, could you ensure that the analytics page name for the home page is `root`, [as in Terra](https://broadinstitute.slack.com/archives/C015QCR0SKY/p1595353169002300)?

devonbush:
> Using `controller_name` and `action_name` might make it easier to get a label for requests, but it won't be the same as the `:as` declaration in `routes.rb`. It would however combine into a unique key for every route. For example, if someone clicks on a study in the home page, that calls `SiteController#view_study`. In this case:
> 
> ```
> controller_name = 'site'
> action_name = 'view_study'
> ```
> 
> We could combine that (and do the replacement Eric suggests) and get `site-view-study`. According to `routes.rb`, this is actually called `site_study_view` - so not exact, but probably close enough. It won't be that much less code, but it probably will be a lot more readable and easier to maintain.

I see where you're coming from -- Rails seems to lean that direction in making the controller name/action much more accessible than the path spec.  But I'd rather have the names directly correspond to the urls.  Otherwise, in order for Vicky or other non-engineers to make sense of the names, we need to maintain a wiki page or something where we map names to URLs.  And it also encourages us to keep our urls sensible.

app/helpers/application_helper.rb
devonbush: yeah, I can do both those

devonbush:
AFter futzing with this some more, I went with Jon's suggestion of controller name action.  I realized that since we're logging appFullPath with every event in addition to pageName, we won't have to maintain a list of correspondences.  So I think this has the best of all worlds -- it's easy to read, it's auto-generated, and it's easily tied to an exact url.

PR #683, opened 2020-08-04
yarn upgrade, and removing prettier-stylelint (SCP-2255)

devonbush:
Fixing dependabot issue https://github.com/broadinstitute/single_cell_portal_core/network/alert/yarn.lock/dot-prop/open

we already have stylelint-prettier, so I believe the prettier-stylelint dependency was superfluous.  This was tested adding a trivial style violation, running `yarn lint` to see it, and then being able to successfully run `yarn lint --fix` to fix it.  

This also includes the monthly run of 'yarn upgrade`.  (Although I reverted the suggestion to move @single-cell-portal/igv to the correct alphabetical order position, since it's position indexed by 'igv' is more descriptive)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,8 +64,7 @@
     "node-fetch": "^2.6.0",
     "node-sass": "^4.13.1",
     "prettier": "^1.19.1",
-    "prettier-eslint": "^9.0.1",
-    "prettier-stylelint": "^0.4.2",
+    "prettier-eslint": "^11.0.0",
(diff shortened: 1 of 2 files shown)


devonbush:
> Looks good. (Updates that derive from Dependabot alerts haven't historically prompted us to create Jira tickets, so I don't think a ticket is required here.)

Ah, there is actually a corresponding ticket, I just forgot to add it to the PR title, thanks Eric

PR #684, opened 2020-08-05
Branding group features and fixing tag overflow for U19 (SCP-2637, SCP-2615)

devonbush:
I tested that the tag width CSS fix did not impact current display of the general space or Alexandria.

### DIFF
--- a/app/javascript/styles/_brand.scss
+++ b/app/javascript/styles/_brand.scss
@@ -61,6 +61,7 @@
   line-height: 22px;
   color: #4A4A4A;
   font-weight: 200;
+  max-width: 450px; // see SCP-2637
 }
 
 .scp-banner-position {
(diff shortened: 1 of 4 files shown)


app/javascript/styles/_brand.scss
```diff
@@ -61,6 +61,7 @@
   line-height: 22px;
   color: #4A4A4A;
   font-weight: 200;
+  max-width: 40vw; // SCP-2637
```
eweitz, line 64: More accessibly:

```suggestion
  max-width: 40vw; // Fixes layout on narrow screens (SCP-2637)
```

app/models/concerns/feature_flaggable.rb
```diff
@@ -22,12 +22,24 @@ def feature_flag_for_instance(instance, flag_key)
...
     def feature_flags_for_instance(instance)
       if instance.nil?
         return FeatureFlag.default_flag_hash
       end
       instance.feature_flags_with_defaults
     end
   end
+
+  # merges feature flags of the passed-in instances from left to right,
+  # using the default flag has if none of the instances is present or suplies a value
```
eweitz, line 35:
```suggestion
  # using the default flag has if none of the instances is present or supplies a value
```

PR #688, opened 2020-08-10
Moving dense expression parsing to ingest pipeline (SCP-2639)

bistline:
Moving parsing of dense expression matrices to `scp-ingest-pipeline`.  Also updating to the latest release: `1.5.4`.  This PR also contains small typo/bugfixes to the upload wizard and setting study gene counts.

This PR satisfies SCP-2639.

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -480,8 +480,7 @@ module Api
               genes.update(parse_status: 'parsing')
               barcodes.update(parse_status: 'parsing')
               @study_file.update(parse_status: 'parsing')
-              job = IngestJob.new(study: @study, study_file: @study_file, user: current_api_user, action: :ingest_expression)
-              job.delay.push_remote_and_launch_ingest
+              ParseUtils.delay.cell_ranger_expression_parse(@study, current_api_user, @study_file, genes, barcodes)
               head 204
             else
               logger.info "#{Time.zone.now}: Parse for #{@study_file.name} as #{@study_file.file_type} in study #{@study.name} aborted; missing required files"
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -479,7 +479,6 @@ module Api
               @study_file.update(parse_status: 'parsing')
               genes.update(parse_status: 'parsing')
               barcodes.update(parse_status: 'parsing')
-              @study_file.update(parse_status: 'parsing')
               ParseUtils.delay.cell_ranger_expression_parse(@study, current_api_user, @study_file, genes, barcodes)
               head 204
             else
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -469,7 +469,8 @@ module Api
             end
           when 'Expression Matrix'
             @study_file.update(parse_status: 'parsing')
-            @study.delay.initialize_gene_expression_data(@study_file, current_api_user)
+            job = IngestJob.new(study: @study, study_file: @study_file, user: current_api_user, action: :ingest_expression)
+            job.delay.push_remote_and_launch_ingest
             head 204
           when 'MM Coordinate Matrix'
             barcodes = @study_file.bundled_files.detect {|f| f.file_type == '10X Barcodes File'}
@@ -478,7 +479,9 @@ module Api
               @study_file.update(parse_status: 'parsing')
               genes.update(parse_status: 'parsing')
               barcodes.update(parse_status: 'parsing')
-              ParseUtils.delay.cell_ranger_expression_parse(@study, current_api_user, @study_file, genes, barcodes)
+              @study_file.update(parse_status: 'parsing')
+              job = IngestJob.new(study: @study, study_file: @study_file, user: current_api_user, action: :ingest_expression)
+              job.delay.push_remote_and_launch_ingest
               head 204
             else
               logger.info "#{Time.zone.now}: Parse for #{@study_file.name} as #{@study_file.file_type} in study #{@study.name} aborted; missing required files"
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/study_files_controller.rb
```diff
@@ -478,7 +479,9 @@ def parse
               @study_file.update(parse_status: 'parsing')
               genes.update(parse_status: 'parsing')
               barcodes.update(parse_status: 'parsing')
```
devonbush, line 481: We don't want to move mtx file parsing over to ingest pipeline yet--the code isn't ready

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -478,7 +479,9 @@ def parse
...
-              ParseUtils.delay.cell_ranger_expression_parse(@study, current_api_user, @study_file, genes, barcodes)
+              @study_file.update(parse_status: 'parsing')
```
eweitz, line 482: This duplicates line 479.

```suggestion
```

More generally, `@study_file.update(parse_status: 'parsing')` is called every time before `IngestJob.new`, and `job.delay.push_remote_and_launch_ingest` is called every time after.


To eliminate duplicate code, can those two calls be moved within the `IngestJob` constructor?

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -469,7 +469,8 @@ def parse
             end
           when 'Expression Matrix'
             @study_file.update(parse_status: 'parsing')
```
devonbush, line 471: should we feature-flag this?  Obviously, it would be a flag that is only used at the server level, not users/branding groups.  But if expression parsing via ingest pipeline goes south, it would be nice to revert without a release.  Not sure the risk of it going south is large enough to be worth the hassle/complexity of a feature flag though

app/controllers/studies_controller.rb
```diff
@@ -947,7 +949,8 @@ def sync_study_file
             @study_file.update(parse_status: 'parsing')
             genes.update(parse_status: 'parsing')
             barcodes.update(parse_status: 'parsing')
```
devonbush, line 951: same, we're not releasing mtx parsing just yet

app/controllers/studies_controller.rb
```diff
@@ -947,7 +949,8 @@ def sync_study_file
...
-            ParseUtils.delay.cell_ranger_expression_parse(@study, current_user, @study_file, genes, barcodes, {sync: true})
+            @study_file.update(parse_status: 'parsing')
```
eweitz, line 952: This duplicates line 949.
```suggestion
```

app/models/papi_client.rb
```diff
@@ -316,13 +316,6 @@ def get_command_line_options(study_file, action)
         taxon = study_file.taxon
         opts += ["--taxon-name", "#{taxon.scientific_name}", "--taxon-common-name", "#{taxon.common_name}",
                  "--ncbi-taxid", "#{taxon.ncbi_taxid}"]
-        if taxon.current_assembly.present?
-          assembly = taxon.current_assembly
-          opts += ["--genome-assembly-accession", "#{assembly.accession}"]
-          if assembly.current_annotation.present?
-            opts += ["--genome-annotation", "#{assembly.current_annotation.name}"]
```
eweitz, line 323: I see these parameters exist, but are now longer used in Ingest Pipeline.  

Could you refresh my memory on how genome assembly and genome annotation will be associated with gene expression matrices that are processed by Ingest Pipeline?

app/controllers/api/v1/study_files_controller.rb
bistline: Ah, good catch.  I thought I had undone that.

app/controllers/api/v1/study_files_controller.rb
bistline: My gut says no - I'm not 100% sure the feature flagging framework will work with this, and the complexity of doing that would be such that I would worry that will have its own bugs/issues.  Our existing test infrastructure has shown that dense parsing via ingest pipeline works, and further manual spot checks have shown that the plots before and after are identical.  I even checked the databased for every gene parsed from one of our test matrices and the scores are perfectly aligned before and after.

If we want to have an "escape lever", we should simply have a PR ready to merge that would undo the integration.  We wouldn't even necessarily need to roll back the Docker image - just the controller logic for calling ingest.  But we have ~8 full business days to test this on staging on this sufficiently.  I don't think feature flagging this is the right way to go.

app/controllers/api/v1/study_files_controller.rb
bistline: I agree about the `parse_status` update, but I don't actually want `push_remote_and_launch_ingest` to be invoked when we call `IngestJob.new`.  There is an undocumented but highly useful side effect of being able to instantiate that class without invoking ingest that allows us to spot-check running jobs easily, or even retrieve PAPI-level error messages from failed runs (even days or weeks later, as the job doesn't get purged for quite a while).  If it invoked an ingest run every time `initialize` gets called, that would cause both to runs to fail, or uselessly retry a previously failed run.

app/models/papi_client.rb
bistline: They're associated at the `StudyFile` level, which happens when the matrix is created/updated.  We only ask for the species, but there are defaults that kick in that will return an assembly & annotation if we ask for them at the file level.

To my knowledge, I don't think we use the assembly or annotation at the `Gene` level anywhere, but if we did or want to, it's easy enough to get from the parent file.

app/controllers/api/v1/study_files_controller.rb
bistline: I've created SCP-2659 to track this and some related ideas directly - there's a great deal of duplicated code surrounding parsing logic that needs to be refactored that I think is a PR in its own right.

PR #692, opened 2020-08-19
Ingest pipeline 1.5.6 (SCP-2674)

bistline:
Updating to the latest version of `scp-ingest-pipeline` : `1.5.6`.  This also adds an integration test to validate that Ingest Pipeline can correctly check that a new expression matrix in a study has a unique list of cell names.

This PR supports SCP-2674

### DIFF
--- a/test/integration/study_validation_test.rb
+++ b/test/integration/study_validation_test.rb
@@ -288,5 +288,42 @@ class StudyValidationTest < ActionDispatch::IntegrationTest
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
+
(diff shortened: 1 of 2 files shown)


test/integration/study_validation_test.rb
```diff
@@ -288,5 +288,42 @@ class StudyValidationTest < ActionDispatch::IntegrationTest
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
+
+  # validates that additional expression matrices with unique cells can be ingested to a study that already has a
+  # metadata file and at least one other expresison matrix
```
eweitz, line 293:
```suggestion
  # metadata file and at least one other expression matrix
```

PR #696, opened 2020-08-21
Bug fixes for bulk download, file uploads & parsing (SCP-2664, SCP-2668, SCP-2689)

bistline:
Large lists of studies in bulk download requests runs into a limit with database connections & timeouts when trying to look up associations for `StudyFile` objects (specifically, getting the `bucket_id` from the parent `Study` and generating the output pathname for curl downloads).  This is due to how Mongoid pools connections and shares them across different application threads when calling `Parallel.map` to speed up generating signed URLs for `curl` commands.  While each thread can manually grab/release DB connections, this drastically increases operation time and effectively nullifies any performance gains of the parallelization.  This change now precomputes maps of study ids to GCS bucket name, and study file ids to output pathnames to avoid the timeout issues.  These maps are computed once and then passed to downstream method calls.

Also, this change addresses two small issues with file upload (allowing `.gz` plain text files for all parseable file types) and file parsing (removing unnecessary options from local Rails parsers, and correctly setting authorization bearer tokens when calling upstream services when file parsing is initiated from an API call, rather than uploading through the UI).

This PR satisfies SCP-2664, SCP-2668, and SCP-2689.


bistline:
> Looks good -- ideally this would have been three separate PRs since the fixes, while in the same functional area, are totally independent of each other. But it is a Friday :)

Normally I would totally agree, but SCP-2664 and SCP-2689 were so small and don't require any CI I figured I'd let them tag along with SCP-2668.

PR #697, opened 2020-08-25
Adding standalone Swagger UI (SCP-2694)

bistline:
As part of the Rails 6 upgrade process, we need to replace the `swagger_ui_engine` gem as it does not work in Rails 6 yet, and it is not clear when it will be updated according to this [PR](https://github.com/zuzannast/swagger_ui_engine/pull/40) in their repo.  Rather than replacing it with another gem to serve the Swagger UI (there aren't any that work in Rails 6 yet), we have opted for the route of using the [standalone Swagger UI](https://www.npmjs.com/package/swagger-ui-dist) bundle.  This can be also loaded from a [CDN](https://cdnjs.com/libraries/swagger-ui), which reduces the amount of server-side JS to compile during deployments, requires only 2 static HTML files with minimal configuration, and integrates perfectly with our existing `swagger_blocks` API implementation.

This PR satisfies SCP-2694.


package.json
```diff
@@ -18,11 +18,11 @@
     "@fortawesome/react-fontawesome": "^0.1.8",
     "@rails/webpacker": "4.2.2",
     "@reach/router": "^1.3.3",
+    "@single-cell-portal/igv": "2.6.4-beta.1",
```
devonbush, line 21: we'd decided to alphabetize this under 'i' for igv, so it can be tracked in the same place as the igv dependency.  But if it's making it too much of a hassle to add dependencies (since it will always complain about them not being sorted), I'm ok putting it here

app/controllers/api/v1/api_docs_controller.rb
```diff
@@ -111,6 +111,14 @@ class ApiDocsController < ActionController::Base
...
         render json: Swagger::Blocks.build_root_json(SWAGGERED_CLASSES)
       end
+
+      def swagger_ui
+
+      end
+
+      def oauth2_redirect
+
+      end
```
eweitz, line 121: Can these empty methods be removed?

```suggestion
```

package.json
eweitz: Same.  I suspect the old placement and its now one are both the result of slightly different tools.

app/views/api/v1/api_docs/oauth2_redirect.html
```diff
@@ -0,0 +1,68 @@
+<!doctype html>
```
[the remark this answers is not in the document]
eweitz: Seconded -- I drafted the same feedback before seeing this.

package.json
bistline: This is likely an artifact of running `yarn add` and `yarn remove` several times as I was trying out different `swagger-ui` packages.  I'll fix that.

app/controllers/api/v1/api_docs_controller.rb
bistline: No - they are required as per standard Rails routing.  In order for a Rails server to respond to a request, there are three prerequisites.

- An entry in `routes.rb` describing the url, method, and mapping to controller/action
- A corresponding method declaration in the aforementioned controller
- Either a corresponding view template, or a `render` call in the controller action describing what to render

app/views/api/v1/api_docs/oauth2_redirect.html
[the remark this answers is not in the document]
bistline: Can do

app/views/api/v1/api_docs/swagger_ui.html.erb
```diff
@@ -0,0 +1,65 @@
+<!-- HTML for static distribution bundle build -->
```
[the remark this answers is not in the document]
bistline: I'll add the links to the static pages in their Github repository.

app/views/api/v1/api_docs/oauth2_redirect.html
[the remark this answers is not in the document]
bistline: 01d440a

app/views/api/v1/api_docs/swagger_ui.html.erb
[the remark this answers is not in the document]
bistline: 01d440a

package.json
bistline: 300fdaf

PR #703, opened 2020-08-31
Integrating new user- and admin-level error logs into ingest_job.rb (SCP-2703)

bistline:
Error logging for failed ingest pipeline jobs is now split into two separate log files, one that is intended for users, and one intended for administrators.  The user-level errors now only contain the actual error message, where as the administrator-level errors include stack traces, PAPI event logs, and debug-level logging messages.  These emails are now sent to both users and administrators separately, allowing for more concise user communication while providing administrators with greater context for parse failures.

For more information regarding the contents of these log files, please refer to the associated [PR](https://github.com/broadinstitute/scp-ingest-pipeline/pull/141) in scp-ingest-pipeline.

This PR satisfies SCP-2703.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -236,8 +236,10 @@ class IngestJob
       DeleteQueueJob.new(self.study_file).delay.perform
       Study.firecloud_client.delete_workspace_file(self.study.bucket_id, self.study_file.bucket_location) unless self.persist_on_fail
       subject = "Error: #{self.study_file.file_type} file: '#{self.study_file.upload_file_name}' parse has failed"
-      email_content = self.generate_error_email_body
-      SingleCellMailer.notify_user_parse_fail(self.user.email, subject, email_content, self.study).deliver_now
+      user_email_content = self.generate_error_email_body
+      SingleCellMailer.notify_user_parse_fail(self.user.email, subject, user_email_content, self.study).deliver_now
+      admin_email_content = self.generate_error_email_body(email_type: :detailed)
+      SingleCellMailer.notify_admin_parse_fail(self.user.email, subject, admin_email_content).deliver_now
     else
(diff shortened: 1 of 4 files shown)


app/models/ingest_job.rb
```diff
@@ -487,35 +490,46 @@ def generate_success_email_array
...
-  def generate_error_email_body
-    error_contents = self.read_parse_logfile(self.error_filepath)
-    warning_contents = self.read_parse_logfile(self.warning_filepath)
-    message_body = "<p>'#{self.study_file.upload_file_name}' has failed during parsing.</p>"
+  def generate_error_email_body(email_type: :user)
+    case email_type
+    when :user
+      error_contents = self.read_parse_logfile(self.user_error_filepath)
+      message_body = "<p>'#{self.study_file.upload_file_name}' has failed during parsing.</p>"
+    when :detailed
```
eweitz, line 507: We refer to this type of email / log a few different ways:
* detailed (here)
* admin (elsewhere in this PR)
* dev (in [Ingest Pipeline code](https://github.com/broadinstitute/scp-ingest-pipeline/pull/141))

Having one (and only one) word for this type of email / log in both `single_cell_portal_core` and `scp-ingest-pipeline` would help.  The other type of email / log is more consistently referred to as a "user" type, which is nice.

bistline: I agree about the `:detailed/:dev` discrepancy, but I would push back about the `admin` name.  That is referring to email delivery, and our convention of `user` vs `admin` describes the audience, rather than the function.  I'd like to keep that consistent.  However, I'll make the changes in `ingest_job.rb` regarding the log file naming so that it matches with `scp-ingest-pipeline`.

eweitz: Helpful context -- I agree `admin` makes sense in that case.  Thanks for consolidating `:detailed/:dev`.

PR #704, opened 2020-08-31
adding checks & enforcement of SCP-specific User-Agent headers (SCP-2691)

bistline:
The `single-cell-portal` [PyPI package](https://pypi.org/project/single-cell-portal/) will begin reporting installed versions of both itself and `scp-ingest-pipeline` as a `User-Agent` header on all requests to the Single Cell Portal REST API.  This is being done in an effort to catch cases where a user is trying to make requests using an outdated client that may no longer be compatible with a given instance of SCP.

The API will now check for the presence of the `scp-ingest-pipeline` UA header and attempt to compare the version against what is configured for that instance.  If both client & server are declaring that they are using a [tagged release](https://github.com/broadinstitute/scp-ingest-pipeline/releases) of `scp-ingest-pipeline` (i.e. both versions conform with [semantic versioning](https://semver.org/) formats), then the portal will enforce that the _major version_ match.  If one or the other is using a non-tagged release, such as a development build, then the entire version tag must match.

Successful matches will allow the request to proceed, whereas invalid versions will result in a `400 Bad Request` with the following JSON response:

```
{ "error": "scp-ingest-pipeline: [ client request version ] incompatible with host, please use [host configured version]" }
```

This PR satisfies SCP-2691.

### DIFF
--- a/app/controllers/api/v1/concerns/ingest_aware.rb
+++ b/app/controllers/api/v1/concerns/ingest_aware.rb
@@ -1,0 +1,58 @@
+module Api
+  module V1
+    module Concerns
+      module IngestAware
+        extend ActiveSupport::Concern
+        INGEST_IMAGE_NAME = 'scp-ingest-pipeline'
+        MANAGE_STUDY_NAME = 'single-cell-portal'
+        SCP_PACKAGES = [INGEST_IMAGE_NAME, MANAGE_STUDY_NAME]
+
+        included do
+          before_action :validate_scp_user_agent!
+        end
+
+        # deny connections that declare they are from the single-cell-portal PyPI package and are using a
+        # version of scp-ingest-pipeline that is not the current major release
+        #
+        # Examples:
+        # 'User-Agent': 'single-cell-portal/0.1.3rc1 (manage-study) scp-ingest-pipeline/1.5.6 (ingest_pipeline.py)'
+        # If the configured version of scp-ingest-pipeline is... (the response will be)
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/concerns/ingest_aware.rb
```diff
@@ -0,0 +1,58 @@
...
+            ingest_pipeline_version = ingest_image_attributes[:tag]
+            request_ingest_version = scp_package_headers[INGEST_IMAGE_NAME]
+            if ingest_pipeline_version.include?('.') && request_ingest_version.include?('.')
+              # both server & client are using a tagged release of scp-ingest-pipeline, so only compare major versions
+              # otherwise fall back to checking entire tag for non-production releases
+              ingest_pipeline_version = ingest_pipeline_version.split('.').first.try(:to_i)
+              request_ingest_version = request_ingest_version.split('.').first.try(:to_i)
+            end
+            render json: {error: "scp-ingest-pipeline: #{scp_package_headers[INGEST_IMAGE_NAME]} incompatible with host, " + \
+                                 "please use #{ingest_image_attributes[:tag]}"},
```
eweitz, line 39: Let's make this more easily actionable.  If I'm a user and I see this, I want something I can copy-and-paste in my terminal.  Something like:

```suggestion
                                 "please use the current major version (e.g. run \"pip install #{ingest_image_attributes[:tag]} --upgrade\" in your terminal)"},
```

(My syntax is worth double-checking, but a trivially-runnable command is the gist.)

PR #707, opened 2020-09-03
* SCP-2588 Log more properties for clicks

knapii-developments:
This PR add more properties for links that are clinked for mix panel.


test/js/metrics-api.test.js
```diff
@@ -99,6 +99,49 @@ describe('Library for client-side usage analytics', () => {
...
+
+    // Mock the user context, to mimic auth'd state
+    const userContext = { accessToken: 'test' }
+    jest.spyOn(UserProvider, 'useContextUser')
+      .mockImplementation(() => {
+        return userContext
+      })
+    // const event = <a href="#" className='class-name-1 class-name-2' id="link-id"> 'Text that is linked' </a>
+    const target = {
+        classList:['class-name-1', 'class-name-2'],
```
eweitz, line 129:
```suggestion
        classList: ['class-name-1', 'class-name-2'],
```

test/js/metrics-api.test.js
```diff
@@ -99,6 +99,49 @@ describe('Library for client-side usage analytics', () => {
...
+        text: 'dif Text that is linked',
+        id: "link-id"
+    }
+    logClickLink(target)
+    console.log(global.fetch.mock.calls)
```
eweitz, line 134:
```suggestion
```

app/javascript/lib/metrics-api.js
```diff
@@ -78,8 +77,12 @@ export function logClick(event) {
 /**
  * Log click on link, i.e. anchor (<a ...) tag
  */
-function logClickLink(target) {
-  const props = { text: target.text }
+export function logClickLink(target) {
+  const props = {
+    text: target.text,
+    classList: 'classList' in target? [...target.classList] : '',
```
eweitz, line 83: Blocking: to keep the logged type the same and avoid possible future Mixpanel validation issues, please use `[]` instead of `''` for empty `classList`.

Non-blocking: `Array.from` is more readable than spread.

```suggestion
    classList: 'classList' in target? Array.from(target.classList) : [],
```

test/js/metrics-api.test.js
```diff
@@ -99,6 +99,49 @@ describe('Library for client-side usage analytics', () => {
...
+    jest.spyOn(global, 'fetch').mockImplementation(() => {
+      mockFetchPromise
+    })
+
+    // Mock the user context, to mimic auth'd state
+    const userContext = { accessToken: 'test' }
+    jest.spyOn(UserProvider, 'useContextUser')
+      .mockImplementation(() => {
+        return userContext
+      })
```
eweitz, line 126: I think this block could be omitted without issue, given we're not testing auth-related code.

eweitz (changes requested):
Looking good overall.  I noted a trivial blocker and suggested a few non-blocking refinements.

eweitz (approved):

bistline (approved):

PR #714, opened 2020-09-11
Logging manage-study usage in Mixpanel, identify updates (SCP-2691, SCP-2734)

bistline:
Tracking usage of `single-cell-portal` PyPI package (i.e. manage-study) via `User-Agent` headers in Mixpanel.  This change also fixes an issue with `503` errors from Bard when logging usage.  The issue identified was authenticated SCP users who have not registered for Terra do no show up in SAM, and result in a `503` from Bard when it queries upstream as it receives an error.  Now, only authenticated SCP users who have also registered for Terra are identified via their Bearer token.  All other instances fall back to either the browser-based UUID cookie, or the new User model-based `metrics_uuid`.  These values will be unified where possible to avoid double-counting.

This PR satisfies SCP-2691 and SCP-2734.

### DIFF
--- a/app/controllers/api/v1/concerns/ingest_aware.rb
+++ b/app/controllers/api/v1/concerns/ingest_aware.rb
@@ -32,7 +32,7 @@ module Api
                 appFullPath: request.fullpath
             }
             if api_user_signed_in?
-              MetricsService.merge_identities_in_mixpanel(current_api_user) if @user.registered_for_firecloud
+              MetricsService.merge_identities_in_mixpanel(current_api_user)
               MetricsService.log('manage-study', mixpanel_log_props, current_api_user)
             end
             ingest_image_attributes = AdminConfiguration.get_ingest_docker_image_attributes
--- a/app/controllers/api/v1/concerns/ingest_aware.rb
+++ b/app/controllers/api/v1/concerns/ingest_aware.rb
@@ -32,7 +32,7 @@ module Api
                 appFullPath: request.fullpath
             }
             if api_user_signed_in?
-              MetricsService.merge_identities_in_mixpanel(current_api_user) if !@user.registered_for_firecloud
+              MetricsService.merge_identities_in_mixpanel(current_api_user) if @user.registered_for_firecloud
               MetricsService.log('manage-study', mixpanel_log_props, current_api_user)
             end
             ingest_image_attributes = AdminConfiguration.get_ingest_docker_image_attributes
--- a/app/controllers/api/v1/concerns/ingest_aware.rb
+++ b/app/controllers/api/v1/concerns/ingest_aware.rb
@@ -26,6 +26,12 @@ module Api
         def validate_scp_user_agent!
           scp_package_headers = extract_scp_user_agent_headers(request)
           if scp_package_headers.any?
+            # log usage to Mixpanel
+            mixpanel_log_props = {
+                user_agent: request.headers['User-Agent'],
+                appFullPath: request.fullpath
+            }
+            MetricsService.log('manage-study', mixpanel_log_props, current_api_user) if api_user_signed_in?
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/concerns/ingest_aware.rb
```diff
@@ -26,6 +26,15 @@ module IngestAware
         def validate_scp_user_agent!
           scp_package_headers = extract_scp_user_agent_headers(request)
           if scp_package_headers.any?
+            # log usage to Mixpanel
+            mixpanel_log_props = {
+                user_agent: request.headers['User-Agent'],
+                appFullPath: request.fullpath
+            }
+            if api_user_signed_in?
+              MetricsService.merge_identities_in_mixpanel(current_api_user) if !@user.registered_for_firecloud
```
devonbush, line 35: Let's move the `if !registered` check to inside the merge_identities method, so you don't have to duplicate it below.  I think it also makes sense there because it's because of the external metrics service dependencies that we need it, not something related to ingest/omniauth

app/javascript/lib/metrics-api.js
```diff
@@ -243,11 +245,14 @@ export function log(name, props={}) {
...
-  if (accessToken === '') {
-    // User is unauthenticated / unregistered / anonynmous
+  // only report 'authenticated' users if signed in and also registered for Terra
+  // reporting non-Terra users to Bard results in 503 errors
+  if (accessToken === '' || !registeredForTerra) {
+    // User is unauthenticated / unregistered / anonynmous / not registered for Terra
     props['distinct_id'] = userId
     delete init['headers']['Authorization']
     props['authenticated'] = false
+    props['registeredForTerra'] = registeredForTerra
```
devonbush, line 255: do we want to add another prop here like 'SCPauthenticated' to easily reflect whether the user is signed into SCP?

app/lib/metrics_service.rb
```diff
@@ -89,13 +101,14 @@ def self.log(name, props={}, user)
...
-    user_id = user.id
+    user_id = user.get_metrics_uuid
 
-    if access_token.nil?
+    if access_token.nil? || !user.registered_for_firecloud
       # User is unauthenticated / unregistered / anonynmous
       props['distinct_id'] = user_id
       headers.delete('Authorization')
       props['authenticated'] = false
+      props['registeredForTerra'] = false
```
devonbush, line 111: same question about whether we want another property here about whether the user is signed into SCP

devonbush (changes requested):
cool, this is some nice handling of nasty external complexity.  One minor style suggestion and one question

app/javascript/lib/metrics-api.js
eweitz: Signed-in status is tracked by the `authenticated` prop.  (I've updated our [lexicon](https://mixpanel.com/report/2120588/view/19059/lexicon#transformations/event-properties) to make that easier to learn.)

eweitz (changes requested):
Per our chat, a logic update is needed, given expected `identify` events aren't seen in Mixpanel.

app/controllers/api/v1/concerns/ingest_aware.rb
bistline: After talking with Eric, we realized that the logic is backwards - we should only ever call `/identify` in Bard if the user _is_ registered for Terra, not the inverse.

To the point, however, I can move it in there at the end and skip the call the Bard if the user isn't registered.

bistline: 9615ad1

eweitz (approved):
Looks good!  Kudos -- dealing with novel user agents and auth in analytics is quite complicated.

devonbush (approved):

PR #716, opened 2020-09-14
Improve observability for application speed (SCP-2564)

eweitz:
This improves observability for high-level application speed, enabling us to answer questions like:

* What percent of users experience a slow search?
* How does search speed correlate with other metrics?
* Are violin plots in Study Overview slow?
* How long does it take from the user starting to load Study Overview until the "Clusters" scatter plot is shown?

And many others.

It does this by instrumenting performance logging for client-side events like searching and chart plotting.  Client-side performance logging enables richer analytics for events that can span multiple HTTP requests, require per-user insight, etc.  

Three performance properties are used:
* **`perfTime`**: Time between event start and end.
* **`perfTime:frontend`**: Duration of front-end processing.  Used in compound events, those beginning `user-action`.  These events summarize multiple lower-level event.
* **`perfTime:backend`**: Duration of back-end processing; used in compound events.

New `user-action` events provide a simple way to get `perfTime` for journeys.  These are useful for high-level timing data.

The diagram below gives examples of these new events and properties.

Client-side_performance_monitoring_for_Single_Cell_Portal_2020-09-16

For example, when a user searches for a gene in Study Overview (bottom of diagram), three events are logged to Mixpanel via Bard:

1.  A `search` event is logged after it completes. `perfTime` here is time between submitting search and receiving search response.  
2.  A `plot:violin` event is logged after the violin plots finish rendering.  perfTime` is the time between the `search` event completing and the violin plot being interactive.  
3.  The `user-action:search:site-study` event is logged last.  `perfTime` is the whole time between when the user e.g. clicks to submit the gene search and sees an interactive violin plot.   `perfTime:backend` is the time between the `search` event starting and ending, and `perfTime:frontend` matches `perfTime` for `plot:violin`.

Additionally, all requests sent via `scpApi` now also return `perfTime` for Mixpanel logging.

SCP-2736 tracks proposals to refine and extend this client-side performance logging.

This satisfies SCP-2564.


knapii-developments (approved):

app/assets/javascripts/application.js
```diff
@@ -158,6 +158,68 @@ $(document).on('hidden.bs.modal', function(e) {
...
+  var perfTimeJourney = Math.round(endTime - startTime);
+  var perfTime = Math.round(endTime - startTimeFrontend);
+  var perfTimeBackend = perfTimeJourney - perfTime;
+
+  var logProps = {
+    'perfTime': perfTime,
+    'perfTime:journey': perfTimeJourney,
+    'perfTime:backend': perfTimeBackend,
+    currentTab: $('#view-tabs .study-nav.active').text().trim().toLowerCase(),
+    genes: $('#search_genes').val().split(' '),
```
devonbush, line 209: This line makes me a little nervous since it will error if the search_genes field isn't on the page yet/anymore, which restricts the places we can use this function.  So maybe `$('#search_genes').val() ? $('#search_genes').val().split(' ') : ''`

app/assets/javascripts/application.js
```diff
@@ -158,6 +158,68 @@ $(document).on('hidden.bs.modal', function(e) {
...
+ *
+ * See overview diagram at:
+ * https://app.lucidchart.com/invitations/accept/3b086049-89bc-4db2-96c3-815ccc438bd5
+ *
+ * @param {String} plotType Type of plot, e.g. scatter, violin, box, dot
+ */
+function logPlot(plotType) {
+
+  var endTime = performance.now();
+  var startTimeFrontend = window.SCP.perfTimeStartPlotRender;
```
devonbush, line 172: how about creating a sub-scope within window.SCP for perfTime stuff?  So this could be window.SCP.perf.startPlotRender?

app/javascript/lib/scp-api-metrics.js
```diff
@@ -144,11 +146,38 @@ export function logFilterSearch(facet, terms) {
...
+ * in more granular events, but in manner that's easier to find and less
+ * implementation-specific.  See also: logPlot in javascripts/application.js.
+ */
+export function logUserAction(lastEvent, perfTime) {
+  var isScatter = lastEvent === 'plot:scatter';
+  var pageName = window.SCP.analyticsPageName;
+  var isStudyOverview = pageName === 'site-study';
+
+  // Consider using this construct more widely
+  var isFullyInteractive = typeof window.SCP.fullyInteractive !== 'undefined';
```
devonbush, line 162: what does this mean?

app/javascript/lib/scp-api-metrics.js
```diff
@@ -144,11 +146,38 @@ export function logFilterSearch(facet, terms) {
...
+
+  if (isScatter && !isFullyInteractive && isStudyOverview) {
+    log(`user-action:page:view:${pageName}`, {perfTime});
+    window.SCP.fullyInteractive = true;
+  }
+
+  var isGeneSearchEffect = lastEvent.includes('plot') && !isScatter;
+  if (isGeneSearchEffect && isStudyOverview) {
+    log(`user-action:search:${pageName}`, {perfTime});
+  }
```
devonbush, line 172: hmmm... it seems like this function requires a lot of hard-coding event names and logic.  Let's find a time to brainstorm other possible approaches

app/views/site/render_gene_expression_plots.js.erb
```diff
@@ -1,3 +1,17 @@
+<%
+=begin
+%>
```
devonbush, line 3: StackOverflow suggests 
```
<%#
   comment stuff
   more comment
%>
```
as a more concise and IDE-friendly way of doing block comments in erb

devonbush (changes requested):
Good work on some complex stuff.  Lots of style quibbles.  the only thing I'm really concerned about is the amount of hardcoding/complex logic in logUserAction().  It's not clear to me what the path is for extending that to other user actions without that function getting really complex.

app/assets/javascripts/application.js
```diff
@@ -158,6 +158,68 @@ $(document).on('hidden.bs.modal', function(e) {
...
+ * Logs properties about rendered Study Overview plot to Mixpanel via Bard.
+ *
+ * See overview diagram at:
+ * https://app.lucidchart.com/invitations/accept/3b086049-89bc-4db2-96c3-815ccc438bd5
+ *
+ * @param {String} plotType Type of plot, e.g. scatter, violin, box, dot
+ */
+function logPlot(plotType) {
+
+  var endTime = performance.now();
```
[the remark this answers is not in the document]
eweitz: Good point.  Updated in a190ec89 to use just `performance`, like we use just `document`.

app/assets/javascripts/application.js
```diff
@@ -158,6 +158,68 @@ $(document).on('hidden.bs.modal', function(e) {
...
+    // startTime = window.performance.timing.navigationStart;
+    //
+    // In other words, for plots that are triggered by merely loading the page,
+    // we consider the "start time" to be the "navigation start" -- e.g.
+    // when they clicked the link to get to this page.  Clicking the link
+    // (or otherwise starting navigation) from that previous page was the
+    // *user action* that triggered this plot.  It most accurately reflects
+    // how long the user waits between doing something and seeing something
+    // they can meaningfully interact with.
+    const perfData = window.performance.timing;
```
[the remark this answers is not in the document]
eweitz: Yup, this one is deprecated and that one is [not supported in IE or Safari](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming#Browser_compatibility).

I see [no IE users, but 9% Safari users](https://mixpanel.com/report/2120588/view/19059/insights#report/9955232/browsers-for-home-page-scp) for home page views, per new logging by Bard client.  We [don't support Safari](https://github.com/broadinstitute/single_cell_portal_core/blob/45da8ce2f5b8f8d171ff6471aecdd0ad3adc3529/app/assets/javascripts/application.js#L430) (due to plot rendering issues, it seems), but the new API would make it costlier to support if we wanted.  I can't find any intent to deprecate `performance.timing` from Chrome (and they usually show `console.warn`s well before removing APIs), so I suspect this will work for at least several more years.

If you prefer the other API, let me know.

app/assets/javascripts/application.js
```diff
@@ -158,6 +158,68 @@ $(document).on('hidden.bs.modal', function(e) {
...
+ * @param {String} plotType Type of plot, e.g. scatter, violin, box, dot
+ */
+function logPlot(plotType) {
+
+  var endTime = performance.now();
+  var startTimeFrontend = window.SCP.perfTimeStartPlotRender;
+
+  var startTime;
+  if (typeof window.SCP.perfTimeStartPlotTrigger === 'undefined') {
+    // Plot trigger is page navigation, e.g. for "Clusters" scatter plot
```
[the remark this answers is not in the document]
eweitz: Done in a2f157e.  We [ignore multiline comment style](https://github.com/broadinstitute/single_cell_portal_core/blob/master/.eslintrc.js#L61) per line rules, so let's codify the suggestion next time it comes up.

app/assets/javascripts/application.js
eweitz: Fixed in a2f157e.  Using this function beyond Study Overview is currently out-of-scope.  To make that part of the docstring clearer, I've added a TODO right nearby to generalize how we get these props.

Intended use aside, you're right that something was off.  The standard way to get the Mixpanel `genes` prop is to use the just-introduced `formatTerms` function.  That includes refined handling for your suggestion.

app/javascript/lib/scp-api-metrics.js
eweitz: I just added a comment in a2f157e saying it means the "action's main effects are shown and ready for input".  That's about as precise as I can make it while keeping the definition flexible enough to handle our varied actions (page views, searches, etc.).

app/views/site/render_gene_expression_plots.js.erb
eweitz: Much better!  Thanks, done in a2f157e.

app/javascript/lib/scp-api-metrics.js
eweitz: Helpful chat!  Per that, I'll wait to review your upcoming refinement that preserves functionality diagrammed above.

app/assets/javascripts/application.js
eweitz: Per our chat, it sounds like your proposed refinement would address this, and I'll review when it's ready.

app/javascript/lib/scp-api-metrics.js
eweitz: Thanks for addressing this in #718.

app/javascript/lib/scp-api-metrics.js
eweitz: Thanks for addressing this in #718 -- fantastic approach.

app/views/site/_study_visualize.html.erb
```diff
@@ -56,6 +56,14 @@
 </div>
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
   <% if @study.can_visualize_clusters? %>
+    if (typeof window.SCP.pageIsFullyInteractive === 'undefined') {
+      window.SCP.startPendingEvent('user-action:page:view:site-study',
+                                 {},
+                                 'plot:',
+                                 true);
+    }
+    window.SCP.pageIsFullyInteractive = true;
```
devonbush, line 65: I'm not sure I understand this variable name -- how do we know the page is fully interactive at this point?  the data hasn't rendered yet

Is what we're looking for 'initialPageLoadDone'?

eweitz: Good point.  Neither name is quite right.  In 4d89e91, I renamed it to `pushedPendingPageViewEvent`.

app/javascript/lib/metrics-api.js
```diff
@@ -266,3 +278,72 @@ export function log(name, props={}) {
...
+    startTime,
+    complete: triggerEventProps => {
+      const perfTime = performance.now() - startTime
+      props.perfTime = Math.round(perfTime)
+      if (triggerEventProps && triggerEventProps.perfTime) {
+        // For now we just assume that triggered events always have a backend
+        // and frontend component and naively measure the backend time as
+        // (totalTime - frontendTime)
+        const frontendTime = Math.round(triggerEventProps.perfTime)
+
```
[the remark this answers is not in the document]
eweitz: See update in bbaf29c.  I've corrected the code to _add_ page load time to `perfTime` (i.e. `totalTime`) when the pending event is triggered `fromPageLoad`.  This now also extends `backendTime`, as it's `totalTime - frontendTime`.  This aligns with the new function and restores the original timing.

app/javascript/lib/scp-api.js
```diff
@@ -354,27 +361,33 @@ export function getBrandingGroup() {
...
-        return json
+        return [json, perfTime]
       }
     } else {
-      return response
+      return [response, perfTime]
     }
   }
-  return response
+  return [response, perfTime]
```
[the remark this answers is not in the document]
eweitz: Agreed, I've added this as something to address in SCP-2736.

eweitz:
This build will likely fail, due to an independent issue being investigated.  All JavaScript tests (which are the only ones relevant for these changes) [all pass](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/jobs/386322511#L1954).

app/javascript/lib/metrics-api.js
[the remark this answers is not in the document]
eweitz: Further simplified in 9f95f95 -- thanks again for the nice refinement.

PR #718, opened 2020-09-16
adding pending event infrastructure to track performance

devonbush:
Thanks for letting me attempt this refactor --  I had a really clear vision of how I thought this could work, and so I wanted to make sure I didn't waste any of your time as I inevitably realized that my initially very clear vision needed lots of tweaks to actually work. :)

I like the way this generally ended up code-wise, you'll want to pull it and make sure it meets your expectations of what it's actually logging.  I tested it with both study gene search and study overview refresh, and it seems to log sensible things that match the lucidchart.  The main difference is that the perfTime:frontend and perfTime:backend are now properties of the user-action event, which has a `perfTime` to correspond to what you previously had as perfTime:journey.

Let me know your thoughts, thanks


eweitz (changes requested):
Elegant refactor!  This approach is more maintainable and extensible.

One blocker: it breaks multi-gene search in Study Overview.  Console after searching two genes there:

```
dot-plot-issue-reproduce-prod-scp1052-on-dev:61 Uncaught ReferenceError: log is not defined
    at renderMorpheusDotPlot (scp-dot-plot.self-47d8b327d51a8a2b9991f79eaea6ec9cf80f13a5d34de0cea5c0a55e405e708c.js?body=1:211)
    at Object.success (<anonymous>:52:17)
    at fire (jquery.js:3496)
    at Object.fireWith [as resolveWith] (jquery.js:3626)
    at done (jquery.js:9786)
    at XMLHttpRequest.<anonymous> (jquery.js:10047)
    at XMLHttpRequest.nrWrapper (dot-plot-issue-reproduce-prod-scp1052-on-dev:61)
```

There are also a few minor functional and stylistic issues, but I'll take those on.

eweitz (approved):
OK, fixed.  (The fix was trivial.)  I'll fix the minor issues I mentioned, over in #716.  

Thanks again for the superb refinement!

PR #722, opened 2020-09-17
removing synth populator test from test suite

devonbush:
This may be causing the build failures we are seeing


eweitz (approved):
The build failed with these PR changes, so it seems this isn't the cause of the build failure.  

Approved, though, in case I'm missing something and that helps this move faster.

knapii-developments (approved):

PR #725, opened 2020-09-17
Making study tests more robust (SCP-2743)

devonbush:


app/models/big_query_client.rb
```diff
@@ -19,4 +19,14 @@ def initialize
...
   end
 
+  # clears the entire BQ table.  This is only intended for utility use in tests/development
+  # it will no-op in production
+  def self.clear_bq_table
+    if Rails.env.production?
+      return
+    end
+    client = BigQueryClient.new.client
+    query = "DELETE FROM #{CellMetadatum::BIGQUERY_TABLE} WHERE 1 = 1"
```
eweitz, line 29: Scary, even with the safeguards on line 25 and elsewhere.

Do we have a backup strategy for BigQuery?

app/models/ingest_job.rb
```diff
@@ -535,7 +535,7 @@ def generate_error_email_body(email_type: :user)
       end
     end
 
-    if error_contents.empty? || email_type == :dev
+    if !error_contents.present? || email_type == :dev
```
eweitz, line 538: Naive questions, after [reading up](https://gist.github.com/pythonicrubyist/8114720) a bit:

1. Is this refinement intended to handle when `error_contents` is `nil`?
2. As a rule of thumb, is `!error_contents.present?` preferred over `foo.empty?`?

eweitz (approved):
Looks good.  I asked some questions, but see no blockers.

app/models/big_query_client.rb
devonbush: Since BQ isn't the ultimate source of truth, it really doesn't need a backup.  The 'backup' is just reingesting the metadata files.  Currently, reingesting all metadata files wouldn't take much more than an hour

app/models/ingest_job.rb
devonbush: yes, this is to fix https://sentry.io/organizations/broad-institute/issues/1895999172/?project=1424198&query=is%3Aunresolved&statsPeriod=14d

eweitz: Got it -- helpful, thanks!

test/api/search_controller_test.rb
```diff
@@ -105,14 +108,9 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
-    facet_queries = []
-    facets.each do |facet|
-      if facet.filters.any?
-        facet_queries << [facet.identifier, facet.filters.map {|f| f[:id]}.join(',')]
-      end
-    end
-    facet_query = facet_queries.map {|query| query.join(':')}.join('+')
+
+    # find all human studies from metadata
+    facet_query = "species:#{HOMO_SAPIENS_FILTER[:id]}"
```
bistline, line 113: I just realized this is wrong; it should read 

`"species:#{HOMO_SAPIENS_FILTER[:id]}+disease:#{NO_DISEASE_FILTER[:id]}"`

bistline (changes requested):
The facet query string is what broke the test.

app/models/big_query_client.rb
```diff
@@ -19,4 +19,14 @@ def initialize
     self.client = Google::Cloud::Bigquery.new
   end
 
+  # clears the entire BQ table.  This is only intended for utility use in tests/development
+  # it will no-op in production
+  def self.clear_bq_table
+    if Rails.env.production?
```
bistline, line 25: I might invert this and move the delete call into the block that checks if the environment is `test` or `development`.  I can't see us wanting to do this on `staging` or other deployed instances.

bistline (approved):
Looks good, non-blocking suggestion about protecting against accidental BQ wipes.

app/models/big_query_client.rb
devonbush: Interesting -- I was actually thinking it might be useful if we ever decide to clean out/reset staging.  I could see this going either way, and since deleting the BQ table on staging wouldn't be that big of a deal (we'd just need to reparse the metadata files for the convention studies), I'll leave it

PR #727, opened 2020-09-18
Refactored handling of bundled expression files in bulk download (SCP-2747)

bistline:
Addressing issues with MongoDB connection timeouts when parallelizing generating signed URLs for bulk download curl configuration files.  This change rescopes the original MongoDB query to include bundled expression files (specifically, 10X genes/barcodes files) in the original request so that no additional queries/associations need to be run during the parallelized downstream processes.  These bundled files are still paired with their parent sparse matrices and appear in the same subfolder once downloaded.

This PR satisfies SCP-2747.


eweitz (approved):

devonbush (approved):
nice, like the simplicity

PR #730, opened 2020-09-23
Direct user to re-authenticate when access tokens are missing (SCP-2729)

bistline:
SCP stores OAuth access tokens in the `User` model after a user signs in to the portal, which is used in a variety actions.  However, there are corner cases where a user can still be considered "signed in", but the access token has been purged from the database.  This previously would result in an unrecoverable error where the user could not use the portal until their session times out after 30 minutes.  This change now gracefully handles missing access tokens, and will instruct the user to sign out/in to generate new refresh & access tokens.

Preview of the modal dialog shown to users:
logout_prompt_updated

This PR satisfies SCP-2729.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -168,7 +168,7 @@
       <div class="modal-body">
         <div class="spinner-target" id="generic-modal-spinner"></div>
       </div>
-      <div class="modal-footer">
+      <div class="modal-footer" id="generic-modal-footer">
         <button class="close" data-dismiss="modal">×</button>
       </div>
     </div>
@@ -197,13 +197,18 @@
 
     // error handling for auth errors when getting user access tokens
     if (window.SCP.userAccessToken === '' && <%= user_signed_in? %>) {
-        var errorText = '<p>There was a problem retrieving some credentials associated with your account - ' +
-            'please sign out and back in again to address the issue.</p>'
-        var signOutBtn = "<p><a class='btn btn-primary' data-method='delete' href='/single_cell/users/sign_out'>" +
-            "<span class='fas fa-sign-out-alt fa-fw'></span> Sign Out</a></p>"
-        var errorMsg = errorText + signOutBtn
-        // open modal as an alert type to persist message until user clears or signs out
-        showMessageModal(null, errorMsg)
+        var errorTitle = 'Re-sign in again';
+        var errorText = '<p class="text-danger">There was a problem retrieving some credentials associated with your account - ' +
+            'please sign out and back in again to address the issue.</p>';
+        var signOutBtn = "<span class='pull-right'><a class='btn btn-primary' data-method='delete' href='/single_cell/users/sign_out'>" +
+            "<span class='fas fa-sign-out-alt fa-fw'></span> Sign Out</a></span>";
+        var errorMsg = errorText + signOutBtn; // add empty paragraph for padding
+        $('#generic-modal-spinner').html(errorMsg + "<p>&nbsp;</p>");
+        $('#generic-modal-spinner').removeClass('spinner-target');
+        $('#generic-modal-title').html(errorTitle);
+        $('#generic-modal-title').removeClass('text-center');
+        $('#generic-modal-footer').remove();
+        $('#generic-modal').modal('show');
     }
 
     // variable used mostly for testing
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -197,7 +197,7 @@
 
     // error handling for auth errors when getting user access tokens
     if (window.SCP.userAccessToken === '' && <%= user_signed_in? %>) {
-        var errorText = '<p>There was an issue retrieving the access token associated with your account - ' +
+        var errorText = '<p>There was a problem retrieving some credentials associated with your account - ' +
             'please sign out and back in again to address the issue.</p>'
         var signOutBtn = "<p><a class='btn btn-primary' data-method='delete' href='/single_cell/users/sign_out'>" +
             "<span class='fas fa-sign-out-alt fa-fw'></span> Sign Out</a></p>"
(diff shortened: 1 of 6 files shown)


knapii-developments (approved):

app/views/layouts/application.html.erb
```diff
@@ -195,6 +195,22 @@
 </div>
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
 
+    // error handling for auth errors when getting user access tokens
+    if (window.SCP.userAccessToken === '' && <%= user_signed_in? %>) {
+        var errorTitle = 'Re-sign in again';
```
devonbush, line 200: 'Re-sign in again' is seems redundant.  how about "Please refresh your login"?

app/views/layouts/application.html.erb
```diff
@@ -195,6 +195,22 @@
...
+        var errorText = '<p class="text-danger">There was a problem retrieving some credentials associated with your account - ' +
```
devonbush, line 201: Would it be simpler to just say "Your session has expired"?  'problem retrieving some credentials' sounds like a network error.

app/views/layouts/application.html.erb
```diff
@@ -195,6 +195,22 @@
...
+            'please sign out and back in again to address the issue.</p>';
```
devonbush, line 202:
```suggestion
            'please sign out and back in again to continue using Single Cell Portal.</p>';
```

test/helper_tests/application_helper_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    # store access token to restore later
+    access_token_hash = user.access_token
+    client_access_token = get_user_access_token(user)
+    expected_token = access_token_hash.dig(:access_token)
+    assert_equal expected_token, client_access_token,
+                 "Access tokens do not match; #{expected_token} != #{client_access_token}"
+    refute client_access_token.nil?, "Should have retrieved a value for access token"
+
+    # simulate issue with access token retrieval by clearing values
+    user.update(access_token: nil)
```
devonbush, line 17:
```suggestion
    user.update!(access_token: nil)
```

test/helper_tests/application_helper_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    user.reload
+    new_token = get_user_access_token(user)
+    assert new_token.nil?, "Access token should be nil, but found #{new_token}"
+
+    # clean up and ensure consistency
+    user.update(access_token: access_token_hash)
```
devonbush, line 23:
```suggestion
    user.update!(access_token: access_token_hash)
```

devonbush (changes requested):
Looks good -- really like the new test.  Just some wordsmithing suggestions and test tweaks.

app/views/layouts/application.html.erb
bistline: That was the suggestion from UX but I like your's better.

app/views/layouts/application.html.erb
bistline: While I take the point, that's actually incorrect in this case - the portal & client still thinks they have a valid user session, but we cannot retrieve their access token because it was either deleted or expired and we cannot regenerate a new one due to issues with the refresh token.  This was why I used the phrase "there was a problem retrieving credentials" because that's what is actually happening.  How about this:

```
There appears to be an issue with you session - please sign out and back in again to continue using Single Cell Portal.
```

app/views/layouts/application.html.erb
eweitz: I'd opt for "Please sign in again", for what it's worth.

app/views/layouts/application.html.erb
```diff
@@ -195,6 +195,22 @@
...
+        var signOutBtn = "<span class='pull-right'><a class='btn btn-primary' data-method='delete' href='/single_cell/users/sign_out'>" +
+            "<span class='fas fa-sign-out-alt fa-fw'></span> Sign Out</a></span>";
+        var errorMsg = errorText + signOutBtn; // add empty paragraph for padding
+        $('#generic-modal-spinner').html(errorMsg + "<p>&nbsp;</p>");
+        $('#generic-modal-spinner').removeClass('spinner-target');
+        $('#generic-modal-title').html(errorTitle);
+        $('#generic-modal-title').removeClass('text-center');
+        $('#generic-modal-footer').remove();
+        $('#generic-modal').modal('show');
+    }
```
eweitz, line 212: I don't consider it blocking, but if it's easy, I'd set `window.SCP.userSignedIn = <%= user_signed_in? %>;` here, move this into a new JavaScript file, then `import` and invoke this code in `packs/application.js`.

That'd better decouple JS and Rails, giving IDE checks for errors, and some front-end testability outside the UI test suite.

eweitz (approved):

app/views/layouts/application.html.erb
bistline: Yeah, that makes sense to me.  I'll give that a shot - doesn't seem too complicated to do, and I agree it will be better for future maintainability.

app/views/layouts/application.html.erb
```diff
@@ -195,6 +195,22 @@
 </div>
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
 
+    // error handling for auth errors when getting user access tokens
+    if (window.SCP.userAccessToken === '' && <%= user_signed_in? %>) {
+        var errorTitle = 'Please sign in again';
+        var errorText = '<p class="text-danger">There appears to be an issue with you session - ' +
```
devonbush, line 201: you -> your

I like this more general wording.   Maybe make it even shorter 'Your session is no longer available' or 'Your session is no longer valid' ?

devonbush (approved):
one typo fix, and one wordsmithing suggestion -- feel free to ignore the wordsmithing at this point though, I think we're close enough, and this isn't a common error

PR #734, opened 2020-09-29
Fixing bundling errors from sync action (SCP-2727)

bistline:
When syncing study files that require "bundling" (meaning creation of `StudyFileBundle` objects), there was not code in place to create these objects which resulted in orphaned files that could never parsed, and required direct database intervention to unblock users.  This change now integrates creating `StudyFileBundle` objects directly into the `FileParseService` class which will automatically detect and create bundles when necessary, irrespective of whether the file being parsed is the parent or a child/bundled file.  This includes changes to how **Coordinate Labels** files are associated (pointing back to the cluster file, not the `ClusterGroup` object parsed from the file), and refactoring to views/javascript to handle this change.

Note - the Javascript in `app/views/studies/_shared_sync_functions.js.erb` is admittedly not up to par, but refactoring for maintainability and conformity to new project standards proved too large an undertaking and is now tracked in SCP-2766.

This PR satisfies SCP-2727


app/assets/javascripts/application.js
```diff
@@ -1131,3 +1131,12 @@ function extractIdentifierFromId(domId) {
...
     return idParts.join('-');
 }
+
+function appendOptionsToDropdown(options, dropdown) {
+    $(options).each(function(index, element) {
+        dropdown.append($('<option />', {
+            value: element[1],
+            text: element[0]
+        }));
+    });
```
eweitz, line 1141: Best practice in cases like this is to not write to DOM on each iteration, but instead construct a larger HTML block via iteration, then after all iterations finish, append the full HTML block once.  Fewer DOM writes improves speed.

That's not a blocking concern, because the number of iterations is likely always low, but it's worth noting.

app/lib/file_parse_service.rb
```diff
@@ -24,19 +24,23 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
       when 'Cluster'
         job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_cluster, reparse: reparse,
                             persist_on_fail: persist_on_fail)
         job.delay.push_remote_and_launch_ingest
+        # check if there is a coordinate label file waiting to be parsed
+        # must reload study_file object as associations have possibly been updated
+        study_file.reload
+        if study_file.study_file_bundle.try(:completed?)
+          study_file.bundled_files.each do |coordinate_file|
+            run_at = 1.minutes.from_now # give cluster parse job time to initiate
```
eweitz, line 38: How reliable is this hard-coded time?  Would the cluster parse job ever now initiate within a minute?

Not a blocking concern; just a reflexive question I'd ask for almost any `sleep`-like code.

app/views/studies/sync_study.html.erb
```diff
@@ -183,45 +183,34 @@
 
     // pre-select all MM coordinate matrix & BAM associations, if present
     $(document).ready(function() {
-        var allMatrices = gatherFilesByType('MM Coordinate Matrix');
-        var allBams = gatherFilesByType('BAM');
         var disabledCss = {
             'pointer-events' : 'none'
         };
+
+        function appendOptionsAndSetDisabled(optionsArray, targetSelect, targetId, parentSelect) {
```
eweitz, line 190: Please add a brief comment above this function signature, to help lazy engineers like me :).

eweitz (approved):

app/lib/file_parse_service.rb
bistline: This is just to avoid a race condition where the coordinate file parse can start before the parent cluster file job is fully invoked.  If that happens, it will fail the coordinate file job because the cluster file isn't labeled as "parsing".  I could have done a manual update of the parent file there, but since a job is about to be launched via `IngestJob`, it seemed cleaner/easier to just wait a little and allow the normal operations to take care of themselves.  So to answer the question, unless something totally unforeseen happens, 1 minute is more than enough time to wait (a few seconds is really all that's needed).

app/assets/javascripts/application.js
bistline: That's a great point - I'll make that edit.

app/lib/file_parse_service.rb
```diff
@@ -76,11 +80,12 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
-        bundle = study_file.study_file_bundle
-        matrix = bundle.parent
-        genes = bundle.bundled_files.detect {|f| f.file_type == '10X Genes File' }
-        if genes.present? && matrix.present? && bundle.completed?
+        study.delay.send_to_firecloud(study_file) if study_file.is_local?
+        study_file.reload
+        if study_file.study_file_bundle.try(:completed?)
+          bundle = study_file.study_file_bundle
+          matrix = bundle.parent
+          genes = bundle.bundled_files.detect {|f| f.file_type == '10X Genes File' }
```
devonbush, line 88:
```suggestion
          sibling_file_type = study_file.file_type == '10X Genes File' ? '10X Barcodes File' : '10X Genes File'
          sibling_file = bundle.bundled_files.detect {|f| f.file_type == sibling_file_type }
```
If you do that, I believe you can consolidate the Genes and Barcodes logic into a single `when`.  (and with something a bit more complex, you could probably also consolidate the Coordinate Matrix `when`, although that might not be worth it)

app/models/cluster_group.rb
```diff
@@ -107,12 +107,10 @@ def has_coordinate_labels?
...
   def coordinate_labels_options
-    # must retrieve study file where options have been set, so search options hash for string value of cluster_group id
-    study_file = StudyFile.find_by('options.cluster_group_id' => self.id.to_s)
     {
-        font_family: study_file.coordinate_labels_font_family,
-        font_size: study_file.coordinate_labels_font_size,
-        font_color: study_file.coordinate_labels_font_color
+        font_family: self.study_file.coordinate_labels_font_family,
+        font_size: self.study_file.coordinate_labels_font_size,
+        font_color: self.study_file.coordinate_labels_font_color
```
devonbush, line 113: is the self necessary here?  I'd check rubocop on this

app/views/studies/sync_study.html.erb
bistline: In looking back at this function, I've realized that it in fact may not be needed at all, or at least most of what is happening doesn't need to.  I'll refactor and comment before getting further review.

app/models/delete_queue_job.rb
```diff
@@ -138,9 +138,11 @@ def perform
   # remove a study_file from a study_file_bundle, and clean original_file_list up as necessary
   def remove_file_from_bundle
     bundle = object.study_file_bundle
-    bundle.original_file_list.delete_if {|file| file['file_type'] == object.file_type} # this edits the list in place, but is not saved
-    object.update(study_file_bundle_id: nil)
-    bundle.save
+    if bundle.present?
```
devonbush, line 141: nice safety check

devonbush (changes requested):
mostly style quibbles, a few genuine questions

app/models/study_file.rb
```diff
@@ -45,6 +45,13 @@ class StudyFile
       'taxon_id' => Taxon.all.map {|t| [t.common_name, t.id.to_s]}
   }
 
+  # map of 'options.[key]' for handling study file bundles, via upload/sync pages
+  BUNDLE_KEY_OPTS = {
+      'MM Coordinate Matrix' => 'matrix_id',
+      'BAM' => 'bam_id',
+      'Cluster' => 'cluster_file_id'
+  }
```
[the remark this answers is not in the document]
bistline: It actually originally did but only moved here because of the weird `NoMethodError` issue I was having because of the typo.  I can move it back and rename.

app/models/cluster_group.rb
bistline: It's not technically, but I'm rather pedantic about `self`.

app/lib/file_parse_service.rb
bistline: Nice idea - I'll give that a shot.

app/lib/file_parse_service.rb
```diff
@@ -24,19 +24,23 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
+        end
       when 'Coordinate Labels'
-        # we need to create the bundle here as it doesn't exist yet
-        parent_cluster = ClusterGroup.find_by(id: study_file.options[:cluster_group_id])
-        if parent_cluster.present?
-          parent_cluster_file = parent_cluster.study_file
-          file_list = StudyFileBundle.generate_file_list(parent_cluster_file, study_file)
-          StudyFileBundle.find_or_create_by(study_id: study.id, bundle_type: parent_cluster_file.file_type,
-                                            original_file_list: file_list)
+        if study_file.study_file_bundle.try(:completed?) # use :try as bundle may or may not exist at this point
```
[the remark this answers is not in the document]
bistline: Sure

app/assets/javascripts/application.js
```diff
@@ -1131,3 +1131,12 @@ function extractIdentifierFromId(domId) {
     idParts.pop();
     return idParts.join('-');
 }
+
```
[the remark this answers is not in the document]
bistline: Can do

app/lib/file_parse_service.rb
bistline: b46fb50

app/models/study_file.rb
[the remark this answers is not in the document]
bistline: b46fb50

app/lib/file_parse_service.rb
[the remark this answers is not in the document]
bistline: b46fb50

app/assets/javascripts/application.js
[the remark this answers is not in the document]
bistline: b46fb50

app/views/studies/sync_study.html.erb
bistline: 71d66e3

app/lib/file_parse_service.rb
bistline: b46fb50

app/assets/javascripts/application.js
bistline: 71d66e3

app/lib/file_parse_service.rb
```diff
@@ -46,43 +51,24 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
-        genes = bundle.bundled_files.detect {|f| f.file_type == '10X Genes File' }
-        if genes.present? && matrix.present? && bundle.completed?
-          genes.update(parse_status: 'parsing')
-          matrix.update(parse_status: 'parsing')
+        study.delay.send_to_firecloud(study_file) if study_file.is_local?
+        study_file.reload
+        if study_file.has_completed_bundle?
+          bundle = study_file.study_file_bundle
+          matrix = bundle.parent
+          bundle.study_files.update_all(parse_status: 'parsing')
```
devonbush, line 71: oh nice, even cleaner than I had it!

devonbush (approved):

PR #735, opened 2020-09-29
Improve discoverability via Ideogram for related genes (SCP-2363)

eweitz:
This enhances gene search in Study Overview, by giving a way to discover related genes and a visual way to explore them.

It shows an ideogram with the searched gene and related genes plotted across the genome. Two classes of related genes are shown: paralogous genes and interacting genes.  Paralogs are evolutionarily similar genes in the same organism.  Interacting genes are genes that are immediately upstream or downstream of the queried gene in a biological pathway.  Users can click on the related genes to trigger a new gene search.

Before, users had to know _a priori_ which genes are related to a gene they searched.  Now, they have an easy way to discover such relevant genes.

This refines an earlier prototype, #552.  In addition to now showing a second type of related gene, it also incorporates Ideogram.js tools like image and annotation download, and accounts for many edge cases in study data and user interaction.  It also moves the widget and integrates other suggestions from product and UX.

The [design doc](https://docs.google.com/document/d/1TltRnD4SDdkqry3AnOiWqbLiaYAv2M2Zkm-GNLbdcGs/edit) has more detail, including architecture diagrams, cavaets, and considerations for security, reliability, testing, and analytics.  A separate PR this sprint will instrument analytics per SCP-2720.

Screenshot:
Ideogram_for_related_genes_SCP_2020-09-29

Architecture:
Architecture_of_ideogram for_related_genes_2020-09-29

This satisfies SCP-2363.


app/models/study.rb
```diff
@@ -1132,6 +1132,22 @@ def unique_genes
...
+    # Get gene record for searched gene string
+    gene = self.genes.by_name_or_id(searched_gene, self.expression_matrix_files.map(&:id))
+
+    # Get matrices containing this gene.  (Same gene name can be in multiple matrices.)
+    study_file_ids = Gene.where(study_id: self.id, name: gene['name']).pluck(:study_file_id)
+
+    # Get species for each matrix.
+    # TODO (): Handle when a searched gene maps to multiple species
+    taxon_ids = StudyFile.where(:_id.in => study_file_ids).pluck(:taxon_id)
+    scientific_name = Taxon.where(:_id.in => taxon_ids).pluck(:scientific_name)[0]
```
devonbush, line 1146:
```suggestion
    scientific_name = Taxon.where(:_id.in => taxon_ids).pluck(:scientific_name).first
```

app/models/study.rb
```diff
@@ -1132,6 +1132,22 @@ def unique_genes
     Gene.where(study_id: self.id, :study_file_id.in => self.expression_matrix_files.map(&:id)).pluck(:name).uniq
   end
 
+  # Get species scientific name for a searched gene.  E.g. PTEN -> Homo sapiens.
+  def infer_species(searched_gene)
+    # Get gene record for searched gene string
+    gene = self.genes.by_name_or_id(searched_gene, self.expression_matrix_files.map(&:id))
+
+    # Get matrices containing this gene.  (Same gene name can be in multiple matrices.)
+    study_file_ids = Gene.where(study_id: self.id, name: gene['name']).pluck(:study_file_id)
```
devonbush, line 1141:
```suggestion
    study_file_ids = genes.where(name: gene['name']).pluck(:study_file_id)
```

this gets the same thing, but will be able to take advantage of rails caching if available

app/models/study.rb
```diff
@@ -1132,6 +1132,22 @@ def unique_genes
     Gene.where(study_id: self.id, :study_file_id.in => self.expression_matrix_files.map(&:id)).pluck(:name).uniq
   end
 
+  # Get species scientific name for a searched gene.  E.g. PTEN -> Homo sapiens.
+  def infer_species(searched_gene)
+    # Get gene record for searched gene string
+    gene = self.genes.by_name_or_id(searched_gene, self.expression_matrix_files.map(&:id))
```
devonbush, line 1138: this is going to cause a double-search for the gene, since I assume this same search logic is happening for the expression viewer.  The performance hit doesn't seem to be large, but it might be worth consolidating the logic if possible just so there aren't bugs if the genes-by-study lookup logic changes

app/views/site/_view_gene_expression_title_bar.html.erb
```diff
@@ -5,4 +5,17 @@
...
   </div>
-</div>
\ No newline at end of file
+</div>
+
+<script>
+  // Initializes Ideogram for related genes
+
+  document.querySelector('.study-lead').parentElement.innerHTML +=
+    '<div id="related-genes-ideogram-container" style="visibility: hidden; position: absolute; z-index: 1000; width: 100%; background: #FFF;"></div>';
```
devonbush, line 14: same comment about static css file.  IIRC correctly David Mohs and Terra are big fans of inline CSS for almost all cases (which makes more sense for Terra given that it is exclusively built out of components), but I think we should be consistent within SCP  of sticking to external CSS files, unless and until we make a decision to migrate the product as a whole over to inline

app/views/site/render_gene_expression_plots.js.erb
```diff
@@ -81,6 +85,8 @@ console.log('render status of box plot at end: ' + $('#expression-plots').data('
 console.log('render status of scatter plot at end: ' + $('#expression-plots').data('scatter-rendered'));
 console.log('render status of reference plot at end: ' + $('#expression-plots').data('reference-rendered'));
```
devonbush, line 87: the changes in this file are nice -- they show that you encapsulated the ideogram logic cleanly.

config/initializers/content_security_policy.rb
```diff
@@ -20,9 +20,9 @@
                     'https://www.googleapis.com', 'https://s3.amazonaws.com', 'https://data.broadinstitute.org', 'https://portals.broadinstitute.org',
                     'https://us.input.tcell.insight.rapid7.com', 'https://api.tcell.io', 'https://us.browser.tcell.insight.rapid7.com',
                     'https://us.agent.tcell.insight.rapid7.com', 'https://us.jsagent.tcell.insight.rapid7.com', 'https://accounts.google.com',
-                    'https://bam.nr-data.net',
-                    'https://terra-bard-dev.appspot.com', 'https://terra-bard-alpha.appspot.com', 'https://terra-bard-prod.appspot.com',
-                    'https://rest.ensembl.org']
+                    'https://bam.nr-data.net', 'https://terra-bard-dev.appspot.com', 'https://terra-bard-alpha.appspot.com',
+                    'https://terra-bard-prod.appspot.com', 'https://rest.ensembl.org', 'https://eutils.ncbi.nlm.nih.gov', 'https://mygene.info',
+                    'https://webservice.wikipathways.org']
```
devonbush, line 25: I'm sure these sites are safe, but I think our practice is that changes to the CSP move the ticket to "Moderate risk", and we should be sure to run these third parties by infoSec.

devonbush (changes requested):
Cool stuff.  I think it needs some work to handle narrower screens (1000px).  I've put a screenshot of what it looks like on my monitor.  (although then when I widened the window, the help description text didn't get any wider, even when I refreshed the page). 
image

<<begin ux-opinions that probably should be ignored in favor of Vicky/Joy/Tim/Sarah/James>>
 I'd prefer to find somewhere other than top-center to put this -- putting it where it is makes it read like Google's 'did you mean X?'-- like we're telling them that they should have searched something else.  But I understand there aren't a lot of good options.  If you can't put it below the graph, maybe put it in a show hide link "See related genes" so it doesn't pull so much focus from the search 
results? 
<< end ux opinions>>

config/initializers/content_security_policy.rb
eweitz: Good point.  I just requested security review [here](https://broadworkbench.atlassian.net/browse/SCP-2363?focusedCommentId=35169&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-35169), per the new CSP policy we floated last week.

app/models/study.rb
bistline: Actually, in this case this is incorrect.  Calling `genes` (which is an alias for `self.genes`) will return a `Mongoid::Relations::Targets::Enumerable`, which eager-loads all `Gene` documents for a given study.  The performance on this is actually quite poor.  Calling `Gene.where(...)` simply returns a `Mongoid::Criteria` which will lazy-load any documents matching the query and is vastly more performant.

app/models/study.rb
bistline: The scope `by_name_or_id` is really only intended for merging data from multiple matrices into a single `Hash` for visualization purposes.  If the intent of this method is purely to return the scientific name of the taxon that a given `Gene` maps to, I agree with Devon that this is unnecessary overhead.  We already search for genes at least twice when doing gene expression views, so adding a third is not going to help with overall render times.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+    annotationHeight: 7,
+    dataDir: 'https://unpkg.com/ideogram@1.24.0/dist/data/bands/native/',
+    showTools: true,
+    onClickAnnot,
+    onLoad() {
+      // Handles edge case: when organism lacks chromosome-level assembly
+      if (window.ideogram.chromosomesArray.length === 0) return
+
+      const searchInput = document.querySelector('#search_genes').value.trim()
+      const geneSymbol = searchInput.split(/[, ]/).filter(d => d !== '')[0]
```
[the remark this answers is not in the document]
eweitz: This feature is only for single-gene search.  For multi-gene search, right: we omit the related genes ideogram due to cluttering.

> If they click on a related gene, it looks like then we clear out the other genes they searched on. I wonder if users would instead want to add the related gene to their list in that case?

Yup, click-related-to-search is the core UX here.  Navigating to a new search result upon clicking a related result is a common search UI pattern.  Modes to add related genes to a list seem exciting and worth considering later, but for this MVP I'd like to keep the existing design's short interaction path.

app/models/study.rb
```diff
@@ -1132,6 +1132,22 @@ def unique_genes
     Gene.where(study_id: self.id, :study_file_id.in => self.expression_matrix_files.map(&:id)).pluck(:name).uniq
   end
 
+  # Get species scientific name for a searched gene.  E.g. PTEN -> Homo sapiens.
+  def infer_species(searched_gene)
```
bistline, line 1136: It seems to me that this method is doing more work than it needs to.  If the goal is simply to return a scientific name for a searched gene, we can probably do that easier.  The `Gene` model has an instance method called [`taxon`](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/gene.rb#L56) that is a getter for the expression matrix `Taxon` instance.  We could rewrite this method as a one-liner that would return an array of all possible species:

```
def infer_species(gene_name)
  Gene.where(study_id: self.id, :study_file_id.in => self.expression_matrix_files.pluck(:id), name: gene_name).map {|gene| gene.taxon.try(:scientific_name)}.uniq
end
```

bistline (changes requested):
I think the `infer_species` method can be greatly simplified.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
+/** Handle clicks on Ideogram annotations */
+function onClickAnnot(annot) {
+  document.querySelector('#search_genes').value = annot.name
```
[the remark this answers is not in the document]
eweitz: Yes, I think so. Browsers have integrated jQuery's best parts.

config/initializers/content_security_policy.rb
eweitz: Security review [passed](https://broadworkbench.atlassian.net/browse/SCP-2363?focusedCommentId=35224&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-35224) -- AppSec agrees this is secure.

app/models/study.rb
eweitz: Done [here](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/models/study.rb#L1135), including some readability and consistency refinements.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+  document.querySelector('#perform-gene-search').click()
```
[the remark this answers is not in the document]
eweitz: I concur with Devon, but for different reasons -- this approach muddies analytics.  I'll to address it in "Analytics for related genes" (SCP-2720).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+  const ideoMiddle = document.getElementById('_ideogramMiddleWrap')
+  const renderTargetTabContent =
+    document.querySelector('#render-target .tab-content')
+  const distTabs = document.querySelectorAll('.expression-plot')
+
+  if (window.ideogram.chromosomesArray.length === 0) {
+    // Handles rare edge case: gene is from a matrix that has
+    // an associated organism, but the organism has no chromosome-level
+    // genome assembly.  Example: axolotl study SCP499.
+    ideoContainer.style.display = 'none'
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/assets/javascripts/scp-related-genes-ideogram.js#L65) -- I opted to the latter standard method, per my other comment.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+    $('#related-genes-ideogram-container').html('')
+    return
+  } else {
+    // Handles theoretical edge case: multi-species study when only one
+    // organism lacks a chromosome-level genome assembly (say, a study on
+    // mouse and axolotl)
+    ideoContainer.style.display = ''
+  }
+
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/javascript/styles/_relatedGenesIdeogram.scss).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+  // Move plots down to make space for Ideogram, per UX recommendation
+  distTabs.forEach(distTab => {
+    distTab.style.position = 'relative'
+    distTab.style.top = '100px'
+  })
+
+  // Move Ideogram to its final location
+  renderTargetTabContent.prepend(ideoContainer)
+
+  // Show Ideogram
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/javascript/styles/_relatedGenesIdeogram.scss).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Resize ideogram (specifically, the legend) after resizing the viewport
+ *
+ * Called from render_gene_expression_plots.js.erb
+*/
+function resizeRelatedGenesIdeogram() { // eslint-disable-line
+
+  // Handles old studies, where matrices lack species
+  if (window.SCP.organism === '') return
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/assets/javascripts/scp-related-genes-ideogram.js#L20).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+
+  const ideoLegend = document.getElementById('_ideogramLegend')
+  const ideoRect = document.getElementById('_ideogram').getBoundingClientRect()
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/javascript/styles/_relatedGenesIdeogram.scss).  I've also upstreamed most of this legend styling code; it's more a concern of the component than the app.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+  ideoLegend.style.left = `${ideoRect.x - 160}px`
+}
+
+/**
+ * Initiates Ideogram for related genes
+ *
```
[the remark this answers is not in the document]
eweitz: Done.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+  ideoContainer.style.height = '100px'
+  ideoMiddle.style.borderBottom = '1px solid #EEE'
+  ideoMiddle.style.borderLeft = '1px solid #DDD'
+  ideoMiddle.style.borderRight = '1px solid #DDD'
+  ideoMiddle.style.overflowY = 'hidden'
+}
+
+/**
+ * Resize ideogram (specifically, the legend) after resizing the viewport
+ *
```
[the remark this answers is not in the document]
eweitz: Done.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+function onClickAnnot(annot) {
+  document.querySelector('#search_genes').value = annot.name
+  document.querySelector('#perform-gene-search').click()
+  window.ideogram.plotRelatedGenes([annot.name])
+}
+
+/**
+ * Displays Ideogram after getting gene search results in Study Overview
+ *
+ * Called from render_gene_expression_plots.js.erb
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/assets/javascripts/scp-related-genes-ideogram.js#L75).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+    chrWidth: 9,
+    chrHeight: 100,
+    chrLabelSize: 12,
+    annotationHeight: 7,
+    dataDir: 'https://unpkg.com/ideogram@1.24.0/dist/data/bands/native/',
+    showTools: true,
+    onClickAnnot,
+    onLoad() {
+      // Handles edge case: when organism lacks chromosome-level assembly
+      if (window.ideogram.chromosomesArray.length === 0) return
```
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/assets/javascripts/scp-related-genes-ideogram.js#L29).

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,110 @@
...
+
+      const searchInput = document.querySelector('#search_genes').value.trim()
+      const geneSymbol = searchInput.split(/[, ]/).filter(d => d !== '')[0]
+
+      this.plotRelatedGenes(geneSymbol)
+    }
+  }
+
+  window.ideogram = Ideogram.initRelatedGenes(ideoConfig, window.uniqueGenes)
```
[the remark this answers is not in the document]
eweitz: Agreed, [done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/assets/javascripts/scp-related-genes-ideogram.js#L111).

app/models/study.rb
```diff
@@ -1132,6 +1132,22 @@ def unique_genes
...
+    # Get gene record for searched gene string
+    gene = self.genes.by_name_or_id(searched_gene, self.expression_matrix_files.map(&:id))
+
+    # Get matrices containing this gene.  (Same gene name can be in multiple matrices.)
+    study_file_ids = Gene.where(study_id: self.id, name: gene['name']).pluck(:study_file_id)
+
+    # Get species for each matrix.
+    # TODO (): Handle when a searched gene maps to multiple species
```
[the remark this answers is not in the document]
eweitz: Thanks for the catch -- I've noted the ticket [here](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/models/study.rb#L1135).

app/models/study.rb
eweitz: Incorporated as part of other refinements; see [here](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/models/study.rb#L1135).

app/models/study.rb
eweitz: I've updated the query per Jon's other comment; see [here](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/models/study.rb#L1135).

app/views/site/_view_gene_expression_title_bar.html.erb
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/blob/1731b32e9d0f2a7fbd6d89c177b5eebc95ca0641/app/javascript/styles/_relatedGenesIdeogram.scss).  While the code is more elegant, it loses at least as much cohesion.  External CSS shines for styles intended for reuse.  For single-use components like this, the case for external CSS is less clear.  That said, I'll commit to the principle -- you make a good point about consistency.

eweitz:
This is ready for another review.

In addition to fixing the show-stopper legend bug Devon posted -- which, spookily, I directly saw him reproduce but could never quite do so myself -- I've also fixed a raft of styling edge-cases and CSP issues that I found while addressing review comments.  I've also tweaked the "Related genes" legend header to be slightly smaller, per Devon.

app/controllers/site_controller.rb
```diff
@@ -408,6 +408,10 @@ def view_gene_expression
     @cluster_annotations = load_cluster_group_annotations
     @top_plot_partial = @selected_annotation[:type] == 'group' ? 'expression_plots_view' : 'expression_annotation_plots_view'
     @y_axis_title = load_expression_axis_title
+    matrix = @study.expression_matrix_files.first
```
bistline, line 411: What is this local variable `matrix` being used for?  I don't see it accessed anywhere else.

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -0,0 +1,112 @@
...
+  putIdeogramInPlotTabs(ideoContainer)
+
+  // Make Ideogram visible
+  ideoContainer.classList = 'show-related-genes-ideogram'
+}
+
+/**
+ * Initiates Ideogram for related genes
+ *
+ * This is only done in the context of single-gene search in Study Overview
```
devonbush, line 78: great comment, thanks

bistline (approved):
Looks great - just a question about a local variable that appears to be unused.

devonbush (approved):

app/controllers/site_controller.rb
eweitz: Good catch, fixed in fc12bf2.

PR #736, opened 2020-09-30
fixing bug where spaces could sometimes be included in gene names (SCP-2712)

devonbush:


app/javascript/components/search/genes/GeneKeyword.js
```diff
@@ -48,11 +48,13 @@ export default function GeneKeyword({ placeholder, helpTextContent }) {
 
   /** Converts any current typed free text to a gene array entry */
   function syncGeneArrayToInputText() {
-    if (!inputText) {
+    const inputTextTrimmed = inputText.trim().replace(',', '')
```
bistline, line 51: Shouldn't `replace()`  take a regex rather than `','`?  That will only catch the first instance.  It is unclear to me if this would be the entire search string or just the last token.

bistline (approved):
Looks fine, modulo the question about the difference between the regex/explicit string for `replace()`

app/javascript/components/search/genes/GeneKeyword.js
eweitz, line 51: Per Jon, this would handle rare edge cases (typos, really) like "BRCA1,,".

```suggestion
    const inputTextTrimmed = inputText.trim().replace(/,/g, '')
```

eweitz (approved):
Looks good!  Please add Jira ticket (SCP-2712?) to title.

app/javascript/components/search/genes/GeneKeyword.js
devonbush: good point.  it doesn't matter here since the user is prevented from ever entering more than one comma by the keydown handler clearing the inputText after a comma is entered.  But it can't hurt to be safe--i'll update it

devonbush: fixed -- as noted, the edge case isn't possible because the inputText is cleared as soon as a single comma is written, but it can't hurt to be safe!

PR #744, opened 2020-10-05
Adding initial raw counts model (SCP-2780)

devonbush:
I wasn't happy with how either the subclass approach or the StudyFileTerm approach worked in practice.  The subclass approach would have required a few too many updates to existing code to handle assigning the appropriate class at creation, and the StudyFileTerm approach seemed too loosey-goosey -- it was too opaque which fields belonged to which file types (even with as few fields as we had).
So I went with the embedding approach--I like that it's really obvious from the file names and relationships what's going on.  I could also see this being a `belongs_to` instead of `embed_one`, but an extra table seemed unnecessary for this relationship.

This is a draft since it doesn't yet reflect all the properties we want to track, but wanted to circulate for opinions


devonbush:
For reference, here's what the StudyFileTerm class ended up looking like to support the same behavior
```class StudyFileTerm

  ###
  #
  # SETTINGS & FIELD DEFINITIONS
  #
  ###

  include Mongoid::Document
  include Mongoid::Timestamps
  include Mongoid::Paperclip

  FIELDS_FOR_TYPE = HashWithIndifferentAccess.new({
    'Expression Matrix': ['units', 'library_preparation_protocol'],
  })

  VALUES_FOR_NAME = HashWithIndifferentAccess.new({
    'units': ['UMI-corrected raw counts', 'raw counts'],
    'biosample_input_type': ['whole cell', 'single nuclei', 'bulk'],
    'multimodality': ['CITE-seq', 'Patch-seq'],
  })

  belongs_to :study, index: true
  belongs_to :study_file, index: true

  field :value, type: String
  field :name, type: String

  validate :ensure_fields_for_type
  validate :ensure_values_for_name

  def ensure_fields_for_type
    file_type = study_file.file_type
    if FIELDS_FOR_TYPE[file_type].exclude?(name)
      errors.add("Field #{name} is not supported for #{file_type} files")
    end
  end

  def ensure_values_for_name
    if VALUES_FOR_NAME[name].present? && VALUES_FOR_NAME[name].exclude?(value)
      errors.add("Value #{value} is not supported for term name of #{name}")
    end
  end

end

```

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,26 @@
+class ExpressionFileInfo
```
bistline, line 1: I'm personally not really sold on this approach.  If our end goal is a way to attach more data to a study file in an amorphous manner, but in the context of "we're going to want to search on this data later to find files", I've generally been advised by people that know Mongo that embedding documents is not the way to go.  Indexing is painful and expensive, as it needs to be defined in the parent model, so you end up with lots of large, sparse indexes because you're accounting for the whole nested structure of the child document.  Furthermore, if we extend this analogy to its logical end, we'll end up with multiple subclasses for `StudyFile` that will compound this issue further, and make querying on these attributes (which in my mind is the main use case) a pain.

The one benefit I do see to this implementation is that the subclassing makes for very clean and simple validations.  While nice, we already have compound/dependent validations in many of our models, and that complexity hasn't been an issue for us so far.  And the drawbacks to indexing/querying I feel outweigh any gains.

bistline (commented):
I'm not really convinced that subclassing/embedding documents in `StudyFile` instances will get us to the functionality we eventually want.  I think there are simpler ways to achieve this.

app/models/expression_file_info.rb
devonbush: Sounds like we're actually on the same page for the most part.  We agree that StudyFileTerm is the optimal structure for search, and that the structure in this PR (probably minus the `embed`) is the best structure from a non-search application perspective. 

Where I think I might not have communicated effectively is that StudyFileTerm was never intended to be a primary source of application data -- it was designed to be a supplementary search index.  Sort of like how BQ was not designed to be the final source of truth for cell metadata -- the raw files are -- BQ just makes them searchable in reasonable time.  My goal for this PR is that we continue to use the existing, solid ways we have of storing file properties as model properties, rather than putting them in a generic key-value mapping.  Then in Q1, when it comes time to implement search, if needed, we can add StudyFileTerm as a search-cache layer on top (or elasticSearch if it ends up being that we're also responsible for indexing top consortia data repo data).  So, essentially, in Q1 we just add an extra hook on study/file save that copies over any properties we're interested in searching on to the StudyFileTerm table.   

There are two things I want to avoid:  (1) implementing StudyFileTerm only to find out in Q1 that we need to implement elasticSearch anyway to support cell-level top consortia search. (2) start getting in a habit of "if it needs to be searchable, it has to go in a key-value pile (and nowhere else)".   SCP has done a fantastic job of keeping its data well-structured, even while using tools (ruby, mongo) that require a lot of discipline to avoid footgunning.  And so I don't want to start going down the key-value rabbit hole just yet... 

So I propose we model the data as in this PR (minus the embed), and then come time to implement search, decide between (1) searching as-is (2) implementing studyFileTerm as a layer on top of this (3) implementing StudyFileTerm as a replacement for this and other places where we have searchable model properties, and (4) implementing ElasticSearch

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,26 @@
...
+  include Mongoid::Document
+  field :library_construction_protocol, type: String
+  field :units, type: String
+  field :is_raw_counts, type: Boolean, default: false
+
```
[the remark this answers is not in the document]
devonbush: done (at top of file)

app/models/study_file.rb
```diff
@@ -54,6 +54,7 @@ class StudyFile
   belongs_to :taxon, optional: true
   belongs_to :genome_assembly, optional: true
   belongs_to :study_file_bundle, optional: true
+  embeds_one :expression_file_info
```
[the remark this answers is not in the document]
devonbush: reciprocal embedded_in added.  The relationship is optional by default, so we don't need to add that.

knapii-developments (approved):

app/models/expression_file_info.rb
eweitz: Insightful question-and-answer, especially to this reviewer who has been less deeply involved.

@devonbush, could you add a distilled version of this summary as a file-overview comment atop the file?

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,42 @@
...
+  UNITS_VALUES = [nil, 'UMI-corrected raw counts', 'raw counts']
+  validates :units, inclusion: {in: UNITS_VALUES}
+
+  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'Whole cell', 'single nuclei', 'bulk']
+  validates :biosample_input_type, inclusion: {in: BIOSAMPLE_INPUT_TYPE_VALUES}
+
+  MULTIMODALITY_VALUES = [nil, 'CITE-seq', 'Patch-seq']
+  validates :multimodality, inclusion: {in: MULTIMODALITY_VALUES}
+
+  LIBRARY_CONSTRUCTION_VALUES = [nil,
```
eweitz, line 23: Could you add a code comment on why we opt for putting these controlled-lists directly in code, in contrast to the controlled lists in metadata convention that exist as external JSON Schema?

The rationale makes sense to me at the moment, I think: 

1. These fields aren't currently integrated with cell metadata validation
2. Nor are they currently intended for all the search use cases we have for cell metadata

But that will quickly fade from my memory, and without a contextual explanation this point will confuse me when re-visiting the code.

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,42 @@
...
+                                 'Smart-seq2/Fluidigm C1',
+                                 'Mars-seq',
```
eweitz, line 25: Per https://science.sciencemag.org/content/343/6172/776 and https://www.nature.com/articles/s41596-019-0164-4:
```suggestion
                                 'MARS-seq',
```

And similarly for other references.

(Either MARS-seq or MARS-Seq -- the first part is always all-capitalized.)

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,42 @@
...
+  field :biosample_input_type, type: String
+  field :multimodality, type: String
+  field :is_raw_counts, type: Boolean, default: false
+
+  # note that species and reference genome/annotation live at the study_file level, not here
+
+  UNITS_VALUES = [nil, 'UMI-corrected raw counts', 'raw counts']
+  validates :units, inclusion: {in: UNITS_VALUES}
+
+  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'Whole cell', 'single nuclei', 'bulk']
```
eweitz, line 17:
```suggestion
  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'whole cell', 'single nuclei', 'bulk']
```

app/models/expression_file_info.rb
```diff
@@ -0,0 +1,42 @@
...
+                                 '10x 3\' v2',
+                                 '10x 3\' v3',
+                                 '10x 5\' v2',
+                                 '10x 5\' v3',
+                                 'CEL-seq2',
+                                 'Drop-seq',
+                                 'SCRB-seq',
+                                 'ATAC-seq',
+                                 'ChIP-seq',
+                                 'muthylomics']
```
eweitz, line 40: Blocker, a typo:

```suggestion
                                 'methylomics']
```

eweitz (changes requested):
Looks good!  A suggested a few refinements.  The only blocker I see is a trivial typo.

eweitz (approved):
Approved.  I went ahead and fixed the trivial blocking typo.

The other changes would be nice, but I don't consider them blocking.

eweitz:
Please add Jira ticket to title, for auditability.

app/models/expression_file_info.rb
devonbush: That's a great point -- I actually hadn't thought about this deeply yet.  I'll touch base with Jean as to how frequently we expect these lists to change

devonbush: After discussing with Jean, we agreed these lists won't change frequently in the short term, so we can create a separate ticket for designing an SCP controlled-list architecture.

app/models/expression_file_info.rb
devonbush: Hmmm... agree that the discussion would be nice to preserve.  I'm not sure the top of this file (a relatively obscure model) is the right place for a general discussion about the search architecture for SCP study file.  Instead I've summarized it and linked it from https://docs.google.com/document/d/1G69e80BsqUhlwRc5pqmc0WF-AfgE4DR2TdvH1Z5nPiw/edit

eweitz: That's reasonable!  Even a differently-focused overview comment (and that link, perhaps) atop this file would help, too.

PR #746, opened 2020-10-07
Download Agreements hotfix (SCP-2788)

bistline:
Implementing `DownloadAgreement` and `DownloadAcceptance`, an interface for creating and tracking user acceptance of any additional study owner-provided terms before downloading data from a study.  An example would be a researcher that wants to distribute their data in advance of publication, but would like to ensure that other users do not take the data and publish any results ahead of the original publication.  This new functionality would block downloads both client- and server-side until the user has accepted the terms, at which time downloads are re-enabled for that user account.

This PR is a "hotfix" to merge the requested functionality into `master` for deployment ahead of the next regularly scheduled release on 2020.10.15 (see #745 for the `development` merge).

This PR satisfies SCP-2788.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ before_install:
   - sudo unzip vault_1.0.1_linux_amd64.zip
   - sudo mv vault /usr/local/bin
 script:
-  - bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json -s secret/kdux/scp/staging/scp_service_account.json -r secret/kdux/scp/staging/read_only_service_account.json -e test -n single-cell-portal-test
+  - bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json -s secret/kdux/scp/staging/scp_service_account.json -r secret/kdux/scp/staging/read_only_service_account.json -e test -n single-cell-portal-staging
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1144,3 +1144,29 @@ function appendOptionsToDropdown(options, selectElement) {
     });
     selectElement.append(optionsArray);
 }
+
+// dynamically disable click events on DOM elements using jQuery selector
+// will not allow pointer events on mouse, and will grey out element (50% opacity)
+function disableElements(selector) {
+    var disabledCss = {
+        'pointer-events' : 'none',
+        'opacity': '0.5'
+    };
+    var cursorCss = {
+        'cursor' : 'not-allowed'
+    };
+    selector.css(disabledCss).parent().css(cursorCss);
+}
+
+// dynamically re-enable click events on DOM elements using jQuery selector
+// restores pointer events and resets opacity to 100%
+function enableElements(selector) {
+    var enalbedCss = {
+        'pointer-events': 'auto',
+        'opacity': '1.0'
+    }
+    var cursorCss = {
+        'cursor' : 'auto'
+    };
+    selector.css(enalbedCss).parent().css(cursorCss);
+}
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1146,31 +1146,16 @@ function appendOptionsToDropdown(options, selectElement) {
 }
 
 /**
- * dynamically disable click events on DOM elements using jQuery selector
+ * dynamically enable/disable click events on DOM elements using jQuery selector
  * will not allow pointer events on mouse, and will grey out element (50% opacity)
  */
-function disableElements(selector) {
-    var disabledCss = {
-        'pointer-events': 'none',
-        'opacity': '0.5'
+function setElementsEnabled(selector, enabled= true) {
+    var elementCss = {
+        'pointer-events': enabled ? 'auto' : 'none',
+        'opacity': enabled ? '1.0' : '0.5'
     };
-    var cursorCss = {
-        'cursor': 'not-allowed'
+    var parentCss = {
+        'cursor': enabled ? 'auto' : 'not-allowed'
     };
-    selector.css(disabledCss).parent().css(cursorCss);
-}
-
-/**
- * dynamically re-enable click events on DOM elements using jQuery selector
- * restores pointer events and resets opacity to 100%
- */
-function enableElements(selector) {
-    var enalbedCss = {
-        'pointer-events': 'auto',
-        'opacity': '1.0'
-    }
-    var cursorCss = {
-        'cursor': 'auto'
-    };
-    selector.css(enalbedCss).parent().css(cursorCss);
+    selector.css(elementCss).parent().css(parentCss);
 }
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1145,28 +1145,32 @@ function appendOptionsToDropdown(options, selectElement) {
     selectElement.append(optionsArray);
 }
 
-// dynamically disable click events on DOM elements using jQuery selector
-// will not allow pointer events on mouse, and will grey out element (50% opacity)
+/**
+ * dynamically disable click events on DOM elements using jQuery selector
+ * will not allow pointer events on mouse, and will grey out element (50% opacity)
+ */
 function disableElements(selector) {
     var disabledCss = {
-        'pointer-events' : 'none',
+        'pointer-events': 'none',
         'opacity': '0.5'
     };
     var cursorCss = {
-        'cursor' : 'not-allowed'
+        'cursor': 'not-allowed'
     };
     selector.css(disabledCss).parent().css(cursorCss);
 }
 
-// dynamically re-enable click events on DOM elements using jQuery selector
-// restores pointer events and resets opacity to 100%
+/**
+ * dynamically re-enable click events on DOM elements using jQuery selector
+ * restores pointer events and resets opacity to 100%
+ */
 function enableElements(selector) {
     var enalbedCss = {
         'pointer-events': 'auto',
         'opacity': '1.0'
     }
     var cursorCss = {
-        'cursor' : 'auto'
+        'cursor': 'auto'
     };
     selector.css(enalbedCss).parent().css(cursorCss);
 }
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -19,6 +19,7 @@ module Api
                            :get_study_submission, :sync_submission_outputs]
       before_action :check_study_edit_permission, only: [:sync_submission_outputs]
       before_action :set_study_file, only: [:download_data, :stream_data]
+      before_action :check_download_agreement, only: [:download_data, :stream_data]
       before_action :get_download_quota, only: [:download_data, :stream_data]
 
       swagger_path '/site/studies' do
@@ -979,6 +980,12 @@ module Api
       def check_study_detached
         if @study.detached?
           head 410 and return
+        end
+      end
+
+      def check_download_agreement
+        if @study.has_download_agreement? && !@study.download_agreement.user_accepted?(current_api_user)
+          head 403 and return
         end
       end
 
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -738,60 +738,10 @@ class SiteController < ApplicationController
 
   # method to download files if study is public
   def download_file
-    # make sure user is signed in
-    if !user_signed_in?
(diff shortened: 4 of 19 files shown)


app/assets/javascripts/application.js
```diff
@@ -1144,3 +1144,29 @@ function appendOptionsToDropdown(options, selectElement) {
     });
     selectElement.append(optionsArray);
 }
+
+// dynamically disable click events on DOM elements using jQuery selector
+// will not allow pointer events on mouse, and will grey out element (50% opacity)
```
eweitz, line 1149: Using `/** comment */` style enables editors to show function docs upon hover.
```suggestion
// dynamically disable click events on DOM elements using jQuery selector
// will not allow pointer events on mouse, and will grey out element (50% opacity)
```

```suggestion
/** 
* Dynamically disable click events on DOM elements using jQuery selector
* will not allow pointer events on mouse, and will grey out element (50% opacity)
*/
```

app/assets/javascripts/application.js
```diff
@@ -1144,3 +1144,29 @@ function appendOptionsToDropdown(options, selectElement) {
...
+function disableElements(selector) {
+    var disabledCss = {
+        'pointer-events' : 'none',
+        'opacity': '0.5'
+    };
+    var cursorCss = {
+        'cursor' : 'not-allowed'
```
eweitz, line 1156:
```suggestion
        'cursor': 'not-allowed'
```

app/assets/javascripts/application.js
```diff
@@ -1144,3 +1144,29 @@ function appendOptionsToDropdown(options, selectElement) {
...
+
+// dynamically re-enable click events on DOM elements using jQuery selector
+// restores pointer events and resets opacity to 100%
+function enableElements(selector) {
+    var enalbedCss = {
+        'pointer-events': 'auto',
+        'opacity': '1.0'
+    }
+    var cursorCss = {
+        'cursor' : 'auto'
```
eweitz, line 1169:
```suggestion
        'cursor': 'auto'
```

app/controllers/site_controller.rb
```diff
@@ -734,6 +748,8 @@ def download_file
     elsif !@study.can_download?(current_user)
       redirect_to merge_default_redirect_params(view_study_path(accession: @study.accession, study_name: @study.url_safe_name), scpbr: params[:scpbr]),
                   alert: 'You do not have permission to perform that action.' and return
+    elsif @study.has_download_agreement? && !@study.download_agreement.user_accepted?(current_user)
```
eweitz, line 751: This multi-part check is done several times in various files.  Wrapping it in a simple helper function would help maintainability.

eweitz (approved):
Looks good.  A suggested a few refinements for maintainability.

Does this account for bulk download via global study search?  It's not urgent given the feature's current closed-beta nature, but worth keeping in mind as it becomes open-beta.

app/controllers/site_controller.rb
bistline: Agreed, but the issue is that since we are attempting to intercept and redirect in that method, if this is pulled out into a helper method, the `and return` no longer functions in the manner that we need (execution continues in the main method, which is the exact opposite of what we want).  I had tried this previously and wasn't able to figure out hte correct way to do it.  While there likely is a solution to the problem, it didn't seem feasible to spend additional time to figure it out given the nature/timeline of this feature.

bistline:
> Looks good. I suggested a few refinements for maintainability.
> 
> Does this account for bulk download via global study search? It's not urgent given the feature's current closed-beta nature, but worth keeping in mind as it becomes open-beta.

That's a good catch - it does not.  The amount of users that have access to that feature is small, but you are correct in that it should be enforced there.

> Looks good. I suggested a few refinements for maintainability.
> 
> Does this account for bulk download via global study search? It's not urgent given the feature's current closed-beta nature, but worth keeping in mind as it becomes open-beta.

It does not - good catch.  Since I need to incorporate some other suggestions, I'm going to see if there's an easy way to do that.

devonbush:
> > Looks good. I suggested a few refinements for maintainability.
> > Does this account for bulk download via global study search? It's not urgent given the feature's current closed-beta nature, but worth keeping in mind as it becomes open-beta.
> 
> That's a good catch - it does not. The amount of users that have access to that feature is small, but you are correct in that it should be enforced there.

Yeah, we decided we didn't need to worry about bulk download since that's currently only available in branded spaces--this study isn't in a branded space.  But now I'm realizing that we're about to release advanced search opt-in, so soon lots of people will have access.  So this hotfix can go out, but we'll need to patch bulk download for this before we release advanced search opt-in

app/assets/javascripts/application.js
```diff
@@ -1144,3 +1144,33 @@ function appendOptionsToDropdown(options, selectElement) {
     });
     selectElement.append(optionsArray);
 }
+
```
devonbush, line 1147: I think it's cleaner (and preferred in standard js libraries) to consolidate these two methods into a single setElementsEnabled(boolean) method.  Even if it doesn't save lines of code in the implementation, it makes the API simpler to remember and the do/undo connection between the two even more explicit

app/assets/javascripts/application.js
```diff
@@ -1144,3 +1144,33 @@ function appendOptionsToDropdown(options, selectElement) {
...
+    };
+    selector.css(disabledCss).parent().css(cursorCss);
+}
+
+/**
+ * dynamically re-enable click events on DOM elements using jQuery selector
+ * restores pointer events and resets opacity to 100%
+ */
+function enableElements(selector) {
+    var enalbedCss = {
```
devonbush, line 1168: 'enabledCss' typo

app/controllers/api/v1/site_controller.rb
```diff
@@ -982,6 +983,12 @@ def check_study_detached
         end
       end
 
+      def check_download_agreement
```
devonbush, line 986: nice!

app/controllers/site_controller.rb
devonbush: Agree that the pulling out the `return` would change scopes in an undesirable way.  But I'm imagining a simple helper method in study.rb `download_agreement_needs_acceptance?(user)` that just returns `has_download_agreement? && !download_agreement.user_accepted?(current_user)`. Would that work?

app/models/download_agreement.rb
```diff
@@ -0,0 +1,34 @@
+class DownloadAgreement
```
devonbush, line 1: Put a comment at the top of this file outlining the general use case for this class

app/models/download_acceptance.rb
```diff
@@ -0,0 +1,28 @@
+class DownloadAcceptance
```
devonbush, line 1: Put a comment outlining general use for the class

app/models/download_acceptance.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  belongs_to :download_agreement
+  field :email, type: String
```
devonbush, line 6: I assume the reason you track email rather than user is so we still have a record even if the user is deleted?  If so, add a comment here explaining why we track email instead of user_id

app/models/study.rb
```diff
@@ -883,6 +887,10 @@ def check_embargo?
     self.embargo.nil? || self.embargo.blank? ? false : Date.today <= self.embargo
   end
 
+  def has_download_agreement?
```
devonbush, line 890: this is nice encapsulation

app/views/site/_study_download_data.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render partial: 'study_download_agreement' if @download_agreement.present? %>
```
devonbush, line 1: Do we still want to render the agreement if it's expired?

bin/run_tests.sh
```diff
@@ -117,10 +117,12 @@ else
                     test/models/branding_group_test.rb
                     test/models/synthetic_branding_group_populator_test.rb
                     test/models/admin_configuration_test.rb
+                    test/integration/synthetic_study_populator_test.rb
```
devonbush, line 120: I don't think this line belongs in this PR

.travis.yml
```diff
@@ -17,4 +17,4 @@ before_install:
   - sudo unzip vault_1.0.1_linux_amd64.zip
   - sudo mv vault /usr/local/bin
 script:
-  - bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json -s secret/kdux/scp/staging/scp_service_account.json -r secret/kdux/scp/staging/read_only_service_account.json -e test -n single-cell-portal-staging
+  - bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json -s secret/kdux/scp/staging/scp_service_account.json -r secret/kdux/scp/staging/read_only_service_account.json -e test -n single-cell-portal-test
```
devonbush, line 20: does this change belong in this PR?

devonbush (changes requested):
This is great stuff -- love the model and the tests and the overall structure.  Just a few questions, comment additions, and stray lines from other PRs that I suspect should be cleaned out.

app/assets/javascripts/application.js
bistline: 27de521

app/assets/javascripts/application.js
bistline: 27de521

bistline:
> > > Looks good. I suggested a few refinements for maintainability.
> > > Does this account for bulk download via global study search? It's not urgent given the feature's current closed-beta nature, but worth keeping in mind as it becomes open-beta.
> > 
> > 
> > That's a good catch - it does not. The amount of users that have access to that feature is small, but you are correct in that it should be enforced there.
> 
> Yeah, we decided we didn't need to worry about bulk download since that's currently only available in branded spaces--this study isn't in a branded space. But now I'm realizing that we're about to release advanced search opt-in, so soon lots of people will have access. So this hotfix can go out, but we'll need to patch bulk download for this before we release advanced search opt-in

@devon, see 3ef5f2d - it covers bulk downloads now.

app/models/download_acceptance.rb
bistline: Yes, that is the case - it's also the same thing we do in `TosAcceptance`.

app/views/site/_study_download_data.html.erb
bistline: No, great catch.

bin/run_tests.sh
bistline: This was already on `development` when I created this branch to begin working.  I was going to remove it, but given that this particular commit does not touch any production code (it is completed confined to the `test` environment, and is only invoked in CI via Travis) it seemed simpler to leave it in rather than remove it.

.travis.yml
bistline: See above

app/controllers/site_controller.rb
bistline: I did a little reading this morning and found a possible solution that I'm going to try - we can extract this logic into a helper method and then call `; return if performed?`.  This leverages `ActionController::Metal#performed?` which checks if `render` or `redirect_to` has been called.

app/assets/javascripts/application.js
bistline: Since I have other refactoring to do I'll fold this in as well.

app/controllers/site_controller.rb
bistline: 6449efb

app/assets/javascripts/application.js
bistline: ef48f0b

app/assets/javascripts/application.js
bistline: ef48f0b

app/models/download_acceptance.rb
bistline: 2e76f7c

app/models/download_agreement.rb
bistline: 2e76f7c

app/models/download_acceptance.rb
bistline: 2e76f7c

.travis.yml
bistline: 7539771

bin/run_tests.sh
bistline: 7539771

app/views/site/_study_download_data.html.erb
bistline: ef48f0b

devonbush (approved):

PR #747, opened 2020-10-07
fixing synth populate for study description

devonbush:
newly-populated Synthetic studies were unviewable due to not having the StudyDetail created.  

I could also see fixing this by adding an on-create trigger to study.rb that auto-creates the StudyDetail object if the description is specified and no StudyDetail already exists.  But given the amount of triggers already in study.rb, I thought it was best to keep it simple and just update the populator.

### DIFF
--- a/app/lib/synthetic_study_populator.rb
+++ b/app/lib/synthetic_study_populator.rb
@@ -49,6 +49,7 @@ class SyntheticStudyPopulator
     end
 
     study = Study.new(study_config['study'])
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Looks good.  Please add ticket Jira title, for auditability.

bistline (approved):
I think your idea of having a conditional method to auto-populate blank `study_detail` objects is worth investigating.

PR #748, opened 2020-10-08
enable opt in for advanced search (SCP-2778)

devonbush:
Implements the opt-in for advanced search as shown in the demo.  See https://docs.google.com/document/d/1xPJGm7oAaLSzgHz6rJyo8pKE9K-Wut5-T3v-_tN12IQ/edit for more background

This also consolidates the UserProvider and FeatureFlagsProvider into UserProvider to simplify the frontend models, and lay the groundwork for using the new current_user API controller to read user information. (even though the UserProvider does not yet use the current_user API controller to retrieve user information just yet)


app/controllers/api/v1/current_user_controller.rb
```diff
@@ -0,0 +1,42 @@
+module Api
+  module V1
+    class CurrentUserController < ApiBaseController
+      # collection of API methods related to the current user object.
+      # for security reasons, the methods in this controller should be extremely tightly controlled
+      include Concerns::Authenticator
+      include Concerns::StudyAware
+      include Swagger::Blocks
+
+      before_action :set_current_api_user!
```
bistline, line 10: You want `:authenticate_api_user!` instead of `:set_current_api_user!` - this automatically returns `401` and halts execution.

app/controllers/api/v1/current_user_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      ALLOWABLE_UPDATE_FIELDS = ['feature_flags']
+      # to ensure users can't arbitrarily change their own flags, restrict the list
+      ALLOWABLE_UPDATE_FEATURE_FLAGS = ['faceted_search']
+      # updates the current user object -- only allowed fields are able to be changed
+      def update
+        user = current_api_user
+
+        if user.nil?
```
bistline, line 19: See above - call `: authenticate_api_user!` and then you don't need this block.

app/controllers/api/v1/current_user_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+          # anonymous users can't change feature flags
+          head 403 and return
+        end
+
+        updated_user = params[:current_user]
+        # if there are any fields included that are not allowed to be updated, return error
+        if (ALLOWABLE_UPDATE_FIELDS - updated_user.keys).present?
```
bistline, line 26: I think that there's a cleaner way of doing this.  It seems to me that since you're trying to save values, but want validations to be enforced, we should be specifying those at the model level (or in the `feature_flaggable` concern), and then all you need to do is check if the `user` model saved.

app/models/feature_flag.rb
```diff
@@ -12,6 +12,9 @@ class FeatureFlag
   field :default_value, type: Boolean, default: false
```
bistline, line 12: I think we should add a `validates_uniqueness_of` validation for `name` - this will prevent us from accidentally creating more than one feature flag with the same name.

eweitz (changes requested):
Kudos on swift work for a significant new feature!

I suggested some areas for code clean-up, and noted one blocker.

Also, a non-blocking idea maybe for later: it'd help to add an analytics property to all SCP logged events to record whether the user has opted into advanced search.  Among other analyses, this would let us measure any _lack of_ advanced search engagement among users who have opted in, and evaluate different UIs for faceted search.

app/models/concerns/feature_flaggable.rb
```diff
@@ -12,6 +12,42 @@ def feature_flags_with_defaults
     FeatureFlag.default_flag_hash.merge(self.feature_flags ? self.feature_flags : {}).with_indifferent_access
   end
 
+  # updates the feature_flags on the instance to the updated_flag_hash,
+  # with saftey checks to ensure values are valid and allowed
+  # suitable for updating flags from user input.
+  # throws an error if the update fails
+  # returns the value from the update! operation
+  def update_feature_flags_safe!(updated_flag_hash, allowed_flag_names)
```
bistline, line 20: I think this should be implemented as a validation rather than a special save function.  It will make the controller logic much simpler.

app/javascript/lib/metrics-api.js
```diff
@@ -69,36 +69,47 @@ export function logPageView() {
...
+    logClickButton(target.closest('button')[0])
+  } else if (target.closest('input').length) {
+    logClickInput(target.closest('input')[0])
   } else {
     // Perhaps uncomment when Mixpanel quota increases
     // logClickOther(target)
   }
 }
 
+function getNameForClickTarget(target) {
```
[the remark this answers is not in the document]
devonbush: Sorry, should have put a comment explaining.  All that's happening here is that if a user clicks on an element that has a data-analytics-name *on the element itself* the click is tracked to that.  This solves the problem of handling clicks on the (?) icon which were getting recorded with name: ""   The tabs you refer to don't have a data-analytics-name on the tabs themselves, and so they will still be recorded as the text they contain

eweitz (approved):

knapii-developments (approved):
Thanks for walking me through this

bistline (approved):
This is fine for now, but I still feel that most of the logic in this controller surrounds validating input values and trying to save the user record.  That really belongs in the model, or more appropriately, the `feature_flaggable` concern.  If we begin adding more functionality to this controller, we should take the time to refactor that.

I'm envisioning something structured like the example below.  It leverages strong parameters, and included validations that encapsulate the logic and constants in the concern:

```
class CurrentUserController < ApiBaseController
  ...

  def update
    user = current_api_user
    if user.update(current_user_params)
      render json: { feature_flags: user.feature_flags }
    else
      render json: { errors: user.errors }, status: 422
    end
  end

  def current_user_params
    params.require(:current_user).permit(feature_flags_attributes: [:faceted_search])
  end
end

module FeatureFlaggable
  extend ActiveSupport::Concern
  ALLOWABLE_UPDATE_FIELDS = ['feature_flags']
  ALLOWABLE_UPDATE_FEATURE_FLAGS = ['faceted_search']
  ...

  included do
    validate :check_feature_flags_for_instance

    def check_feature_flags_for_instance
      if self.is_a?(User)
        [logic from current_user_controller]
      end
    end
  end
end 
```

eweitz (approved):

devonbush:
merging, as test failure is unrelated

PR #749, opened 2020-10-08
Addressing potential race condition re: updating billing project users (SCP-2792)

bistline:
The test `FireCloudClientTest#test_update_billing_project_members` has been failing sporadically with the following error:
```
[test/integration/fire_cloud_client_test.rb:484]:
Did not successfully remove singlecelltest@gmail.com from list of billing project members
```
Repeated attempts to reproduce this error have failed, yet some [builds](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/188450072) continue to have issues.  Given that the the following assertion does _not_ throw an error:
```
user_deleted = @fire_cloud_client.delete_user_from_billing_project(project_name, user_role, @test_email)
assert user_deleted == 'OK', "Did not delete user from project: #{user_deleted}"
```
A logical conclusion is that there is occasionally latency between the `"OK"` response from the successful deletion of a user from a billing project and the removal of that user from the project (since this is simply a Google group under the hood, and FireCloud makes upstream API calls to do the actual delete).  This update attempts to address that issue by checking to see if the billing project has not yet updated, and introduces a 1 second delay before retrieving the user list again.  Given the nature of this error, and the inability to reproduce it consistently, this is only a "best guess" at the actual problem.

This PR satisfies SCP-2792.

### DIFF
--- a/test/integration/fire_cloud_client_test.rb
+++ b/test/integration/fire_cloud_client_test.rb
@@ -451,7 +451,9 @@ class FireCloudClientTest < ActiveSupport::TestCase
     assert projects.any?, 'Did not find any billing projects'
 
     # select a project (only valid projects, not in the compute blacklist)
-    project_name = projects.select {|p| p['creationStatus'] == 'Ready' && !FireCloudClient::COMPUTE_BLACKLIST.include?(p['projectName'])}.sample['projectName']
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Thorough explanation -- seems plausible.  If true, I'd consider this an upstream bug.

devonbush (approved):

PR #750, opened 2020-10-08
Fix a UI flow for gene search in Study Overview (SCP-2772)

eweitz:
This fixes a bug in gene search within studies.  For context, users can search a gene in Study Overview via one of four flows:

1. Type gene name, press Enter
2. Type gene name, click button
3. Type part of gene name, pick from autocomplete, press Enter
4. Type part of gene name, pick from autocomplete, click button

Flow 1 was broken and is now fixed.  

The bug was caused by #710, which fixed a minor server-load bug triggered by flows 1 and 3, but was only tested in flow 3 (which I've grown so accustomed to I conflated with flow 1).  These changes preserve the server-load fix while also, more importantly, ensuring flow 1 works.

The user interaction that triggered search is now also logged.  This will be extended to help facilitate analytics for Ideogram for related genes (SCP-2720).

I manually tested the four flows above.  I'll also be talking with folks about automated test strategy for this complex legacy UI.

This satisfies SCP-2772.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -261,9 +261,6 @@ function initializeAutocomplete(selector) {
         if (event.keyCode === $.ui.keyCode.TAB && $(this).autocomplete("instance").menu.active) {
             // allow user to select terms with TAB key
             event.preventDefault();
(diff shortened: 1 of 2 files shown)


bistline (approved):

devonbush (approved):
nice

PR #754, opened 2020-10-13
add synthetic raw counts files (SCP-2535)

jlchang:
Synthetic raw counts dataset with study_info.json for dense matrix (with expression_file_info), metadata and 2D cluster file. This synthetic data set also includes a corresponding sparse matrix and a 3D cluster file.

This satisfies SCP-2535.


bistline (approved):
Looks good - my only nit is that the new raw_counts files don't have a terminal newline.  It's not critical, but `POSIX` defines a "line" as a series of strings terminating with a newline character.  Without that terminal newline, the final line of the file might get ignored in some circumstances.

eweitz (changes requested):
It's good to have some more representative synthetic data.

Could you make the file names more intuitive?  E.g. something like "`<species_common_name>_<x>k_cells_<y>k_genes`" for the expression matrices?

jlchang:
> Looks good - my only nit is that the new raw_counts files don't have a terminal newline. It's not critical, but `POSIX` defines a "line" as a series of strings terminating with a newline character. Without that terminal newline, the final line of the file might get ignored in some circumstances.

manually added terminal newline to files in synthetic dataset `6c25a28f`

jlchang:
> It's good to have some more representative synthetic data.
> 
> Could you make the file names more intuitive? E.g. something like "`<species_common_name>_<x>k_cells_<y>k_genes`" for the expression matrices?

Renamed in `6c25a28f`

jlchang:
> Agree with Eric about the descriptive file names. Also the study name and folder name should be more descriptive than `raw_counts_1`. At least include the species and some other salient fact. e.g. `human_10x_with_raw`

renamed study folder in `e3529d19`

db/seed/synthetic_studies/human_raw_counts/study_info.json
```diff
@@ -0,0 +1,29 @@
+{
+  "study": {
+    "name": "raw_1",
```
devonbush, line 3: Update the study name here -- "Human raw counts"

devonbush (approved):
Approve pending the update of the study name in study_info.json

eweitz (approved):

db/seed/synthetic_studies/human_raw_counts/study_info.json
jlchang: Thanks for catching that! fixed `7be9c110`

PR #755, opened 2020-10-13
Speed up and refine UI in related genes ideogram (SCP-2363)

eweitz:
This improves performance and includes several suggested UI refinements in the new related genes ideogram.  Compared to #735, the feature is over 2x faster and more polished.

Speed optimizations include:
* [Parallelizing requests for paralogs and interacting genes](https://github.com/eweitz/ideogram/pull/243)
* [Making fewer requests and using faster APIs](https://github.com/eweitz/ideogram/pull/244)
* Starting genomics API calls sooner

The last one is SCP-specific and accounts for most of the code changes seen here.  In brief: if the expression matrices for the current study are all for the same taxon, then we can 1) avoid search-time, gene-specific taxon lookups and thus 2) start related genes API calls sooner.  See diagram below.

Optimizing_related_genes_ideogram_SCP_2020-10-13

User interface improvements include:
* Softer color for chromosome labels and borders
* No more extraneous scrolling for cursor-placement edge cases
* Legend icon rotation matches chromosome annotations
* Z-index fixes for autocomplete in gene search UI [flow 1](#750), and ideogram tooltips relative to loading modal

Case sensitivity in related genes search has also been relaxed.  So now in single-species studies, like we do elsewhere, searching for "gad1" will return results even if the gene name is technically "GAD1".

A look at the latest:
ideogram_for_related_genes_SCP_2020-10-13

This relates to SCP-2363.

### DIFF
--- a/app/views/site/_view_gene_expression_title_bar.html.erb
+++ b/app/views/site/_view_gene_expression_title_bar.html.erb
@@ -8,14 +8,13 @@
 </div>
 
 <script>
-  // Initializes Ideogram for related genes
 
-  document.querySelector('.study-lead').parentElement.innerHTML +=
-    '<div id="related-genes-ideogram-container" class="hidden-related-genes-ideogram"></div>';
-
-  // TODO (SCP-2769): Handle when a searched gene maps to multiple species
-  window.SCP.taxon = '<%= @taxons.first %>';
-
-  createRelatedGenesIdeogram();
+  // Initializes Ideogram for related genes, if it hasn't already been
+  // started for the current searched gene.
+  if (window.SCP.taxons.length > 1) {
+    // TODO (SCP-2769): Handle when a searched gene maps to multiple species
+    window.SCP.taxon = '<%= @gene_taxons.first %>';
+    createRelatedGenesIdeogram();
+  }
 
 </script>
--- a/app/views/site/_view_gene_expression_title_bar.html.erb
+++ b/app/views/site/_view_gene_expression_title_bar.html.erb
@@ -13,8 +13,8 @@
   // started for the current searched gene.
   if (window.SCP.taxons.length > 1) {
     // TODO (SCP-2769): Handle when a searched gene maps to multiple species
-    window.SCP.taxon = '<%= @gene_taxons.first %>';
-    createRelatedGenesIdeogram();
+    const currentTaxon = '<%= @gene_taxons.first %>';
+    createRelatedGenesIdeogram(currentTaxon);
   }
 
 </script>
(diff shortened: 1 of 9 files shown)


eweitz:
The [build error](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/189768106#L2796) seems unrelated to these changes.

```
  1) Error:
SummaryStatsUtilsTest#test_should_get_ingest_run_counts:
TypeError: no implicit conversion of nil into String
    app/lib/summary_stats_utils.rb:72:in `block in ingest_run_count'
    app/lib/summary_stats_utils.rb:71:in `each'
    app/lib/summary_stats_utils.rb:71:in `ingest_run_count'
    test/integration/lib/summary_stats_utils_test.rb:104:in `block in <class:SummaryStatsUtilsTest>'
```

app/views/site/_view_gene_expression_title_bar.html.erb
```diff
@@ -8,14 +8,13 @@
...
-  document.querySelector('.study-lead').parentElement.innerHTML +=
-    '<div id="related-genes-ideogram-container" class="hidden-related-genes-ideogram"></div>';
-
-  // TODO (SCP-2769): Handle when a searched gene maps to multiple species
-  window.SCP.taxon = '<%= @taxons.first %>';
-
-  createRelatedGenesIdeogram();
+  // Initializes Ideogram for related genes, if it hasn't already been
+  // started for the current searched gene.
+  if (window.SCP.taxons.length > 1) {
```
devonbush, line 14: I'm not sure I understand the flow between this and the above setting of window.SCP.taxons.  Why aren't window.SCP.taxon and window.SCP.taxons set at the same time?  And is window.SCP.taxon just the first taxon in the window.SCP.taxons array?  That seems like a needless variable

devonbush: I also don't understand why we would only create the ideogram container if there are more than one taxon

devonbush (changes requested):
it's probably doing really cool stuff that my after-lunch brain is incapable of comprehending, but I don't understand all the complexity around needing both window.SCP.taxon and window.SCP.taxons.  At the very least, I'd suggest renaming window.SCP.taxon to window.SCP.inferredTaxon or window.SCP.inferredTaxonIndex (which would then point to a spot in the taxons array)

app/views/site/_view_gene_expression_title_bar.html.erb
devonbush: ok, now I think I've figured it out.  oof, that is some tricky state-passing between rails templates and JS.  I'm trying to think of ways to clean it -- maybe have createRelatedGenesIdeogram() take a parameter of the taxon to render?   that saves a global variable

devonbush (approved):
approved given the minor update we just chatted about - changing the global window.SCP.taxon to a method parameter

app/views/site/_view_gene_expression_title_bar.html.erb
eweitz: Tricky indeed.  Now mitigated per your suggestion in 0ba699c.

PR #758, opened 2020-10-13
Fix expression threshold slider in inferCNV ideogram (SCP-2745)

eweitz:
This fixes a bug -- seemingly caused by auto-formatting -- in which adjusting the "Expression threshold" slider in inferCNV ideograms would change the visualization as expected, but not be reflected in the slider widget itself.

This satisfies SCP-2745.

### DIFF
--- a/app/assets/javascripts/scp-infercnv-ideogram.js
+++ b/app/assets/javascripts/scp-infercnv-ideogram.js
@@ -75,7 +75,7 @@ function addMarginControl() {
 function updateThreshold(event) {
   let newThreshold
 
-  const expressionThreshold = parseInt(event.target.value)
+  window.expressionThreshold = parseInt(event.target.value)
 
   adjustedExpressionThreshold = Math.round(expressionThreshold/10 - 4)
   const thresholds = window.originalHeatmapThresholds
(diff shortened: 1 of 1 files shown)


bistline (approved):

app/assets/javascripts/scp-infercnv-ideogram.js
```diff
@@ -99,7 +99,7 @@ function updateThreshold(event) {
 
 /** Create slider to adjust expression threshold for "gain" or "loss" calls */
 function addThresholdControl() {
-  if (typeof(expressionThreshold) === 'undefined') {
+  if (typeof window.expressionThreshold === 'undefined') {
```
devonbush, line 102: this variable is only accessed in this file, so a window variable seems like overkill.  Add it to the `let` declarations at the top of this file?  Or if that doesn't work (I don't know the fine details of how scoping works in stuff that's required in by the sprockets pipeline), at least scope this to window.SCP

devonbush (changes requested):

app/assets/javascripts/scp-infercnv-ideogram.js
eweitz: Good call with `let` -- done in fda6d7a.

devonbush (approved):

PR #760, opened 2020-10-15
Only show ideogram if related genes are found (SCP-2816)

eweitz:
This refines handling of Ensembl REST API outages by only showing the new related genes ideogram if results are found.  This resiliency enhancement preserves incremental rendering -- a significant performance optimization.

This satisfies SCP-2816.

### DIFF
--- a/app/assets/javascripts/scp-related-genes-ideogram.js
+++ b/app/assets/javascripts/scp-related-genes-ideogram.js
@@ -17,6 +17,8 @@ function onClickAnnot(annot) {
   const ideogram = this // eslint-disable-line
   document.querySelector('#search_genes').value = annot.name
 
+  setFoundRelatedGenes(false)
+
   // Enable merge of related-genes log props into search log props
   // This helps profile the numerator of click-through-rate
   const event = {}
@@ -71,9 +73,6 @@ function showRelatedGenesIdeogram() { // eslint-disable-line
   }
 
   putIdeogramInPlotTabs(ideoContainer)
-
-  // Make Ideogram visible
-  ideoContainer.classList = 'show-related-genes-ideogram'
 }
 
 /**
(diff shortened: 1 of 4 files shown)


bistline (approved):

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -103,6 +102,35 @@ function onPlotRelatedGenes() {
   window.SCP.log('ideogram:related-genes', props)
 }
 
+/**
+ * Set inner visibility of ideogram
+ *
+ * @param found Boolean Whether related genes have been found
+ */
+function setFoundRelatedGenes(found) {
```
devonbush, line 110: Bootstrap has the built-in class 'hidden' that should simplify this and not require new css.  It looks like you want to start with the wrapper having the 'hidden' class, and then on finding related changes, toggle the 'hidden' class appropriately

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -103,6 +102,35 @@ function onPlotRelatedGenes() {
...
+  const innerFoundClass = (found) ? 'found-related-genes' : ''
+  const innerSelect = '#related-genes-ideogram-container #_ideogramInnerWrap'
+  document.querySelector(innerSelect).classList = innerFoundClass
+
+  if (found) {
+    const ideoContainer =
+      document.querySelector('#related-genes-ideogram-container')
```
devonbush, line 117: I don't understand why you need to change classes on two different elements to make the ideogram visible.  Shouldn't showing/hiding the container be enough?

devonbush (changes requested):
good quick work adding the event handler in ideogram (I presume that's why the version bumped).  Just some css questions

jlchang (approved):
Don't know enough javascript to be able to assess but I appreciate being included!

app/assets/javascripts/scp-related-genes-ideogram.js
eweitz: Good to know!  Done in 9e52773.

app/assets/javascripts/scp-related-genes-ideogram.js
eweitz: Good catch, done in 9e52773.

eweitz:
This needs more work, based on Ensembl REST API behavior seen today.

When Ensembl is down, these changes work well.  When Ensembl is _slow but up_, though, these changes degrade UX compared to staging.  I set up a [monitoring dashboard](https://broadinstitute.slack.com/archives/CBEHTH601/p1602797026197300) to measure Ensembl availability and performance.

Anecdotally, Ensembl is more often slow than down, so what’s on staging seems overall better.  I'll try refining this to optimize around Ensembl availability scenarios, and close it if I don't have something reviewable by Monday.

What's on staging now is viable for production -- this should not block release.

devonbush:
Ok, agree that we can hold off on releasing this.  Let's set aside some time to discuss the most robust solution

eweitz:
I'll close this.  After a period of instability last week, monitoring indicates Ensembl has had excellent availability since Friday.

I plan to soon open a separate PR for logging to better answer "How often is Ensembl _up but slow_", among other analyses.

eweitz:
Ensembl API is currently degraded.  Errors are sometimes thrown slowly (10+ seconds), but often quickly (~0.5 seconds).

PR #764, opened 2020-10-16
 * SCP-2826 Update UX to reflect updates to MTX parsing

knapii-developments:
The upload wizard needs to be updated not that we allow unsorted MTX files. Without this change, users will be misinformed and perform extra work to sort matrixes. 

The new UI now looks like:
Screen Shot 2020-10-16 at 8 45 00 AM

### DIFF
--- a/app/views/studies/initialize_study.html.erb
+++ b/app/views/studies/initialize_study.html.erb
@@ -52,8 +52,7 @@
           </div>
           </br>
           <div class="row">
-            <p class="col-sm-12">* Before upload, <strong>MM Coordinate Matrix file must be <%= link_to 'sorted', 'https://github.com/broadinstitute/single_cell_portal/wiki/Sorting-Matrix-Market-Coordinate-Matrices', target: :_blank %></strong>.
-              Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
+            <p class="col-sm-12">* Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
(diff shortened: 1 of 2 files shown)


app/views/studies/initialize_study.html.erb
```diff
@@ -52,8 +52,7 @@
           </div>
           </br>
           <div class="row">
-            <p class="col-sm-12">* Before upload, <strong>MM Coordinate Matrix file must be <%= link_to 'sorted', 'https://github.com/broadinstitute/single_cell_portal/wiki/Sorting-Matrix-Market-Coordinate-Matrices', target: :_blank %></strong>.
-              Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
+            <p class="col-sm-12">* Additionally, you will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
```
eweitz, line 55: Removing the first bullet point gives us an opportunity to trim even a bit more copy:

```suggestion
            <p class="col-sm-12">* You will need to upload the <%= link_to 'genes', 'https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices', target: :_blank %> (.csv or .tsv) and
```

eweitz (approved):

bistline (changes requested):
So glad we don't need to require this anymore!  There's an additional thing that needs updating though before this is done.  When the user changes the "File Type" dropdown to `MM Coordinate Matrix`, there is some javascript that fires here: https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/views/studies/_initialize_expression_form.html.erb#L75

```
$('#expression_form_<%= study_file._id %>').find('.file-type').on('change', function() {
  if ($(this).val() === 'MM Coordinate Matrix') {
    $('#expression_form_<%= study_file._id %>').append("<%= j render partial: 'sort_warning' %>")
 } else {
   $('#expression_form_<%= study_file._id %>').find($('.sort-warning')).remove();
   $('#expression_form_<%= study_file._id %>').find('.initialize-bundle-target').empty();
 }
});
```

All we need from this function now is this:

```
if ($(this).val() === 'Expression Matrix') {
  $('#expression_form_<%= study_file._id %>').find('.initialize-bundle-target').empty();
}
```

devonbush (approved):
approve, pending Jon's comment

bistline (approved):

PR #768, opened 2020-10-19
adding annotation to study file and updating synth populate script (SCP-2824)

devonbush:
@bistline  This likely still needs work to make sure the test passes -- I'm not sure if the needed genome/annotations are populated.  But it works in development, so I thought I'd put it up so you can grab the model/populator changes.

Also, since this doesn't impact UI/UX, it might as well go on development, and then we can merge development into raw_counts_ingest

### DIFF
--- a/app/lib/synthetic_study_populator.rb
+++ b/app/lib/synthetic_study_populator.rb
@@ -61,14 +61,6 @@ class SyntheticStudyPopulator
     file_infos = study_config['files']
     file_infos.each do |finfo|
       infile = File.open("#{synthetic_study_folder}/#{finfo['filename']}")
-      taxon_id = nil
-      if finfo['species_scientific_name'].present?
-        taxon = Taxon.find_by(scientific_name: finfo['species_scientific_name'])
-        if taxon.nil?
-          throw "You must populate the species #{finfo['species_scientific_name']} to ingest the file #{finfo['filename']}. Stopping populate"
-        end
-        taxon_id = taxon.id
-      end
 
       study_file_params = {
         file_type: finfo['type'],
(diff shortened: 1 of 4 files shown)


app/lib/synthetic_study_populator.rb
```diff
@@ -61,35 +61,25 @@ def self.add_files(study, study_config, synthetic_study_folder, user)
...
         name: finfo['name'] ? finfo['name'] : finfo['filename'],
         upload: infile,
         use_metadata_convention: finfo['use_metadata_convention'] ? true : false,
         status: 'uploading',
-        study: study,
-        taxon_id: taxon_id
+        study: study
       }
 
+      study_file_params = study_file_params.merge(process_genomic_file_params(finfo))
```
bistline, line 74: Small nit - you can call `study_file_params.merge!(process_genomic_file_params(finfo))` instead.

app/lib/synthetic_study_populator.rb
```diff
@@ -61,35 +61,25 @@ def self.add_files(study, study_config, synthetic_study_folder, user)
...
-        if taxon_id.nil?
-          throw "You must specify a species in the study_info.json for Expression Matrix files"
-        end
         exp_finfo_params = finfo['expression_file_info']
         if exp_finfo_params.present?
           exp_file_info = ExpressionFileInfo.new(
             is_raw_counts: exp_finfo_params['is_raw_counts'] ? true : false,
             units: exp_finfo_params['units'],
-            library_construction_protocol: exp_finfo_params['library_construction_protocol']
+            library_construction_protocol: exp_finfo_params['library_construction_protocol'],
```
bistline, line 82: Extraneous comma at the end

bistline (approved):
Small non-blocking styles comments.

app/lib/synthetic_study_populator.rb
devonbush: oooh, nice, I hadn't used that Ruby method before

eweitz (approved):

PR #769, opened 2020-10-19
Avoid scroll upon clicking tabs in Study Overview (SCP-2804)

eweitz:
This removes a jarring automatic scroll upon clicking the top tabs -- Summary, Explore, Download, etc. -- in Study Overview.

The culprit was updating `window.location.href`.  Simply using `pushState` does the same, so removing the unnecessary code fixed the bug.

This satisfies SCP-2804.


bistline (approved):

devonbush (approved):
love fixes that result in less overall code :)

PR #770, opened 2020-10-19
Fixing bulk download size updating

devonbush:
Also fixes a minor issue where the click target for the download button was smaller than the visible button. Updated the download button test suite to use our latest react style.


eweitz (approved):
Looks good!  Please add Jira ticket to title, for auditability

As a passing thought, this PR is a decent use case for code coverage metrics.  The changed test looks equivalent -- and nicely shorter -- but I'd be even more confident if a bot commented here there was no change in test coverage.

bistline (approved):

PR #771, opened 2020-10-20
updating to scp-ingest-pipeline:1.7.0, alexandria_convention:2.2.0

bistline:
Updating to the latest versions of `scp-ingest-pipeline` and `alexandria_convention`.

This PR supports SCP-2702, SCP-2806.


devonbush (approved):

jlchang (approved):
Didn't realize we had been populating biosample_type as a facet. Thanks for pushing this change, Jon.

bistline:
> Didn't realize we had been populating biosample_type as a facet. Thanks for pushing this change, Jon.

That's only for the test, where we call the SearchFacetPopulator.  We haven't done that in production.

PR #773, opened 2020-10-22
Raw counts information download via manifest (SCP-2811, SCP-2817)

devonbush:
This enables download of a manifest file via bulk download, as shown in demo.  As discussed, the file/display format will likely be subject to change over the coming month as we get more feedback, but I think it makes sense to merge this code now into the raw counts branch, so that it's easier to iterate, and since this has a number of infrastructure improvements.

1.  Totats -- in order to support getting the manifest files, I had to significantly improve the robustness of the totat generator and checks.  Totats are now scoped to a specific path, and are 8-char alphanumeric so the search space is much larger.  Also, I've consolidated the totat operations between study download and bulk download -- both now use the API endpoint, and totat authentication is contained within the authenticator concern.

2. I've added window.SCP.ScpApi as an entry to scp-api.js.  This gives our rails templates and other javascript code easy access to the v1/api via the same javascript infrastructure we have for advanced search.  I think this will really let us simplify our JS as we start adding more API endpoints that may be shared across rails templates and react components.

3.  I've added FactoryBot (https://github.com/thoughtbot/factory_bot) for easier construction of single-use studies and other objects in tests.  The syntax for building a factory takes some getting used to, but using them makes tests much cleaner.  The tests only need to specify the attributes they care about, rather than a bunch of extra stuff just so the object passes validation, and we can auto-add uniqueness to names and descriptions for easier cleanup/debug.  

4.  Bulk download now gives error messages on being forbidden to download one of the accessions.  The errors are put in the cfg.txt file, which is not deleted if the curl execution of it fails, so the user can read the message.  Eventually, we'll want the authorization check to also be done when the auth_code is requested, so the user can be alerted in the UI, but this should be good enough for now.

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -592,6 +592,7 @@ module Api
       def generate_manifest
         manifest_obj = BulkDownloadService.generate_study_manifest(@study)
         # prettify the string so it's more human-readable after download
+        response.headers['Content-Disposition'] = 'attachment; filename=manifest.json'
         render json: JSON.pretty_generate(manifest_obj)
       end
 
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -6,9 +6,10 @@ module Api
       include Concerns::FireCloudStatus
       include Swagger::Blocks
 
-      before_action :authenticate_api_user!
+      before_action :authenticate_api_user!, except: [:generate_manifest]
       before_action :set_study, except: [:index, :create]
-      before_action :check_study_permission, except: [:index, :create]
+      before_action :check_study_permission, except: [:index, :create, :generate_manifest]
+      before_action :check_study_view_permission, only: [:generate_manifest]
 
       respond_to :json
 
@@ -538,6 +539,62 @@ module Api
         @synced_study_files.delete_if {|file| bundled_file_ids.include?(file.id)}
       end
 
+      swagger_path '/studies/{id}/manifest' do
+        operation :post do
+          key :tags, [
+              'Studies'
+          ]
+          key :summary, 'Get a Study Manifest file'
+          key :description, 'Return a file summarizing the study and each of the files within it'
+          key :operationId, 'generate_manifest'
+          parameter do
+            key :name, :id
+            key :in, :path
+            key :description, 'ID of Study to generate manifest'
+            key :required, true
+            key :type, :string
+          end
+          parameter do
+            key :name, :file_types
+            key :in, :query
+            key :description, 'File types to include in the manifest (default is all)'
+            key :required, false
+            key :type, :array
+          end
+          response 200 do
+            key :description, 'Manifest file'
+            schema do
+              key :title, 'JSON object of Study and study file'
+              # Once file format is finalized, complete this documentation
+            end
+          end
+          response 401 do
+            key :description, ApiBaseController.unauthorized
+          end
+          response 403 do
+            key :description, ApiBaseController.forbidden('edit Study')
+          end
+          response 404 do
+            key :description, ApiBaseController.not_found(Study)
+          end
+          response 410 do
+            key :description, ApiBaseController.resource_gone
+          end
+          response 406 do
+            key :description, ApiBaseController.not_acceptable
+          end
+          response 500 do
+            key :description, 'Server error when attempting to generate study manifest'
+          end
+        end
+      end
+
+      def generate_manifest
+        manifest_obj = BulkDownloadService.generate_study_manifest(@study)
+        # prettify the string so it's more human-readable after download
+        render json: JSON.pretty_generate(manifest_obj)
+      end
+
       private
 
       def set_study
@@ -551,6 +608,10 @@ module Api
 
       def check_study_permission
         head 403 unless @study.can_edit?(current_api_user)
+      end
+
+      def check_study_view_permission
+        head 403 unless (@study.public || @study.can_view?(current_api_user))
       end
 
       # study params whitelist
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -611,8 +611,18 @@ module Api
         head 403 unless @study.can_edit?(current_api_user)
       end
 
+      # checks the view permissions, either for the current_api_user or a totat, if given
       def check_study_view_permission
-        head 403 unless (@study.public || @study.can_view?(current_api_user))
+        user = current_api_user
+        totat = params[:auth_code]
+        if totat
+          user = User.verify_totat(totat, "#{@study.accession}-view")
+          # if a bad totat is given, fail the request even if there is a valid API user
+          if !user
+            head 403
+          end
+        end
+        head 403 unless (@study.public || @study.can_view?(user))
       end
 
       # study params whitelist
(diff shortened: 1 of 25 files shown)


eweitz (changes requested):
Good stuff.  Thank you for refining the TOTAT code as part of this!

Most of my suggestions are straightforward.  The only one I think requires more work is using accessions instead of IDs for the new API endpoint.  Accessions are better for maintainability even for just internal developers, but I think the case for them is even better for new SCP REST API endpoints.

app/views/site/_study_download_data.html.erb
```diff
@@ -96,55 +97,48 @@
...
           // Get a time-based one time authentication token (totat),
           // then show the download command
-          $.ajax({
-              url: '/single_cell/totat',
-              type: 'post',
-              dataType: 'json',
-              success: function (response) {
-                  var totat = response.totat;
-                  var timeInterval = response.time_interval; // token expires after this many seconds
+          window.SCP.ScpApi.fetchAuthCode().then(
```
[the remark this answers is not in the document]
eweitz: See [other comment](https://github.com/broadinstitute/single_cell_portal_core/pull/773#discussion_r512073035) for context.

app/lib/bulk_download_service.rb
```diff
@@ -184,4 +208,52 @@ def self.generate_output_path_map(study_files)
...
+  # generate a study_info.json object from an existing study_file
+  def self.generate_study_file_manifest(study_file)
+    output = {
+      filename: study_file.name,
+      file_type: study_file.file_type
+    }
+
+    if study_file.expression_file_info
+      # explicitly list properties to copy to avoid accidentally exposing internal fields
+      props_to_copy =  [
```
bistline, line 238: I don't understand why you need to list this out if you're passing them all directly to the output.  This is only going to create a place that will get out of sync with the model (especially since I'm renaming `library_construction_protocol` to `library_preparation_protocol` in my branch already).  

You can achieve the same thing with this:
```
study_file.expression_file_info.attributes.each {|key, value| output[key] = value}
```

config/routes.rb
```diff
@@ -37,6 +37,9 @@
           member do
             post 'sync', to: 'studies#sync_study'
           end
+          member do
```
bistline, line 40: This doesn't need it's own `member` block, you can add it to the above one:

```
member do
  get 'manifest', to: 'studies#generate_manifest'
end
````

File: test/api/search_controller_test.rb [hunk 4]
```diff
@@ -252,11 +252,9 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
     assert_response :unauthorized
 
-    execute_http_request(:get, api_v1_search_bulk_download_path(
-        auth_code: ' ', accessions: study.accession, file_types: file_types)
-    )
-
-    assert_response :unauthorized
+    execute_http_request(:get, api_v1_search_bulk_download_path(accessions: study.accession, file_types: file_types))
+    # response should be successful with no auth_code, since bearer token is valid
+    assert_response :success
```
bistline, line 257: The response should definitely be `401` without an auth code.  If we're responding `200` then we aren't enforcing totats/auth_codes to be present.  Maybe I missed something, but this seems like an obvious error.  If we don't enforce this then users can spam bulk download and rack up enormous egress charges.

test/integration/lib/bulk_download_service_test.rb
```diff
@@ -119,29 +122,56 @@ def setup
...
 
     # clean up
     download_acceptance.destroy
     download_agreement.destroy
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
 
+  test 'should generate study manifest file' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+    study = FactoryBot.create(:detached_study, name: "#{self.method_name}")
```
bistline, line 154: Handy

bistline (changes requested):
I like the `FactoryBot` change, and the `totat` updates seem like they enhance security.  Just some standard refactoring, but I have a quibble about accessions vs. IDs for the API, and it seems like we have a potential issue with requests w/o totats being honored.

app/controllers/api/v1/studies_controller.rb
```diff
@@ -538,6 +539,56 @@ def sync_study
         @synced_study_files.delete_if {|file| bundled_file_ids.include?(file.id)}
       end
 
+      swagger_path '/studies/{id}/manifest' do
```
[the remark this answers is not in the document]
eweitz: Moving this endpoint under a Swagger resource other than [`Studies`](https://singlecell.broadinstitute.org/single_cell/api/swagger_docs/v1#/Studies) (e.g. [`Site`](https://singlecell.broadinstitute.org/single_cell/api/swagger_docs/v1#/Site)) and refactoring it to use accessions would satisfy both our concerns, I think. This endpoint is read-only and not administrative (it's idempotent and not in [Admin](https://singlecell.broadinstitute.org/single_cell/admin)), and `Site` already has another download-related endpoint.

Given the different perspectives here, I don't consider my suggestion blocking.

The topic of accessions and database IDs in the SCP REST API is important and involved, but not urgent -- ideal for engineering strategy discussion. I've added it to the agenda for our November meeting.

app/lib/bulk_download_service.rb
```diff
@@ -14,17 +14,39 @@ class BulkDownloadService
...
     curl_configs = ['--create-dirs', '--compressed']
     # Get signed URLs for all files in the requested download objects, and update user quota
     Parallel.map(study_files, in_threads: 100) do |study_file|
       client = FireCloudClient.new
       curl_configs << self.get_single_curl_command(file: study_file, fc_client: client, user: user,
                                                    study_bucket_map: study_bucket_map, output_pathname_map: output_pathname_map)
     end
+    studies = study_files.map(&:study).uniq
+    study_manifest_paths = studies.map do |study|
+      "#{Rails.application.routes.url_helpers.manifest_api_v1_study_path(study)}"
```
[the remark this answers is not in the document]
bistline: since you have the include now, this isn't needed.  Furthermore, you can call `manifest_api_v1_study_url(study)` to get the fully-qualified URL.  If that barks because of a missing host/protocol, then add this:

```
url_params = {id: study.id}.merge(ApplicationController.default_url_options)
manifest_api_v1_study_url(url_params)
```

bistline (changes requested):
Almost there - a couple refinements regarding url helpers, and one of the calls to `BulkDownloadService.generate_study_files_tsv` hasn't been updated to the new form.

bistline (approved):

eweitz (approved):
I approve given tomorrow's release, but please follow up on a prior [requested change](https://github.com/broadinstitute/single_cell_portal_core/pull/773#discussion_r512060280) for maintainability when you can.

bistline:
> I approve given tomorrow's release, but please follow up on a prior [requested change](https://github.com/broadinstitute/single_cell_portal_core/pull/773#discussion_r512060280) for maintainability when you can.

This isn't being released tomorrow, it's merging into `raw-counts-ingest`.

devonbush:
yup, no hurry to merge this, and given it's complexity, I'm going to give it plenty of time.  @eweitz per your refactor suggestion, I just realized that now that the API endpoint isn't being used by the application proper, the content headers shouldn't reference an attachment.  So I don't think the two methods now share enough code to make the refactor worthwhile. (the shared code is just 'render function output')

PR #774, opened 2020-10-22
SCP-2843: Remove bundled files from bucket & save copy on fail

bistline:
When an ingest run fails, if the file being ingested is the "parent" file of a `StudyFileBundle`, the "bundled" files were neither being cached in the `parse_logs/:study_file_id` directory for QA, nor being deleted from their original locations.  This leaves users in a partially blocked state where they need to manually remove the files from the bucket, and then re-sync the study and delete the "orphaned" `StudyFile` records.  This workaround is also completely non-intuitive, and most users likely would not know how to correct the issue.  This change correctly removes all files from the bundle on ingest failure & caches copies along with the error log for QA to investigate, if necessary.

This PR satisfies SCP-2843.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -251,6 +251,9 @@ class IngestJob
       self.study_file.update(parse_status: 'failed')
       DeleteQueueJob.new(self.study_file).delay.perform
       ApplicationController.firecloud_client.delete_workspace_file(self.study.bucket_id, self.study_file.bucket_location) unless self.persist_on_fail
+      self.study_file.bundled_files.each do |bundled_file|
+        ApplicationController.firecloud_client.delete_workspace_file(self.study.bucket_id, bundled_file.bucket_location) unless self.persist_on_fail
+      end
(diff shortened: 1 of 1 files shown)


app/models/ingest_job.rb
```diff
@@ -391,9 +394,20 @@ def set_subsampling_flags
...
   def create_study_file_copy
     begin
-      ApplicationController.firecloud_client.execute_gcloud_method(:copy_workspace_file, 0, self.study.bucket_id,
-                                                   self.study_file.bucket_location,
-                                                   self.study_file.parse_fail_bucket_location)
+      ApplicationController.firecloud_client.execute_gcloud_method(:copy_workspace_file, 0,
+                                                                   self.study.bucket_id,
+                                                                   self.study_file.bucket_location,
+                                                                   self.study_file.parse_fail_bucket_location)
+      if self.study_file.is_bundled? && self.study_file.is_bundle_parent?
```
jlchang, line 401: Thanks for noticing and adding to this bugfix that we'd also want to cache the associated bundle files on ingest failure!

jlchang (approved):

devonbush (approved):

PR #775, opened 2020-10-22
Proof of concept: Refactor default view for Explore tab in Study Overview (SCP-2815)

eweitz:
This decouples JavaScript and Rails in the top-level template for the "Explore" tab in Study Overview.  General idea:

* Extract Rails variables interpolated in JavaScript.  Move them to top of `.js.erb`, create new `.js` file for plain JavaScript.

This improves testability.  It progresses towards fast, reliable tests of Study Overview that can run with mock data.  This will help ensure upcoming spatial UI features are robust.

The code is a proof-of-concept.  In a subsequent PR, I'd like to refine use of `nonced_javascript_include_tag` and `config.assets.check_precompiled_asset = false`, e.g. to perhaps `embed` plain `.js` files -- unless reviewers know a solution off-hand.  I didn't find anything suitable after searching for some time.

The code merges into a long-lived branch, `spatial-transcriptomics`, to land spatial code not stable enough for mainline.

More context is available in the [Study Overview refactor for spatial UI](https://docs.google.com/document/d/1nnr1NNeejP5I1XApeK9BcKxZLCRbQeNJItGYyeilqm8/edit#) design doc.  This code is a first step in that.

This relates to SCP-2815.

### DIFF
--- a/app/assets/javascripts/study-overview.js
+++ b/app/assets/javascripts/study-overview.js
@@ -6,15 +6,12 @@
 
 const study = window.SCP.study
 
-if (typeof window.SCP.pushedPendingPageViewEvent === 'undefined') {
-  window.SCP.startPendingEvent('user-action:page:view:site-study',
-    { speciesList: window.SCP.taxons },
-    'plot:',
-    true)
 
-  // Flag to ensure we only log this pending event once per page view
-  window.SCP.pushedPendingPageViewEvent = true
-}
+window.SCP.startPendingEvent('user-action:page:view:site-study',
+  { speciesList: window.SCP.taxons },
+  'plot:',
+  true)
+
 
 /** Draws the scatter plot for the default Explore tab view */
 function renderScatter() {
--- a/app/assets/javascripts/study-overview.js
+++ b/app/assets/javascripts/study-overview.js
@@ -1,0 +1,130 @@
+/**
+* @fileoverview Default view of Explore tab in Study Overview
+*
+* Shows "Clusters" and sometimes "Genomes", etc.
+*/
+
+if (typeof window.SCP.pushedPendingPageViewEvent === 'undefined') {
+  window.SCP.startPendingEvent('user-action:page:view:site-study',
+    { speciesList: window.SCP.taxons },
+    'plot:',
+    true)
+
+  // Flag to ensure we only log this pending event once per page view
+  window.SCP.pushedPendingPageViewEvent = true
+}
+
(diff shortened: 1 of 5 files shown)


app/views/site/_study_visualize.html.erb
```diff
@@ -55,118 +55,18 @@
   </div>
 </div>
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
-  <% if @study.can_visualize_clusters? %>
-    if (typeof window.SCP.pushedPendingPageViewEvent === 'undefined') {
-      window.SCP.startPendingEvent('user-action:page:view:site-study',
-                                 {},
-                                 'plot:',
-                                 true);
+  window.SCP.study = {
```
bistline, line 58: I like how this encapsulates the proposed API endpoint for the study overview.

bistline (approved):
I really like where this is going.  In reference to your proposal for reviewing the use of nonced JS tags, I am in full support of this - inline script tags should ideally only be used as a last resort, rather than as a matter of convenience.

app/assets/javascripts/study-overview.js
```diff
@@ -0,0 +1,132 @@
...
+const study = window.SCP.study
+
+if (typeof window.SCP.pushedPendingPageViewEvent === 'undefined') {
+  window.SCP.startPendingEvent('user-action:page:view:site-study',
+    { speciesList: window.SCP.taxons },
+    'plot:',
+    true)
+
+  // Flag to ensure we only log this pending event once per page view
+  window.SCP.pushedPendingPageViewEvent = true
```
devonbush, line 16: let's not use global variables for things like this.  Maybe we can create a parameter "onceOnly" on startPendingEvent (and maybe also on "log") and then metrics-api can can keep a module-scoped list to handle making sure that events like this only fire/start once.  (And I realize the global wasn't added on this change, but now is a good time to get rid of it)

devonbush (changes requested):
looks good, just one refactor that we should tackle while we're touching this code

app/assets/javascripts/study-overview.js
eweitz: [Done](23f9eb0).  The global made sense initially, but it seems one of our refactors has made `window.SCP.pushedPendingPageViewEvent` obsolete.  So I simply removed it, and manually verified it only logs once.

devonbush (approved):

PR #776, opened 2020-10-23
Specifying expression file info during upload/sync (SCP-2807, SCP-2808)

bistline:
This update adds all necessary `ExpressionFileInfo` fields to upload and sync forms, as well as the REST API and Swagger documentation.  This will force users to specify raw counts specific information with all new expression matrix uploads.

This PR satisfies SCP-2807, SCP-2808.


app/models/expression_file_info.rb
```diff
@@ -11,17 +11,16 @@ class ExpressionFileInfo
...
   # note that species and reference genome/annotation live at the study_file level, not here
 
-  UNITS_VALUES = [nil, 'UMI-corrected raw counts', 'raw counts']
-  validates :units, inclusion: {in: UNITS_VALUES}
+  UNITS_VALUES = ['UMI-corrected raw counts', 'raw counts']
+  validates :units, inclusion: {in: UNITS_VALUES}, allow_blank: true
 
-  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'whole cell', 'single nuclei', 'bulk']
-  validates :biosample_input_type, inclusion: {in: BIOSAMPLE_INPUT_TYPE_VALUES}
+  BIOSAMPLE_INPUT_TYPE_VALUES = ['whole cell', 'single nuclei', 'bulk']
```
eweitz, line 17: Question about user-facing term consistency:

Does "biosample input type" differ from the newly-optional `biosample_type` field in the metadata convention?  If so, how?  This seems like a potential source of user confusion.

Perhaps we should omit "input" from this field name throughout our code.

app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="row expression-file-info-fields">
+  <div class="col-sm-2">
+    <%= f.label :is_raw_counts, 'Raw Count Matrix?' %><br />
```
eweitz, line 3: For consistency:

```suggestion
    <%= f.label :is_raw_counts, 'Raw Counts Matrix?' %><br />
```

eweitz (approved):
Looks good.  I noted a nit and a perhaps-out-of-scope question.

app/views/studies/_expression_file_info_fields.html.erb
eweitz: If it indeed should be labeled "raw count", I'd suggest we rename this throughout our code.  Tedious, but best to be consistent very early and before we release.

app/models/expression_file_info.rb
jlchang: The "biosample_type" and "biosample_input_type" are distinct. The biosample_type describes the context of the biosample. The biosample can then be bulk processed or processed as whole cell or single nuclei. 

The "biosample_input_type" is determined by "biosample_type" + an experimental processing activity.

devonbush (approved):
Looks good.  Agree that standardizing on Raw Count vs. Raw Counts will be helpful -- I've probably been using both randomly up to now

app/javascript/styles/_brand.scss
```diff
@@ -156,3 +156,7 @@
   font-size: 18px;
   font-weight: 300;
 }
+
+ul.pager li.dimmed {
```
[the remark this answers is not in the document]
bistline: Yes, good suggestion.

app/views/studies/_expression_file_info_fields.html.erb
bistline: 7ff60c2

app/javascript/styles/_brand.scss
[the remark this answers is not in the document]
bistline: 7ff60c2

bistline:
Build failure is in synthetic study populator, and was triggered by a previous build failure that did not clean up after itself properly.  Marking as a false negative.

PR #777, opened 2020-10-23
Adding code coverage, repo badges (SCP-1970, SCP-2851)

bistline:
This update adds code coverage via [Codecov](https://codecov.io/), as well as badges showing current coverage and build status on the default branch.  This also includes a substantive refactoring of how tests are invoked via `bin/run_tests.sh` to execute all unit/integration tests in a single command, thus allowing for the correct calculation of coverage.  This has an added benefit of reducing the overhead for individual test runs, which decreases runtime by several minutes (modulo delays in file parsing via scp-ingest-pipeline, which are not predictable).

Note: This PR disables the test `'should fail all local parse jobs'` in `test/integration/study_validation_test.rb` as the new testing framework via `rake test` has exposed an I/O issue that causes most downstream tests to fail for unknown reasons.  The issue appears isolated to this single test, which only covers negative testing of parsing gene lists.  With the test removed, subsequent complete test runs pass without issue reliably.  Fixing the issue is now tracked in SCP-2886.

This PR satisfies SCP-1970, SCP-2851.


devonbush (approved):

eweitz (changes requested):
This is a good start!  I'm glad to see better observability for code quality in the works.

We'll need wider measurement for Codecov to be useful.  It currently measures < 3% of essential code.

By "essential code", I mean what is currently tested and our Rails models and React UI -- two areas that are business-critical, tested via CI, and not slated for major refactor.

Model code is 9.1K lines (`sloc app/models/ | grep Source`) and React code is 2.6K lines (`sloc app/javascript/{lib,components}/ | grep Source`).  This code coverage instrumentation currently measures 259 lines of code.

It's worth investigating how to include those areas in code coverage measurement.  Rails and React can likely not be integrated into the same report -- but even including just the models would make this useful.  (It would also be useful to  _somehow_ publicly coverage for our React code.)  If that research yields no quick wins, I think we should omit the coverage badge and open tickets for those TODO items.

eweitz (approved):
Fantastic work -- from 259 lines to 20,968 lines measured!  The fact that you got JavaScript coverage in the same report is impressive and really useful.  This is a major improvement for SCP code quality observability.

app/controllers/api/v1/api_base_controller.rb
```diff
@@ -18,7 +18,7 @@ class ApiBaseController < ActionController::API
       # this is needed to get stack traces of view errors on the console in development
       # otherwise, e.g. errors in study_search_results_objects.rb would just be swallowed and returned as 500
       rescue_from StandardError do |exception|
-        ErrorTracker.report_exception(exception, current_api_user, params.to_unsafe_hash)
```
devonbush, line 21: I don't understand why you don't want to send the current user here?

db/seeds.rb
```diff
@@ -136,14 +168,33 @@
...
+bq_data = JSON.parse bq_seeds.read
+bq_data.each do |entry|
+  entry['CellID'] = SecureRandom.uuid
+  entry['study_accession'] = study.accession
+  entry['file_id'] = metadata_file.id.to_s
+end
+tmp_file = File.new('tmp_bq_seeds.json', 'w+')
+tmp_file.write bq_data.map(&:to_json).join("\n")
+table = ApplicationController.big_query_client.dataset(CellMetadatum::BIGQUERY_DATASET).table(CellMetadatum::BIGQUERY_TABLE)
+job = table.load(tmp_file, write: 'append', format: :json)
```
devonbush, line 196: given that this is 'append', will we ever delete cells from the test BQ instance?  it seems like this might as well write the table fresh, to mirror how we clean out the test mongoDB with each run

test/api/search_controller_test.rb
```diff
@@ -264,24 +277,15 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
   test 'should return preview of bulk download files and total bytes' do
     puts "#{File.basename(__FILE__)}: #{self.method_name}"
 
-    study = Study.find_by(name: "Test Study #{@random_seed}")
+    study = Study.find_by(name: "Testing Study #{@random_seed}")
```
devonbush, line 280: is there a reason for this name change?  I like 'Test Study' better as a name (although either is fine)

devonbush (changes requested):
Looks great!  Just a couple questions

app/controllers/api/v1/api_base_controller.rb
bistline: Good catch, this is a regression.  When I was searching for the cause of all the testing errors I had thought this was part of the cause, but it was a red herring, and I forgot to undo this change.

db/seeds.rb
bistline: I agree, but for some reason when I didn't include the "append" mode, the job fails every time.  I'll try it without again but I couldn't get it to work correctly without this.

test/api/search_controller_test.rb
bistline: This is the first study created in `db/seeds.rb`, and it is the same study that has the synthetic BQ seeds loaded in.

db/seeds.rb
devonbush: makes sense.  The data is scoped by study/file so it shouldn't be a problem -- definitely something we can save for later

devonbush (approved):

PR #778, opened 2020-10-27
preparation protocol updates from Jean (SCP-2809)

devonbush:
Note that this does not handle any renaming of variables from construction->preparation.  I didn't want to step on Jon's toes since I think he's in this code too.  See https://docs.google.com/document/d/1tyXF_IMrgnmEgQZRph_5ryIoTFBNB9Z6YEyoQKbQz9k/edit#


app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
                                  '10x 5\' v2',
                                  '10x 5\' v3',
                                  'CEL-seq2',
                                  'Drop-seq',
-                                 'SCRB-seq',
-                                 'ATAC-seq',
-                                 'ChIP-seq',
-                                 'methylomics']
+                                 'inDrop',
+                                 'Mars-seq',
```
eweitz, line 32: Per similar comment in earlier PR, "MARS" is how it's rendered in literature:

```suggestion
                                 'MARS-seq',
```

eweitz (changes requested):
Looks good, but for one user-facing nit.

app/models/expression_file_info.rb
jlchang: Eric is correct - many thanks for catching that!

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
+                                 'sci-RNA-seq',
+                                 'Seq-Well S^3',
+                                 'Seq-Well v1',
+                                 'Smart-like',
+                                 'Smart-seq2/Fluidigm C1',
+                                 # non-scRNAseq Assays
+                                 # scATACseq
```
jlchang, line 39:
```suggestion
                                 # single cell ATAC-seq assays
```

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
+                                 'dsc-ATAC-seq',
+                                 'dsci-ATAC-seq',
+                                 'scATAC-seq/10x',
+                                 'scATAC-seq/Fluidigm',
+                                 'sci-ATAC-seq',
+                                 'scTHS-seq',
+                                 'snATAC-seq',
+                                 # spatial transcriptomics
```
jlchang, line 47:
```suggestion
                                 # spatial transcriptomics assays
```

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
+                                 '10x Visium',
+                                 'MERFISH',
+                                 'osmFISH',
+                                 'SeqFISH+',
+                                 'Slide-seq',
+                                 'smFISH',
+                                 'Drop-ChIP']
```
jlchang, line 54: For consistency with comments above to indicate assay types would add, as a comment (above Drop-ChIP):
single cell ChIP-seq assays

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
   validates :units, inclusion: {in: UNITS_VALUES}
 
-  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'whole cell', 'single nuclei', 'bulk']
+  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'Whole cell', 'Single nuclei', 'Bulk']
   validates :biosample_input_type, inclusion: {in: BIOSAMPLE_INPUT_TYPE_VALUES}
 
   MULTIMODALITY_VALUES = [nil, 'CITE-seq', 'Patch-seq']
   validates :multimodality, inclusion: {in: MULTIMODALITY_VALUES}
 
   LIBRARY_CONSTRUCTION_VALUES = [nil,
```
jlchang, line 23: We ultimately want `LIBRARY_CONSTRUCTION_VALUES` to be the same as library_preparation_protocol from the metadata convention (ie. if we accept values in both places, we would be checking the data matches for data integrity) not sure if we want/need these names to be the same in code right now. For the user facing text in the expression matrix upload wizard, we should definitely use "Library preparation protocol".

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
...
-                                 'ATAC-seq',
-                                 'ChIP-seq',
-                                 'methylomics']
+                                 'inDrop',
+                                 'Mars-seq',
+                                 'sci-RNA-seq',
+                                 'Seq-Well S^3',
+                                 'Seq-Well v1',
+                                 'Smart-like',
+                                 'Smart-seq2/Fluidigm C1',
```
jlchang, line 37: 'Smart-seq2/plate-based' is missing, please add

jlchang (changes requested):
Mostly minor suggestions, but the list is missing "Smart-seq2/plate-based" which must be added.

I note that `LIBRARY_CONSTRUCTION_VALUES` collected with expression matrices may ultimately be values we report for metadata convention `library_preparation_protocol`. Should the name be changed now (user-facing we should definitely use `Library preparation protocol`), is there value in being able to see "at-a-glance" whether the value was collected via expression matrix ingest or through the metadata file?

app/models/expression_file_info.rb
```diff
@@ -14,29 +14,43 @@ class ExpressionFileInfo
   UNITS_VALUES = [nil, 'UMI-corrected raw counts', 'raw counts']
   validates :units, inclusion: {in: UNITS_VALUES}
 
-  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'whole cell', 'single nuclei', 'bulk']
+  BIOSAMPLE_INPUT_TYPE_VALUES = [nil, 'Whole cell', 'Single nuclei', 'Bulk']
```
jlchang, line 17: These changes require the corresponding values for the synthetic studies also need updating:
db/seed/synthetic_studies/human_raw_counts/study_info.json#L21
db/seed/synthetic_studies/chicken_raw_counts/study_info.json#L21

app/models/expression_file_info.rb
bistline: > We ultimately want `LIBRARY_CONSTRUCTION_VALUES` to be the same as library_preparation_protocol from the metadata convention (ie. if we accept values in both places, we would be checking the data matches for data integrity) not sure if we want/need these names to be the same in code right now. For the user facing text in the expression matrix upload wizard, we should definitely use "Library preparation protocol".

I've made this change on my PR: #776

bistline (approved):

eweitz (approved):

jlchang (approved):
Thanks for making the updates!

knapii-developments (approved):

PR #779, opened 2020-10-28
Enhance related genes analytics for load times, tooltips (SCP-2850)

eweitz:
This improves Mixpanel logging instrumentation for the related genes ideogram.  It gives more insight on load times and tooltips.

For example, we'll be able to measure time of first plot of related genes, and between first and last plot of related genes.  This will help assess impact of occasionally slow or down upstream services (mostly Ensembl).  That will inform improvements around how we display the ideogram (or not) when there are no related genes.  

This satisfies SCP-2850.

### DIFF
--- a/app/assets/javascripts/scp-related-genes-ideogram.js
+++ b/app/assets/javascripts/scp-related-genes-ideogram.js
@@ -76,18 +76,39 @@ function showRelatedGenesIdeogram() { // eslint-disable-line
   ideoContainer.classList = 'show-related-genes-ideogram'
 }
 
-/**
- * Get summary of related-genes ideogram that was just loaded or clicked
- */
-function getRelatedGenesAnalytics(ideogram) {
-  const props = Object.assign({}, ideogram.relatedGenesAnalytics)
-
+/** Refine analytics to use DSP-conventional names */
+function conformAnalytics(props, ideogram) {
   // Use DSP-conventional name
-  props['perfTime'] = props.time
-  delete props['time']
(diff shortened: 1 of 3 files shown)


bistline (approved):

app/assets/javascripts/scp-related-genes-ideogram.js
```diff
@@ -76,18 +76,39 @@ function showRelatedGenesIdeogram() { // eslint-disable-line
...
+/** Log hover over related genes ideogram tooltip */
+function onWillShowAnnotTooltip(annot) {
+  // Ideogram object; used to inspect ideogram state
+  const ideogram = this // eslint-disable-line
+  let props = ideogram.getTooltipAnalytics(annot)
+
+  // `props` is null if it is merely analytics noise.
+  // Accounts for quick moves from label to annot, or away then immediately
+  // back to same annot.  Such action flickers tooltip and represents a
+  // technical artifact that is not worth analyzing.
```
devonbush, line 99: would a good long-term (outside the scope of this PR) be to change the event to 'didShowTooltip' so that the artifacts can get filtered out before this is called?

devonbush (approved):

app/assets/javascripts/scp-related-genes-ideogram.js
eweitz: I like the gist!  I'll try incorporating something like that upstream, so clients don't need to filter out this noise.

PR #780, opened 2020-10-28
removing unneeded ideogram setup code (SCP-2842)

devonbush:
Per Eric, the ideogram initialization code is duplicated by other parts of ideogram.  this also revealed some broken code paths collapsing gene expression by mean/median.  I've done my best to test, but could use some extra eyes once this hits staging since this is stuff I don't know very well

### DIFF
--- a/app/assets/javascripts/scp-dot-plot.js
+++ b/app/assets/javascripts/scp-dot-plot.js
@@ -244,16 +244,17 @@ function renderMorpheusDotPlot(
 }
 
 /** High-level function called from _expression_plots_view.html.erb */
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Looks good, and good find on the dot plot bug.

The sooner dot plots get into global search results for multiple genes, the better.  Then we'll be able to do much easier automated tests ([like with React violin plots](https://github.com/broadinstitute/single_cell_portal_core/blob/development/test/js/search/study-violin-plot.test.js#L22)), and prevent such bugs from going unnoticed again.  The ongoing refactor of the "Explore" tab will help dot plot testability, too.

bistline (approved):

PR #785, opened 2020-11-02
Adding cluster and data array factories

devonbush:
This is the factories from the raw-counts-ingest branch, and then I added two new ones for cluster and data_array.  These are going to be handy for testing the new visualization/explore endpoints

### DIFF
--- a/test/api/site/explore_test.rb
+++ b/test/api/site/explore_test.rb
@@ -25,7 +25,7 @@ class ExploreControllerTest < ActionDispatch::IntegrationTest
                                        cell_data: {
                                          x: [1, 4 ,6],
                                          y: [7, 5, 3],
-                                         text: ['A', 'B', 'C']
+                                         cells: ['A', 'B', 'C']
                                        })
   end
 
--- a/test/api/site/explore_test.rb
+++ b/test/api/site/explore_test.rb
@@ -1,0 +1,49 @@
+require 'api_test_helper'
+
+class SearchControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+  include Requests::JsonHelpers
+  include Requests::HttpHelpers
+
+  HOMO_SAPIENS_FILTER = { id: 'NCBITaxon_9606', name: 'Homo sapiens' }
+  NO_DISEASE_FILTER = { id: 'MONDO_0000001', name: 'disease or disorder' }
+
+  setup do
+    @user = User.find_by(email: 'testing.user.2@gmail.com')
+    OmniAuth.config.mock_auth[:google_oauth2] = OmniAuth::AuthHash.new({
+                                                                           :provider => 'google_oauth2',
+                                                                           :uid => '123545',
+                                                                           :email => 'testing.user@gmail.com'
+                                                                       })
+    sign_in @user
+    @user.update_last_access_at! # ensure user is marked as active
+
+    @basic_study = FactoryBot.create(:detached_study, name: 'Basic Explore')
+    @basic_study_cluster_file = FactoryBot.create(:study_file,
+                                                  name: 'clusterA.txt',
+                                                  file_type: 'Cluster',
+                                                  study: @basic_study)
+    @cluster_group = FactoryBot.create(:cluster_group_with_cells,
+                                       study_file: @basic_study_cluster_file,
+                                       cell_data: {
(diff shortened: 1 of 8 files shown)


bistline (approved):
Really like where this is heading.  Once I'm finished with the `jb-code-coverage` PR, we should investigate where else we can integrate this functionality to make tests faster and more reliable.

test/api/site/explore_test.rb
```diff
@@ -0,0 +1,49 @@
+require 'api_test_helper'
+
+class SearchControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+  include Requests::JsonHelpers
+  include Requests::HttpHelpers
+
+  HOMO_SAPIENS_FILTER = { id: 'NCBITaxon_9606', name: 'Homo sapiens' }
+  NO_DISEASE_FILTER = { id: 'MONDO_0000001', name: 'disease or disorder' }
```
jlchang, line 9: Can we talk through the intent of NO_DISEASE_FILTER? I suspect its inclusion in this test is related to its function in search_controller_test.rb and I want to make sure that what we're doing is still appropriate.

test/factories/cluster_group.rb
```diff
@@ -0,0 +1,42 @@
...
+    study { study_file.study }
+    # create a study but mark as detached, so a Terra workspace is not created
+    factory :cluster_group_with_cells do
+      transient do
+        # cell_data is a hash of three (or 4) arrays: cells, x and y and z
+        cell_data {
+          {
+            x: [1, 2, 3],
+            y: [1, 2, 3],
+            text: ['cellA', 'cellB', 'cellC']
```
eweitz, line 17: `cells` seems like the intended field here.

`text` is similar, but as seen [in `load_cluster_group_data_array_points`](https://github.com/broadinstitute/single_cell_portal_core/blob/4661694410edc44c99af0ffde77d153b9a0195b7/app/controllers/site_controller.rb#L1562), it's an array comprised of cell names and other text. `render_cluster` responses reveal it's used only for tooltips in Plotly.  It's also the likeliest candidate for removal from our server-side-constructed Plotly object -- as it's the largest element in the cluster response, and ought to be computable client-side at runtime.  So relying on `cells` would also be more robust to performance-related refactoring.

test/factories/cluster_group.rb
```diff
@@ -0,0 +1,42 @@
+FactoryBot.define do
+  # gets a study object, defaulting to the first user found.  Auto-appends a unique number to the name, and a note
+  # in the description of the study, to aid in test DB uniqueness and cleanup efforts
+  factory :cluster_group do
+    cluster_type { '2d' }
```
eweitz, line 4: Given `{name: :z, type: 'coordinates'}` below, it seems `cluster_type` should be parameterized.

test/api/site/explore_test.rb
devonbush: Ooops, this was a copy-paste error -- this doesn't belong here.  I just copied the test structure of this test from search_controller (as you guessed).  I'll delete

test/api/site/explore_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    @basic_study_cluster_file = FactoryBot.create(:study_file,
+                                                  name: 'clusterA.txt',
+                                                  file_type: 'Cluster',
+                                                  study: @basic_study)
+    @cluster_group = FactoryBot.create(:cluster_group_with_cells,
+                                       study_file: @basic_study_cluster_file,
+                                       cell_data: {
+                                         x: [1, 4 ,6],
+                                         y: [7, 5, 3],
+                                         text: ['A', 'B', 'C']
```
eweitz, line 28: See comment beginning "`cells` seems like the intended field here".

eweitz (changes requested):
This looks useful!

After renaming a field as described below, I think this will be ready.  I also suggested a parameterization, but I could also see deferring that as an enhancement.

test/factories/cluster_group.rb
eweitz: Thanks for correcting my misunderstanding -- this already is parameter, and '2d' is the default.  What you have here seems ideal.

eweitz (approved):

eweitz:
For auditability, please append a related Jira ticket in the title.

PR #784, opened 2020-10-30
Add API endpoint for "Clusters" plot data (draft)

eweitz:
Work in progress, surfacing for early feedback if anyone is interested.

PR #786, opened 2020-11-03
basic cluster rendering endpoint (SCP-2583)

devonbush:
I believe this is now ready for merge -- not because the endpoints are complete, but because they are in reasonable form and do not cause any regressions.  I'll continue working on the actual implementation of the 'explore' endpoint, but I think that work can be in a separate PR

This pulls in a lot of stuff from Eric's draft PR, and consolidates them in an api/v1/visualization module.  This has some slight changes from the API design doc, as I realized we should just match the pattern we used for the expression_controller for now (that powers the updated global gene search).  

In order to minimize risk (again, this is a very short quarter, with very ambitious goals), I've done everything possible to reduce the amount of regression testing we need.  For example, although many methods in cluster_rendering_service are copies of SiteController methods, I have not updated old code paths to use the new methods.  I'd like to create a separate ticket for that update/removal of the old methods, so that we can tackle that after we've completed the spatial MVP, to minimize the amount of regression testing we need to do before then.


eweitz (commented):
Looking good so far!

It needs more to be viable, but this nicely incorporates #784, refines its roughest edges, and adds substantial new code.  Also, good strategy call on mainlining early behind a flag.  I'll branch off this today to integrate these new endpoints into the spatial UI for the default "Explore" view.

app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+          else
+            render json: {error: "Unknown expression data type: #{data_type}"}, status: 400
+          end
+
+        end
+
+        def render_violin
+          cluster = ClusterRenderingService.get_cluster_group(params, @study)
+          annot_params = ExpressionRenderingService.parse_annotation_legacy_params(params, @study, cluster)
+          annotation = ExpressionRenderingService.get_selected_annotation(@study,
```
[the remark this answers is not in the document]
devonbush: Agree, though I would like something to indicate that they are visualization-focused, rather than CRUD. ClusterVizService?   or should we name them ClusterService, but put them in a visualization module?

test/api/site/explore_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    @basic_study_cluster_file = FactoryBot.create(:study_file,
+                                                  name: 'clusterA.txt',
+                                                  file_type: 'Cluster',
+                                                  study: @basic_study)
+    @cluster_group = FactoryBot.create(:cluster_group_with_cells,
+                                       study_file: @basic_study_cluster_file,
+                                       cell_data: {
+                                         x: [1, 4 ,6],
+                                         y: [7, 5, 3],
+                                         text: ['A', 'B', 'C']
```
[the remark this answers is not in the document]
devonbush: ah, yeah, this will get fixed once the change from the other branch comes through

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+              key :name, :annotation_scope
+              key :in, :query
+              key :description, 'Scope of annotation.   One of “study” or “cluster”.'  # TODO: Use enum
+              key :required, true
+              key :type, :string
+            end
+            parameter do
+              key :name, :subsample
+              key :in, :query
+              key :description, 'Number of cells to use as subsample threshold.  Omit parameter if not subsampling (i.e. if threshold is “all cells”).' # TODO: Use enum
```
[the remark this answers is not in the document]
devonbush: I'll take care them -- thanks

app/models/study.rb
```diff
@@ -1039,6 +1060,26 @@ def default_annotation_type
     end
   end
 
+  # helper to return default annotation name
+  def default_annotation_name
+    if self.default_options[:annotation].nil?
```
bistline, line 1065: We might want to use `.blank?` here instead of `.nil?`.  I know that existing code calls `.nil?` but I think `.blank?` is safer.

config/routes.rb
```diff
@@ -37,7 +37,12 @@
           member do
             post 'sync', to: 'studies#sync_study'
           end
-          resources :expression_data, only: [:show], param: :data_type
+          get 'explore', to: 'visualization/explore#show'
+          resources :expression, controller: 'visualization/expression', only: [:show], param: :data_type
+          resources :clusters, controller: 'visualization/clusters',
```
bistline, line 42: Fancy

bistline (commented):
I like where this is going.  It appears to largely be a relocation of existing code into new classes/controllers, so I don't have much by way of comments in terms of content.  But the structure makes sense to me.

app/controllers/api/v1/visualization/expression_controller.rb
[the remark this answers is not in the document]
eweitz: Makes sense!  "ClusterService, but put them in a visualization module" strikes me as a bit better of the two options.

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -5,7 +5,8 @@ module StudyAware
         extend ActiveSupport::Concern
 
         def set_study
-          @study = Study.find_by(accession: params[:study_id])
+          # allow either accession or id
+          @study = Study.find_by(accession: params[:study_id]) || Study.find_by(id: params[:study_id])
```
bistline, line 9: slightly easier way to do this: 
```suggestion
          @study = Study.any_of({accession: params[:study_id]},{id: params[:study_id]}).first
```

app/controllers/api/v1/studies_controller.rb
```diff
@@ -541,7 +541,8 @@ def sync_study
       private
 
       def set_study
-        @study = Study.find_by(id: params[:id])
+        # enable either id or accession as url param
+        @study = Study.find_by(id: params[:id]) || Study.find_by(accession: params[:id])
```
bistline, line 545:
```suggestion
        @study = Study.any_of({accession: params[:study_id]},{id: params[:study_id]}).first
```

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+          if cluster.is_3d?
+            range = ClusterVizService.set_range(cluster, coordinates.values)
+            if cluster.has_range?
+              aspect = ClusterVizService.compute_aspect_ratios(range)
+            end
+          end
+
+          titles = ClusterVizService.load_axis_labels(cluster)
+
+          axes_full = {
```
bistline, line 148: Consistent formatting of Hashes with rest of controller
```suggestion
          axes_full = {
            titles: titles,
            ranges: range,
            aspects: aspect
          }
```

app/lib/cluster_viz_service.rb
```diff
@@ -0,0 +1,252 @@
...
+  def self.set_range(cluster, inputs)
+    # select coordinate axes from inputs
+    domain_keys = inputs.map(&:keys).flatten.uniq.select {|i| [:x, :y, :z].include?(i)}
+    range = Hash[domain_keys.zip]
+    if cluster.has_range?
+      # use study-provided range if available
+      range = cluster.domain_ranges
+    else
+      # take the minmax of each domain across all groups, then the global minmax
+      vals = inputs.map {|v| domain_keys.map {|k| RequestUtils.get_minmax(v[k])}}.flatten.minmax
```
bistline, line 119: Since this has abstracted out of `site_controller.rb`, we need to get it up to par with what is on `development` now:

```
# take the minmax of each domain across all groups, then the global minmax
raw_values = []
inputs.each do |input|
  domain_keys.each do |domain|
    domain_range = RequestUtils.get_minmax(input[domain])
    # RequestUtils.get_minmax will discard NaN/nil values that were ingested
    # only add domain range to list if we have a valid minmax
    raw_values << domain_range if domain_range.any?
  end
end
aggregate_range = raw_values.flatten.minmax
# add 2% padding to range
padding = (aggregate_range.first - aggregate_range.last) * 0.02
absolute_range = [aggregate_range.first + padding, aggregate_range.last - padding]
range[:x] = absolute_range
range[:y] = absolute_range
range[:z] = absolute_range
```

test/factories/user.rb
```diff
@@ -0,0 +1,18 @@
...
+  factory :user do
+    transient do
+      random_seed { SecureRandom.alphanumeric(4).upcase }
+    end
+    email { "test.user.#{random_seed}@test.edu" }
+    password { "test_password" }
+    factory :api_user do
+      api_access_token {
+                         {
+                           access_token: "test-api-token-#{random_seed}",
```
bistline, line 12: Nice catch!  I've found I needed to do something similar in my branches.

bistline (changes requested):
Small style suggestions, but one request to update `ClusterVizService.set_range` to match what is on `development` now, otherwise SCP-2863 will come back again.

eweitz (approved):
Thanks for the updates -- this looks good to me.

After this is merged, I'll open a PR against `development` for the finished UI refactor and refinements to the "Clusters" endpoint.

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -0,0 +1,172 @@
...
+        # see the 'show' method for documentation of the return type
+        def index
+          begin
+            cluster = @study.default_cluster
+            if cluster.nil?
+              render json: {error: 'No default cluster exists'}, status: 422 and return
+            end
+            render json: self.class.get_cluster_viz_data(@study, cluster, params)
+          rescue => e
+            byebug
```
bistline, line 92: I think this is probably a debug leftover - if there are errors we should be reporting them to Sentry and also the user.

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -0,0 +1,172 @@
...
+            cluster_param_docs.each { |doc| parameter(doc) }
+            cluster_response_docs.each { |doc| response doc[:code], doc }
+          end
+        end
+
+        def show
+          # this endpoint requires a cluster to be specified -- index is used to fetch the default
+          cluster = @study.cluster_groups.find_by(name: params[:cluster_name])
+          if cluster.nil?
+            render json: {error: "No cluster named #{params[:cluster_name]} could be found"}, status: 422 and return
```
bistline, line 120: If the cluster cannot be found, the status should be `404` and not `422`.  `422` is for an "Unprocessable Entity" which is typically reserved for POST/PUT requests that contain validation errors.

bistline (changes requested):
Just some debug leftovers, and a status code change.  Otherwise it looks good.

bistline (approved):

PR #791, opened 2020-11-06
Adding spatial to studyFile (SCP-2874)

devonbush:
Also enables synthetic study populator to populate detached studies, which is sometimes useful for quick testing


eweitz (approved):
Thanks for the quick work!

For auditability, please add a Jira ticket to the title.

bistline (approved):

PR #793, opened 2020-11-09
Temporarily disable Burp (SCP-2878)

eweitz:
This temporarily disables Burp.  Merging #733 broke the build due to an expired license.  I'm working with AppSec to enact a proper solution; this patch from AppSec patches the problem for now.

This relates to SCP-2878.


devonbush (approved):
test failure is unrelated -- looks like an issue with one of the changes I pushed earlier

bistline (approved):

PR #794, opened 2020-11-10
UI refinements from demo, and making advanced search linkable (SCP-2821)

devonbush:
Now you can default advanced search to on (even overriding a user feature flag), by including `advancedSearch` in the query params.  e.g. `https://singlecell.broadinstitute.org/single_cell?advancedSearch`    This will enable reliable linking to the feature from wikis, etc...

### DIFF
--- a/app/javascript/components/HomePageContent.js
+++ b/app/javascript/components/HomePageContent.js
@@ -11,12 +11,13 @@ import StudySearchProvider, { StudySearchContext } from 'providers/StudySearchPr
 import SearchFacetProvider from 'providers/SearchFacetProvider'
 import UserProvider from 'providers/UserProvider'
 import ErrorBoundary from 'lib/ErrorBoundary'
+import * as queryString from 'query-string'
 
 /** include search controls and results */
-export function StudySearchView() {
+export function StudySearchView({advancedSearchDefault}) {
   const studySearchState = useContext(StudySearchContext)
   return <>
-    <SearchPanel searchOnLoad={true}/>
+    <SearchPanel advancedSearchDefault={advancedSearchDefault} searchOnLoad={true}/>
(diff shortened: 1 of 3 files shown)


app/javascript/components/HomePageContent.js
```diff
@@ -26,6 +27,9 @@ const LinkableSearchTabs = function(props) {
   // since the Reach router doesn't own the home path
   const location = useLocation()
   const showGenesTab = location.pathname.startsWith('/single_cell/app/genes')
+  const queryParams = queryString.parse(location.search)
+  // the queryParams object does not support the more typical hasOwnProperty test
+  const advancedSearchDefault = ('advancedSearch' in queryParams)
```
eweitz, line 32: Nice -- a readability plus.

bistline (approved):

app/javascript/styles/_search.scss
```diff
@@ -325,7 +325,8 @@ nav.search-links, #search-panel {
 .advanced-opts {
   vertical-align: top;
   display: inline-block;
-  padding-top: 0.65em;
+  padding: 0.65em 0.25em;
+  border-radius: 1em;
 }
 
 .new-feature {
```
eweitz, line 332: The "BETA" mark is too close to the preceding "Advanced search" text.  Also (and I don't feel too strongly about this) having the mark vertically extend above the text looks odd to me.

```suggestion
.new-feature {
  margin-left: 0.4em;
  vertical-align: sub;
```

eweitz (changes requested):
Good refinements!  

Once the BETA mark has some margin, I think this is ready.  (My vertical alignment suggestion is more subjective.)

eweitz (approved):
Thanks for clarifying my understanding -- what you had in your changes looks good!

PR #797, opened 2020-11-10
Complete refactor for default "Explore" view (SCP-2857)

eweitz:
This matures a proof-of-concept shown in #775.  It finishes decoupling Rails and JS in default "Explore" view as diagrammed in that PR, and integrates new JSON API endpoints from #786.

The refactor has no user-facing changes, but does front-end plumbing needed to add side-by-side plots for spatial transcriptomics.

This satisfies SCP-2857.


bistline (approved):

app/javascript/lib/scp-api.js
```diff
@@ -120,6 +120,67 @@ export function setMockOrigin(origin) {
...
+  } else {
+    return `${delimiter}${param}=${encodeURIComponent(value)}`
+  }
+}
+
+/**
+* Returns initial content for the "Explore" tab in Study Overview
+* @param {String} studyAccession Study accession
+*/
+export async function fetchExploreInitialization(studyAccession, mock=false) {
```
devonbush, line 138: I'm not sure I understand 'initialization' in this function name.  I would mirror the api name more closely, so that we're not tying scp-api.js function names to specific UI elements.  "fetchStudyExploreVizData" or somesuch

app/javascript/lib/scp-api.js
```diff
@@ -120,6 +120,67 @@ export function setMockOrigin(origin) {
...
+  const [exploreInit, perfTime] =
+    await scpApi(apiUrl, defaultInit(), mock, false)
+
+  return exploreInit
+}
+
+/**
+ * Returns an object with scatter plot data for a cluster in a study
+ *
+ * This endpoint is volatile, so intentionally not documented in Swagger.
```
devonbush, line 149: this endpoint is now documented in swagger :)

app/javascript/lib/scp-api.js
```diff
@@ -120,6 +120,67 @@ export function setMockOrigin(origin) {
...
+ * @param {String} studyAccession Study accession
+ * @param {String} cluster Name of cluster, as defined at upload
+ * @param {String} annotation Full annotation name, e.g. "CLUSTER--group--study"
+ * @param {String} subsample Subsampling threshold, e.g. 100000
+ * @param {String} consensus Statistic to use for consensus, e.g. "mean"
+ *
+ * Example:
+ * https://localhost:3000/single_cell/api/v1/studies/SCP56/clusters/Coordinates_Major_cell_types.txt?annotation_name=CLUSTER&annotation_type=group&annotation_scope=study
+ */
+export async function fetchScatter(
```
devonbush, line 162: same as above, I would call this fetchStudyClusterData.  It's up to the UI whether it gets turned into a scatter plot or a nice spreadsheet, so let's make the API function neutral :)

app/javascript/lib/scp-api.js
```diff
@@ -120,6 +120,67 @@ export function setMockOrigin(origin) {
...
+  const annotScopeParam = encodeParam('annotation_scope', annotScope)
+  const annotationParams = annotNameParam + annotTypeParam + annotScopeParam
+
+  const subsampleParam = encodeParam('subsample', subsample)
+  const consensusParam = encodeParam('consensus', consensus)
+
+  const params = annotationParams + subsampleParam + consensusParam
+  const apiUrl = `/studies/${studyAccession}/clusters/${cluster}${params}`
+  // don't camelcase the keys since those can be cluster names,
+  // so send false for the 4th argument
```
devonbush, line 178: great comment, thanks

app/javascript/lib/scp-api.js
```diff
@@ -120,6 +120,67 @@ export function setMockOrigin(origin) {
   mockOrigin = origin
 }
 
+/** Constructs and encodes a URL parameter, if it has a value */
+function encodeParam(param, value, isFirst=false) {
```
devonbush, line 124: I like that you pulled this function out, but I think it's duplicative of functionality in the query-string module we're already using elsewhere. https://www.npmjs.com/package/query-string

app/javascript/lib/study-overview.js
```diff
@@ -0,0 +1,338 @@
+/**
+* @fileoverview Default view of Explore tab in Study Overview
+*
+* Shows "Clusters" and sometimes "Genomes", etc.
+*/
+
+/**
+ * Render cluster code
+ *
+ * TODO: Move this to render-cluster.js, or some such
```
devonbush, line 10: ticket?  or rename from TODO to 'perhaps'?

app/javascript/lib/study-overview.js
```diff
@@ -0,0 +1,338 @@
...
+    }
+  })
+
+  window.SCP.scatterPlotLayout = layout
+}
+
+/**
+ * End render cluster code
+ */
+
```
devonbush, line 169: when I see comments like the above, I think 'I wonder if that means the stuff above/below should be in different files?'

app/javascript/lib/study-overview.js
```diff
@@ -0,0 +1,338 @@
...
+    'plot:',
+    true)
+
+  $('#cluster-plot').data('rendered', false)
+
+  const baseCamera = {
+    'up': { 'x': 0, 'y': 0, 'z': 1 },
+    'center': { 'x': 0, 'y': 0, 'z': 0 },
+    'eye': { 'x': 1.25, 'y': 1.25, 'z': 1.25 }
+  }
```
devonbush, line 305: baseCamera feels like it should be a constant declared outside of this function?

devonbush (changes requested):
looks good.  the only change I feel strongly about is the use of query-params rather than rolling your own query string generator.  The other stuff is just my opinions on comments/naming, which I don't feel too strongly about.

app/javascript/lib/study-overview.js
eweitz: Quite likely -- I added a TODO in b24e07f so this can be refactored as other views (e.g. single-gene) get implemented.

app/javascript/lib/study-overview.js
eweitz: Same, that's my intent.  See other comment about just-ticketed work.

app/javascript/lib/study-overview.js
eweitz: Thanks for the catch!  Ticketed, and added to code comment, in b24e07f.

app/javascript/lib/scp-api.js
eweitz: Good call!  [Done](b24e07f).

app/javascript/lib/scp-api.js
eweitz: That works -- I went with a concisely mirrored name, see b24e07f.

app/javascript/lib/scp-api.js
eweitz: I went with `fetchExplore` to concisely mirror the API name; see b24e07f.

app/javascript/lib/scp-api.js
eweitz: Thanks, [updated](b24e07f).

devonbush (approved):

PR #804, opened 2020-11-16
returning spatial and cluster options in api calls (SCP-2898)

devonbush:
this change is meant to be entirely backwards-compatible, it just adds a list of spatial clusters to the API endpoints.  

As part of testing this, I also started down the path of creating gene and annotation factories, so we can more easily test other vizService methods


eweitz (approved):
Looks good!  For auditability, please add Jira ticket to title.

app/lib/expression_viz_service.rb
```diff
@@ -30,9 +31,16 @@ def self.load_expression_axis_title(study)
     study.default_expression_label
   end
 
-   # helper method to load all possible cluster groups for a study
+   # helper method to load all possible cluster group names for a study
   def self.load_cluster_group_options(study)
-    study.cluster_groups.map(&:name)
+    non_spatial_file_ids = StudyFile.where(study: study, :is_spatial.ne => true).pluck(:id)
```
bistline, line 36:
```suggestion
    non_spatial_file_ids = StudyFile.where(study: study, :is_spatial.ne => true, file_type: 'Cluster').pluck(:id)
```

app/lib/expression_viz_service.rb
```diff
@@ -30,9 +31,16 @@ def self.load_expression_axis_title(study)
...
+    ClusterGroup.where(study: study, :study_file_id.in => non_spatial_file_ids).pluck(:name)
+  end
+
+  # helper method to load spatial coordinate group names
+  def self.load_spatial_options(study)
+    spatial_file_ids = StudyFile.where(study: study, is_spatial: true).pluck(:id)
```
bistline, line 42:
```suggestion
    spatial_file_ids = StudyFile.where(study: study, is_spatial: true, file_type: 'Cluster').pluck(:id)
```

test/integration/lib/expression_viz_service_test.rb
```diff
@@ -0,0 +1,44 @@
...
+                                                  file_type: 'Expression Matrix',
+                                                  study: @basic_study)
+    @cluster_group = FactoryBot.create(:cluster_group_with_cells,
+                                       study_file: @basic_study_cluster_file,
+                                       cell_data: {
+                                         x: [1, 4 ,6],
+                                         y: [7, 5, 3],
+                                         cells: ['A', 'B', 'C']
+                                       },
+                                       annotations: [{name: 'foo', type: 'group', values: ['bar', 'bar', 'baz']}])
```
bistline, line 23: `ClusterGroup` documents have a field called `cell_annotations`, not `annotations`.  Also, the `values` array holds _unique_ values, so it cannot have repeated entries.

bistline (changes requested):
Looks good, the only blocker appears to be in one of the `FactoryBot` objects.  The `cluster_group_with_cells` factory has invalid attributes/values.

test/integration/lib/expression_viz_service_test.rb
devonbush: this is correct.  The `cluster_group_with_cells` factory defines a transient property 'annotations', which it then processes to create both the annotation data array, and the cell_annotations property you mention

devonbush:
the factory you mention is correct -- I'm wondering if there is a better name than 'annotations' for the property so it's not as confusing which are transient properties and which aren't.  maybe `annotation_input`?

test/integration/lib/expression_viz_service_test.rb
bistline: Ah - understood.  Thanks for the clarification.

bistline (approved):

PR #808, opened 2020-11-17
adding logging for non-compliant metadata uploads (SCP-2889)

devonbush:
[major updates to PR following demo today]
This now feature-flags whether the convention is required
If the convention is required, here's the UI the users see 

image

We now auto-add studyAccession to all frontend events that are from a page with a specific study

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -78,6 +78,13 @@ class FileParseService
       when 'Gene List'
         ParseUtils.delay.initialize_precomputed_scores(study, study_file, user)
       when 'Metadata'
+        # log convention compliance -- see SCP-2890
+        if !study_file.use_metadata_convention
+          MetricsService.log('file-upload:metadata:non-compliant', {
+            studyAccession: study.accession,
+            studyFileName: study_file.name
+        }, user)
+        end
         job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_cell_metadata, reparse: reparse,
                             persist_on_fail: persist_on_fail)
         job.delay.push_remote_and_launch_ingest
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -80,10 +80,13 @@ class FileParseService
       when 'Metadata'
         # log convention compliance -- see SCP-2890
         if !study_file.use_metadata_convention
+          if User.feature_flag_for_instance(user, 'convention_required')
+            raise 'Metadata files must comply with the SCP metadata convention'
+          end
           MetricsService.log('file-upload:metadata:non-compliant', {
             studyAccession: study.accession,
             studyFileName: study_file.name
-        }, user)
+          }, user)
         end
         job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_cell_metadata, reparse: reparse,
                             persist_on_fail: persist_on_fail)
--- a/app/views/studies/_initialize_metadata_form.html.erb
+++ b/app/views/studies/_initialize_metadata_form.html.erb
@@ -4,7 +4,7 @@
   <%= f.hidden_field :_id, value: study_file._id %>
   <%= hidden_field_tag :selector, '#metadata_form' %>
   <%= hidden_field_tag :partial, 'initialize_metadata_form' %>
-  <% convention_required = true || User.feature_flag_for_instance(current_user, 'convention_required') %>
+  <% convention_required = User.feature_flag_for_instance(current_user, 'convention_required') %>
 	<div class="form-group row">
 		<div class="col-sm-4">
 			<%= f.label :name %><br />
--- a/app/views/studies/_initialize_metadata_form.html.erb
+++ b/app/views/studies/_initialize_metadata_form.html.erb
@@ -4,6 +4,7 @@
   <%= f.hidden_field :_id, value: study_file._id %>
   <%= hidden_field_tag :selector, '#metadata_form' %>
   <%= hidden_field_tag :partial, 'initialize_metadata_form' %>
+  <% convention_required = true || User.feature_flag_for_instance(current_user, 'convention_required') %>
 	<div class="form-group row">
 		<div class="col-sm-4">
 			<%= f.label :name %><br />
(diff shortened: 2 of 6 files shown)


bistline (approved):

app/views/studies/_initialize_metadata_form.html.erb
```diff
@@ -22,13 +22,17 @@
...
+        <%= render partial: 'metadata_convention_help_popover', locals: {id: f.object.id.to_s} %>
+      <% end %>
+      <br />
+      <%= f.label :use_metadata_convention_true, 'data-analytics-name': 'metadata-convention-optin' do %>
         <%= f.radio_button :use_metadata_convention, true, checked: true, disabled: !study_file.new_record?%>
         Yes
       <% end %>
       &nbsp;
-      <%= f.label :use_metadata_convention_false do %>
+      <%= f.label :use_metadata_convention_false, 'data-analytics-name': 'metadata-convention-optout' do %>
```
eweitz, line 35: Good call logging from the client-side, too.  When convention becomes require for upload via UI, the unconventional uploads will essentially disappear from logs sent via `file_parse_service.rb` and `ingest_job.rb`.  

But this will log users who click "No", see "Convention required, contact us for help", and abandon upload.  That's the important cohort.

Given its importance, please add `studyAccession` to the Mixpanel props sent in logs from the upload wizard.  Currently, it's unavailable; the closest we have is study ID embedded in `appFullPath`, e.g. `/single_cell/studies/5f836edeb067340dbce3a79e/upload`.

Adding `studyAccession` to props sent from upload wizard would let analysts link upload UI events to other events in the same study.

eweitz (changes requested):
Looking good so far!  I suggested a more unified server-side approach, but I defer to you on whether it's in scope.

The only blocker I see is the need to add `studyAccession` to the props in client-side logs from the upload wizard.  My comment below has details.

For auditability, please add a Jira ticket to the PR title.

app/lib/file_parse_service.rb
```diff
@@ -78,6 +78,13 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
       when 'Gene List'
         ParseUtils.delay.initialize_precomputed_scores(study, study_file, user)
       when 'Metadata'
+        # log convention compliance -- see SCP-2890
+        if !study_file.use_metadata_convention
+          MetricsService.log('file-upload:metadata:non-compliant', {
+            studyAccession: study.accession,
+            studyFileName: study_file.name
+        }, user)
+        end
```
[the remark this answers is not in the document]
devonbush: I agree that bespoke events aren't great.  The problem is that MixPanel custom alerts don't seem to let you filter on properties -- they require thresholds based on the total number of events of a given name.  So we need a unique name for this alert

app/views/studies/_initialize_metadata_form.html.erb
devonbush: oh, I misunderstood the ticket.  I thought this was temporary instrumentation until we removed this option.  I didn't realize we were going to keep the option apparently selectable, even after it was forbidden.  That seems like a dark pattern that will make people angry "you thought you had an easier option, but you were wrong! And now we're going to contact you for even thinking about going the easy route"   I wonder if there's another way to instrument that -- sending events when someone goes to the tab and then navigates away without uploading?  Sending special email alerts when someone uploads a metadata file that is clearly not attempting to be convention compliant?

app/lib/file_parse_service.rb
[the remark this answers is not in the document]
eweitz: Gotcha, helpful to know!  So while the unified approach would have benefits, it's clearly not in scope here.

app/lib/file_parse_service.rb
```diff
@@ -78,6 +78,16 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
       when 'Gene List'
         ParseUtils.delay.initialize_precomputed_scores(study, study_file, user)
       when 'Metadata'
+        # log convention compliance -- see SCP-2890
+        if !study_file.use_metadata_convention
+          if User.feature_flag_for_instance(user, 'convention_required')
+            raise 'Metadata files must comply with the SCP metadata convention'
```
bistline, line 84: Are we sure we want to raise an error here?  I'm not sure that would be handled, and would likely show the "Something went wrong" page for users.

bistline (approved):
Looks fine, although I have a question about raising an error if the feature flag isn't present.  I think we're going to need a better mechanism for this rather than just calling `raise`.

app/views/studies/_initialize_metadata_form.html.erb
eweitz: The UI and analytics for SCP-2890 should optimize for learning from that important cohort, the "silent decliners".  These are users who don't upload because they suspect their data isn't conventional, and also don't have strong enough incentives to contact us.  The ticket assumes they'll represent a large proportion of study creators.

You're right a simple "No" option is a bit crude.  At the same time, the update's merely disabled "No" option gives no [affordance](https://en.wikipedia.org/wiki/Affordance) that information is available on hover.  

To address both of those problems, what do you think about a clearer, non-disabled option -- "No, this an issue for my study and I could use help"?  Its obvious clickability would give users an intuitive way to find important information -- "Please contact us at [scp-support@broadinstitute.zendesk.com](mailto:scp-support@broadinstitute.zendesk.com) so we can help or make any updates to the convention" -- while also not confusing them that uploading unconventional metadata is an (easy UI) option.

It'd also preserve the essence of the business goal: giving us a way to reach out to silent decliners, if we desire.  The clearer wording also lets us select for silent decliners who indicate they want help.

eweitz (approved):
Great points on the need for more thoughtful UI, and the feature flags are an awesome addition.

The disabled "No" poses usability challenges, and I propose a solution that addresses your concerns in my latest reply in [our client-side logging thread](https://github.com/broadinstitute/single_cell_portal_core/pull/808/files#r526085185).  If that's unpersuasive, though, I don't consider my concern blocking.

For auditability, please add a Jira ticket to the PR title.

app/lib/file_parse_service.rb
devonbush: Good point -- I could see this going either way.  The reason I put this here is that it seemed like the best way to make sure you couldn't get around the requirement by using, e.g. manage-study, or javascript form shenanigans.  You're right that I don't expect the error to be pretty, but I think that's ok since you shouldn't get this error unless you're doing something weird

eweitz (approved):
Looks good!  Nice UI refinements -- I think they address all the subtle concerns we've realized so far.

PR #811, opened 2020-11-18
refining factories, adding viz service tests (SCP-2898)

devonbush:
@bistline  let me know if these factories look like they'd be useful to your testing efforts,  and what additions we should plan on.  

Obviously, the services and controllers need more comprehensive tests on all their methods, but this is a step in the right direction

### DIFF
--- a/test/api/visualization/cluster_controller_test.rb
+++ b/test/api/visualization/cluster_controller_test.rb
@@ -62,7 +62,7 @@ class ClusterControllerTest < ActionDispatch::IntegrationTest
 
     execute_http_request(:get, api_v1_study_clusters_path(@empty_study))
     assert_equal 404, response.status
-    assert_equal {"error"=>"No default cluster exists"}, json
+    assert_equal({"error"=>"No default cluster exists"}, json)
 
     puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
   end
--- a/test/api/visualization/cluster_controller_test.rb
+++ b/test/api/visualization/cluster_controller_test.rb
@@ -1,6 +1,6 @@
 require 'api_test_helper'
 
-class ExploreControllerTest < ActionDispatch::IntegrationTest
+class ClusterControllerTest < ActionDispatch::IntegrationTest
   include Devise::Test::IntegrationHelpers
   include Requests::JsonHelpers
   include Requests::HttpHelpers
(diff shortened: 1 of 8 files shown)


test/factories/data_array.rb
```diff
@@ -9,10 +9,25 @@
...
+      # convenience allowing specification of a CellMetadatum, rather than id and name individually
+      cell_metadatum { nil }
+    end
+
     study { study_file.study }
     cluster_name { cluster_group.try(:name) || study_file.try(:name) }
     cluster_group_id { cluster_group.try(:id) }
-    linear_data_id { cluster_group.try(:id) || gene.try(:id) }
-    linear_data_type { cluster_group_id ? 'ClusterGroup' : 'Gene' }
+    linear_data_id { cluster_group.try(:id) || gene.try(:id) || cell_metadatum.try(:id) || study.id }
```
bistline, line 20: Nice - I was struggling with this exact issue and I think this encapsulates it nicely.

test/factories/cluster_group.rb
```diff
@@ -6,48 +6,48 @@
...
-          {name: :z, type: 'coordinates'},
-          {name: :cells, type: 'cells'}
+          {name: :x, type: 'coordinates', array_name: 'x'},
+          {name: :y, type: 'coordinates', array_name: 'y'},
+          {name: :z, type: 'coordinates', array_name: 'z'},
+          {name: :cells, type: 'cells', array_name: 'text'}
         ].each do |input_type|
-          if evaluator.cell_data[input_type[:name]]
+          if evaluator.cell_input[input_type[:name]]
             FactoryBot.create(:data_array,
```
bistline, line 40: This works nicely, but there is a problem in that you can define a cluster with `x`, `y`, and `z` coordinates but it will still be marked as `2d`.  We probably need to parameterize that, or make it auto-detect and set that attribute.

test/api/visualization/cluster_controller_test.rb
```diff
@@ -15,35 +15,54 @@ class ExploreControllerTest < ActionDispatch::IntegrationTest
...
+
+    puts "#{File.basename(__FILE__)}: #{self.method_name} successful!"
+  end
+
+  test 'should 404 with no default cluster' do
+    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+
+    execute_http_request(:get, api_v1_study_clusters_path(@empty_study))
+    assert_equal 404, response.status
+    assert_equal {"error"=>"No default cluster exists"}, json
```
bistline, line 65: This is throwing an error in CI, I think because you're declaring the expected output in `assert_equal`:
```suggestion
    expected_error = {"error"=>"No default cluster exists"}
    assert_equal expected_error, json
```

bistline (changes requested):
There's a problem with one of the assertions (see [build](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/202462525) for info), and there's a small issue with `2d` vs. `3d` cluster factories, but otherwise this is great!

test/factories/cluster_group.rb
devonbush: yup, that's how it's set up now,
`cluster_type { '2d' }` defines the default for cluster_type as `'2d'`, but if you specify `cluster_type` it will override the default.

bistline: Ah, understood - I wasn't sure you could specify that dynamically.

eweitz (approved):
I approve, assuming Jon does and tests pass.

test/api/visualization/cluster_controller_test.rb
devonbush: yup, good catch, needs parenthesis so ruby can parse what's the arguments

bistline (approved):

PR #815, opened 2020-11-20
Improve speed and robustness of side-by-side scatter plots (SCP-2881)

eweitz:
This adds support for side-by-side scatter plots in the default "Explore" tab of Study Overview.  It improves speed and robustness compared to #797.

This satisfies SCP-2881.


app/javascript/lib/study-overview.js
```diff
@@ -98,7 +98,14 @@ function calculatePlotRect() {
...
 
   const height = $(window).height() - 250
-  const width = ($('#plots-tab').width() - 80) / numPlots
+
+  // Accounts for expanding "View options" after page load
+  let baseWidth = $('#plots-tab').width()
+
+  // Accounts for when "Summary" tab is selected by default
+  if (baseWidth === 0) baseWidth = $(window).width()
+
```
devonbush, line 107: I think always having ifs followed by {} is a convention we'd like to keep

app/javascript/lib/study-overview.js
```diff
@@ -98,7 +98,14 @@ function calculatePlotRect() {
...
+  const width = (baseWidth - 80) / numPlots
```
devonbush, line 108: can you add a comment about where the 80 comes from?

app/javascript/lib/study-overview.js
```diff
@@ -179,36 +167,47 @@ function renderScatterPlot(target, rawPlot) {
...
-  window.SCP.scatterPlotLayout = layout
 }
 
-/** Fetch and draw scatter plot for default Explore tab view */
-async function drawScatterPlot() {
-  const spinnerTarget = $('#plots')[0]
+/** draws scatter plot */
+async function drawScatterPlot(accession, cluster, plotIndex) {
+  const plotId = `cluster-plot-${plotIndex}`
+  const legendId = `cluster-legend-${plotIndex}`
```
devonbush, line 175: typically it's good form to not have parallel indexes like this, and if you need two things to be related, you either create an object for them or use subselectors within a shared parent container (e.g. the legend DOM element would be found by $(parentContainerId).find('.legend-container')

I don't think that's important to fix here though, since this code will be getting migrated to React

app/javascript/lib/study-overview.js
```diff
@@ -179,36 +167,47 @@ function renderScatterPlot(target, rawPlot) {
...
+
+  let plotClass = '' // For only 1 plot (study without spatial data)
+  if (window.SCP.numPlots > 1) {
+    plotClass = ' plot-left'
+    if (plotIndex !== 0) {
+      plotClass = ' plot-right'
+    }
+  }
```
devonbush, line 183: I think this assigning of CSS classes could be avoided by using the :first-of-type selector to accomplish the styling you want

app/javascript/lib/study-overview.js
```diff
@@ -179,36 +167,47 @@ function renderScatterPlot(target, rawPlot) {
...
+
+  $('#plots .panel-body').append(`
+    <div class="row${plotClass}">
+      <div id="${plotId}"></div>
+      <div id="${legendId}"></div>
+    </div>`)
+
+  const plotJqDom = $(`#${plotId}`)
```
devonbush, line 191: I think the convention for delineating that a variable holds a jquery object is prefixing the variable name with $.  So $plotElement would be a more standard name for this variable

devonbush (changes requested):
good stuff -- this looks cleaner than what was there before.  Just a few style questions

bistline (approved):

app/javascript/lib/study-overview.js
eweitz: I used Jon's suggestion, which removed this line.  But I've updated `.eslintrc.js` to warn in such cases (555d1f9c).

app/javascript/lib/study-overview.js
eweitz: [Done](555d1f9c)!

app/javascript/lib/study-overview.js
eweitz: Good point, I added a comment to this effect in 555d1f9.

app/javascript/lib/study-overview.js
eweitz: Nice!  I did essentially this in 555d1f9.

app/javascript/lib/study-overview.js
eweitz: Good point, done in 555d1f9.

app/javascript/lib/study-overview.js
```diff
@@ -98,7 +98,14 @@ function calculatePlotRect() {
   const numPlots = window.SCP.numPlots
 
   const height = $(window).height() - 250
-  const width = ($('#plots-tab').width() - 80) / numPlots
+
+  // Accounts for expanding "View options" after page load
+  let baseWidth = $('#plots-tab').width()
```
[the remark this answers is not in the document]
eweitz: Ah!  Thanks, done in 555d1f9.  I overlooked how we used that plug-in elsewhere.

app/javascript/styles/_studyOverview.scss
```diff
@@ -0,0 +1,8 @@
+.dual-plot {
+  float: left;
+  margin-left: 0;
+}
```
devonbush, line 4: I don't think this will work in the case of >2 plots, only the last one will get left margin.  I think the structure you want is

```
.dual-plot { 
  float: right;
  margin-left: 60px;
}
.dual-plot:first-of-type {
  float: left;
  margin-left: 0;
}
```

Incidentally, is 'dual-plot' the right class name for this, given that more than two are possible?  Should it be "multi-plot"?

devonbush (approved):
Looks good, one lingering non-critical question.

app/javascript/styles/_studyOverview.scss
eweitz: Thanks, that makes sense.

For at least this quarter, only <= 2 plots are supported in the default view for the Explore tab.  But your suggestion would ease support for > 2 if/when we pursue that, so I'll incorporate that.

PR #816, opened 2020-11-24
Caching API visualization endpoints (SCP-2899)

bistline:
As the JSON API in SCP is subclassed from `ActionController::API`, and not `ActionController::Base`, we cannot use the current caching mechanism of "action caching" employed in the rest of the portal.  This update implements caching for API visualization endpoints by manually writing/reading from the standard `Rails.cache` instance, mimicking the functionality provided by the `caches_action` method from `ActionPack::ActionCaching`.  Cache clearing is still controlled via `CacheRemovalJob`, which removes entries from the file-based cache depending on a provided key.

This new API caching is implemented as a concern, which can be included into any API controller to allow caching for responses via the `:check_api_cache!` and `:write_api_cache!` methods (they must be declared in the included controller to execute).  However, since cache clearing is dependent on functionality from the `StudyFile` class, it should only be used in controllers/endpoints that relate to files (such as cluster & expression rendering).

This PR satisfies SCP-2899.

### DIFF
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -6,6 +6,7 @@ module Api
       class ClustersController < ApiBaseController
         include Concerns::Authenticator
         include Concerns::StudyAware
+        include Concerns::ApiCaching
         include Swagger::Blocks
 
         VALID_SCOPE_VALUES = ['study', 'cluster']
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -16,6 +16,8 @@ module Api
         before_action :set_current_api_user!
         before_action :set_study
         before_action :check_study_view_permission
+        before_action :check_api_cache!
+        after_action :write_api_cache!
 
         # shared
         cluster_param_docs = [{
--- a/app/controllers/api/v1/visualization/expression_controller.rb
+++ b/app/controllers/api/v1/visualization/expression_controller.rb
@@ -6,6 +6,7 @@ module Api
       class ExpressionController < ApiBaseController
         include Concerns::Authenticator
         include Concerns::StudyAware
+        include Concerns::ApiCaching
         include Swagger::Blocks
 
         before_action :set_current_api_user!
--- a/app/controllers/api/v1/visualization/expression_controller.rb
+++ b/app/controllers/api/v1/visualization/expression_controller.rb
@@ -12,7 +12,8 @@ module Api
         before_action :set_current_api_user!
         before_action :set_study
         before_action :check_study_view_permission
-
+        before_action :check_api_cache!
+        after_action :write_api_cache!
 
         # Returns the specified expression data for the gene within the given study, optimized for rendering
         # by the SCP UI.
--- a/app/controllers/api/v1/concerns/api_caching.rb
+++ b/app/controllers/api/v1/concerns/api_caching.rb
@@ -11,11 +11,6 @@ module Api
         # character regex to convert into underscores (_) for cache path setting
         PATH_REGEX =/(\/|%20|\?|&|=)/
 
-        included do
-          before_action :check_api_cache!
-          after_action :write_api_cache!
-        end
-
         # check Rails cache for JSON response based off url/params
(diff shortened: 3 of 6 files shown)


app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -6,6 +6,7 @@ module Visualization
       class ExpressionController < ApiBaseController
         include Concerns::Authenticator
         include Concerns::StudyAware
+        include Concerns::ApiCaching
```
devonbush, line 9: Same concern about ordering

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -6,6 +6,7 @@ module Visualization
       class ClustersController < ApiBaseController
         include Concerns::Authenticator
         include Concerns::StudyAware
+        include Concerns::ApiCaching
```
devonbush, line 9: Do we have to worry about the order of this include, so that it's before/after the check_study_view_permission? since it renders out the response on a cache hit, I think this needs to be included after we call before_action :check_study_view_permission so that the auth check happens first

devonbush (changes requested):
I think this might introduce a security hole, so I want to make sure we're careful about how this interfaces with the auth checks in the controllers

app/controllers/api/v1/visualization/clusters_controller.rb
bistline: That's a really great point - let me do some reading/testing to see what order concern methods fire in based off of includes.

bistline: As I suspected, you're correct - the order of declaration sets order of precedence for before/around/after filters:

https://stackoverflow.com/questions/34609237/order-of-before-filters-in-rails-controller-and-concerns

Also, I put in some logging statements in each concern, and it backed up that assertion:

```
Started GET "/single_cell/api/v1/studies/SCP16/clusters/Cluster%201?annotation_name=Cluster&annotation_scope=study&annotation_type=group" for 172.17.0.1 at 2020-11-25 10:18:53 -0500
  Parameters: {"annotation_name"=>"Cluster", "annotation_scope"=>"study", "annotation_type"=>"group", "study_id"=>"SCP16", "cluster_name"=>"Cluster 1", "cluster"=>{}}
calling check_api_cache!
calling set_current_api_user!
calling current_api_user
calling set_study
calling check_study_view_permission
calling write_api_cache!
Writing to API cache: _single_cell_api_v1_studies_SCP16_clusters_Cluster_1_annotation_name_Cluster_annotation_scope_study_annotation_type_group_cluster_name_Cluster_1
Completed 200 OK in 1054ms (Views: 1.1ms)
```

Based off of this, I think the "correct" thing to do is to modify the concern to not automatically run the before/after filters on all included modules.   This way, we can control what/when methods fire in each controller like we do for `StudyAware` and `Authenticator`.

bistline: 46b4817

app/controllers/api/v1/visualization/expression_controller.rb
bistline: 46b4817

eweitz (approved):

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -15,6 +16,8 @@ class ClustersController < ApiBaseController
         before_action :set_current_api_user!
         before_action :set_study
         before_action :check_study_view_permission
+        before_action :check_api_cache!
+        after_action :write_api_cache!
```
devonbush, line 20: it's a bummer we have to explicitly specify these to get the ordering we want.  Can you get the same effect by just doing the include down here?

devonbush (approved):
Looks good, just one question which isn't critical

app/controllers/api/v1/visualization/clusters_controller.rb
bistline: As we move forward with extracting logic out into the API, I imagine both of these controllers getting much larger.  With that in mind, I decided I wanted these methods declared explicitly so that we could control which endpoints get cached using `:only` or `:except`.  There quite possibly could be methods that we don't want cached in here.

PR #819, opened 2020-11-30
Global gene search dot plots

devonbush:
Ok, this ended up being bigger than expected, but there's lots of good stuff here.  I'm going to add some frontend tests tomorrow, but I think those are not critical for getting this code reviewed and on staging.
1. As shown in demo, global gene search results render as dotplots if they match more than one gene to a study
image

1b. The dotplot legend has been updated with feedback from Eric and Jean -- Jean is still considering/researching whether there might be more easily comprehensible text
2. The cluster/annotation selection UI has been factored out so violin plot and dotplots share the same controller code.
2b. relatedly, I've created a components/visualization directory, and moved our react viz components there.  as usual, git mv did not behave since  StudyViolinPlot is changed considerably.
3. Swagger docs for study-related API controller responses got consolidated
4. I've added minitest/hooks which lets us run setup/teardown code at the test class level, which is nice for self-contained classes of tests that all share studies/users.
5. Because we're increasingly writing code that's both important to test, and the test time can be considerable, I've added a new test helper 'TestInstrumentor'.  Including it in a test suite will result in test output that breaks down the time for each test and the total suite. Note that the total suite time also reflects setup/teardown.  It also lists the line number for each test which makes it really easy to copy into the command line for re-running):

```
SUITE ClustersControllerTest starting
clusters_controller_test.rb:52 test_enforces_view_permissions starting
clusters_controller_test.rb:52 test_enforces_view_permissions completed (0.617 sec)
.clusters_controller_test.rb:70 test_show_should_return_visualization_information starting
clusters_controller_test.rb:70 test_show_should_return_visualization_information completed (1.67 sec)
.SUITE ClustersControllerTest completed (14.47 sec)
```


app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -51,6 +53,28 @@ def render_violin
           render json: render_data, status: 200
         end
 
+        def render_heatmap
+          cluster = RequestUtils.get_cluster_group(params, @study)
```
bistline, line 57: This method is in the `ClusterVizService` now.
```suggestion
          cluster = ClusterVizService.get_cluster_group(params, @study)
```

app/lib/request_utils.rb
```diff
@@ -30,4 +30,5 @@ def self.sanitize_search_terms(terms)
     inputs = terms.is_a?(Array) ? terms.join(',') : terms.to_s
     SANITIZER.sanitize(inputs).encode('ASCII-8BIT', invalid: :replace, undef: :replace)
   end
+
```
bistline, line 33: Extra whitespace

bistline (changes requested):
Just the one broken reference - looks good!

app/lib/cluster_viz_service.rb
```diff
@@ -30,6 +30,60 @@ def self.load_cluster_group_annotations(study, cluster, current_user)
...
+    {
+      default_cluster: study.default_cluster&.name,
+      default_annotation: ExpressionVizService.get_selected_annotation(study, nil, nil, nil, nil),
+      annotations: ClusterVizService.available_annotations(study, nil, user),
+      subsample_thresholds: subsample_thresholds
+    }
+  end
+
+  # returns a flat array of annotation objects, with name, scope, annotation_type, and values for each
+  def self.available_annotations(study, cluster, current_user, annotation_type=nil)
```
bistline, line 46: Probably another candidate for `AnnotationVizService`.

app/lib/cluster_viz_service.rb
```diff
@@ -30,6 +30,60 @@ def self.load_cluster_group_annotations(study, cluster, current_user)
...
+    end
+    annotations.concat(cluster_annots)
+    if current_user.present?
+      user_annotations = UserAnnotation.viewable_by_cluster(current_user, cluster)
+      annotations.concat(user_annotations)
+    end
+    annotations
+  end
+
+  def self.available_annotations_by_cluster(cluster, annotation_type=nil)
```
bistline, line 75: This is cluster-specific, so I think should stay here.  But on the whole, study-level methods that are centered around annotations can be moved.

app/lib/expression_viz_service.rb
```diff
@@ -445,4 +441,61 @@ def self.calculate_median(genes, cell)
...
+      headers = %w(Name Description)
+      cols = cells.size
+      cells.each do |cell|
+        headers << cell
+      end
+      rows = []
+      genes.each do |gene|
+        row = [gene['name'], ""]
+        case collapse_by
+        when 'z-score'
```
bistline, line 466: I know this is just a copy-paste, but since we don't expose these in the UI, do we want to leave them?  Or should we add them as collapse metrics since we have code for it?

config/routes.rb
```diff
@@ -38,11 +38,22 @@
             post 'sync', to: 'studies#sync_study'
           end
           get 'explore', to: 'visualization/explore#show'
+          get 'explore/cluster_options', to: 'visualization/explore#cluster_options'
+
           resources :expression, controller: 'visualization/expression', only: [:show], param: :data_type
```
bistline, line 43: I would prepend all of this with `namespace :visualization`, and then we can remove the `controller:` parameter as the controller inheritance will match (e.g. `Api::V1::Visualization::ExpressionController` in `app/controllers/api/v1/visualization/expression_controller.rb`).

test/api/visualization/annotations_controller_test.rb
```diff
@@ -0,0 +1,98 @@
...
+require 'test_instrumentor'
+
+class AnnotationsControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+  include Requests::JsonHelpers
+  include Requests::HttpHelpers
+  include Minitest::Hooks
+    include TestInstrumentor
+
+  before(:all) do
```
bistline, line 11: VERY cool - no more `setup` and `teardown`!  This should be a major time savings.

test/api/visualization/annotations_controller_test.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  after(:all) do
+    @basic_study.destroy
+    @empty_study.destroy
+    @user.destroy
+    @user2.destroy
+  end
+
+  test 'methods should check view permissions' do
```
bistline, line 51: I've noticed this test doesn't have a positive (e.g. `200`) assertion anywhere - we should add that.

test/api/visualization/clusters_controller_test.rb
```diff
@@ -36,56 +40,65 @@ class ClusterControllerTest < ActionDispatch::IntegrationTest
...
+    # even though this study is only used in one test, create it here so teardown
+    # can be properly managed
+    @slash_study = FactoryBot.create(:detached_study, name: 'Cluster Slash Study')
   end
 
-  teardown do
+  after(:all) do
     delete_study_and_ensure_cascade(@basic_study)
-    @empty_study.destroy
+    delete_study_and_ensure_cascade(@empty_study)
```
bistline, line 52: This will need to get removed as my [PR](#828) deletes this helper.

test/api/visualization/clusters_controller_test.rb
```diff
@@ -36,56 +40,65 @@ class ClusterControllerTest < ActionDispatch::IntegrationTest
...
+    delete_study_and_ensure_cascade(@slash_study)
     @user.destroy
+    @user2.destroy
   end
 
-  test 'should get basic cluster, with metadata annotation by default' do
-    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+  test 'enforces view permissions' do
```
bistline, line 58: Again, this lacks a positive assertion.

test/api/visualization/explore_controller_test.rb
```diff
@@ -1,58 +1,87 @@
...
+  after(:all) do
     @basic_study.destroy
+    @empty_study.destroy
     @user.destroy
+    @user2.destroy
   end
 
-  test 'should get basic study visualization data' do
-    puts "#{File.basename(__FILE__)}: #{self.method_name}"
+  test 'should enforce view permissions' do
```
bistline, line 46: Same lack of positive assertion.

test/api_test_helper.rb
```diff
@@ -19,10 +19,16 @@ def json
   module HttpHelpers
```
bistline, line 19: I like both of these updates!

test/test_instrumentor.rb
```diff
@@ -0,0 +1,27 @@
...
+# messages will be written to stdout of the format
+#
+# SUITE ExploreControllerTest starting
+# explore_controller_test.rb:72 test_should_handle_invalid_study_id starting
+# explore_controller_test.rb:72 test_should_handle_invalid_study_id completed (0.925 sec)
+# .explore_controller_test.rb:78 test_should_enforce_view_privileges starting
+# explore_controller_test.rb:78 test_should_enforce_view_privileges completed (2.249 sec)
+# .SUITE ExploreControllerTest completed (14.294 sec)
+
+module TestInstrumentor
```
bistline, line 11: LOVE - we should add this _everywhere_.

bistline (changes requested):
Overall, this looks great!  Main areas of possible improvement:
1. More explicit Swagger documentation for array/object responses
1. Possibly moving purely annotation-specific methods into `AnnotationVizService` for better consistency
1. Some new tests that ensure permission enforcement lack a positive assertion.

app/javascript/components/visualization/DotPlot.js
```diff
@@ -0,0 +1,141 @@
...
+
+    // Create mapping of selected annotations to colorBrewer colors
+    const annotColorModel = {}
+    annotColorModel[annotation.name] = {}
+    const sortedAnnots = annotation['values'].sort()
+
+    // Calling % 27 will always return to the beginning of colorBrewerSet
+    // once we use all 27 values
+    $(sortedAnnots).each((index, annot) => {
+      annotColorModel[annotation.name][annot] = window.colorBrewerSet[index % 27]
```
eweitz, line 132: Better: use `getColorBrewerColor(index)`.  See example in `study-overview.js`, or `scatter-plot.js` as of #829.

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState, useEffect } from 'react'
```
eweitz, line 1: Per our chat: I updated this module in #829, and will plan to merge my updates over there.

app/controllers/api/v1/visualization/annotations_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+
+        swagger_path '/studies/{accession}/annotations' do
+          operation :get do
+            key :tags, [
+                'Visualization'
+            ]
+            key :summary, 'Get all annotations for the study'
+            key :description, 'Get all annotations for the study, with name, values, type, scope, and cluster_name if applicable'
+            key :operationId, 'study_annotations_path'
+            parameter({
```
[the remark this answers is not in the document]
devonbush: ah, that's a remnant of when I had these factored out in a different way.  I found that calling the method with an object, rather than a block, made it easier to pull out shared params between methods.  Agree that since that's not what's happening here, block-passing matches rest of the codebase better.

app/controllers/api/v1/visualization/annotations_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+                end
+              end
+            end
+            extend SwaggerResponses::StudyControllerResponses
+          end
+        end
+
+        #
+        def index
+          render json: ClusterVizService.available_annotations(@study, nil, current_api_user)
```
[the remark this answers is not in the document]
devonbush: Oh man, you hit on the issue that caused me the most waffling on this entire PR.  I tried a few different configs of clust/annot controllers/services, and they were all vaguely unsatisfying.  I can't even say this current config is the least bad, it just happened to be the one at which I finished the PR. The core problem is that it will never be really obvious which service a given method belongs in.  e.g. if I'm grabbing a list of all annotations of a cluster, is that a cluster or annotation service method? Either way, the controllers will be calling methods in the other service (e.g. get annotation will always need to call a get_cluster method which belongs in the ClusterVizService).  The good news is that it's really easy to rearrange methods between service classes, so I'd propose we table this until we have a few more of these controllers and can see what patterns emerge

app/controllers/api/v1/visualization/annotations_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+            annot_params[:name] = nil
+          end
+          cluster = nil
+          if annot_params[:scope] == 'cluster'
+            if params[:cluster].blank?
+              render(json: {error: 'You must specify the cluster for cluster scoped annotations'}, status: 404) and return
+            end
+            cluster = study.cluster_groups.by_name(params[:cluster])
+          end
+          ExpressionVizService.get_selected_annotation(study,
```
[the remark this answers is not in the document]
devonbush: you're right, and it definitely doesn't belong in ExpressionVizService -- it's just getting the annotation info, no expression info.  I take back my earlier comment (see, I waffle on this constantly :)  will move to annotation service

app/controllers/api/v1/visualization/annotations_controller.rb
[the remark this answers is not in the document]
bistline: I see the issue, and I definitely wondered about some methods as a read them.  You could just move everything annotation-specific into `AnnotationVizService`, except methods that are definitively cluster-specific.  

Strawman for `AnnotationVizService`

- self.load_cluster_group_annotations(study, cluster, current_user) -- from `ClusterVizService`
- self.get_study_annotation_options(study, user) -- from `ClusterVizService`
- self.available_annotations(study, cluster, current_user, annotation_type=nil) -- from `ClusterVizService`
- self.annotation_cell_values_tsv(study, cluster, annotation) -- from `ClusterVizService`
- def self.parse_annotation_legacy_params(study, params) -- from `ExpressionVizService`
- def self.get_selected_annotation(study, cluster, annot_name, annot_type, annot_scope) -- from `ExpressionVizService`

It's not perfect, but it feels to me like those are definitely annotation-based, as opposed to cluster- or expression-based.  I could also see `annotation_cell_values_tsv` staying in `ClusterVizService` as well.

app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -86,7 +88,44 @@ def show
...
+            key :operationId, 'api_v1_studies_explore_cluster_options_path'
+            parameter do
+              key :name, :study_id
+              key :in, :path
+              key :description, 'ID of Study'
+              key :required, true
+              key :type, :string
+            end
+            response 200 do
+              key :description, 'JSON of annotation list, default_cluster, and subsampling thresholds'
```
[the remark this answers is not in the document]
devonbush: yup, I was just being lazy while this was in flux.  will specify

app/javascript/components/visualization/ClusterControls.js
```diff
@@ -0,0 +1,149 @@
...
+  return (
+    <div>
+      <div className="form-group">
+        <label>Load cluster</label>
+        <Select options={clusterOptions}
+          value={{ label: renderParams.cluster, value: renderParams.cluster }}
+          onChange={ cluster => setRenderParams({
+            userUpdated: true,
+            annotation: annotationList.annotations
+              .filter(annot => annot.cluster_name == cluster.value)[0],
```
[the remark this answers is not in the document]
devonbush: agree, will pull this out

app/lib/expression_viz_service.rb
devonbush: Let's leave them in this for now, since eventually this might get used by a UX feature that does expose them

app/javascript/components/visualization/ClusterControls.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/** takes the server response and returns annotation options suitable for react-select */
+function getAnnotationOptions(annotationList, clusterName) {
+  return [{
+      label: 'Study Wide',
+      options: annotationList.annotations
+        .filter(annot => annot.scope == 'study')
+    }, {
+      label: 'Cluster-Based',
```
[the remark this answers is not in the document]
eweitz: Good point, I just commented in SCP-2944 accordingly.

app/javascript/components/visualization/DotPlot.js
```diff
@@ -0,0 +1,141 @@
...
+      setActiveTab: () => {},
+      getWidth: () => $(target).width(),
+      getHeight: () => $(target).height(),
+      getTabCount: () => 1
+    },
+    tools
+  }
+
+  // Set height if specified, otherwise use default setting of 500 px
+  if (dotHeight !== undefined) {
```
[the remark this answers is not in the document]
devonbush: yup, this block is unneeded given the dfault param, good catch

bistline (approved):
I noticed we only have an `AnnotationsControllerTest` and not a `AnnotationVizService` test, but for now this will suffice.

PR #821, opened 2020-12-03
Allowing cluster names with / in API requests (SCP-2920)

bistline:
Cluster files that have been synced into a study that are in sub-folders in the GCP bucket end up with slashes (`/`) in their names.  While this is disallowed at the UI level if a user attempts to enter this manually, it is not blocked at the server or database level.  This update now correctly encodes the slash at the UI level via `encodeURIComponent` to pass to the corresponding API endpoint so that it can be decoded and properly and load the corresponding cluster.  There are currently 135 clusters across 40 studies that are affected, all of which are the default (or only) cluster.

This update also improves database cleanup in the `cluster_controller_test.rb` test suite.

This PR satisfies SCP-2920.


devonbush (approved):

app/javascript/lib/scp-api.js
```diff
@@ -159,15 +159,15 @@ export async function fetchCluster(
   // Digest full annotation name to enable easy validation in API
   const [annotName, annotType, annotScope] = annotation.split('--')
   const paramObj = {
-    annotation_name: annotName,
+    annotation_name: encodeURIComponent(annotName),
```
eweitz, line 162: Blocker: this would cause double-encoding bugs.  We use `queryString.stringify` to encode these parameters, and that function [strictly URI-encodes components by default](https://github.com/sindresorhus/query-string#strict).

```suggestion
    annotation_name: annotName,
```

For a demo, go to an SCP page and run the following in DevTools Console, and note the URL that gets requested.

Code on production now:
```
// Annotation name contains a slash.  Requested URL includes: bar%2Fbaz -- already URI-encoded
cluster = await SCP.API.fetchCluster('SCP1', 'foo', 'bar/baz', 'study', 'group')
```

With `annotation_name: encodeURIComponent(annotName),`:
```
// Annotation name contains a slash.  Requested URL includes: foo%252Fbar -- a double-encoding bug
cluster = await SCP.API.fetchCluster('SCP1', 'foo', 'bar/baz', 'study', 'group')
```

eweitz (changes requested):
Thanks for the quick work.  I see one lurking bug, but otherwise no issues.

Also, does this new back-end sanitization properly handle URLs where path segments and query strings have already been run through `encodeURIComponent` on the front-end?  In the back of my mind, I wonder if it'd cause a different class of double-encoding bug.

Given the pervasive impact (thank you for the analysis!) I support releasing a fix to production as soon as this is merged.

app/javascript/lib/scp-api.js
bistline: Good catch!  I didn't realize that `queryString.stringify` also did this.  I can remove that function call there.  

As far as other potential double-encodings, my testing showed that this is not the case - in fact, providing a non-encoded value for `cluster_name` does not in fact load the correct cluster, and will return `404`.  I can add a negative test in `cluster_controller_test.rb` that will assert this for better coverage/confidence.

bistline: 4e950b, 0aa6082

eweitz (approved):

test/api/visualization/cluster_controller_test.rb
```diff
@@ -39,6 +39,10 @@ class ClusterControllerTest < ActionDispatch::IntegrationTest
   end
 
   teardown do
+    StudyFile.where(study_id: @basic_study.id).destroy_all
+    DataArray.where(study_id: @basic_study.id).destroy_all
```
devonbush, line 43: I don't understand why you can't trust the study destroy to delete these via cascade?  and if you can't, then lets move these subsidiary deletes to a helper method `delete_study_and_ensure_cascade` because otherwise it will be too easy to forget them in other test cleanups

devonbush (changes requested):
test cleanup question -- was fine for the hotfix but I want to make sure I understand why we have the extra delete steps

test/api/visualization/cluster_controller_test.rb
bistline: Deleting a study only deletes the `Study` object - there are no callbacks to remove other data.  We do this in normal operation through the `DeleteQueueJob` class since deletes take a while.  But I take the point - I do have to remember to add these cleanup tasks so that I can ensure the DB is clean between runs, so I'll refactor that into a helper method.

bistline: 0b7abb0

bistline:
Build failure appears to be a red herring: https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/206396012#L2466

Issue is apparently a missing date object for ingest run, which hasn't happened before.  Marking this as a false negative.

devonbush (approved):

PR #822, opened 2020-12-03
Enable live control over action/view caching in local development (SCP-2923)

bistline:
Because SCP employs caching for "visualization" endpoints, making changes to code in local development will not always show updated results due to previous requests being already cached.  This requires developers to constantly run `Rails.cache.clear` in the Rails console in order to remove exiting caches.  This is both tedious and error-prone, and drastically slows down productivity.  Other options are even more disruptive that require commenting out methods, or changing configuration values and restarting the server, which can take minutes to complete.

This update now allows developers to turn on/off caching dynamically by running the `bin/rails dev:cache` command locally in the terminal (_not_ the Rails console).  This will only affect the `development` environment, and will not be used in deployed environments.  This command will place/remove a file in the `tmp` directory called `caching-dev.txt`.  The presence or absence of this file will control caching behavior, and does not require any restarts or reloads.

This PR satisfies SCP-2923.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,29 @@ in package.json), The following snippet can be used to update dependencies not c
 
 https://gist.github.com/pftg/fa8fe4ca2bb4638fbd19324376487f42
 
+## CACHING IN LOCAL DEVELOPMENT
+Single Cell Portal employs action caching for all data visualization requests, such as viewing cluster, violin, or dot 
+plots.  As such, making code changes during local development may not always reflect the most recent changes as a particular 
+request may have already been cached by the server.
+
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Thanks, this seems like another nice DX improvement.  Looking forward to trying it out.

bistline:
Same issue with ingest runs and a missing timestamp: https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/206396012#L2466

This looks like it's only affecting staging/test, so I'm again marking as a false positive.

knapii-developments (approved):

devonbush (commented):
nice README updates

devonbush (approved):

PR #827, opened 2020-12-08
Fixing Codecov integration on PRs (SCP-2927)

bistline:
Merges `codecov` bash upload invocations into single call to merge all coverage data into a single report, rather than using multiple reports separated by flags (`-F`).  This will prevent `codecov` from clobbering existing reports and "losing" coverage data of either the Ruby or Javascript test suites.

Note: the flags still appear in this PR due to multiple reports appearing on the default branch, but should disappear after merge.

This PR satisfies SCP-2927.


eweitz (approved):
Nice example of less code doing more.

devonbush (approved):

PR #828, opened 2020-12-08
Optimizing performance for destroying Study objects (SCP-2940)

bistline:
Due to how Rails handles callbacks on associations when invoking the `destroy` method, doing this for a `Study` object has severe performance implications both in local development and on deployed hosts.  Calling `study.destroy` can essentially grind performance to a halt as the ORM performs expensive collections walks collections with potentially millions of documents, and then calls `destroy` or `delete` on individual associated documents.  In production, this can take over an hour, during which time the portal may become unstable.  Normal study deletions from the UI/API do not incur this penalty, as they rely on the `DeleteQueueJob` class to manage deletes in a background process that runs on a nightly basis, after the database has been reindexed to optimize performance.

This update now removes all `dependent: :destroy` or `dependent: :delete` association callbacks in the `Study` class on the largest collections containing "parsed" data, such as `CellMetadatum`, `Gene`, `ClusterGroup`, and `DataArray`.  A new `before_destroy` callback now ensures that all previously dependent associations preemptively delete data in a performant fashion, and then the parent `Study` object is destroyed, which will remove all other dependent records.  In addition, calling `study.destroy` will now delete convention metadata from BigQuery, without the need to run `DeleteQueueJob`.

Note: this update does not change how `DeleteQueueJob` is invoked or managed - this only applies to calling `study.destroy` directly.   

This PR satisfies SCP-2940.


app/models/study.rb
```diff
@@ -1635,6 +1622,18 @@ def self.delete_all_and_remove_workspaces
     end
   end
 
+  # helper method that mimics DeleteQueueJob.delete_convention_data
+  # referenced from ensure_cascade_on_associations to prevent orphaned rows in BQ on manual deletes
```
devonbush, line 1626: I like this method a lot, should we update DeleteQueueJob to call this method instead of its own?

devonbush (approved):
looks great, appreciate the deep dive you took here.  Just one refactor question

app/models/study.rb
bistline: Originally I thought "no", but on reflection maybe it should.  I'll check it out.

app/models/study.rb
```diff
@@ -1959,4 +1959,26 @@ def prevent_firecloud_attribute_changes
       end
     end
   end
+
+  # delete all records that are associate with this study before invoking :destroy to speed up performance
```
eweitz, line 1963:
```suggestion
  # delete all records that are associated with this study before invoking :destroy to speed up performance
```

test/integration/download_agreement_test.rb
```diff
@@ -21,8 +20,7 @@ class DownloadAgreementTest < ActionDispatch::IntegrationTest
   end
 
   teardown do
-    @study = Study.find_by(name: "Download Agreement #{@random_seed}")
-    delete_study_and_ensure_cascade(@study)
+    Study.find_by(name: "Download Agreement #{@random_seed}").destroy
```
eweitz, line 23: More succinct _and_ faster?  Cool.

eweitz (approved):

PR #829, opened 2020-12-09
Spatial visualizations for single genes (SCP-2855, SCP-2884)

eweitz:
This adds side-by-side scatter plots for the single-gene view of the Explore tab.  It enhances spatial transcriptomics support.

Here's how it looks with real Slide-seq data:
Spatial_transcriptomics_single_gene_side-by-side_scatter_plots_SCP_2020-12-09
_Provenance note: The underlying data is not yet scientifically polished, but is our most representative spatial dataset._

As outlined in #815 and prior PRs linked there, adding the above plots in a maintainable, feasilbly-testable manner required refactoring legacy embedded Rails (ERB) templates into a JS module and REST-API-endpoint structure.

For understandability, there are now clearer connections between module names and UI functionality.  E.g. the abstruse `kernel-functions.js` was renamed to the more obvious `violin-plot.js`.  Function names have been similarly updated.  Compared to #815, `study-overview.js` has been moved to `explore-default.js` and `explore-single.js`, per SCP-2884.  File overviews atop each module summarize inter-module organization and relationships to the UI.

This satisfies SCP-2855 and SCP-2884.


app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -146,11 +152,29 @@ def self.get_cluster_viz_data(study, cluster, url_params)
...
+          if gene_name.blank?
+            # For "Clusters" tab in default view of Explore tab
+            coordinates = ClusterVizService.load_cluster_group_data_array_points(study, cluster, annotation, subsample, colorscale)
+          else
+            # For single-gene view of Explore tab
+            gene = study.genes.by_name_or_id(gene_name, study.expression_matrix_files.map(&:id))
+            title = ExpressionVizService.load_expression_axis_title(study)
+            if is_annotated_scatter
+              # For "Annotated scatter" tab, shown in first tab for numeric annotations
+              coordinates = ExpressionVizService.load_annotation_based_data_array_scatter(study, gene, cluster, annotation, subsample, title)
```
devonbush, line 171: is there a reason we don't pass in the color scale here?

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+  if (numRows > 1) {height = height/numRows - 10}
+
+  // Get width, and account for expanding "View Options" after page load
+  const baseWidth = $('#render-target .tab-content').actual('width')
+  const horizontalPad = 80 // Accounts for empty space to left and right
+  let width = (baseWidth - horizontalPad) / numColumns
+
+  // Ensure plots aren't too small
+  if (height < 100) {height = 100}
+  if (width < 100) {width = 100}
```
devonbush, line 145: I think using Math.max is the more conventional way to set min values like this, but either way is fine
```suggestion
  height = Math.max(height, 100)
  width = Math.max(width, 100)
```

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+    return trace
+  })
+}
+
+/** Get height and width for a to-be-rendered cluster plot */
+function calculatePlotRect(numRows, numColumns) {
+  // Get height
+  const $ideo = $('#_ideogram')
+  const ideogramHeight = $ideo.length ? $ideo.height() : 0
+  const verticalPad = 225 // Accounts for page header, search area, nav tabs
```
devonbush, line 134: hardcoding this (and the horizontalPad below) means that this plot function can't be used on any other page? Maybe make them default params or at least pull them into file-wide constants to make it clearer where the limit is coming from?

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+  // Get width, and account for expanding "View Options" after page load
+  const baseWidth = $('#render-target .tab-content').actual('width')
+  const horizontalPad = 80 // Accounts for empty space to left and right
+  let width = (baseWidth - horizontalPad) / numColumns
+
+  // Ensure plots aren't too small
+  if (height < 100) {height = 100}
+  if (width < 100) {width = 100}
+
+  return { height, width }
```
devonbush, line 147: I'm not a plotly expert, but there is a lot of managing what I would expect to be handled by CSS in this function.  This is probably beyond the scope of this PR, since I think you're just porting over stuff from other files, but I'd love to see us move to a setup where we can just tell plotly "fill up the container you got rendered in", and then let our CSS manage the size of that container

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+    }
+  })
+}
+
+/**
+ * Load and draw scatter plot.  Handles clusters and spatial groups, numeric
+ * and group annotations.
+ *
+ * @param {Object} clusterParams Parameters for `/clusters` plot API endpoint
+ * @param {Object} frame Plot UI properties not from plot API data
```
devonbush, line 202: I'm not sure I understand why you're calling this param 'frame' instead of 'frameConfig'.  'frame' sounds like the value should be a DOM container.  Or maybe containerConfig?

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+  spinner.stop()
+}
+
+/**
+ * Load and draw scatter plots for reference clusters, gene + ref clusters
+ *
+ * @param {Object} study Study object returned by clusters API endpoint
+ * @param {String} gene Searched gene name, e.g. "TP53"
+ * @param {Boolean} hasReference Whether each plot has a paired reference
+ */
```
devonbush, line 276: This function seems to mix the business logic of deciding which plots to render for a given study/gene combo with the rendering logic of 'how to show plots in a grid'.  And I'm not sure either of those belongs in this file, since both may be somewhat page-specific.  For now, I'd just move this function to one of the explore-* js files, so that it's clearer that this function has opinions about what plots it will draw.

app/javascript/lib/scp-api.js
```diff
@@ -122,8 +122,10 @@ export function setMockOrigin(origin) {
 
 /** Constructs and encodes URL parameters; omits those with no value */
 function stringifyQuery(paramObj) {
-  const stringified = queryString.stringify(paramObj, {skipEmptyString: true})
-  return `?${stringified}`;
+  // Usage and API: https://github.com/sindresorhus/query-string#usage
+  const options = { skipEmptyString: true, skipNull: true }
```
devonbush, line 126: nice, good catch

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+    isAnnotatedScatter = true
+  }
+
+  $('#search_annotation').val(annotation)
+  $('#gene_set_annotation').val(annotation)
+
+  let rawPlot = await fetchCluster(
+    accession, cluster, annotation, subsample, null, gene, isAnnotatedScatter
+  )
+  rawPlot = Object.assign(rawPlot, frame)
```
devonbush, line 243: wait, the comment above says the frame is not related to the cluster opts, but now you're merging them?  And maybe rawPlot shoudl be named 'plotConfig' or 'plotOpts' since it's not an actual plot object or container

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+
+  // Consider putting into a dictionary instead of a list
+  window.SCP.scatterPlots.push(rawPlot)
```
devonbush, line 246: this variable is only accessed inside this file except when it's cleared by the respective pages.  So we could scope it to this file and then just add a 'clearScatterPlots' function, and save a global

devonbush (approved):
Looks good -- as I said, nice work getting a bunch of new functionality with only ~200 lines of new code (many of which are just you formatting old code in a more readable way). Lots of style quibbles, but nothing blocking.  The only potential blocker would be if colorscale is a needed param for the annotated scatter

app/views/site/_view_gene_expression.html.erb
```diff
@@ -34,107 +34,8 @@
   <div id="expression-plots"></div>
   <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
 
-      var baseCamera = {
-          "up":{"x":0,"y":0,"z":1},
-          "center":{"x":0,"y":0,"z":0},
-          "eye":{"x":1.25,"y":1.25,"z":1.25}
-      };
+      window.SCP.gene = '<%= @gene %>';
```
bistline, line 37: Always love seeing updates like this

bistline (approved):
No changes other than what Devon already highlighted.

app/controllers/api/v1/visualization/clusters_controller.rb
eweitz: It's omitted because the front-end does not support changing color scale for "Annotated scatter" plots.  (The "Scatter" view option only applies to non-annotated scatter plots for numeric annotations.)

app/javascript/lib/scatter-plot.js
eweitz: Good suggestion with default parameters, done.

app/javascript/lib/scatter-plot.js
eweitz: Agreed!

app/javascript/lib/scatter-plot.js
eweitz: I've renamed these props `apiParams` and `props` to make them a bit more generically applicable.

app/javascript/lib/scatter-plot.js
eweitz: Good point.  This function was a vestige from earlier refactoring.  I moved it to a new module, `explore.js`.

app/javascript/lib/scatter-plot.js
eweitz: To help clarify (or at least not mislead), I renamed `frame` to `props`, and renamed the return variable from `fetchCluster` to `fetchedData`.

app/javascript/lib/scatter-plot.js
eweitz: Done, and similarly for `window.SCP.violinPlots`.  

The globals were convenient for debugging Plotly issues, but using e.g. `document.getElementById('scatter-plot-0').data` and `document.getElementById('scatter-plot-0'.layout` is a decent substitute.

app/javascript/lib/study-overview/explore-default.js
```diff
@@ -0,0 +1,108 @@
...
+export default async function exploreDefault() {
+  window.SCP.scatterPlots = []
+
+  window.SCP.startPendingEvent('user-action:page:view:site-study',
+    { speciesList: window.SCP.taxons },
+    'plot:',
+    true)
+
+  // if tab position was specified in url, show the current tab
+  if (window.location.href.split('#')[1] !== '') {
```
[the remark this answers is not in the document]
eweitz: Good point!  Done, and refactored a bit further.

app/javascript/lib/scatter-plot.js
```diff
@@ -0,0 +1,308 @@
...
+}
+
+/**
+ * Load and draw scatter plots for reference clusters, gene + ref clusters
+ *
+ * @param {Object} study Study object returned by clusters API endpoint
+ * @param {String} gene Searched gene name, e.g. "TP53"
+ * @param {Boolean} hasReference Whether each plot has a paired reference
+ */
+export async function scatterPlots(study, gene=null, hasReference=false) {
```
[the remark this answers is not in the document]
eweitz: Agreed, I followed up [over there](https://github.com/broadinstitute/single_cell_portal_core/pull/829#discussion_r540063892).

PR #833, opened 2020-12-10
fixing coordinate labels bug. (SCP-2942)

devonbush:
This fixes the support issue 69589  from Chris.  It also updates the synthetic populator to be able to handle coordinate label files for easier testing.

### DIFF
--- a/app/models/parse_utils.rb
+++ b/app/models/parse_utils.rb
@@ -410,7 +410,7 @@ class ParseUtils
       end
 
       raw_header_data = coordinate_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
-      header_data = study.sanitize_input_array(raw_header_data)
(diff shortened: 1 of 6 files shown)


eweitz (approved):

app/models/parse_utils.rb
```diff
@@ -410,7 +410,7 @@ def self.initialize_coordinate_label_data_arrays(study, coordinate_file, user, o
       end
 
       raw_header_data = coordinate_data.readline.encode('UTF-8', invalid: :replace, undef: :replace, replace: '').split(/[\t,]/).map(&:strip)
-      header_data = study.sanitize_input_array(raw_header_data)
+      header_data = self.sanitize_input_array(raw_header_data)
```
bistline, line 413: Ahhh yes...  this particular issue has raised its head many times due to various refactors.  Good catch.

bistline (approved):

PR #834, opened 2020-12-11
Change "Multimodality" to "Modality", update options (SCP-2924)

eweitz:
This changes the label of the "Multimodality" dropdown menu to "Modality", and updates the list of available options.  The particular values, order, and default for the list have been specified (and refinements verified) by Jean.  I've tested it manually locally.

Expression_file_upload_wizard_form_raw_counts_modality_2020-12-10

After this last PR for `raw-counts-ingest` is merged, I plan to merge `raw-counts-ingest` into `development`.

This supports SCP-2924.

### DIFF
--- a/app/models/expression_file_info.rb
+++ b/app/models/expression_file_info.rb
@@ -6,7 +6,7 @@ class ExpressionFileInfo
   field :library_preparation_protocol, type: String
   field :units, type: String
   field :biosample_input_type, type: String
-  field :multimodality, type: String
+  field :modality, type: String
   field :is_raw_counts, type: Boolean, default: false
 
   # note that species and reference genome/annotation live at the study_file level, not here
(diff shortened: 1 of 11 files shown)


app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -25,9 +25,9 @@
                  {}, class: 'form-control' %>
```
jlchang, line 25: library_preparation_protocol should not have a default value, can we add `include_blank: true` so the default is blank?

jlchang (changes requested):
Notice a pre-existing issue (we're defaulting to `10x 3' v1` for library preparation protocol instead of leaving it blank. Requesting we make that change so study owners look at the list of library preparation protocol values and don't blithely accept the default (There are many 10x options and v1 is least likely to be the right one).

Also, Modality will be a list of values, rather than a single value. Don't know if it is needful to make that change now or trivial to amend later when we implement multiselect.

app/models/expression_file_info.rb
```diff
@@ -6,7 +6,7 @@ class ExpressionFileInfo
   field :library_preparation_protocol, type: String
   field :units, type: String
   field :biosample_input_type, type: String
-  field :multimodality, type: String
+  field :modality, type: String
```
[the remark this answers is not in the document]
eweitz: It's OK to wait.  This can be handled by a migration.  I tried setting it to `Array` here and elsewhere and encountered issues, so it seems reasonable to bundle the data type refinement with the UI refinement for later.

app/views/studies/_expression_file_info_fields.html.erb
eweitz: The field is marked as required, so I've added a prompt to show "Select one..." by default.

devonbush (approved):

knapii-developments (approved):

PR #836, opened 2020-12-14
Raw counts ingest (SCP-2924)

devonbush:
Still waiting final version of ingest-pipeline, but otherwise this is good to go


test/models/study_file_test.rb
```diff
@@ -60,7 +60,7 @@ def setup
       taxon_id: Taxon.new.id,
       expression_file_info: ExpressionFileInfo.new(
         units: 'bad_value',
-        library_construction_protocol: 'Mars-seq',
+        library_preparation_protocol: 'Mars-seq',
```
eweitz, line 63: To make this validation issue more obvious:

```suggestion
        library_preparation_protocol: 'Mars-seq', # Incorrect case
```

eweitz (approved):
Looks good.  Thanks for fixing my test oversight from #834.

knapii-developments (approved):

knapii-developments:
Approved once build passes.

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -583,7 +583,9 @@ def study_file_params
                                            :x_axis_max, :y_axis_min, :y_axis_max, :z_axis_min, :z_axis_max, :species, :assembly,
                                            options: [:cluster_group_id, :font_family, :font_size, :font_color, :matrix_id,
                                                      :submission_id, :bam_id, :analysis_name, :visualization_name, :cluster_name,
-                                                     :annotation_name])
+                                                     :annotation_name],
+                                           expression_file_info_attributes: [:id, :_destroy, :library_preparation_protocol, :units,
+                                                                             :biosample_input_type, :is_raw_counts])
```
bistline, line 588: We're missing `modality` here.

app/models/study_file.rb
```diff
@@ -82,6 +84,8 @@ class StudyFile
   field :remote_location, type: String, default: ''
   field :options, type: Hash, default: {}
 
+  accepts_nested_attributes_for :expression_file_info
```
bistline, line 87: This is a duplicate

app/views/site/_study_download_data.html.erb
```diff
@@ -96,55 +97,48 @@
...
-              type: 'post',
-              dataType: 'json',
-              success: function (response) {
-                  var totat = response.totat;
-                  var timeInterval = response.time_interval; // token expires after this many seconds
+          window.SCP.API.fetchAuthCode().then(
+            function (response) {
+              var authCode = response.authCode;
+              var timeInterval = response.timeInterval; // token expires after this many seconds
+              debugger;
```
bistline, line 104: I'm assuming this is a holdover from local testing?

app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  <div class="col-sm-2">
+    <%= f.label :units %><br />
+    <%= f.select :units, options_for_select(ExpressionFileInfo::UNITS_VALUES, f.object.units),
+                 {include_blank: true},
+                 class: 'form-control raw-counts-select counts-unit-dropdown', disabled: !f.object.is_raw_counts %>
+  </div>
+  <div class="col-sm-3">
+    <%= f.label :biosample_input_type, "Biosample Input Type <i class='text-danger'>*</i>".html_safe %><br />
+    <%= f.select :biosample_input_type, options_for_select(ExpressionFileInfo::BIOSAMPLE_INPUT_TYPE_VALUES, f.object.biosample_input_type),
+                 {}, class: 'form-control' %>
```
bistline, line 18: Without a prompt value, this will force the user to make a selection.  Since `biosample_input_type` is not required we need to allow blanks at both the UI and model level.
```suggestion
                 {prompt: 'Select one...'}, class: 'form-control' %>
```

db/migrate/20201021141827_clear_old_totats.rb
```diff
@@ -0,0 +1,18 @@
...
+  class UserMigrator
+    include Mongoid::Document
+    store_in collection: 'users'
+    field :totat_t_ti, type: String
+    field :totat
+  end
+
+  def self.up
+    UserMigrator.where(:totat.exists => true).each do |user|
+      user.update(totat: nil, totat_t_ti: nil)
```
bistline, line 12: Since `totat_t_ti` has been removed from the `User` model, we can't use it here in the migration.  It errors out and prevents downstream migrations from running.  If you want to remove the field from the document, you need to call `user.remove_attribute(:totat_t_ti)`.  That will fail gracefully if it doesn't exist.

test/factories/study.rb
```diff
@@ -1,27 +1,23 @@
...
   factory :study do
     transient do
-      auto_suffixes { true }
       # If specified, the created study will be added to the passed-in array after creation
       # this enables easy managing of a central list of studies to be cleaned up by a test suite
       test_array { nil }
+      name_prefix { 'FactoryBot Study' }
+      description_prefix { ' ' }
+    end
+    name { name_prefix + " #{SecureRandom.alphanumeric(5)}" }
```
bistline, line 12: I like this enhancement.

test/integration/lib/bulk_download_service_test.rb
```diff
@@ -115,29 +117,69 @@ def setup
...
+      name: 'metadata2.tsv'
+    )
+
+    manifest_obj = BulkDownloadService.generate_study_manifest(study)
+    # just test basic properties for now, we can add more once the format is finalized
+    assert_equal study.name, manifest_obj[:study][:name]
+    assert_equal 2, manifest_obj[:files].count
+
+    tsv_string = BulkDownloadService.generate_study_files_tsv(study)
+    tsv = ::CSV.new(tsv_string, col_sep: "\t", headers: true)
```
bistline, line 176: Slick!

bistline (changes requested):
A couple small things to change, but on the whole very good.

db/migrate/20201021141827_clear_old_totats.rb
devonbush: yeah, this has been updated to not use the User model anymore -- it's using the UserMigrator class, which is defined in this migration, and so we can guarantee it has the totat_t_ti field

app/controllers/api/v1/study_files_controller.rb
devonbush: good catch

bistline (approved):

PR #837, opened 2020-12-15
Adding url-safe token for some visualization endpoints (SCP-2928)

devonbush:
This enables generation of links that can be passed to morpheus and still be authenticable like https://localhost:3000/single_cell/api/v1/studies/SCP179/expression/heatmap?genes=agpat2%2Capoe&url_safe_token=CGB12WkrqhGWYsuAbGz5XiSwxaAnTAZ7 , but not unnecessarily exposing risk (since putting tokens in URLs is generally a bad idea).

This PR is based off the raw_counts_ingest branch because it builds off the totat changes from that branch.  But the behavior it fixes is on development now

I would support renaming the User field to url_safe_token too, just to be clearer that that's not the 'real' token.  But I think that's a refactor for later.

### DIFF
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -8,6 +8,10 @@ module Api
           {controller: 'search', action: 'bulk_download'},
           {controller: 'studies', action: 'generate_manifest'}
         ]
+        URL_SAFE_TOKEN_ALLOWED_ACTIONS = [
+          {controller: 'expression', action: 'show'},
+          {controller: 'annotations', action: 'cell_values'}
+        ]
         def authenticate_api_user!
           head 401 unless api_user_signed_in?
         end
@@ -57,6 +61,11 @@ module Api
                 # update last_access_at
                 user.update_last_access_at!
                 return user
+              end
+            else
+              if URL_SAFE_TOKEN_ALLOWED_ACTIONS.include?({controller: controller_name, action: action_name}) && params[:url_safe_token]
+                url_safe_token = params[:url_safe_token]
+                return User.find_by(authentication_token: url_safe_token)
               end
             end
           end
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -33,14 +33,7 @@ module Api
         def get_current_api_user
           user = nil
           if TOTAT_REQUIRED_ACTIONS.include?({controller: controller_name, action: action_name})
-            # check for a valid totat/action
-            if !params[:auth_code].present?
-              return nil
-            end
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -8,6 +8,10 @@ module Authenticator
           {controller: 'search', action: 'bulk_download'},
           {controller: 'studies', action: 'generate_manifest'}
         ]
+        URL_SAFE_TOKEN_ALLOWED_ACTIONS = [
```
bistline, line 11: I might combine `URL_SAFE_TOKEN_ALLOWED_ACTIONS` and `TOTAT_REQUIRED_ACTIONS` together into the same variable as it feels like they're doing the same thing.  You could add in the url param as well and have a `ALLOWED_ACTIONS_BY_PARAM` with `controller`, `action`, and `param_name` attributes for each entry.

app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -58,6 +62,11 @@ def get_current_api_user
                 user.update_last_access_at!
                 return user
               end
+            else
```
bistline, line 65: It really feels like the cyclomatic complexity of this method is ballooning.  We might want to extract some logic into helper functions as it's already at 40 lines now.

bistline (approved):
Looks good, but I feel like `get_current_api_user` in `authenticator.rb` is getting complicated and could stand to be refactored some.

app/helpers/application_helper.rb
```diff
@@ -361,6 +361,15 @@ def get_user_access_token(user)
     end
   end
 
+   # Return a scope-limited access token that can be used in URLs (e.g. in urls passed to Morpheus)
+  def get_url_safe_access_token(user)
```
eweitz, line 365: It'd help to have a brief code comment on the difference between a URL safe access token and a TOTAT.

eweitz (approved):
Looks good.  I suggested code-commentary to distinguish from similar existing code.  I found myself wondering "Why not just use a TOTAT here?" (and will when I encounter this code 6+ months from now).

I see no blockers, though!

devonbush:
Great comments -- thanks @eweitz  and @bistline  The overall sentiment I get from both of you (and I agree with it) is that there's a too much overlap between what totats and url-safe-tokens are doing.  I'm going to create an infrastructure ticket to merge totat and url_safe_token capability. There will be some subtlety in figuring out how they will generate and support repeated calls for some operations but not others, but I think it will be worth it to consolidate.

app/controllers/api/v1/concerns/authenticator.rb
devonbush: I like keeping these separate for now since they're different authentication mechanisms.  But 100% agree there's some overlap that we should address in a systematic way

PR #838, opened 2020-12-16
Integration tests for raw counts data (SCP-2951)

bistline:
This update adds a full end-to-end integration test for adding raw counts & processed expression matrices to a study to ensure that users can repeat cells across raw/processed matrix files.  This uses the existing `SyntheticStudyPopulatorTest` as it already contained the necessary data, and only needed to wait for the ingest runs to complete before validating the cells have been ingested.

This also contains refactoring to `all_cells_array` and associated methods in the `Study` model.  Since cells can now be repeated across expression matrix files, de-duplication is required.  In addition, success emails for raw counts matrix files now report the number of cells ingested, as there are no genes created from raw counts files.

This PR satisfies SCP-2951.


eweitz (approved):

app/models/ingest_job.rb
```diff
@@ -470,9 +470,16 @@ def generate_success_email_array
 
     case file_type
     when /Matrix/
-      genes = Gene.where(study_id: self.study.id, study_file_id: self.study_file.id).count
-      message << "Gene-level entries created: #{genes}"
-      mixpanel_log_props.merge!({:numGenes => genes})
+      # since genes are not ingested for raw counts matrices, report number of cells ingested
+      if self.study_file.is_raw_counts_file?
+        cells = self.study.expression_matrix_cells(self.study_file).count
```
devonbush, line 475: FYI: I *think* these 'self's are unneeded, but I'm not sure.  (and don't mind if they're left in even if they're unneeded).  Rubocop has a really good cop for detecting unecessary 'self' statements, so sometime after this release, we should run it and clean out all the extraneous selfs

app/models/study.rb
```diff
@@ -1230,29 +1230,29 @@ def all_cells
   # return an array of all single cell names in study, will check for master list of cells or concatenate all
   # cell lists from individual expression matrices
   def all_cells_array
-    vals = []
     arrays = DataArray.where(study_id: self.id, linear_data_type: 'Study', linear_data_id: self.id, name: 'All Cells').order_by(&:array_index)
     if arrays.any?
-      arrays.each do |array|
-        vals += array.values
-      end
+      arrays.pluck(:values).reduce(&:+)
```
devonbush, line 1235: 1,000 points from me for using 'pluck'. :)

app/models/study.rb
```diff
@@ -1230,29 +1230,29 @@ def all_cells
...
-        vals += array.values
-      end
+      expression_cells = self.expression_matrix_cells(file)
+      all_cells += expression_cells
     end
-    vals
+    all_cells.uniq
+  end
+
+  # return the cells found in a single expression matrix
```
devonbush, line 1251: nice, I like how you pulled this out

test/integration/synthetic_study_populator_test.rb
```diff
@@ -24,16 +29,16 @@ class SyntheticStudyPopulatorTest < ActionDispatch::IntegrationTest
   teardown do
     # this will also destroy the dependent annotation and assembly
     @taxon.destroy!
+    @study = Study.find_by(name: SYNTH_STUDY_INFO[:name])
     if @study.present?
-      Study.find_by(name: SYNTH_STUDY_INFO[:name]).destroy_and_remove_workspace
+      @study.destroy_and_remove_workspace
```
devonbush, line 34: should this just be '@study.destroy'  now?

test/integration/synthetic_study_populator_test.rb
```diff
@@ -24,16 +29,16 @@ class SyntheticStudyPopulatorTest < ActionDispatch::IntegrationTest
   teardown do
     # this will also destroy the dependent annotation and assembly
     @taxon.destroy!
+    @study = Study.find_by(name: SYNTH_STUDY_INFO[:name])
```
devonbush, line 32:
```suggestion
    study = Study.find_by(name: SYNTH_STUDY_INFO[:name])
```
I don't think this needs to be an instance variable

devonbush (approved):
Good stuff, that extra test is going to be really helpful.  Some style suggestions, nothing blocking

app/models/ingest_job.rb
bistline: Agreed, we had said previously that we wanted to do a full pass of `rubocop`.  And you're correct that they aren't needed - I'm just pedantic about `self`.  But it would be good if we were consistent across the codebase.

test/integration/synthetic_study_populator_test.rb
bistline: Agreed

PR #847, opened 2020-12-18
switching to find_or_create for test users

devonbush:


eweitz (approved):
Looks good!  Nit: the "find_or_create" part of the PR title seems no longer accurate.

For auditabilty, please add the Jira ticket to the PR title.

bistline:
@devon I've encountered this error as well, and it seems to only happen in CI runs.  Given that we get our daily server report emails without issue, I don't think this is an actual bug, but just weirdness with `PAPI`.

bistline (approved):

PR #848, opened 2020-12-18
Prevent spamming of `UploadCleanupJob` on error (SCP-2960)

bistline:
If an error occurs when pushing a file to a study bucket, it was possible for a feedback loop to arise that would endlessly create new instances of `UploadCleanupJob`, since that class would also attempt to push the file to the bucket if it couldn't find it.  This update now removes that piece of functionality, and `UploadCleanupJob` will now only attempt to clean up local copies of files on the filesystem once they have successfully pushed to the bucket.  This update also enhances email notifications of failed pushes, as well as failures trying to launch an `IngestJob` (which is where this particular bug first arose).

In addition, this also adds the standalone `DelayedJobAccessor` module for retrieving instances of either `UploadCleanupJob` and `IngestJob` from the `Delayed::Job` queue, matching them on a provided model instance (like `Study` or `StudyFile`).  Finally, this update removes a large number of unnecessary `extend` declarations for `ErrorTracker`.

This PR satisfies SCP-2960.


app/mailers/single_cell_mailer.rb
```diff
@@ -27,11 +27,13 @@ def batch_users_email(email_params, user)
   end
 
   def notify_admin_upload_fail(study_file, error)
-    @users = User.where(admin: true).map(&:email)
+    dev_email_config = AdminConfiguration.find_by(config_type: 'QA Dev Email')
+    emails = dev_email_config.present? ? [dev_email_config.value] : User.where(admin: true).map(&:email)
```
devonbush, line 31: this line could use a comment or break up the ternary into an if/else for clarity

app/models/upload_cleanup_job.rb
```diff
@@ -7,26 +7,27 @@
 
 class UploadCleanupJob < Struct.new(:study, :study_file, :retry_count)
   extend ErrorTracker
+  MAX_RETRIES = 2
 
   def perform
-    retries = retry_count + 1
+    self.retry_count += 1
```
devonbush, line 13: I don't see you initializing this to zero anywhere?   is this some ruby magic where nil gets auto-converted to zero?   That could probably use a comment (even if it is an established rubyism, it throws off folks who spend most of their time in other languages)

devonbush (approved):
nice, like the extra logging and the ability find the job for a given file.  Since we have a lot of file-specific jobs, I could see that being useful in a lot of other places too.

app/models/upload_cleanup_job.rb
```diff
@@ -7,26 +7,27 @@
...
     # make sure file or study isn't queued for deletion first
     if study_file.nil?
       Rails.logger.info "#{Time.zone.now}: aborting UploadCleanupJob due to StudyFile already being deleted."
     elsif study_file.queued_for_deletion || study.queued_for_deletion
-      Rails.logger.info "#{Time.zone.now}: aborting UploadCleanupJob for #{study_file.bucket_location}:#{study_file.id} in '#{study.name}', file queued for deletion"
+      Rails.logger.info "#{Time.zone.now}: aborting UploadCleanupJob for #{study_file.bucket_location}:#{study_file.id} in '#{study.accession}', file queued for deletion"
```
eweitz, line 18: Nice.

app/models/upload_cleanup_job.rb
```diff
@@ -42,26 +43,33 @@ def perform
...
-            Delayed::Job.enqueue(UploadCleanupJob.new(study, study_file, retries), run_at: run_at)
+            # remote file was not found, so schedule a new cleanup job to check again
+            # file may be pushing in another thread, and attempting to push here creates infinite recursion if errors are encountered
+            if self.retry_count <= MAX_RETRIES
+              interval = retry_count * 2
+              run_at = interval.minutes.from_now
+              Rails.logger.info "#{Time.zone.now}: remote file MISSING for #{study_file.bucket_location}:#{study_file.id}, scheduling new UploadCleanupJob for #{run_at}"
+              Delayed::Job.enqueue(UploadCleanupJob.new(study, study_file, self.retry_count), run_at: run_at)
+            else
+              Rails.logger.error "#{Time.zone.now}: file #{study.accession}:#{study_file.bucket_location}:#{study_file.id} has failed to push to #{study.bucket_id}"
```
eweitz, line 54: Prepending all logs with `#{Time.zone.now}: ` is tedious to write, and slightly distracting to read.  Given all logs should have a timestamp, can this be abstracted to a common function?

E.g. in some other .rb file:

```
def log(message, type="info")
  if type == 'info'
    Rails.logger.info "#{Time.zone.now}: #{message}"
  elif type == 'error'
    Rails.logger.error "#{Time.zone.now}: #{message}"
    # Maybe also log to Sentry
  elif ... other log levels
end
```

Then:
```suggestion
              log("remote file MISSING for #{study_file.bucket_location}:#{study_file.id}, scheduling new UploadCleanupJob for #{run_at}")
              Delayed::Job.enqueue(UploadCleanupJob.new(study, study_file, self.retry_count), run_at: run_at)
            else
              log("file #{study.accession}:#{study_file.bucket_location}:#{study_file.id} has failed to push to #{study.bucket_id}", 'error')
```

app/views/single_cell_mailer/notify_admin_parse_launch_fail.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <li>Workspace: <%= link_to @study.workspace_url, @study.workspace_url %></li>
+  <li>Bucket:  <%= link_to @study.google_bucket_url, @study.google_bucket_url %></li>
+  <li>File: <%= @study_file.upload_file_name %></li>
+  <li>File Type: <%= @study_file.file_type %></li>
+  <li>Uploaded/synced on: <%= @study_file.created_at.to_s(:db) %></li>
+  <li>File still local? <%= @study_file.is_local? %></li>
+</ul>
+
+<p>
+  Once the error has been resolved, please relaunch the file from the Rails console with the following commands:
```
eweitz, line 19: This will help -- thanks in advance!

eweitz (approved):
Looks good!  I suggested a few maintainability improvements, but see no blockers.

app/mailers/single_cell_mailer.rb
bistline: Gotcha

app/models/upload_cleanup_job.rb
bistline: It's set to 0 when the job is queued - see [here](https://github.com/broadinstitute/single_cell_portal_core/blob/jb-upload-cleanup-limit/app/models/study.rb#L1503), for instance.  But I agree we could add a `retry_count ||= 0` at the top to ensure that.

app/models/upload_cleanup_job.rb
bistline: Good catch!  I've been slowly removing all those duplicate timestamps when I find them because they're automatically prepended in deployed environments like `staging` and `production`.  I'll remove these as well.

bistline:
> nice, like the extra logging and the ability find the job for a given file. Since we have a lot of file-specific jobs, I could see that being useful in a lot of other places too.

I'm actually going to pursue this some as I'm chasing down a regression, and having this generalized will actually help.

devonbush (approved):
looks good to me pending resolving those test failures

bistline:
Issues with failed builds appear to be similar to the `No such file or directory: getcwd` errors we have seen in the past.  Running a build with only the failing tests enabled [pass](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/212027101), so the issue appears to be related to the Travis environment, or resource limitations.

devonbush (approved):

PR #850, opened 2020-12-18
Addressing raw counts UI papercuts (SCP-2963)

bistline:
This update addresses several small UI and form validation issues associated with the upload wizard:

- Enforces validation of all `expression_file_info` fields, except `units` if matrix is processed
- Adjusts layout of dropdown menus for better utilization of space
- Adds "Required *" asterisks to all required fields for expression files
- Changes content of "initialized" tooltip to reflect what is actually being checked (parsed data, not uploaded files)

This PR satisfies SCP-2963.

### DIFF
--- a/app/views/studies/_expression_file_info_fields.html.erb
+++ b/app/views/studies/_expression_file_info_fields.html.erb
@@ -1,32 +1,32 @@
 <div class="row expression-file-info-fields">
-  <div class="col-sm-2">
+  <div class="col-lg-2">
     <%= f.label :is_raw_counts, 'Is this a raw counts matrix?' %><br />
     <%= f.radio_button :is_raw_counts, 1, checked: f.object.is_raw_counts, class: 'raw-counts-radio'  %>
     <%= f.label :is_raw_counts, 'Yes', class: 'radio-pad' %>
     <%= f.radio_button :is_raw_counts, 0, checked: !f.object.is_raw_counts, class: 'raw-counts-radio' %>
     <%= f.label :is_raw_counts, 'No', class: 'radio-pad'  %>
   </div>
-  <div class="col-sm-2">
+  <div class="col-lg-2">
     <%= f.label :units %><br />
     <%= f.select :units, options_for_select(ExpressionFileInfo::UNITS_VALUES, f.object.units),
                  {include_blank: true},
                  class: 'form-control raw-counts-select counts-unit-dropdown', disabled: !f.object.is_raw_counts %>
   </div>
-  <div class="col-sm-3">
-    <%= f.label :biosample_input_type, "Biosample Input Type <i class='text-danger'>*</i>".html_safe %><br />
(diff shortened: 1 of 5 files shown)


eweitz (approved):

app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -1,32 +1,32 @@
...
                  {include_blank: true},
                  class: 'form-control raw-counts-select counts-unit-dropdown', disabled: !f.object.is_raw_counts %>
   </div>
-  <div class="col-sm-3">
-    <%= f.label :biosample_input_type, "Biosample Input Type <i class='text-danger'>*</i>".html_safe %><br />
+  <div class="col-lg-2">
+    <%= f.label :biosample_input_type, label_with_asterisk('Biosample Input Type') %><br />
     <%= f.select :biosample_input_type, options_for_select(ExpressionFileInfo::BIOSAMPLE_INPUT_TYPE_VALUES, f.object.biosample_input_type),
-                 {prompt: 'Select one...'}, class: 'form-control' %>
+                 {}, class: 'form-control' %>
```
jlchang, line 18: Curious why we removed the 'Select one...' prompt here but leave it in place for Species. The prompt seems useful to indicate selecting before file upload.

app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -1,32 +1,32 @@
...
 
   </div>
-  <div class="col-sm-3">
-    <%= f.label :library_preparation_protocol, "Library Preparation Protocol <i class='text-danger'>*</i>".html_safe %><br />
+  <div class="col-lg-2">
+    <%= f.label :library_preparation_protocol, label_with_asterisk('Library Preparation Protocol') %><br />
     <%= f.select :library_preparation_protocol,
                  options_for_select(ExpressionFileInfo::LIBRARY_PREPARATION_VALUES, f.object.library_preparation_protocol),
-                 {prompt: 'Select one...'}, class: 'form-control' %>
+                 {}, class: 'form-control' %>
```
jlchang, line 25: Same question here about the 'Select one' prompt...

jlchang (approved):
I thought the 'Select one...' prompts were helpful to guide study owners to supply those info. Just a non-blocking comment.

app/views/studies/_expression_file_info_fields.html.erb
bistline: The issue with that is that the `Select one...` prompt actually is a blank value in the dropdown menu.  If the user does not select a value and begins an upload, it will delete the file as it fails validation.  That seems like a bad UX as they may have entered a lot of other information.  We can do client-side validation similar to how we validate `Species` values, but that's more work.  Happy to do that though if we feel that UX is preferable.

devonbush (approved):

PR #852, opened 2021-01-06
Ensure deletion of study files regardless of state (SCP-2992)

bistline:
A recent update to the codebase changed validations for `ExpressionFileInfo` documents, which are embedded in `StudyFile` documents for expression matrix files.  This caused an unintended issue with `DeleteQueueJob` which ended up preventing the deletion of expression matrices that had been added before the validation change was made, as the embedded documents were now invalid, which prevented saving the parent document.  This then broke the normal garbage collection process in the portal for these newly invalid documents, which resulted in the UX of files simply not deleting, without any error being presented to the user.

This update now unsets all embedded documents before attempting to queue a file for deletion.  Furthermore, it has been written in a manner that will be forward-compatible: any new embedded document types that are added in the future will not need to have new cleanup code written.  Each embedded document will now be unset prior to deleting the parent document, which sidesteps the validation issue entirely.

This PR satisfies SCP-2992.

### DIFF
--- a/app/models/delete_queue_job.rb
+++ b/app/models/delete_queue_job.rb
@@ -90,10 +90,6 @@ class DeleteQueueJob < Struct.new(:object)
         nil
       end
 
-      # remove any embedded documents from study_file before queueing for deletion as validation errors can occur
-      unset_embedded_documents(object)
-      object.reload
-
       # if this is a parent bundled file, delete all other associated files and bundle
       if object.is_bundle_parent?
         object.bundled_files.each do |file|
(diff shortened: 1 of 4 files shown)


app/models/delete_queue_job.rb
```diff
@@ -90,6 +90,10 @@ def perform
         nil
       end
 
+      # remove any embedded documents from study_file before queueing for deletion as validation errors can occur
+      unset_embedded_documents(object)
+      object.reload
+
```
devonbush, line 96: Rather than hunt down possible validation errors and remove them in advance (and there might be other ways validation errors can sneak into our files besides embedded docs), I think it will be cleaner to change lines 104 to use a pattern like 

```
object.assign_attributes({ queued_for_deletion: true, ... })
 object.save!(validate: false)
```

Although ordinarily we should never skip validations, here that's expressly what we want "I don't care if this document is valid or not, since I'm deleting it anyway"

devonbush (changes requested):
Great tests!

eweitz (approved):

app/models/delete_queue_job.rb
bistline: That's a fair point.

bistline: 8c390fb

devonbush (approved):

PR #854, opened 2021-01-07
Release 0.67.0

bistline:


eweitz (approved):

devonbush:
ugh, the 'getcwd' problem returns.

devonbush (approved):

bistline:
> ugh, the 'getcwd' problem returns.

Yes, but with the updates I made to `delayed_job_config.rb`, we're actually getting a stack trace now.  It has something to do with `ExpressionFileInfo` documents and form rendering - possible a race condition in initializing objects.  So we possibly have an avenue to pursue.

PR #856, opened 2021-01-08
Fixing server-side TCell integration (SCP-3017)

bistline:
Client-side integration for TCell has been working, but server-side traffic is being dropped due to authentication issues.  This update fixes auth issues by using a separate API key for server-side traffic (using the client-side key responds `401` or `403`, depending on the environment).  This also changes the configuration method for TCell to an initializer, which dynamically configures TCell on boot if the agent is installed and necessary ids/keys are configured.  This is more secure as our keys no longer exist inside the project directory as files.  All necessary boot/config scripts have been updated to reflect these changes, including information about TCell in the README.

All associated vault secrets have been updated with the new keys, including developer environments.

This PR also restores changes from #820 that were lost due to accidental merge conflict resolution from `raw-counts-ingest` being merged back into `development`.

This PR satisfies SCP-3017.


eweitz (approved):
Looks good.

devonbush (approved):

PR #859, opened 2021-01-11
Resize heatmap upon "View Options" click, modularize JS (SCP-2968)

eweitz:
This fixes a UI layout bug in our use of the Morpheus.js heatmap widget.  Previously, upon clicking "View Options", the heatmap would not resize, and overflow into "View Options".  Now, it resizes like other widgets.

Also, heatmap JavaScript has been modularized, like dot plots and other widgets in the Explore tab.

To test:
* Go to a study with visualizations fully initialized, e.g. [here on production](https://singlecell.broadinstitute.org/single_cell/study/SCP1) or [here on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP197).
* In "Explore" tab, search two genes, e.g. "Sergef Setd6"
* Click "Heatmap" tab
* Click "View Options"
* Verify that heatmap resizes

See below for demos of the bug and the fix.

https://user-images.githubusercontent.com/1334561/104224175-99b73f80-5412-11eb-8121-2074967035e6.mov

https://user-images.githubusercontent.com/1334561/104224198-a340a780-5412-11eb-8ae0-e679e8f2d4dd.mov

This satisfies SCP-2968.

### DIFF
--- a/app/assets/javascripts/scp-heatmap.js
+++ b/app/assets/javascripts/scp-heatmap.js
@@ -1,0 +1,75 @@
+/**
+ * Render Morpheus heatmap
+ */
+function renderMorpheus(
+  dataPath, annotPath, selectedAnnot, selectedAnnotType, target, annotations,
+  fitType, heatmapHeight, colorScaleMode
+) {
+  console.log(
+    `render status of ${target} at start: ${$(target).data('rendered')}`
+  )
+  $(target).empty()
+  console.log(`scaling mode: ${colorScaleMode}`)
+  const config = {
+    name: 'Heatmap',
+    dataset: dataPath,
+    el: $(target),
+    menu: null,
+    colorScheme: { scalingMode: colorScaleMode }
+  }
+
+  // set height if specified, otherwise use default setting of 500 px
+  if (heatmapHeight !== undefined) {
+    config.height = heatmapHeight
+  } else {
+    config.height = 500
+  }
+
+  // fit rows, columns, or both to screen
+  if (fitType === 'cols') {
+    config.columnSize = 'fit'
+  } else if (fitType === 'rows') {
+    config.rowSize = 'fit'
+  } else if (fitType === 'both') {
+    config.columnSize = 'fit'
+    config.rowSize = 'fit'
+  } else {
+    config.columnSize = null
+    config.rowSize = null
+  }
+
+  // load annotations if specified
+  if (annotPath !== '') {
+    config.columnAnnotations = [{
+      file: annotPath,
+      datasetField: 'id',
+      fileField: 'NAME',
+      include: [selectedAnnot]
+    }]
+    config.columnSortBy = [
+      { field: selectedAnnot, order: 0 }
+    ]
+    config.columns = [
+      { field: 'id', display: 'text' },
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/scp-heatmap.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * Render Morpheus heatmap
+ */
+function renderMorpheus(
```
bistline, line 4: Always love to see refactoring like this.

app/views/site/_view_gene_expression_heatmap.html.erb
```diff
@@ -60,6 +60,9 @@
     $('#dot-plot').data('rendered', false);
     function drawHeatmap(height) {
         $(window).off('resizeEnd');
+
+        $(window).on('resizeEnd', () => { $(window).trigger('resize.morpheus')})
```
bistline, line 64: Does the `resize.morpheus` event have anything special associated with it?  It doesn't appear that when testing locally there is a functional difference between `resize` and `resize.morpheus`.

bistline (approved):

app/views/site/_view_gene_expression_heatmap.html.erb
eweitz: Yup, `resize.morpheus` triggers only resizing the heatmap.  Implementation details are [here](https://github.com/cmap/morpheus.js/blob/a4055b4bb1c1e7907295a067275d6dca4b8137cd/src/ui/heat_map.js#L516).

I suppose my naive concern about `resizeEnd`-trigger-`resize` causing an infinitely loop is unfounded given your test.  That said, I lean towards `resize.morpheus` because that's consistent with how other widgets like scatter plots and violin plots handle `resizeEnd`.  They do so via widget-specific resize functions.

devonbush (approved):
Looks good, and works locally.  Some questions about code style, but don't have to be resolved in this PR

bistline:
> also, what's our policy on console.log statements in production code? I'd just as soon have a policy that we don't leave console.log statements in production code. When I need to debug prod javascript, I find it's often easier without extra log statements cluttering the console.

@devonbush I don't think that we have an explicit policy for `console.log` in production.  My personal take would be "less is more" - to only log things that are significant, and would be useful to both a user and a developer when investigating the DOM.

app/assets/javascripts/scp-heatmap.js
eweitz: Regarding function name: I agree `renderMorpheus` isn't ideal, especially now that we have another Morpheus graphic in dot plots.  I'm thinking `renderHeatmap`, consistent with how we omit vendor from other new visualization rendering function names (e.g. `renderScatterPlot` instead of `renderPlotlyScatterPlot`).

Regarding `console.log`: I also lean towards omitting them all.  Given the different perspectives, I'll keep this as is for now.  I've added it as a topic for this Thursday's [engineering strategy](https://docs.google.com/document/d/11eNlwMjwKoUcx4TQflzxw4DuN9q03fNNLphUuAVNIc4/edit#heading=h.4zti8yqhognv) meeting, so we can hopefully reach clear consensus.

eweitz: e0bf891 refines function names.

PR #861, opened 2021-01-11
global gene search collapse genes to violin (SCP-2934, SCP-2933)

devonbush:
image

To test:
1. Populate synthetic study 'human_lymph_node' (and ensure `rails jobs:work` is running)
2. do global gene search for apoe and agpat2
3. Confirm dotplot appears, along with 'Collapse by genes' dropdown.
4. click label or (?) icon to confirm tooltip
5. select 'mean'
6. Observe the graph change to violin plot, with data reflecting cell averages.  Label under graph should now say 'mean expression of agpat2, apoe'
7. Do search for just apoe
8. confirm 'Collapse genes by' dropdown does not appear

We have an issue where the 'collapse_by' variable is called a bunch of different things across our codebase. 'consensus' 'collapse' 'row_centered' (the latter is from morpheus), etc...  When we fix the existing study explore collapse function, we should standardize as best we can.  (I'd lean towards 'collapse_by' because that's what's in the UI, but don't feel strongly)


bistline (approved):
Looks good, although it does not appear that we have a test for the collapse function.  It would be good to add that.

app/javascript/components/visualization/ClusterControls.js
```diff
@@ -141,7 +157,7 @@ export default function ClusterControls({studyAccession, onChange, showSubsample
           styles={ customSelectStyle }/>
       </div>
       <div className="form-group">
-        <label>Subsampling threshold</label>
+        <label>Subsampling</label>
```
eweitz, line 160: Nice refinement

eweitz (changes requested):
Looks good overall -- kudos on the swift work.

Two blockers for trivial user-facing refinements: see comments beginning "Conventionally" and "Let's use".  The other suggestions concern maintainability nits.

Also, I can't tell at a glance whether the gene names shown in the UI use the casing from the study data or the user search, but it'd be best to use the former.  (That's not necessarily in scope, so I don't consider it blocking.)

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -67,12 +72,19 @@ export default function StudyViolinPlot({ study, gene }) {
             className="gene-load-spinner"
           />
         }
+        { isCollapsedView && studyGeneNames.length &&
+          <div className="text-center">
+            <span>{collapseBy} expression of {studyGeneNames.join(', ')}</span>
```
eweitz, line 77: Let's use [`capitalize`](https://esamattis.github.io/underscore.string/#capitalize-string-lowercaserest-false-gt-string) and [`toSentenceSerial`](https://esamattis.github.io/underscore.string/#tosentenceserial-array-delimiter-lastdelimiter-gt-string) from the `underscore.string` package:

```suggestion
            <span>{capitalize(collapseBy)} expression of {toSentenceSerial(studyGeneNames)}</span>
```

eweitz: It'd be nice polish, but I don't feel strongly about it.

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -27,29 +30,31 @@ export default function StudyViolinPlot({ study, gene }) {
...
   useEffect(() => {
-    // do a load from the server if this is the initial load
-    if (!isLoading && !isLoaded) {
+    if (!isLoading) {
       loadData()
     }
-  }, [study.accession, gene])
-
+  }, [study.accession, genes[0], collapseBy])
+  const isCollapsedView = ['mean', 'median'].indexOf(collapseBy) >= 0
```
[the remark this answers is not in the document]
eweitz: We have many pre-existing IE incompatibilities: [arrow functions](https://caniuse.com/?search=arrow), [`fetch`](https://caniuse.com/?search=fetch), and likely a dozen or so more.  

In other words, this would not make us become incompatibility with IE.  That happened long ago!

eweitz (approved):
A few nits remain, but none block!

eweitz (changes requested):
I found a bug when testing locally.  (I count this as a success of the new testing strategy we discussed last week!)

To reproduce:
* Do global gene search for apoe and agpat2
* Select Mean in "Collapse genes by"
* Select cluster2.tsv in "Load cluster"
* Select Median in "Collapse genes by"
* Note cluster.tsv unexpectedly replaces cluster2.tsv in "Load cluster"

A similar bug occurs with "Select annotation".

eweitz (approved):

PR #862, opened 2021-01-12
Adding cleanup tools for tests, local development (SCP-3018)

bistline:
Test failures during CI can at times leave behind workspaces in the `single-cell-portal-test` (or developer-owned, depending on where the tests are run from) Terra billing project that are "orphaned", meaning they no longer have a `Study` object associated with them in any schema.  Cleaning up these workspaces previously had been a purely manual process, either through the Terra UI, or from the Rails console in whichever instance owns the workspaces in question.

This update adds `StudyCleanupTools`, a module that contains methods for deleting studies and workspaces as part of a test run:

1. `StudyCleanupTools.destroy_all_studies_and_workspaces` will delete all studies & workspaces within a given instance of the portal
2. `StudyCleanupTools.delete_all_orphaned_workspaces` will delete all workspaces within a given Terra billing project that do not have a study associated with them in the given instance.  It will ignore any workspaces not created by the service account configured on that instance, as these belong to other users or instances of SCP.

Since both of these methods can potentially be highly destructive if misused, they have multiple safeguards in place to prevent a developer from accidentally invoking them and destroying user data:

1. `StudyCleanupTools.permit_hostname?` will prevent any cleanup methods from being invoked on deployed hosts for staging (**singlecell-staging.broadinstitute.org**) and production (**singlecell.broadinstitute.org**)
2. `StudyCleanupTools.permit_billing_project?` will prevent any cleanup methods from being invoked on Terra billing projects for staging (**single-cell-portal-staging**) and production (**single-cell-portal**)
3. `StudyCleanupTools.permit_environment?` will prevent any cleanup methods from being run in the staging or production schemas
4. `StudyCleanupTools.is_continuous_integration?` will prevent any cleanup methods from firing unless this is a CI run in Travis (governed by the presence of the environment variable `CI` being set to `true`).  This validation only applies to `StudyCleanupTools.delete_all_orphaned_workspaces`.
5. `Study.service_account_created_workspace?` will prevent deleting any Terra workspaces not created by the same service account.  This validation only applies to `StudyCleanupTools.delete_all_orphaned_workspaces`.
6. `StudyCleanupTools.confirm_delete_request?` will prompt a developer running any cleanup commands in the `development` environment to confirm they wish to proceed by typing `Delete All` in the console prompt.

Both cleanup methods can be run in local development, but the developer must manually override the environment validation by passing `allow_dev_env: true` when invoking either method.  In addition, `StudyCleanupTools.delete_all_orphaned_workspaces` will throw an error if this is not a CI run in Travis.  To override this and run in a local instance, a developer must set the value of `CI` before entering the console like this:

```
CI=true bin/rails c
```

Lastly, the developer will be prompted to confirm any cleanup command by typing `Delete All` in the confirmation prompt. 
 This three-pronged safety check for local development will prevent developers from accidentally deleting all of their studies or workspaces for their local instance.  In addition, it will allow a developer to run unit & integration tests locally without worrying about workspaces from their `development` schema being accidentally deleted by these cleanup tools.

Lastly, this update changes all `Rails.env` checks to leverage `ActiveSupport::StringInquirer` equality checks, which simplifies checking which environment a method is being run in.  Instead of checking `Rails.env == 'test'`, we can simply check `Rails.env.test?`

This PR satisfies SCP-3018.


lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,159 @@
...
+  #   - +project_name+: (String) => Name of Terra billing project (defaults to FireCloudClient::PORTAL_NAMESPACE )
+  #   - +allow_dev_env+: (Boolean) => allow execution of this method in development environment (default: false)
+  #
+  # * *raises*
+  #   - (ArgumentError) => if command cannot be run for failing a security check (hostname, project, environment, etc)
+  #
+  # * *returns*
+  #   - (TrueClass) => true if method was performed, otherwise ArgumentError is thrown
+  def self.delete_all_orphaned_workspaces(project_name = FireCloudClient::PORTAL_NAMESPACE, allow_dev_env: false)
+    if validate_hostname! && validate_environment!(allow_dev_env) && validate_billing_project!(project_name) && validate_continuous_integration!
```
eweitz, line 50: I appreciate all of these guards.

eweitz (approved):

lib/study_cleanup_tools.rb
bistline: I tried to make this as "safe" as possible.  I would be interested to hear if you can think of any other safeguards we can put in place for production/staging data, that would still enable a developer to purge their local environment/billing project if need be.

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,159 @@
...
+  # can be run manually in development if allow_dev_env is set to true
+  #
+  # * *raises*
+  #   - (ArgumentError) => if environment is not test/pentest, or is development and allow_dev_env = false
+  #
+  # * *returns*
+  #   - (TrueClass) => true if environment passes validation
+  def self.validate_environment!(allow_dev_env = false)
+    environments = %w(test pentest)
+    allow_dev_env ? environments << 'development' : nil
```
devonbush, line 124: A plain `if` would be easier to read here.   I'm not a fan of using ternaries for anything other than assignment.

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,159 @@
...
+    end
+  end
+
+  # delete all Terra workspaces in a project that do not have a study associated with them in the given env/schema
+  # will ignore any workspaces created by different service accounts/users
+  # by default will only run against test/pentest environment during continuous integration build
+  #
+  # * *params*
+  #   - +project_name+: (String) => Name of Terra billing project (defaults to FireCloudClient::PORTAL_NAMESPACE )
+  #   - +allow_dev_env+: (Boolean) => allow execution of this method in development environment (default: false)
```
devonbush, line 42: thanks!  I'm sure this dev environment mode will come in handy for me

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,159 @@
...
+      raise ArgumentError.new("This is not a continuous integration test run, ENV['CI'] is not true")
+    else
+      true
+    end
+  end
+
+  private
+
+  # is requested project permitted to run a command
+  def self.permit_billing_project?(project_name)
```
devonbush, line 151: this seems like it should just be inlined in validate_billing_project.  `validate_billing_project` has no other logic, so I don't think you're gaining testability

devonbush (changes requested):
Nice!  Love that you pulled this out of study.rb into its own class.  A few style nits.  The ones I feel the most strongly about are cleaning up/out the ternaries--this is code that I want to be as vanilla-readable as possible so nothing gets confused.

lib/study_cleanup_tools.rb
bistline: I've actually seen a lot of different explanations for "bang methods" and what they actually mean/do.  While generally it denotes that you're modifying an object in place, I've also read that it simply denotes that the method is "dangerous".  I'm using it here because rather than returning `false` for a validation, it's actually throwing an error.  I did this because I wanted to be _absolutely sure_ that execution is being halted if something doesn't pass one of these checks.  I added the `return true` statements later because I wanted to actually be able to test these, and having a return value made that possible.

On reflection, you're right - it's mixing metaphors, and is generally confusing.  A method called `validate_[something]` in ActiveRecord or another ORM will return true/false rather than explicitly raising an exception.  I think it would probably make a lot more sense for the parent method to raise an exception rather than the validator.  It just seemed DRYer to have the exceptions in the validations.

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,159 @@
...
+    !DISALLOWED_BILLING_PROJECTS.include?(project_name) && project_name == FireCloudClient::PORTAL_NAMESPACE ? true : false
```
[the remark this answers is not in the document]
bistline: ooh I love learning little Rubyisms like this...

lib/study_cleanup_tools.rb
bistline: Fair point.

lib/study_cleanup_tools.rb
bistline: It's actually used in two places - one in the validation as the check, and also in `destroy_all_studies_and_workspaces` as a final check on whether or not to delete a study.  But given the other refactoring, I think this will all get cleaned up.

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,174 @@
...
+  ##
+
+  # handler to run a security check and raise an error on returning false to halt execution
+  #
+  # * *params*
+  #   - +validation_name+: (String, Symbol) => name of security check to perform
+  #   - +args+: (*Array) => array of arguments for calling method, passed with splat (*)
+  #
+  # * *raises*
+  #   - (RuntimeError) => raise exception if security check returns false to halt parent method execution
```
devonbush, line 100: I like what you're doing in centralizing the error handling code and making the check methods pure boolean logic.  I do think the structure of passing validation methods around as symbols and arglists makes the code hard to read/trace, and is generally something that should only be done if you have to.

How about making this a method that takes a block instead?  So the calling code would be
```halt_on_validation_fail { permit_hostname? }```

and this method's implementation would essentially be
```
  result = yield
  if result != true
     raise RuntimeError
  end
```

Also, I'd rename this to ```raise_exception_unless_true``` to be more explicit about its operation

File: test/lib/study_cleanup_tools_test.rb [hunk 7]
```diff
@@ -0,0 +1,105 @@
...
+    @user = FactoryBot.create(:user, test_array: @@users_to_clean)
+    @basic_study = FactoryBot.create(:study, name_prefix: 'Cleanup Security Checks', test_array: @@studies_to_clean)
+  end
+
+  teardown do
+    Rails.env = "test" # reset this value to prevent other env checks failing later as this will persist across all runs
+  end
+
+  test 'should validate hostname' do
+    assert StudyCleanupTools.permit_hostname?
```
devonbush, line 21: it would be really nice to have a negative test here.  Can you stub the Socket.gethostname method?

test/lib/study_cleanup_tools_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    Rails.env = 'development'
+    assert StudyCleanupTools.permit_environment?(allow_dev_env: true)
+
+    # ensure production-like environments are disallowed
+    Rails.env = 'staging'
+    refute StudyCleanupTools.permit_environment?
+  end
+
+  test 'should validate continuous integration' do
+    assert StudyCleanupTools.is_continuous_integration?
```
devonbush, line 45: this test will fail in local environments if this suite is run on its own, which seems suboptimal.  Is there a way to stub environment variables so that this test can pass when run individually?https://stackoverflow.com/questions/9611276/what-is-the-best-way-to-write-specs-for-code-that-depends-on-environment-variabl

test/lib/study_cleanup_tools_test.rb
```diff
@@ -0,0 +1,105 @@
...
+  test 'should validate environment' do
+    assert StudyCleanupTools.permit_environment?
+
+    # test allowing development environment
+    Rails.env = 'development'
+    assert StudyCleanupTools.permit_environment?(allow_dev_env: true)
+
+    # ensure production-like environments are disallowed
+    Rails.env = 'staging'
+    refute StudyCleanupTools.permit_environment?
```
devonbush, line 41: does this test need to reset the env at the end?  I'd also rather stubbing was used so that the environment isn't left in a weird state if the test fails midway through

test/lib/study_cleanup_tools_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    # negative test with mock workspace JSON
+    not_owned_by_portal = {workspace: {createdBy: @user.email}}
+    refute StudyCleanupTools.service_account_created_workspace?(not_owned_by_portal)
+  end
+
+  test 'should raise error on validation failure' do
+
+    # positive tests - known goods should return nil as there is no declared return value
+    # no arguments
```
devonbush, line 61: Yeah, rewriting halt_on_validation_fail to take a block will let you test it much more independently, because you can just test 
```halt_on_validation_fail { false }``` ```halt_on_validation_fail { true }```
```halt_on_validation_fail { nil }```
etc...

File: test/test_helper.rb [hunk 11]
```diff
@@ -74,7 +74,14 @@ module SelfCleaningSuite
   after(:all) do
     puts "#{self.class}: Cleaning up #{@@studies_to_clean.count} studies, #{@@users_to_clean.count} users"
     [@@studies_to_clean, @@users_to_clean].each do |entity_list|
```
devonbush, line 76: nice

devonbush (changes requested):
I like the centralization of error handling that you did.  I think the code structure can be refined

lib/study_cleanup_tools.rb
bistline: That's a great suggestion - I like the idea of invoking the method directly rather than calling `self.send(method, *args).  It definitely is much more explicit.

test/lib/study_cleanup_tools_test.rb
bistline: Great point.

test/lib/study_cleanup_tools_test.rb
bistline: Yeah, I had struggled with that.  It definitely works in CI, but locally it will fail unless you run the test with `CI=true bin/rake`.

test/lib/study_cleanup_tools_test.rb
bistline: the `teardown` declaration resets it.  Since `permit_environment?` calls `Rails.env` directly, I'm not sure this can be stubbed, but I'll take a crack at it.

test/lib/study_cleanup_tools_test.rb
bistline: 32c1bc3

test/lib/study_cleanup_tools_test.rb
bistline: 32c1bc3

test/lib/study_cleanup_tools_test.rb
bistline: 32c1bc3

test/lib/study_cleanup_tools_test.rb
bistline: 32c1bc3

lib/study_cleanup_tools.rb
bistline: 32c1bc3

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,172 @@
...
+  #   - +validation_name+: (String) => name of validation being run (for error message)
+  #   - +&block+: (Proc) => validation method to execute that returns true/false
+  #
+  # * *raises*
+  #   - (RuntimeError) => raise exception if security check returns false to halt parent method execution
+  def self.raise_exception_unless_true(validation_name, &block)
+    valid_command = yield
+    if !valid_command
+      # report validation that failed and method caller
+      error_message = "halting execution as #{validation_name} failed validation when calling #{self.name}##{caller_locations.first.label}"
```
devonbush, line 105: this is firmly in the realm of 'nit'.  But you're not actually halting execution here--you're just throwing the exception. (which might be caught further upstream).   So I'd remove 'halting execution' from the error_message.

lib/study_cleanup_tools.rb
```diff
@@ -0,0 +1,172 @@
...
+  #   - +project_name+: (String) => Name of Terra billing project (defaults to FireCloudClient::PORTAL_NAMESPACE )
+  #   - +allow_dev_env+: (Boolean) => allow execution of this method in development environment (default: false)
+  #
+  # * *raises*
+  #   - (RuntimeError) => if command cannot be run for failing a security check (hostname, project, environment, etc)
+  #
+  # * *returns*
+  #   - (Integer) => count of workspaces that were deleted
+  def self.delete_all_orphaned_workspaces(project_name = FireCloudClient::PORTAL_NAMESPACE, allow_dev_env: false)
+    raise_exception_unless_true(:permit_environment?) { permit_environment?(allow_dev_env) }
```
devonbush, line 55: I'd just as soon have these be passed as strings rather than symbols, so it's clearer that they're just used for messaging purposes

devonbush (approved):
looks great -- two nits that I don't feel strongly about so go with what looks right to you.

lib/study_cleanup_tools.rb
bistline: Fair enough.

PR #863, opened 2021-01-14
fixing mean/median consensus view for multi-gene search (SCP-3015, SCP-2988)

devonbush:
In addition to fixing the bug, I did *not* update the dropdown to match the language we agreed on for the global gene search collapse control -- the new labels don't make as much sense in a control that's being shown alongside both single and multi-gene search.  And implementing show/hide for that control is outside the scope of this fix

image

To test:
1. open study explore for a study
2. search for two or more genes
3. then use the hamburger menu to the left of the gene search box to select 'violin-mean' or 'violin-median'
4. Observe that a consensus violin plot appears
5. Select 'dot plot' and observe the display reverts to the previous, uncollapsed view

PR #864, opened 2021-01-15
fixing menu resize bug (SCP-3015)

devonbush:
Fixes bug where opening the 'view options' tab after doing a 'collapse my mean' would not resize the graph.  

To test:
1. Go to explore tab
2. do a search for 2+ genes
3. select "collapse by" -> "mean"
4. after graph renders, select "view options".  Observe that the graphs resize to make room for the menu

### DIFF
--- a/app/javascript/lib/study-overview/explore-single.js
+++ b/app/javascript/lib/study-overview/explore-single.js
@@ -59,7 +59,10 @@ async function renderSingleGenePlots(study, gene) {
 
   const geneList = window.SCP.formatTerms(gene)
   if (geneList.length > 1) {
+    // draw the dotPlot for comparison if this is a multi-gene collapsed by mean
     window.drawHeatmap()
+    // reattach the event handlers, since that call removes them
+    attachEventHandlers(study, gene)
   } else if (geneList.length === 1) {
     // only show ideogram for single gene queries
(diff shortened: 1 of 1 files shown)


app/javascript/lib/study-overview/explore-single.js
```diff
@@ -59,7 +59,10 @@ async function renderSingleGenePlots(study, gene) {
 
   const geneList = window.SCP.formatTerms(gene)
   if (geneList.length > 1) {
+    // draw the dotPlot for comparison if this is a multi-gene collapsed by mean
     window.drawHeatmap()
+    // reattach the event handlers, since that call removes them
+    attachEventHandlers(study, gene)
```
eweitz, line 65: Thinking-out-load: I created `attachEventHandlers` as a way to encapsulate the old practice of attaching-and-reattaching event handlers.  But the underlying practice still seems a bit tedious and error-prone.

I suspect migrating these components to React will mitigate such footguns.

eweitz (approved):
Looks good -- I confirmed in my local SCP that this fixes [the bug](https://github.com/broadinstitute/single_cell_portal_core/pull/863#pullrequestreview-568661529).

bistline (approved):

PR #866, opened 2021-01-19
Allowing export of raw study analytics data (SCP-2946)

devonbush:
Goal is to export a plain tsv of useful study info (available only to admins), so that we can can slice and dice the numbers in our tool of choice.  We could also eventually build a nice interactive reports page on the site, but that's a task for another day.  This adds a link on the report page to download a tsv 

image

Notably, this tries a new pattern for hybrid API/site cookie validation.  This allows a single controller to support both plain links on the site from logged-in users, as well as bearer token auth.  I think this is actually a superior pattern to the URL_SAFE_TOKEN_ALLOWED_ACTIONS pattern we're using to allow embedded morpheus widgets to hit our API.  So if you all approve, after this is merged, I will convert those, and simplify authenticator.rb accordingly.  We should really only have to use totat-like constructs for users who aren't on the site.

To test:
1. log in as admin user
2. go to 'reports'
3. click 'raw study data tsv'
4. observe the downloaded tsv file, and that it matches the studies in the DB

### DIFF
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -3,6 +3,9 @@ module Api
     module Concerns
       module Authenticator
         extend ActiveSupport::Concern
+        include ActionController::Cookies
+        include ActionController::RequestForgeryProtection
+
         OAUTH_V3_TOKEN_INFO = 'https://www.googleapis.com/oauth2/v3/tokeninfo'
         TOTAT_REQUIRED_ACTIONS = [
           {controller: 'search', action: 'bulk_download'},
@@ -11,6 +14,12 @@ module Api
         URL_SAFE_TOKEN_ALLOWED_ACTIONS = [
           {controller: 'expression', action: 'show'},
           {controller: 'annotations', action: 'cell_values'}
+        ]
+        # these are API actions that we allow logged-in site users to access using
+        # regular rails session validation & csrf protection, which can be useful
+        # to save from having to put a token in a URL
+        COOKIE_ALLOWED_ACTIONS = [
+          {controller: 'reports', action: 'show'}
         ]
         def authenticate_api_user!
           head 401 unless api_user_signed_in?
@@ -34,6 +43,11 @@ module Api
           user = nil
           if TOTAT_REQUIRED_ACTIONS.include?({controller: controller_name, action: action_name})
             return find_user_from_totat
+          elsif COOKIE_ALLOWED_ACTIONS.include?({controller: controller_name, action: action_name}) && current_user.present?
+            # this is a request from a user logged into the website
+            verify_authenticity_token
+            verify_same_origin_request
+            return current_user
           else
             api_access_token = extract_bearer_token(request)
             if api_access_token.present?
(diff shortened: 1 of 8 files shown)


app/views/reports/index.html.erb
```diff
@@ -1,5 +1,10 @@
 <h1>Reports <%= link_to "<i class='fas fa-fw fa-chart-bar'></i> Request a new plot".html_safe, report_request_path, class: 'btn btn-primary pull-right', id: 'report-request' %></h1>
 <h3 class="lead">Updated as of <%= Time.zone.now.strftime("%A, %B %-d %Y %r") %> <%= link_to "<i class='fas fa-toggle-on'></i> Toggle Totals".html_safe, '#/', class: 'btn btn-default pull-right', id: 'toggle-column-annots' %></h3>
+<div class="row">
+  <div class="col-md-12">
+    <%= link_to 'Raw study data tsv', api_v1_report_url('studies') %>
```
eweitz, line 5:
```suggestion
    <%= link_to 'Raw study data TSV', api_v1_report_url('studies') %>
```

eweitz (approved):
Looks good.  I confirmed this works in my local SCP instance.

app/controllers/api/v1/concerns/authenticator.rb
```diff
@@ -34,6 +43,11 @@ def get_current_api_user
           user = nil
           if TOTAT_REQUIRED_ACTIONS.include?({controller: controller_name, action: action_name})
             return find_user_from_totat
+          elsif COOKIE_ALLOWED_ACTIONS.include?({controller: controller_name, action: action_name}) && current_user.present?
+            # this is a request from a user logged into the website
+            verify_authenticity_token
```
bistline, line 48: Nice

app/controllers/api/v1/reports_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    # reports endpoint.  No swagger docs since this is admin-user only
+    # This returns tsv files
+    class ReportsController < ApiBaseController
+      include ActionController::RequestForgeryProtection
+
+      before_action :authenticate_api_user!
+
+      def show
+        if !current_api_user.admin?
+          head 403 and return
```
bistline, line 12: Technically, we don't validate that a user is an admin for the reports section.  We only validate that `current_user.acts_like_reporter?`, which returns `true` for admins or any other user who has `reporter` set to `true`.

app/controllers/api/v1/reports_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+
+        report_name = params[:report_name].to_sym
+        if !ReportsService::REPORTS.keys.include?(report_name)
```
bistline, line 16: Since `ReportsService.get_report_data` will raise an error if the requested report doesn't exist, I'd leave this out and do this instead:

```
begin
  render plain: ReportsService.get_report_data(report_name)
rescue ArgumentError => e
  render json: {error: e.message}, status: 422
end
```

and change `reports_service.rb` line 16 to:

```
if report_obj.nil?
  raise ArgumentError.new("Unrecognized report: '#{report_name}'")
end
```

app/lib/reports_service.rb
```diff
@@ -0,0 +1,93 @@
+class ReportsService
```
bistline, line 1: Really like how this has been abstracted.

bistline (changes requested):
Just a refinement of error handling in the controller.

bistline (approved):

PR #867, opened 2021-01-19
* SCP-2989 Handle no filename being passed in download URL

knapii-developments:
This is a PR to address [sentry error](https://sentry.io/organizations/broad-institute/issues/2031252931/?project=1424198&statsPeriod=14d). The error was caused by a user trying to download a file via URL where the did have provided a filename. 

To test:

- Go to "Download" tab on a public study.
- Copy link address on any study download option
- Post link in browser
- Remove "filename= </file name>" from URL. Ex: https://localhost/single_cell/data/public/SCP1/<public study name>. There's a clearer example of this in the URL in the screenshot.
- Hit enter

Screenshot of the error a user will see when the filename is not included in the URL
Screen Shot 2021-01-19 at 1 18 57 PM


app/controllers/application_controller.rb
```diff
@@ -223,7 +223,7 @@ def execute_file_download(study)
     begin
       # get filesize and make sure the user is under their quota
       requested_file = ApplicationController.firecloud_client.execute_gcloud_method(:get_workspace_file, 0, study.bucket_id, params[:filename])
-      if requested_file.present?
+      if requested_file.present? && !requested_file.size.nil?
```
devonbush, line 226:
```suggestion
      if requested_file.present? && requested_file.size.present?
```
is a little more readable

app/controllers/application_controller.rb
```diff
@@ -243,10 +243,15 @@ def execute_file_download(study)
                       alert: 'We are unable to process your download.  Please try again later.' and return
         end
       else
-        # send notification to the study owner that file is missing (if notifications turned on)
-        SingleCellMailer.user_download_fail_notification(study, params[:filename]).deliver_now
-        redirect_to merge_default_redirect_params(view_study_path(accession: study.accession, study_name: study.url_safe_name), scpbr: params[:scpbr]),
-                    alert: 'The file you requested is currently not available.  Please contact the study owner if you require access to this file.' and return
+        if requested_file.size.nil?
```
devonbush, line 246: why don't we move this check to the beginning of the method and just check `params[:filename].blank?` to be more explicit about what we are testing?

That also fits better into the typical controller pattern of "first validate input, then do stuff, then return"

devonbush (changes requested):
good stuff, just some style fixes that I should have been able to spot during pairing, were I not half-distracted by fussy chunkBears.  I'm glad we're moving pairing to a different time next week

eweitz (approved):
Looks good, confirmed via local manual test.

Cavaet: redirect fails is study name is omitted in URL.  E.g. https://localhost:3000/single_cell/data/public/1/the-study-name works, but https://localhost:3000/single_cell/data/public/SCP1/ fails.  I see no cases like the latter in the linked Sentry report.  So while accounting for such cases would be a nice touch, I think the current update suffices.

app/controllers/application_controller.rb
knapii-developments: > Looks good, confirmed via local manual test.
> 
> Cavaet: redirect fails if study name is omitted in URL. E.g. https://localhost:3000/single_cell/data/public/1/the-study-name works, but https://localhost:3000/single_cell/data/public/SCP1/ fails. I see no cases like the latter in the linked Sentry report. So while accounting for such cases would be a nice touch, I think the current update suffices.

Just to explain the problem:
The sentry error was due to the filename missing in the URL. This [line of code](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L225) passes the file name from the URL. When the file name isn't in the URL, the variable `requested_file` is an empty object. As a result, [file_name](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L227) is nil. When trying to calculate the [user_quota](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L228) this causes an error because you can not add an integer to a nil value. This is why we saw the error in Sentry. I'm not sure what behavior would ensue with the URL you tested against. 


> Looks good, confirmed via local manual test.
> 
> Cavaet: redirect fails if study name is omitted in URL. E.g. https://localhost:3000/single_cell/data/public/1/the-study-name works, but https://localhost:3000/single_cell/data/public/SCP1/ fails. I see no cases like the latter in the linked Sentry report. So while accounting for such cases would be a nice touch, I think the current update suffices.

knapii-developments:
> Looks good, confirmed via local manual test.
> 
> Cavaet: redirect fails if study name is omitted in URL. E.g. https://localhost:3000/single_cell/data/public/1/the-study-name works, but https://localhost:3000/single_cell/data/public/SCP1/ fails. I see no cases like the latter in the linked Sentry report. So while accounting for such cases would be a nice touch, I think the current update suffices.

Just to explain the problem:
The sentry error was due to the filename missing in the URL. This [line of code](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L225) passes the file name from the URL. When the file name isn't in the URL, the variable `requested_file` is an empty object. As a result, [file_name](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L227) is nil. When trying to calculate the [user_quota](https://github.com/broadinstitute/single_cell_portal_core/blob/c593b6bbc781dc582dcf402e88e92df11ca62c7b/app/controllers/application_controller.rb#L228) this causes an error because you can not add an integer to a nil value. This is why we saw the error in Sentry. I'm not sure what behavior would ensue with the URL you tested against.

app/controllers/application_controller.rb
knapii-developments: c2bc214

devonbush (approved):

PR #868, opened 2021-01-19
fixing 'no files available' logic in study download (SCP-3027)

devonbush:
Fixes if/else logic so that 'no downloads are available for you' only shows when appropriate (I suspect the logic got corrupted in a merge conflict resolution).

Adds quote-escaping to manifest download path.

updates filename in page description to match the name of downloaded manifest

TO TEST:
1. Log in as a user that does NOT have direct bucket access to a study (e.g. not the user who created the study)
2. Go to the 'download' tab of a study that user has access to, but not direct bucket access
3. click 'bulk download'
4. Observe that the file list is correct
5. download the files, and confirm success

### DIFF
--- a/app/views/site/_study_download_data.html.erb
+++ b/app/views/site/_study_download_data.html.erb
@@ -31,7 +31,7 @@
                   </thead>
                 </table>
               <% end %>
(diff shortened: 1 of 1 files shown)


bistline (approved):

eweitz (approved):
Looks good, confirmed via local SCP test.

devonbush:
merging -- test failure is unrelated

PR #869, opened 2021-01-20
Stagger ingesting expression matrices of same type (SCP-2971)

bistline:
Uploading multiple expression matrices of the same "type" (e.g. raw/processed) in quick succession will currently cause the initial expression matrix to fail validation due to a constraint in previous releases of `scp-ingest-pipeline` that all matrix files must have cell names associated with them.  The second matrix will then ingest normally as the first has been removed.  This constraint is completely unintuitive to the user, and results in unnecessary retries for files that should ingest without issue.

In conjunction with release [1.9.1](https://github.com/broadinstitute/scp-ingest-pipeline/releases/tag/1.9.1) of `scp-ingest-pipeline`, the constraint has been relaxed and will no longer throw an error.  However, a race condition can still exist where two matrix files uploaded at the same time will validate even if they have repeated cells between them, as the the first matrix will not have finished storing the arrays of cell names before the second matrix starts validating.  To address this potential data integrity issue, ingest jobs for matrix files of the same type in a study are now staggered so that only one can validate at any given time.  Once a matrix has finished validating (either by passing or failing), any additional uploaded matrices that are waiting to ingest will then progress through the queue.  If multiple matrices are waiting to ingest, they will launch in the order in which they were uploaded (first in, first out).

This PR satisfies SCP-2971.


eweitz (approved):
Looks good.

This change is an example of how the [`perfTime` metric for the `ingest` event in Mixpanel](https://mixpanel.com/s/4C9vfz) doesn't tell the whole story of user-perceived performance and wait times.  That's because `ingest` `perfTime` won't be affected due to this change, even though we know it will somewhat slow down end-to-end performance.

That isn't something worth resolving here, but it's perhaps worth considering how to better measure end-to-end performance time of "upload and ingest".  Naively, that would measure "time between when user clicks Upload and when ingest completion email is sent".

knapii-developments (approved):

bistline:
> Looks good.
> 
> This change is an example of how the [`perfTime` metric for the `ingest` event in Mixpanel](https://mixpanel.com/s/4C9vfz) doesn't tell the whole story of user-perceived performance and wait times. That's because `ingest` `perfTime` won't be affected due to this change, even though we know it will slow down end-to-end performance in some cases.
> 
> That isn't something worth resolving here, but it's perhaps worth considering how to better measure end-to-end performance time of "upload and ingest". Naively, that would measure "time between when user clicks Upload and when ingest completion email is sent".

That's a really good point - we can have a discrepancy between `perfTime` and total elapsed time due to this change.  Hopefully this is only on the scale of 2-4 minutes, and only in rare cases.  But certainly something we should think about as we instrument more analytics.

app/models/ingest_job.rb
```diff
@@ -126,6 +164,21 @@ def persisted?
     true
   end
 
+  # Get all instance variables associated with job
+  #
+  # * *returns*
+  #   - (Hash) => Hash of all instance variables
+  def attributes
```
devonbush, line 171: Where is this method used?  I can see the value, I just don't see it used

app/models/ingest_job.rb
```diff
@@ -118,6 +124,38 @@ def poll_for_remote(skip_push: false)
...
+  #   - (Boolean) => T/F if file can launch PAPI ingest job
+  def can_launch_ingest?
+    case self.study_file.file_type
+    when /Matrix/
+      # expression matrices currently cannot be ingested in parallel due to constraints around validating cell names
+      # this block ensures that all other matrices have all cell names ingested and at least one gene entry, which
+      # ensures the matrix has validated
+      other_matrix_files = StudyFile.where(study_id: self.study.id, file_type: /Matrix/, :id.ne => self.study_file.id)
+      # only check other matrix files of the same type, as this is what will be checked when validating
+      similar_matrix_files = other_matrix_files.select {|matrix| matrix.is_raw_counts_file? == self.study_file.is_raw_counts_file?}
```
devonbush, line 140: I like that you did this part of the check in ruby instead of in the mongo query -- mongo nested equality/inequality is finicky

app/models/ingest_job.rb
```diff
@@ -118,6 +124,38 @@ def poll_for_remote(skip_push: false)
...
+      similar_matrix_files.each do |matrix_file|
+        if matrix_file.parsing?
+          matrix_cells = self.study.expression_matrix_cells(matrix_file)
+          matrix_genes = Gene.where(study_id: self.study.id, study_file_id: matrix_file.id)
```
devonbush, line 144: So if I understand this, as long as there are at least some cells and genes from the other file in the database, we're okay to begin parsing this file.  Is that because we always upload the full cell name list first, and so once there are genes, we know the list of cells is complete?

devonbush (commented):
Nice!

PR #874, opened 2021-01-22
Restore "Create annotations" (SCP-2956)

eweitz:
This restores, refines, and refactors the "[Create annotations](https://github.com/broadinstitute/single_cell_portal/wiki/Annotations)" feature.  

Due to a refactor done for maintainable spatial transcriptomics support, the "Create annotations" front-end broke.  It had used a JS-in-ERB structure and fetched dynamic `script` tags.  That approach makes front-end debugging and development hard.  Rather than fixing those bugs by adding substantial new code to cement long-term support for the legacy way, I decided to fix the issues by extending the refactor.

With these changes, "Create annotations" is not only fixed but also more maintainable.  Syntax highlighting, definition lookup, and JavaScript breakpoints are trivial.  Creating user annotations is done via JSON REST API.  The JavaScript -- which is more inherently complex than usual -- now has clearer and more consistent names, smaller and more testable functions, and more high-level code explanation for maintainers.

Demo:

https://user-images.githubusercontent.com/1334561/105666592-99558480-5ea7-11eb-8bc3-f61a3ebf61c0.mov

To test:
1. Sign in to SCP
2. Go to "Explore" tab in Study Overview for a study that has visualizations fully initialized, e.g. [SCP197 on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP197) or [SCP1 on production](https://singlecell.broadinstitute.org/single_cell/study/SCP1).
3. Click hamburger menu to left of search box
4. Click "Create Annotations", then "Select Cells"
5. In the "Clusters" scatter plot, drag cursor to trace an outline of cells of interest
6. Enter text into inputs for "Name this annotation" box, and "Label these cells" boxes for both selections (i.e. Selected and Unselected)
7. Click "Create Annotations"
8. Confirm that success message appears, after waiting for annotation to save
9. Click "View Options" at upper right
10. Click "Select annotation" dropdown menu
11. Confirm the text entered for "Name this annotation" appears as an option under the "User annotation" option group
12. Pick your user annotation from the menu
13. Confirm your annotation appears in the scatter plot, along with your two annotation labels

This satisfies SCP-2956.


jlchang:
Why do we require users to provide a label for the unselected cells? It has always seemed a little odd... The flow is somewhat understandable after you've done it once (ie. all cells start off in an "unselected" pool, as you select, you pick subsets of the cells to be in newly defined groups and they need to be named) but I remember it surprising me that I failed to make annotations because I had not supplied a label for the cells that I was not interested in.

app/controllers/api/v1/user_annotations_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+            @study, params[:name], params[:user_data_arrays_attributes],
+            params[:cluster], params[:loaded_annotation],
+            params[:subsample_threshold], current_api_user
+          )
+
+          response_body = {
+            message: message, annotations: annotations
+          }
+
+          render json: response_body, status: status and return
```
bistline, line 57: The `and return` is not necessary here as calling `render` halts execution.  Usually, we only need `and return` when there is an if/else block that calls `redirect_to` as an escape condition to avoid executing any code further in the method.

app/javascript/lib/scp-api-metrics.js
```diff
@@ -129,6 +129,14 @@ export function logSearch(type, searchParams, perfTime) {
   )
 }
 
+/**
+ * Log create user annotation metrics
+ */
+export function logCreateUserAnnotation() {
```
bistline, line 135: Do we still need to report activity to Google Analytics?  I had thought we'd moved all metrics/activity reporting to Mixpanel.

app/javascript/lib/scp-api.js
```diff
@@ -72,6 +80,67 @@ export async function fetchAuthCode(mock=false) {
...
+  const apiUrl = `/site/studies/${studyAccession}/user_annotation`
+  const [jsonOrResponse, perfTime] = await scpApi(apiUrl, init, mock)
+
+  let message = ''
+  let annotations = {}
+  let errorType = null
+
+  // Consider refactoring this when migrating user-annotations.js to
+  // React, so components share more error handling logic and UI
+  if (jsonOrResponse.ok === false) {
```
bistline, line 122: I agree this should be refactored to share error handling logic, but given the scope of this ticket it's fine for now.

eweitz:
Agreed, Jean.  I've opened SCP-3044 to track that UX enhancement.

app/models/user_annotation.rb
```diff
@@ -14,12 +14,39 @@ class UserAnnotation
...
+  include Swagger::Blocks
+
   field :name, type: String
   field :values, type: Array
   field :queued_for_deletion, type: Boolean, default: false
   field :publishing, type: Boolean, default: false
   field :source_resolution, type: Integer
 
+  # Define parameters used in API POST endpoint to create user annotations
+  swagger_schema :UserAnnotation do
```
bistline, line 26: Naming quibble - since this is not the actual `UserAnnotation` document structure as defined above, we should call this something different, like `UserAnnotationInput` to avoid confusion with the actual model.

bistline (approved):
Looks good - I agree there is room for refinement/improvement, but given the scope of the ticket and the work required to complete it, I think this is more than adequate.  I have some style/usage comments that are non-blocking, and a question about reporting.

devonbush (commented):
Looks good, I haven't completed a detailed code style review, but I don't see anything that would block pushing this out as a hotfix tonight

app/controllers/api/v1/user_annotations_controller.rb
eweitz: Done in 296cd75.

app/javascript/lib/scp-api-metrics.js
eweitz: Yup, we want to preserve GA logging until we resolve SCP-2801, then remove all GA logging per SCP-2449.

app/models/user_annotation.rb
eweitz: Good point, I had been waffling around on this.  Done in 296cd75.

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -6,7 +6,8 @@ module StudyAware
 
         def set_study
           # allow either accession or id
-          @study = Study.any_of({accession: params[:study_id]},{id: params[:study_id]}).first
+          id = params[:accession].present? ? params[:accession] : params[:study_id]
```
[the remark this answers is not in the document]
eweitz: A standard name would be ideal, indeed, but I think `accession` (or `study_accession`) would be more developer-friendly.  Accessions are stable and more glanceable and memorable, the new front-end typically uses accessions, and we return `accession` and not `study_id` in API responses.

app/javascript/lib/scatter-plot.js
```diff
@@ -277,6 +277,15 @@ export async function scatterPlot(apiParams, props) {
...
   scatterPlots.push(rawPlot)
 
+  if (scatterPlots.length === 1) {
+    // Set cluster type for "Create Annotations".  Crude, but ensures parity
+    // with prior ERB-centric Study Overview functionality for studies lacking
+    // spatial data.
+    //
+    // TODO (SCP-2962): Support "Create Annotations" for spatial scatter plots
+    window.SCP.firstScatterPlotIs3D = rawPlot.is3D
+  }
```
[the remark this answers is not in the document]
eweitz: Nice, agreed, done in 296cd75.

app/javascript/lib/scp-api.js
```diff
@@ -41,8 +42,20 @@ export function studyNameAsUrlParam(studyName) {
   return studyName.toLowerCase().replace(/ /g, '-').replace(/[^0-9a-z-]/gi, '')
 }
 
+/** Configure an `init` for `fetch` to use POST, and respect any mocking  */
+function configPost(mock=false) {
```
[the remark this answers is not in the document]
eweitz: Good call, done in 296cd75.

config/routes.rb
```diff
@@ -66,6 +65,8 @@
           get 'studies/:accession/download', to: 'site#download_data', as: :site_study_download_data
           get 'studies/:accession/stream', to: 'site#stream_data', as: :site_study_stream_data
 
+          post 'studies/:accession/user_annotation', to: 'user_annotations#create_user_annotation', as: :site_study_create_user_annotation
```
devonbush, line 68: I'd like to stick to REST resource conventions as much as possible, especially within the studies namespace.  I would move this up to the 'studies' resource, that will let you define it with something like
`resources :user_annotations, only: [:create]`. 
This you can rename your api controller method to just `create` and also eliminate the need to modify study_aware, since the study id will get passed the same way as the other newer API calls.  

Might make sense to have a quick call on this -- I know rails routing suggestions are never as simple to follow as they are to make :)

app/views/site/_search_options.html.erb
```diff
@@ -258,21 +253,22 @@
...
+              <div id="create-annotation-panel" class="collapse"></div>
             </div>
           </div>
         </div>
       </div>
       <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
         <% if user_signed_in? %>
 
-          var targetPlotString = '<%= params[:gene_set].nil? && params[:gene].nil? ? 'cluster-plot' : 'scatter-plot' %>';
+          // TODO (SCP-2962): Support "Create Annotations" for spatial scatter plots
```
devonbush, line 264: nice TODO

app/views/site/_search_options.html.erb
```diff
@@ -258,21 +253,22 @@
...
+          var targetPlotId = 'scatter-plot-0';
```
devonbush, line 265: I believe this can be replaced with `window.SCP.getScatterPlots()[0].plotId`, which is nice to avoid hardcoding ID formats, (even if it requires a double-check that window.SCP.getScatterPlots() has length)

app/models/user_annotation.rb
```diff
@@ -14,12 +14,39 @@ class UserAnnotation
...
+    end
+    property :subsample_threshold do
+      key :type, :string
+      key :description, 'Subsample threshold'
+    end
+    property :subsample_annotation do
+      key :type, :string
+      key :description, 'Subsample annotation'
+    end
+  end
```
devonbush, line 48: we discussed moving this into the controller, since API docs don't really belong in models -- is that still true, or are there other concerns?

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,102 @@
...
+    rescue Mongoid::Errors::InvalidValue => e
+      error_context = ErrorTracker.format_extra_context(study, {params: log_params})
+      ErrorTracker.report_exception(e, current_user, error_context)
+      # If an invalid value was somehow passed through the form, and couldn't save the annotation
+      cluster_annotations = ClusterVizService.load_cluster_group_annotations(study, cluster, current_user)
+      message = 'The following errors prevented the annotation from being saved: ' + 'Invalid data type submitted. (' + e.problem + '. ' + e.resolution + ')'
+      Rails.logger.error "Creating user annotation of params: #{user_annotation_params}, invalid value of #{e.message}"
+      [message, cluster_annotations, 400] # Bad request
+
+    rescue NoMethodError => e
```
devonbush, line 82: if there are places where our code is known to not be nil-safe, we should address that directly.   And if there aren't, then catching these separately doesn't add any value -- they'll show up in Sentry and we'll triage them the same either way.  So I would remove this rescue NoMethodError block

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,102 @@
...
+      error_context = ErrorTracker.format_extra_context(study, {params: log_params})
+      ErrorTracker.report_exception(e, current_user, error_context)
+      # If something is nil and can't have a method called on it, respond with an alert
+      cluster_annotations = ClusterVizService.load_cluster_group_annotations(study, cluster, current_user)
+      message = 'The following errors prevented the annotation from being saved: ' + e.message
+      Rails.logger.error "Creating user annotation of params: #{user_annotation_params}, no method error #{e.message}"
+      [message, cluster_annotations, 500] # Internal server error
+
+    rescue => e
```
devonbush, line 91: what does this add beyond our current default API controller exception handling?  I think the error message itself belongs in the controller or the UI, not here.  I think this block should be removed and, to the extent it isn't duplicative of our existing error-handling code, reallocate it to the controller

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+      [message, cluster_annotations, status]
+
+    # Handle other errors in saving user annotation
+    rescue Mongoid::Errors::InvalidValue => e
+      error_context = ErrorTracker.format_extra_context(study, {params: log_params})
+      ErrorTracker.report_exception(e, current_user, error_context)
+      # If an invalid value was somehow passed through the form, and couldn't save the annotation
+      cluster_annotations = ClusterVizService.load_cluster_group_annotations(study, cluster, current_user)
+      message = 'The following errors prevented the annotation from being saved: ' + 'Invalid data type submitted. (' + e.problem + '. ' + e.resolution + ')'
```
devonbush, line 78: I like that you're catching a relatively low-level message here and reformatting it to be more sensible.  Instead of a complex return type for this function though, I think you should just raise a new exception with your more informative message

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,102 @@
...
+        # Method call to create the user data arrays for this annotation
+        user_annotation.initialize_user_data_arrays(user_data_arrays_attributes, subsample_annotation, subsample_threshold, loaded_annotation)
+
+        # Reset the annotations in the dropdowns to include this new annotation
+        cluster_annotations = ClusterVizService.load_cluster_group_annotations(study, cluster, current_user)
+
+        message = "User Annotation: '#{user_annotation.name}' successfully saved. You may now view this annotation via the annotations dropdown."
+        status = 200 # Success
+      else
+        # If there was an error saving, alert the user something broke
```
devonbush, line 64: This if/else construct can be avoided by using `save!` rather than `save`, since all we're doing is just re-throwing the error.  `save!` will throw an exception with the error message if the object could not be saved, and then our regular error-handling code will pick it up.

devonbush (changes requested):
Love the attention to all the error-handling.  And I think we can make that code a lot simpler.

config/routes.rb
eweitz: Thanks for the discussing this through with me.  It's done in the latest code.

app/views/site/_search_options.html.erb
eweitz: Done in latest code.

app/models/user_annotation.rb
eweitz: I had qualms after I looked back, and saw I put this in a model because that's where all 26 other SCP `swagger_schema` declarations are.  But I chatted with Jon, and he agrees with your rationale, so I've made your suggested change.

app/lib/user_annotation_service.rb
eweitz: Done -- latest code refines this aspect of the old `site_controller.rb` handling.

app/lib/user_annotation_service.rb
eweitz: Done -- latest code refines this aspect of the old `site_controller.rb` handling.

app/lib/user_annotation_service.rb
eweitz: Done -- latest code refines this aspect of the old `site_controller.rb` handling.

app/lib/user_annotation_service.rb
eweitz: Done -- latest code refines this aspect of the old `site_controller.rb` handling.

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,54 @@
...
+    user_annotation = UserAnnotation.new(user_id: user_id, study_id: study_id,
+                                          cluster_group_id: cluster_group_id,
+                                          values: data_names, name: name,
+                                          source_resolution: source_resolution)
+
+    # override cluster setter to use the current selected cluster, needed for reloading
+    cluster = user_annotation.cluster_group
+
+    # Save the user annotation, and handle any exceptions
+    if user_annotation.save!
```
devonbush, line 34: you don't need the 'if' here -- `save!` will never return false -- if the save fails, an exception is thrown

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,54 @@
+class UserAnnotationService
+  # Methods to interact with annotation data, beyond just visualization
+
```
devonbush, line 3: this function could use some documentation on what these params are, especially the structure expected by user_data_arrays_attributes and loaded_annotation .  I *think* I know what they mean, but I'm not sure.

app/lib/user_annotation_service.rb
```diff
@@ -0,0 +1,54 @@
...
+      # known use cases where `subsample_annotation` != `loaded_annotation`.
+      subsample_annotation = loaded_annotation
+
+      # Method call to create the user data arrays for this annotation
+      user_annotation.initialize_user_data_arrays(user_data_arrays_attributes, subsample_annotation, subsample_threshold, loaded_annotation)
+
+      # Reset the annotations in the dropdowns to include this new annotation
+      cluster_annotations = ClusterVizService.load_cluster_group_annotations(study, cluster, current_user)
+
+      message = "User Annotation: '#{user_annotation.name}' successfully saved. You may now view this annotation via the annotations dropdown."
```
devonbush, line 48: this message should be assembled in the controller -- that will simplify the return type of this function and let the controller properly own user-facing messaging.

app/controllers/api/v1/user_annotations_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+          message, annotations = UserAnnotationService.create_user_annotation(
+            @study, params[:name], params[:user_data_arrays_attributes],
+            cluster, params[:loaded_annotation],
+            params[:subsample_threshold], current_api_user
+          )
+
+          render json: {message: message, annotations: annotations}
+
+        # Handle other errors in saving user annotation
+        rescue Mongoid::Errors::InvalidValue => e
```
devonbush, line 91: I think I was too vague in my earlier comments about where I think this excpetion handling code belongs.  I LOVE this specific exception catching, but it belongs in the service method.  This controller shouldn't need to know anything about Mongoid-specific errors (or that the service uses Mongoid at all).  So this `rescue Mongoid::Errors::InvalidValue` block should be in the service, and it should handle the exception by re-throwing it as an ArgumentError.  Then the more general rescue block below (which should stay in the controller) will handle rendering it to the client.

app/controllers/api/v1/user_annotations_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+          message = 'The following errors prevented the annotation from being saved: ' + 'Invalid data type submitted. (' + e.problem + '. ' + e.resolution + ')'
+          Rails.logger.error "Creating user annotation of params: #{user_annotation_params}, invalid value of #{e.message}"
+          render json: {error: error}, status: 400 # Bad request
+
+        rescue => e
+          # If a generic unexpected error occurred and couldn't save the annotation
+          error_context = ErrorTracker.format_extra_context(@study, {params: log_params})
+          ErrorTracker.report_exception(e, current_user, error_context)
+
+          if e.summary.include? 'Name' and e.summary.include? 'has already been taken'
```
devonbush, line 103: This likewise feels like specific business logic that belongs in the service layer -- the service layer can catch this, and throw it as an ArgumentError.  Then this controller can do `rescue ArgumentError { tell the user it's their own fault }` and `rescue  <<Everything else>> { tell the user it was something unexpected }`

I hope I'm being clear about the general pattern I want -- service methods should handle the business logic, and re-throw low-level exceptions as the correct type of exception that would make sense to the caller.  E.g. if a caller passes in bad values to a service method, the service should throw ArgumentError, because that's the exception that makes sense to the caller, regardless of what exception caused the service method to realize that one of its arguments was bad.  

Happy to hop on a call to chat--this is something that's relatively important as we continue to convert to this pattern of thin controllers delegating to business services, and so I want to make sure we're on the same page.

devonbush (changes requested):
Sorry, I don't think my previous comments captured the pattern that's important for controller->service delegation and exception handling.  I've tried to explain better below, also happy to chat

app/lib/user_annotation_service.rb
eweitz: Done in 655b981.

app/lib/user_annotation_service.rb
eweitz: Done in 69caefc1.

app/lib/user_annotation_service.rb
eweitz: Good point, done in 69caefc1.

app/controllers/api/v1/user_annotations_controller.rb
eweitz: Done in 655b981.

app/controllers/api/v1/user_annotations_controller.rb
eweitz: Good points, and thanks for the chat -- done in 655b981.

devonbush (approved):

eweitz:
The last two builds have different random failures, preceded by a success.  Given the failures differ from each other and no changes to tested code were made since success, the failures seem like false positives.

PR #877, opened 2021-01-26
* SCP-2763 Send user metric ID to ingest pipeline

knapii-developments:
This PR passes a user's metric ID to Ingest pipeline so that the user identifier is included in Ingest pipeline's logs. The Papi command now looks like:

```
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 600f3b56fa6429c7942c7152
  - "--study-file-id"
  - 60103546fa6429c7942c715f
  - "--user-metrics-uuid"
  - <user uuid>
  - ingest_cell_metadata
  - "--cell-metadata-file"
  - gs://<bucket location>/metadata_example.txt
  - "--study-accession"
  - SCP78
  - "--ingest-cell-metadata"
  ```

This PR has to be release  [pull #203](https://github.com/broadinstitute/scp-ingest-pipeline/pull/203) in ingest pipeline

### DIFF
--- a/app/models/papi_client.rb
+++ b/app/models/papi_client.rb
@@ -98,7 +98,7 @@ class PapiClient < Struct.new(:project, :service_account_credentials, :service)
     study = study_file.study
     accession = study.accession
     resources = self.create_resources_object(regions: ['us-central1'])
-    command_line = self.get_command_line(study_file: study_file, action: action,  user_metrics_uuid: user.metrics_uuid)
+    command_line = self.get_command_line(study_file: study_file, action: action, user_metrics_uuid: user.metrics_uuid)
     labels = {
         study_accession: accession,
         user_id: user.id.to_s,
--- a/app/models/papi_client.rb
+++ b/app/models/papi_client.rb
@@ -98,7 +98,7 @@ class PapiClient < Struct.new(:project, :service_account_credentials, :service)
     study = study_file.study
     accession = study.accession
     resources = self.create_resources_object(regions: ['us-central1'])
-    command_line = self.get_command_line(study_file: study_file, action: action)
+    command_line = self.get_command_line(study_file: study_file, action: action,  user_metrics_uuid: user.metrics_uuid)
     labels = {
         study_accession: accession,
         user_id: user.id.to_s,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by eweitz — APPROVED** [review]

**Review by bistline — CHANGES_REQUESTED** [review]
I think we might possibly have an issue with the `metrics_uuid` not getting assigned which is causing ingest runs not to initiate.  In normal usage, this would always get set when a user signs in.  However, since most of these tests simulate a login without actually performing one, we aren't setting it in tests, and the ingest runs are all erroring out before even launching.

I think we can handle by doing a couple of things:

1. auto-set the `metrics_uuid` field for all users in `db/seeds.rb`
2. in `test/factories/user.rb`, add the following code after line 13:
```
metrics_uuid {SecureRandom.uuid}
```

Without this value being present, I think the output of `PapiClient#get_command_line` is getting munged so that the value for `action` is passed as `metrics_uuid`, and this is causing a failure before the job can even launch.

devonbush:
good points Jon--agree with those changes.  One other thing we could do to mitigate this is to have these methods check if metrics_uuid is null, and if so, just use the dedicated constant ingest_pipeline mixpanel uuid that we are using.  that will make sure these methods never fail due to mixpanel user wonkiness

devonbush (changes requested):
overall looks good, just the cleanups needed to make sure tests work

devonbush (approved):

knapii-developments:
> I think we might possibly have an issue with the `metrics_uuid` not getting assigned which is causing ingest runs not to initiate. In normal usage, this would always get set when a user signs in. However, since most of these tests simulate a login without actually performing one, we aren't setting it in tests, and the ingest runs are all erroring out before even launching.
> 
> I think we can handle by doing a couple of things:
> 
> 1. auto-set the `metrics_uuid` field for all users in `db/seeds.rb`
> 2. in `test/factories/user.rb`, add the following code after line 13:
> 
> ```
> metrics_uuid {SecureRandom.uuid}
> ```
> 
> Without this value being present, I think the output of `PapiClient#get_command_line` is getting munged so that the value for `action` is passed as `metrics_uuid`, and this is causing a failure before the job can even launch.

This has been handled in e19bbb0

bistline (approved):

PR #879, opened 2021-01-27
Updating paperclip gem for test IO errors (SCP-3039)

bistline:
Resolving test IO errors with an update of `paperclip` gem to latest release.  This gem has been deprecated, but the last release from [July 2018](https://github.com/thoughtbot/paperclip/releases/tag/v6.1.0) contains a fix to "Manually close Tempfiles when we are done with them", which could account for the `I/O error: flush error` that was repeatedly triggered during CI runs.  

Because all tests are invoked with a single `bin/rake test` call, the stack is loaded once and all resources are shared across each test suite. This change was introduced in #777 and differered from the previous "manual" invocation of each test suite individually in a specified order, which had a separate overhead/garbage collection for each suite.  This new, streamlined invocation can lead to an overflow if garbage collection does not release all of the temp files created in a timely manner as we simulate file uploads in integration tests.

Eventually, Paperclip will need to be replaced with a more up-to-date library, but this final update will allow us more time to evaluate candidates and perform the migration without constantly having to re-run builds due to spurious errors.

The build associated with this PR was run 5 times in a row without triggering the I/O error.  While not absolute proof, these errors have increased as of late, and were occurring on approximately 1 out of 3 runs, so 5 consecutive green builds is a good indicator this issue is at least resolved for now.

This PR satisfies SCP-3039.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -109,7 +109,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
-    concurrent-ruby (1.1.7)
+    concurrent-ruby (1.1.8)
     connection_pool (2.2.2)
     crass (1.0.6)
     daemons (1.2.6)
@@ -193,7 +193,7 @@ GEM
     http-cookie (1.0.3)
       domain_name (~> 0.5)
     httpclient (2.8.3)
-    i18n (1.8.5)
+    i18n (1.8.7)
       concurrent-ruby (~> 1.0)
     jbuilder (2.7.0)
       activesupport (>= 4.2.0)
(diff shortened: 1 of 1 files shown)


devonbush (approved):
love it -- good digging to find out about that temp file issue they resolved

devonbush:
also, assuming this works, you get +500 engineering foresight points for correctly targeting this gem and championing the issue during sprint planning

bistline:
> love it -- good digging to find out about that temp file issue they resolved

Thanks - it was mostly luck, but I eventually noticed all the tests that could trigger this error usually had something like this in `STDOUT`:

```[paperclip] Link failed with Invalid cross-device link @ rb_file_s_link - (/home/app/webapp/db/seed/synthetic_studies/mouse_brain/coordinate_labels.tsv, /tmp/43b182cd4b52c9d02925712441a2d91220210112-2422-2m1re6.tsv); copying link /home/app/webapp/db/seed/synthetic_studies/mouse_brain/coordinate_labels.tsv to /tmp/43b182cd4b52c9d02925712441a2d91220210112-2422-2m1re6.tsv```

Searching for the `Link failed with Invalid cross-device link @ rb_file_s_link` error eventually led me back to paperclip and the latest update.  We're still seeing the errors in the logs, but the last update apparently fixes garbage collection so that there isn't a handler for that temp file hanging around too long.

It's also worth noting that we don't see this is deployed environments, only test runs.  Looking at `STDOUT` in production/staging doesn't have any of these errors.

knapii-developments (approved):

PR #880, opened 2021-01-27
Fixing bundling button regression re: file type radio buttons (SCP-2993)

bistline:
When the "File Type" form input was switched from a select input to a group of radio buttons, this broke the on-page javascript that would dynamically render buttons for adding bundled genes/barcodes files for MTX matrix files while the upload is in process.  This update restores that functionality by replacing function calls to `$(selector).val()` with `$(selector).prop("checked")` to check if a particular radio is selected to designate the file type.

This PR satisfies SCP-2993.

### DIFF
--- a/app/views/studies/_initialize_expression_form.html.erb
+++ b/app/views/studies/_initialize_expression_form.html.erb
@@ -83,7 +83,7 @@
 
 	<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
       $('#expression_form_<%= study_file._id %>').find('.file-type').on('change', function() {
-        if ($(this).val() === 'Expression Matrix') {
+        if ( $('#expression_form_<%= study_file._id %> #study_file_file_type_expression_matrix').prop("checked") ) {
           $('#expression_form_<%= study_file._id %>').find('.initialize-bundle-target').empty();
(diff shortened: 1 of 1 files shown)


app/views/studies/_initialize_expression_form.html.erb
```diff
@@ -102,7 +102,7 @@
 					// auto-select MM Coordinate Matrix if applicable
           fileExt = fileName.split('.').slice(-1)[0];
           if (fileExt === 'mtx') {
```
devonbush, line 104: we use `#expression_form_<%= study_file._id %>` more than once -- might be good to pull it into a variable at the top of the script.

devonbush (approved):
one non-blocking suggestion -- could be nice, but not worth introducing regressions over if it's at all non-trivial

eweitz (approved):

app/views/studies/_initialize_expression_form.html.erb
bistline: agreed

bistline: dcda9f6

devonbush (approved):

PR #881, opened 2021-01-28
fixing default annotation logic (SCP-3059)

devonbush:
There was a big gap in the automated tests for the annotation service, which was definitely my oversight in rushing to get them out before the end of the year.

@bistline  I'd really appreciate your careful eyes on this to make sure I'm passing through the behavior of study.default_annotation correctly.  

To test:
1. Set a default annotation for a study.  
2. Do a global gene search for that study
3. confirm the default annotation is what renders first

### DIFF
--- a/app/lib/annotation_viz_service.rb
+++ b/app/lib/annotation_viz_service.rb
@@ -1,10 +1,46 @@
 class AnnotationVizService
   # set of utility methods used for interacting with annotation data
-  def self.get_selected_annotation(study, cluster, annot_name, annot_type, annot_scope)
+
+  # Retrieves an object representing the selected annotation. If nil is passed for the last four
+  # arguments, it will get the study annotation instead
+  # Params:
+  # - study: the Study object
+  # - cluster: ClusterGroup object (or nil for study-wide annotations)
+  # - annot_name: string name of the annotation
+  # - annot_type: string type (group or numeric)
+  # - annot_scope: string scope (study, cluster, or user)
+  # Returns:
+  # - See populate_annotation_by_class for the object structure
+  def self.get_selected_annotation(study, cluster=nil, annot_name=nil, annot_type=nil, annot_scope=nil)
     # construct object based on name, type & scope
+    if annot_name.blank?
+      # get the default annotation
+      default_annot = nil
+      if annot_scope == 'study'
+        # get the default study-wide annotation
+        default_annot = study.default_annotation(nil)
+      elsif cluster.present?
+        # get the default annotation for the cluster
+        default_annot = study.default_annotation(cluster)
+      else
+        # get the default annotation for the default cluster
+        default_annot = study.default_annotation
+      end
+
+      if !default_annot.blank?
+        annot_name, annot_type, annot_scope = default_annot.split('--')
+        if cluster.blank?
+          cluster = study.default_cluster
+        end
+      end
+    end
     case annot_scope
     when 'cluster'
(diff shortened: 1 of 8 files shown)


app/lib/annotation_viz_service.rb
```diff
@@ -1,10 +1,46 @@
 class AnnotationVizService
   # set of utility methods used for interacting with annotation data
-  def self.get_selected_annotation(study, cluster, annot_name, annot_type, annot_scope)
+
+  # Retrieves an object representing the selected annotation. If nil is passed for the last four
+  # arguments, it will get the study annotation instead
```
bistline, line 5:
```suggestion
  # arguments, it will get the study's default annotation instead
```

app/lib/annotation_viz_service.rb
```diff
@@ -1,10 +1,46 @@
...
+  # Params:
+  # - study: the Study object
+  # - cluster: ClusterGroup object (or nil for study-wide annotations)
+  # - annot_name: string name of the annotation
+  # - annot_type: string type (group or numeric)
+  # - annot_scope: string scope (study, cluster, or user)
+  # Returns:
+  # - See populate_annotation_by_class for the object structure
+  def self.get_selected_annotation(study, cluster=nil, annot_name=nil, annot_type=nil, annot_scope=nil)
```
bistline, line 14: I think this refactoring is fine - this method can have a case where the request returns `nil`, either through bad parameters, or the study not having a default annotation.  While this isn't ideal, because `populate_annotation_by_class` is type safe, it can safely return `nil` without throwing an exception.

app/lib/annotation_viz_service.rb
```diff
@@ -47,7 +93,7 @@ def self.get_study_annotation_options(study, user)
   end
 
   # returns a flat array of annotation objects, with name, scope, annotation_type, and values for each
-  def self.available_annotations(study, cluster, current_user, annotation_type=nil)
+  def self.available_annotations(study, cluster=nil, current_user=nil, annotation_type=nil)
```
bistline, line 96: I would convert this to kwargs and refactor where used - having numerous `nil` defaults in a row means you have to make method calls like this:
```AnnotationVizService.get_selected_annotation(@study, nil, nil, nil, 'study')```

I would posit that this is cleaner:

```
def self.available_annotations(study, cluster: nil, current_user: nil, annotation_type: nil)
...
AnnotationVizService.get_selected_annotation(@study, annotation_type: 'study')
```

File: test/integration/lib/annotation_viz_service_test.rb [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+                                             annotation_input: [
+                                                 {name: 'species', type: 'group', values: ['dog', 'cat', 'dog']},
+                                                 {name: 'disease', type: 'group', values: ['none', 'none', 'measles']}
+                                             ])
+
+  end
+
+
+  test 'gets the default annotation when no annotation name is specified' do
+    annotation = AnnotationVizService.get_selected_annotation(@basic_study, nil, nil, nil, nil)
```
bistline, line 39: See above comment about this method signature.

bistline (changes requested):
Looks good - I think the logic is fine.  I would change the method signature for `available_annotations` to use keyword arguments, rather than positional ones.

devonbush:
test failures are all the dreaded 'getcwd' -- none relevant

bistline (approved):

eweitz (approved):

PR #882, opened 2021-01-28
Allow testing via `bin/run_tests.sh` outside of Docker (SCP-3019)

bistline:
Currently, if a developer is working in a non-Dockerized setup, running the unit/integration test suites requires multiple manual steps to set up the environment and run tests.  While fine for some, it is cumbersome and easy to miss steps, which can result in false negatives in tests, or prevents tests from running at all.  Our CI setup uses the shell script `bin/run_tests.sh` to automate initializing the environment and running all tests, but it does not work outside of Docker.

This update modifies the shell script to run both inside and outside of Docker by passing a single flag of `-l`.  When present, the script will invoke `rails_local_setup.rb` for the current user (this initializes all vault secrets), and perform other setup tasks specific to the non-Dockerized environment to allow running tests natively in Ruby.  All tests are then invoked using the same syntax of `yarn ui-test` and `bin/rake test`.  Developers can also specify single test suites, or even individual tests to run by using the `-t` and `-R` options, respectively.

Example usage (run from the project root directory, outside of Docker):

```
# run all tests locally
bin/run_tests.sh -l

# run a single Ruby test suite
bin/run_tests.sh -l -t test/integration/fire_cloud_client_test.rb

# run a single test from a test suite
bin/run_tests.sh -l -t test/integration/fire_cloud_client_test.rb -R 'test_firecloud_api_available'

# run all matching tests from a test suite
bin/run_tests.sh -l -t test/integration/fire_cloud_client_test.rb -R /workspace/
```

In addition, this update addresses two small bugs discovered while working on this PR:
1. `RequestUtils. get_base_url` now correctly honors Dockerized vs. non-Dockerized configurations by setting `config. default_url_options` for the development environment
2. Removes incorrect/unused form parameters in `test/integration/study_creation_test.rb` that generated error messages in the logs

_Note: when invoking locally, the script can appear to hang occasionally when trying to compute coverage stats, especially when only running a single test.  It will eventually complete and exit normally.  Developers can also hit CTRL+C to exit early as conve_

This PR satisfies SCP-3019.


devonbush (approved):
Code looks good--and I tested it locally and it worked great.  This is definitely a step forward in letting developers easily run tests in a CI-like environment as a final validation.  
 
I think it's important to note the running time, though.  Running `bin/run_tests.sh -t test/integration/lib/expression_viz_service_test.rb:62` takes ~4 minutes each run, compared to ~15 seconds for `rails test  test/integration/lib/expression_viz_service_test.rb:62`. 

I want to make sure that anyone who is going to be doing semi-regular rails work can run tests individually without this script, so they can get faster feedback.  Rebuilding the world each test run is useful as a final 'all-clear' check right before pushing, but makes regular development a slog.

bistline:
> I want to make sure that anyone who is going to be doing semi-regular rails work can run tests individually without this script, so they can get faster feedback. Rebuilding the world each test run is useful as a final 'all-clear' check right before pushing, but makes regular development much more difficult.

That is an excellent point, and one I had considered.  I opted for the longer, all-in-one approach that is idempotent as opposed to the faster one that has the potential to fail due to setup/cleanup not being run.  There is a lot of setup and cleanup that have to run in order to ensure that each test run starts with an identical state.  Having said that, it could be useful to disentangle all of that complexity into a separate script so that tests could be invoked natively.  My only issue with that approach is some tests rely on DB state, and not properly running cleanups can affect it.

I think some useful follow-on work could be completely disentangling `db/seeds.rb` from all tests, and rely solely on `FactoryBot` and `SelfCleaningSuite` for handling test DB state.  This way, we can reduce the amount of setup needed to simply initializing the environment and starting `delayed_job`, and can achieve the best performance/productivity for developers.

devonbush:
Yup, right on.

eweitz (approved):
> I think some useful follow-on work could be completely disentangling db/seeds.rb from all tests, and rely solely on FactoryBot and SelfCleaningSuite for handling test DB state. This way, we can reduce the amount of setup needed to simply initializing the environment and starting delayed_job, and can achieve the best performance/productivity for developers.

That is also necessary for test parallelization, which would drastically improve CI build times.  Two birds, one stone.

PR #884, opened 2021-02-01
Logging 500 type errors to Mixpanel (SCP-3052)

bistline:
Currently, server-side errors in SCP are only logged to Sentry.  While sufficient for tracking and triaging bug fixes, we lack good analytics into groups/classes of errors due to how Sentry tracks and groups errors.  This update adds Mixpanel as a secondary location for tracking errors that occur server-side in SCP.  This is achieved by adding `rescue_from Exception` handlers in `ApplicationController` and `ApiBaseController`, from which all SCP controllers inherit.  In addition, this update improves the UX for 500 type errors by catching them and rendering an actual Rails layout & template instead of a static HTML 500 page.  This means that future UI updates will be reflected in error pages moving forward, without having to update raw HTML pages.

Originally, this effort was to include 4xx type errors as well, but due to both how these errors are caught in Rails (at the middleware level, as opposed to a controller), and our existing Sentry integration, catching and logging these errors proved impossible.

This PR satisfies SCP-3052.


app/lib/metrics_service.rb
```diff
@@ -124,4 +124,65 @@ def self.log(name, props={}, user)
...
+    # type: error Ruby class
+    # text: error message/details
+    # appId: identifier for Single Cell Portal
+    # env: Rails environment
+    # studyAccession: corresponding SCP study (if present)
+    # authenticated: user auth status
+    # registeredForTerra: status of user Terra registration (if present)
+    # distinct_id: user metrics UUID for identifying users in Bard
+    props = {
+      appPath: request.fullpath,
```
eweitz, line 153: Blocker: `appPath` must omit study name, which exists in some URLs.  Ideally,`appPath` logged from server should align with `appPath` logged from client, to enable easy tracing between client and server event. (Solving the former would aid the latter.)

See [`appPath` and `appFullPath` usage](https://github.com/broadinstitute/single_cell_portal_core/blob/e7882646087c6eed0d9023b4d9dbff150815ee09/app/javascript/lib/metrics-api.js#L267), and relevant function definitions in [`getAnalyticsPageName`](https://github.com/broadinstitute/single_cell_portal_core/blob/e7882646087c6eed0d9023b4d9dbff150815ee09/app/javascript/lib/metrics-api.js#L247) and [`getAppFullPath`](https://github.com/broadinstitute/single_cell_portal_core/blob/e7882646087c6eed0d9023b4d9dbff150815ee09/app/javascript/lib/metrics-api.js#L238).

This special handling exists because workspace name (and, analogously, study name) should not be logged to Mixpanel per DSP policy; see requirement 1 [here](https://docs.google.com/document/d/1di8uwv-nDMNJp83Gs9Q_mlZp3q1DFboXGPLzbsbPyn8/edit#heading=h.n7z2sg51xcwg).  For example, if an error occurred in https://singlecell.broadinstitute.org/single_cell/study/SCP555/personally-identifiable-information-in-title, that special handling prevents the sensitive information in the title from being logged to Mixpanel.

eweitz (changes requested):
Looks good, except for missing handling of an edge case related to Mixpanel and PII.

app/lib/metrics_service.rb
bistline: 849ddd9

eweitz (approved):

app/helpers/application_helper.rb
```diff
@@ -376,11 +376,7 @@ def pluralize_without_count(count, noun, text=nil)
   end
 
   def get_page_name
```
devonbush, line 378: Nice -- not only is it good to have this pulled into a service, but pulling it into MetricsService clarifies that this is a page name that's only used for analytics

devonbush (approved):
great stuff

PR #885, opened 2021-02-02
Log error modal message to Mixpanel (SCP-3053)

bistline:
Currently, server- and client-side errors are logged to Mixpanel when they are caught.  However, there is a large number of potential errors that are averted before a potential exception is thrown (e.g. trying to pull up a study by an accession that does not exist) that are simply shown to the user as an "alert modal", and are not reported anywhere else.  This update adds a `log` call from the `metrics-api.js` library to report the content of the error text to Mixpanel.  This will help identify any usability issues that the portal has been handling silently.

This update also extends server-side error reporting to Mixpanel of explicitly handled exceptions.  Previously only uncaught exceptions were logged to Mixpanel via `MetricsService.report_error()`.

To test:

1. Pull `jb-mixpanel-error-modal` locally and boot development instance
2. When home page loads, navigate to https://localhost/single_cell/study/SCP0 (or https://localhost:3000/single_cell/study/SCP0 if running non-Dockerized)
3. Once modal renders with `Study not found` message, open Dev Tools > Network and observe Mixpanel `POST` event with `event: "error-modal"` title
4. Alternatively, go to https://mixpanel.com/report/2085496/view/19055/dashboards and create a new Insights report using the `error-modal` event

This PR satisfies SCP-3053.


bistline:
The dreaded `No such file or directory - getcwd` error again - no obvious regressions.

eweitz (approved):

eweitz:
More seems needed to make these analytics specifically actionable.  Here's an example `error-modal` event [from Mixpanel](https://mixpanel.com/s/3O4brA):

Screen Shot 2021-02-02 at 10 55 37 PM

The event nicely indicates that the user saw an error message, but omits essential information about what triggered it.  

In this particular case, perhaps the unfound accession (SCP0) could be included in the alert text in a way that makes alerts formulaic and thus searchable, but also specific and reproducible.  For example, instead of alerting:

> Study not found. Please check the name and try again.

Perhaps the alert could say:

> Study not found.  Please check the accession you provided ("SCP0") and try again.

Care would also be needed distinguish different "Study not found" messages to e.g. only include the full trigger if it weren't a study name / PII, i.e. matched the format of an accession or study ID.  Else, perhaps study name in the alert and log could be truncated to obfuscate sensitive information.

Examining other cases of `alert`  calls triggered via Rails controllers, I see many similarly omit key data needed to be able to reproduce the user's issue.  Refining these error modal alerts for observability might be out-of-scope for this task, but seems necessary for this work to be useful for investigating particular problem cases.

bistline:
> Examining other cases of `alert` calls triggered via Rails controllers, I see many similarly omit key data needed to be able to reproduce the user's issue. Refining these error modal alerts for observability might be out-of-scope for this task, but seems necessary for this work to be useful for investigating particular problem cases.

I agree that many of the messages are not as informative as they could be when viewed through the "analytics" lens, but many of these are designed to be informative directly to the user in the context of the action they are trying to perform.  What might seem cryptic to someone looking at an error message in Mixpanel may mean much more to the user as they attempt to perform an action, and then are blocked from doing so.  But your main point still stands.

My only reticence is that some of these error messages take the form of "You do not have permission to perform that action", or something similar.  This is actually a privacy concern - if a user is attempting to traverse a series of routes, and are supplying dummy accession values in the hope of finding specific ones, responding to the effect of "yes this is a valid accession and you can't see it" could expose potentially sensitive information.  An example of the analogous action in Terra (trying to fish for workspace names) displays the following error:

```
Could not display workspace
You are trying to access a workspace that either does not exist, or you do not have access to it.
```

This informs the user that they are attempting an action that is either invalid or unauthorized, but doesn't specify which, nor does it leak any information.  I think a holistic review of error messages with this in mind is warranted, and given the relative minuscule size of this PR so far would not necessarily be "out of scope".

bistline:
Again with the `getcwd` errors - unable to get a clean build after 5 attempts, but no errors were reproducible.

knapii-developments (approved):

PR #889, opened 2021-02-04
Adding permission enforcement tests (SCP-3075)

bistline:
This update adds explicit tests to ensure study/file permissions are honored in both traditional and API Rails controllers.  While positive tests for normal access are in place, negative tests ensuring that unauthorized users cannot perform edit actions on resources they do not have access to were missing.  This covers accessing & editing both studies and files at the integration level.

This update also deals an issue where `delayed_job` was not processing some background jobs quickly enough in tests that caused false negative results.  The `sleep_delay` has been lowered, and some unnecessary assertions were removed.  Also, `test/api/site_controller_test.rb` has been renamed to `test/api/api_site_controller_test.rb` as it shared a class name with `test/controllers/site_controller_test.rb` that caused both test suites to run concurrently and caused additional false negative results.

This PR satisfies SCP-3075.

### DIFF
--- a/config/initializers/delayed_job_config.rb
+++ b/config/initializers/delayed_job_config.rb
@@ -6,7 +6,7 @@ Delayed::Worker.logger = Logger.new(File.join(Rails.root, 'log', "delayed_job.#{
 Delayed::Worker.default_queue_name = :default
 
 if Rails.env.test? || Rails.env.development? # save a little time in testing/dev
-  Delayed::Worker.sleep_delay = 5
+  Delayed::Worker.sleep_delay = 5 # seconds
 end
 
 # Fixing class loader issues with delayed_job, for Rails 5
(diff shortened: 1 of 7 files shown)


config/initializers/delayed_job_config.rb
```diff
@@ -6,7 +6,7 @@
 Delayed::Worker.default_queue_name = :default
 
 if Rails.env.test? || Rails.env.development? # save a little time in testing/dev
-  Delayed::Worker.sleep_delay = 10
+  Delayed::Worker.sleep_delay = 5
```
eweitz, line 9: For rusty or unacquainted maintainers:

```suggestion
  Delayed::Worker.sleep_delay = 5 # minutes
```

eweitz (approved):
Looks good!  I suggested a trivial refinement for maintainability.

config/initializers/delayed_job_config.rb
bistline: It's actually seconds, so I'll add that for clarity.

knapii-developments (approved):

PR #891, opened 2021-02-10
Restoring single study bulk download of DirectoryListings (SCP-3096)

bistline:
Previously, when doing a single-study bulk download request (not via advanced search), the `curl` command would also include any extra directories of data in the study that were registered via `DirectoryListing` objects.  These are only created via sync, and represent entire folders of files that are all of the same type (such as fastq, BAM, or text files) and have at least 10 entries.  Since the refactoring of bulk download into `BulkDownloadService`, this functionality has been lost, and only `StudyFiles` can be downloaded.

This update restores the ability to download _all_ data from a single study via the bulk download modal, or individual folders of files, if present.  Bulk downloads via advanced search will not include these directories, as the potential for exceeding the user's download quota would be quite high, and would fail the entire request before downloading.  In addition, this also adds `DirectoryListing` objects to the `file_supplemental_info.tsv` manifest, as `DirectoryListing` objects can have an associated `Taxon` if the folder contains sequence data.

This PR satisfies SCP-3096.

### DIFF
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -51,6 +51,46 @@ module Api
         @studies = Study.viewable(current_api_user)
       end
 
+      swagger_schema :DirectoryListingDownload do
+        property :name do
+          key :type, :string
+          key :description, 'Name of remote GCS directory containing files'
+        end
+        property :description do
+          key :type, :string
+          key :format, :email
+          key :description, 'Block description for all files contained in DirectoryListing'
+        end
+        property :file_type do
+          key :type, :string
+          key :description, 'File type (i.e. extension) of all files contained in DirectoryListing'
+        end
+        property :download_url do
+          key :type, :string
+          key :description, 'URL to bulk download all files in this directory'
+        end
+        property :files do
+          key :type, :array
+          key :description, 'Array of file objects'
+          items type: :object do
+            key :title, 'GCS File object'
+            key :required, [:name, :size, :generation]
+            property :name do
+              key :type, :string
+              key :description, 'name of File'
+            end
+            property :size do
+              key :type, :integer
+              key :description, 'size of File'
+            end
+            property :generation do
+              key :type, :string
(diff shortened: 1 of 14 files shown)


app/controllers/api/v1/site_controller.rb
```diff
@@ -51,6 +51,46 @@ def studies
         @studies = Study.viewable(current_api_user)
       end
 
+      swagger_schema :DirectoryListingDownload do
+        property :name do
+          key :type, :string
+          key :description, 'Name of remote GCS directory containing files'
```
eweitz, line 57: This acronym is worth expanding on first use in an endpoint:

```suggestion
          key :description, 'Name of remote Google Cloud Storage (GCS) directory containing files'
```

knapii-developments (approved):
Thanks for walking me through this

db/seeds.rb
```diff
@@ -125,6 +125,14 @@
 gene2_cells = gene_2.data_arrays.build(name: gene_2.cell_key, array_type: 'cells', cluster_name: expression_file.name,
                                        array_index: 1, study_id: study.id, study_file_id: expression_file.id, values: all_cell_array)
 gene2_cells.save!
+DirectoryListing.create!(name: 'fastq', file_type: 'fastq',
```
eweitz, line 128: Nice touch.  Having a directory listing added via synthetic study populator would also be helpful, e.g. for trivially setting up manual testing in the UI.

eweitz (approved):
Looks good.  

A few readability nits, noted below, seem worth refining.  My note about testability seems is a bit less trivial, and less important.

db/seeds.rb
bistline: We have a [ticket](https://broadworkbench.atlassian.net/browse/SCP-3061) in the backlog to transition off of `db/seeds.rb` and use `FactoryBot` for seeding test data.  I would argue that whenever we start this work, we can address this more thoroughly.

PR #896, opened 2021-02-18
Better error handling re: uploading Coordinate Labels files (SCP-3101)

bistline:
Coordinate Labels files are files that add text annotations to Cluster files inside the plot area, usually to annotate a specific population inside a scatter plot.  These files cannot be uploaded until the user has added at least one cluster file, and the association is managed through the `StudyFileBundle` class.  However, there is a corner case where the user can add the cluster file, which will then begin an ingest run, and then add the coordinate labels file.  If that cluster file fails its ingest run before the coordinate labels file has finished uploading, it will fail to launch a parse job and become orphaned, as the "parent" of the `StudyFileBundle` that was created was removed.  This then causes two errors: first, the upload wizard or sync page cannot be rendered due to the missing association (trying to reference `id` on a `nil` object), and the coordinate labels file cannot be parsed or deleted, and requires manual intervention at the database level to resolve.

This update fixes the error that prevents forms from rendering, and adds better error handling to detect and rescue from this corner case.  The update flow will now catch this potential error and remove the coordinate labels file, and then show the following modal to the user to notify them of the issue:

automatic_bundle_delete

Also, the `FileParseServiceTest` suite was updated to test this new integration, and was generally refactored for better maintainability by using `FactoryBot`.

This PR satisfies SCP-3101

### DIFF
--- a/app/models/study_file_bundle.rb
+++ b/app/models/study_file_bundle.rb
@@ -10,6 +10,8 @@ class StudyFileBundle
 
   # allowed bundle types
   BUNDLE_TYPES = ['MM Coordinate Matrix', 'BAM', 'Cluster']
+  # file types that must be bundled to be valid
+  REQUIRE_BUNDLE = ['MM Coordinate Matrix', 'BAM', 'BAM Index', 'Coordinate Labels', '10X Genes File', '10X Barcodes File']
   # required keys for file list
   FILE_LIST_KEYS = %w(name file_type)
   # child file type requirements by bundle type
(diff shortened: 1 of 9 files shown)


app/models/study_file_bundle.rb
```diff
@@ -18,6 +20,13 @@ class StudyFileBundle
       'BAM' => ['BAM Index'],
       'Cluster' => ['Coordinate Labels']
   }
+  # parent file type requirements by child file type
+  CHILD_REQUIREMENTS = {
```
devonbush, line 24: This is a tricky variable to name - CHILD_PARENT_REQUIREMENT is a bit long, but might better clarify that the keys are child types, and the values are a single parent type.

test/factories/study.rb
```diff
@@ -25,6 +25,7 @@
     # create a study but mark as detached, so a Terra workspace is not created
     factory :detached_study do
       detached { true }
+      bucket_id { SecureRandom.alphanumeric(16) }
```
devonbush, line 28: can you add a comment here for why it's better to have a filler value here than leaving it null?

devonbush (approved):
This is good stuff, just a couple of minor non-blocking suggestions

knapii-developments (approved):

test/factories/study.rb
bistline: Sure - it was throwing an error in one of my new tests because the mock was trying to access that value, and the `nil` value cause a `MockExpectationError`.

app/models/study_file_bundle.rb
bistline: Yeah I went back and forth on it for a while.  Concision seemed the better route to take.  I can update the comment for more clarity.

test/factories/study.rb
bistline: 7b97748

app/models/study_file_bundle.rb
bistline: 7b97748

PR #897, opened 2021-02-19
Explore react refactor

devonbush:
In my mind, the key 5 files are

* ExploreView.js.    Handles the layout of the explore tab and associated controls
* ExploreTabRouter.js.  The biggest functional change this refactor brings is linkable visualizations -- if the user changes which cluster/annotation they are viewing, that now propagates to the URL.  And this file handles the logic for doing that mapping of URL to props. 
* ExploreDisplayTabs.js   This handles taking all the user input and high-level study information and figuring out the graph layout
* ClusterControls.js. This handles the most important user input--which clusters and annotations they are currently viewing
* ScatterPlot.js   -- Takes a bunch of props, and then handles going out to the server and getting the scatter plot data to render, and then rendering it with plotly.  (there are other files like this for other visualizations too)

The problem that occupied most of my mental energy and caused the most anguish was handling showing the "default cluster" for a given study.  The issue is that when a user navigates to the study page, we don't know yet what the default cluster is.  So the 'cluster' property is initially blank.  Then the ExploreView component will get a response back from the server saying "here is all the cluster info for the study". Meanwhile, in parallel, the ScatterPlot component has fired a request to the server saying 'just give me the default graph".  And figuring out the right way to get these components in sync, without causing re-renders, was a huge pain for me.


devonbush:
> I don't see how this could work, since I believe removeEventListener requires object identity.

Thanks for this, and the comment above -- you were right, this was not working.  I also realized that wrapping this in a useEffect was a much more appropriate approach to ensure the listener only got attached once.

app/models/study_file.rb
```diff
@@ -83,6 +83,8 @@ class StudyFile
   field :z_axis_min, type: Integer
   field :z_axis_max, type: Integer
   field :is_spatial, type: Boolean, default: false
+  # for spatial files, the ids of cluster files that correspond to this file for default display
+  field :spatial_cluster_associations, type: Array, default: []
```
bistline, line 87: I feel like there's a more robust way of managing this association, but given since they're both `StudyFile` instances of type `Cluster`, nothing is coming to mind.  But we should ruminate on this in the hopes that something may shake out.

app/controllers/reports_controller.rb
```diff
@@ -15,163 +15,7 @@ class ReportsController < ApplicationController
   # Calculations in this method must be done reasonably performantly to avoid server timeouts
   # This could eventually be converted to individual REST endpoints for each stat group
   def index
```
bistline, line 17: If we have indeed deprecated this, should we remove it altogether?

bistline (approved):
No blocking edits, mostly just some questions about how things are structured.  I'd love a more in-depth walkthrough of the React changes as they're a little over my head.

test/js/explore/explore-tab-router.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    updateDataParams({ annotation: { name: 'bar2', type: 'numeric', scope: 'user' } })
+    expect(routerNav).toHaveBeenLastCalledWith('?cluster=foo&annotation=bar2--numeric--user#study-visualize', { replace: true })
+  })
+
+  /** This test validates that we are parsing data params on URL links in a certain way
+    * This is IMPORTANT to make sure that we maintain URL consistency over time.
+    * e.g. if a user shares a ink to a certain cluster visualization, we want to be SURE
+    * that link stays stable.
+    * So before making any updates to this test, please consult Devon and/or Vicky about the possible implications
```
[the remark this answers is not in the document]
dmohs: Think of this code as a precisely-specified solution to a problem, infer the problem, then suggest an alternative solution. For even better clarity, start by describing the problem as inferred, which will help establish a baseline understanding.

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -73,9 +126,143 @@ export default function StudyViolinPlot({ study, genes, renderParams, setAnnotat
...
-    </>
+    </div>
+  )
+}
+
+
+/** Formats expression data for Plotly, draws violin (or box) plot */
+function renderViolinPlot(target, results, { plotType, showPoints }) {
+  const traceData = getViolinPropsWrapper(results, plotType, showPoints)
+  const expressionData = [...traceData[0]]
```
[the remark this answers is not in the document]
dmohs: This is an important philosophical point on which we should align as a team. My preference is to assume proficiency and factor that in when onboarding people. I judge this needs to be a faces-to-faces discussion.

app/controllers/reports_controller.rb
devonbush: ah, it's now restored thanks to a revised plotly import

devonbush:
Going to go ahead and merge this -- I recognize there are still outstanding comments -- they're not being ignored!  It just seems like we should have that discussion about the right amount of ES2018 to use, and how, before doing those, and I don't want to hold up this PR (which Jean is waiting on so she can do user interviews) for it

devonbush:
test failures are the dreaded getcwd and the flaky params test Jon is fixing

PR #898, opened 2021-02-22
Pair DirectoryListing of sequence index files with parent (SCP-3097)

bistline:
Currently, `DirectoryListing` objects only pair together files of 10 or more with the exact same file extension into a single list.  This is primarily to allow downloads of individual folders of like files.  However, if a user has created a folder of BAM files and their indexes (.bai), SCP will not allow users to pair these together, and will attempt to create two separate `DirectoryListing` objects.  While this will not break when saving, it will cause bulk downloads of those folders to clobber each other, meaning a user downloading "All Data" will only get one of those folders (whichever is downloaded last).

This update now allows BAM and index files to pair together in a single `DirectoryListing` that will set the file type to the parent (BAM).  In addition, new unit tests are added for coverage of relevant methods in the `DirectoryListing` class, as none were present before.

Also, this update fixes a bug with downloading an individual directory that has underscores "`_`" in the name.  Previously, the download modal would not update with the `curl` command to download that directory due to a parsing issue in JS.

This PR satisfies SCP-3097.

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -260,8 +260,8 @@ class StudiesController < ApplicationController
     @unsynced_directories = @study.directory_listings.unsynced
 
     # split directories into primary data types and 'others'
-    @unsynced_primary_data_dirs = @unsynced_directories.select {|dir| dir.file_type == 'fastq'}
-    @unsynced_other_dirs = @unsynced_directories.select {|dir| dir.file_type != 'fastq'}
+    @unsynced_primary_data_dirs = @unsynced_directories.select {|dir| DirectoryListing::PRIMARY_DATA_TYPES.include?(dir.file_type)}
+    @unsynced_other_dirs = @unsynced_directories.select {|dir| !DirectoryListing::PRIMARY_DATA_TYPES.include?(dir.file_type)}
 
     # now determine if we have study_files that have been 'orphaned' (cannot find a corresponding bucket file)
     @orphaned_study_files = @study_files - @synced_study_files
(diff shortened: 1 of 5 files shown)


app/controllers/studies_controller.rb
```diff
@@ -1223,19 +1223,18 @@ def process_workspace_bucket_files(files)
...
       else
-        # we are now dealing with singleton files or fastqs, so process accordingly (making sure to ignore directories)
-        if DirectoryListing::PRIMARY_DATA_TYPES.any? {|ext| file_extension.include?(ext)} && !file.name.end_with?('/')
-          # process fastq file into appropriate directory listing
-          process_directory_listing_file(file, 'fastq')
+        # we are now dealing with singleton files or sequence data, so process accordingly (making sure to ignore directories)
+        if DirectoryListing::PRIMARY_DATA_TYPES.any? {|ext| file_type.include?(ext)} && !file.name.end_with?('/')
+          # process sequence data file into appropriate directory listing
+          process_directory_listing_file(file, file_type)
         else
```
devonbush, line 1238: This service and the one above seem to share a lot of code.  Is it worth pulling some of this into the Directory Listing model?  Or maybe a DirectoryListingService?  (or maybe even a more generic StudyService)

devonbush (approved):
One non-blocking suggestion.  Good stuff!

app/controllers/studies_controller.rb
bistline: What I'd actually like to do is this: https://broadworkbench.atlassian.net/browse/SCP-3121

Basically, refactor all code that is "sync" related into a service and DRY up everything else, much like we did with the FileParseService and the various viz-related services.  It would improve both test coverage and maintainability.

knapii-developments (approved):

PR #899, opened 2021-02-25
Fixing gene list parsing, adding Mixpanel logging, code cleanup (SCP-2966)

bistline:
Gene list file parsing has been broken since #743 due to a broken reference to an object when the parse methods were refactored out of the `Study` model and into `ParseUtils`.  This update fixes the issue, and also adds reporting to Mixpanel for the two remaining file parse jobs that still run on the server: Coordinate Labels and Gene Lists.  These will show up in Mixpanel now under the `rails-parse` event, since these are not ingest pipeline jobs.

A preview of what the `rails-parse` event looks like in Mixpanel:

gene_list_job

In addition, this update removes all deprecated parsers/methods from `ParseUtils` that have been superseded by scp-ingest-pipeline.

This PR satisfies SCP-2966.

### DIFF
--- a/app/models/parse_utils.rb
+++ b/app/models/parse_utils.rb
@@ -1,222 +1,6 @@
 require 'rubygems/package' # for tar reader
 class ParseUtils
 
-  # parse a 10X gene-barcode matrix file triplet (input matrix must be sorted by gene indices)
-  def self.cell_ranger_expression_parse(study, user, matrix_study_file, genes_study_file, barcodes_study_file, opts={})
-    begin
-      start_time = Time.zone.now
-      # localize files
-      Rails.logger.info "Parsing gene-barcode matrix source data files for #{study.name} with the following options: #{opts}"
-      study.make_data_dir
-      Rails.logger.info "Localizing output files & creating study file entries from 10X CellRanger source data for #{study.name}"
-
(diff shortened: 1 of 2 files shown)


devonbush (approved):
This looks good.  I think there are a couple places where the code might be cleaner by using https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html   (and it might be informative to report both wall-clock and cpu clock time), but that's definitely not critical

bistline:
> This looks good. I think there are a couple places where the code might be cleaner by using https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html instead of class @start_time variables (and it might be informative to report both wall-clock and cpu clock time), but that's definitely not critical

That's an interesting thought - we could refactor it so that the parser just handles the actual parse, and we could have a wrapper method that would handle emails/reporting/error handling.  I'll take a look at that.

bistline:
@devonbush after looking at it, it would probably be more trouble than it's worth for these remaining parse methods.  They're not frequently used, and usually run on the order of seconds.  But in order to use `Benchmark.measure` we'd have to restructure the methods to not encompass error handling/reporting so that they can be called in a block, and at that point we're basically rewriting everything.

knapii-developments (approved):

PR #901, opened 2021-02-26
Setting default empty values for non-existent tokens (SCP-3127)

bistline:
This update adds a default token response (empty hash: `{}`) to `User#valid_access_token` to prevent `NoMethodError` exceptions when trying to generate an OAuth token for a user.  This can happen if a user is signed into the portal, and either their session times out, or a deployment happens while they still have a valid session open.  Both cases result in their cached tokens being deleted, which can cause issues if a request is attempted before they re-authenticate.

This update also refactors `test/models/user_test.rb` to use `FactoryBot` for test maintainability.

This PR satisfies SCP-3127.


devonbush (approved):
nice

knapii-developments (approved):

PR #903, opened 2021-03-02
Limit size of error logfile in admin emails (SCP-3106)

bistline:
When `IngestJob` PAPI runs fail, one of the QA features is that SCP sends an email to the "QA Dev Email" list with the contents of the error log from the ingest pipeline job.  However, in certain instances this logfile can hundreds of MB (if the job was for an expression matrix, and it fails near the end of the run after putting all of the gene names parsed into the log), which will cause Sendgrid to throw an error and not deliver the email.  This reduces visibility of parsing issues for admins as error reports are suppressed.

This update places a cap of 1 MB on logfile contents transcribed into emails to avoid this issue.  The full log still persists in the study bucket at the same location, and admins can still access it from the link in the email.

This PR satisfies SCP-3106.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -502,14 +502,16 @@ class IngestJob
   # * *params*
   #   - +filepath+ (String) => relative path of file to read in bucket
   #   - +delete_on_read+ (Boolean) => T/F to remove logfile from bucket after reading, defaults to true
+  #   - +range+ (Range) => Byte range to read from file
   #
   # * *returns*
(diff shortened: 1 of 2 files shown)


knapii-developments (approved):

bistline:
Marking test failure as a false negative, and opened https://broadworkbench.atlassian.net/browse/SCP-3138 to investigate further.  All other tests passed except this one, which fails often without explanation.

devonbush (approved):

PR #905, opened 2021-03-02
Adding IGV to react explore tab

devonbush:


app/javascript/components/explore/GenomeView.js
```diff
@@ -0,0 +1,251 @@
...
+function getBamTracks(bamAndBaiFiles) {
+  let bam; let bamTrack; let i; let bamFileName
+
+  const bamTracks = []
+
+  for (i = 0; i < bamAndBaiFiles.length; i++) {
+    bam = bamAndBaiFiles[i]
+
+    // Extracts BAM file name from its GCS API URL
+    bamFileName = bam.url.split('/o/')[1].split('?')[0]
```
bistline, line 113: Since we have control over what gets put in `bamAndBaiFiles` from `Study#get_bam_files`, why not add a field called `label` so we don't have to do this?

app/lib/synthetic_study_populator.rb
```diff
@@ -77,67 +80,97 @@ def self.add_files(study, study_config, synthetic_study_folder, user)
...
-            is_raw_counts: exp_finfo_params['is_raw_counts'] ? true : false,
-            units: exp_finfo_params['units'],
-            biosample_input_type: exp_finfo_params['biosample_input_type'],
-            library_preparation_protocol: exp_finfo_params['library_preparation_protocol'],
-            modality: exp_finfo_params['modality']
-          )
-          study_file_params['expression_file_info'] = exp_file_info
-        end
-      end
+      study_file_params.merge!(process_genomic_file_params(study, finfo))
```
bistline, line 85: Much nicer.

app/lib/synthetic_study_populator.rb
```diff
@@ -77,67 +80,97 @@ def self.add_files(study, study_config, synthetic_study_folder, user)
...
+  def self.process_coordinate_file_params(study, file_info)
     params = {}
     if !file_info['is_spatial'].nil?
       params[:is_spatial] = file_info['is_spatial']
     end
+    if file_info['spatial_cluster_associations'].present?
+      # look up the ids for the associations
+      # note that this requires the associated file to have already been added
+      params[:spatial_cluster_associations] = file_info['spatial_cluster_associations'].map do |cluster_file_name|
+        StudyFile.find_by!(study: study, name: cluster_file_name).id.to_s
```
bistline, line 166: This would fail gracefully and therefore avoid the issue of the file not being present yet
```suggestion
        StudyFile.find_by!(study: study, name: cluster_file_name).try(:id).to_s
```

app/models/study.rb
```diff
@@ -1400,6 +1400,7 @@ def get_bam_files
 
     bam_files.each do |bam_file|
       bams << {
+          'name' => bam_file.name,
```
bistline, line 1403: Ah - that would address my earlier concern about `label`.

bistline (approved):
Some small non-blocking edits for maintainability.

app/javascript/components/explore/GenomeView.js
devonbush: hah, yup, I added the field, and then just didn't use it. Good catch

app/lib/synthetic_study_populator.rb
devonbush: I actually want this to fail hard -- the particular way the populator assembles bundle params requires the cluster file to be defined before the spatial one is.

bistline: Fair point.

PR #908, opened 2021-03-05
implementing gene list file upload (SCP-3111)

devonbush:
To test:
 0. Set the 'react_explore' feature flag to true
 1. Visit any study with gene expression data (and the react-explore feature flag set to true).   
 2. go to the explore tab
 3. click the 'upload' icon next to the gene search text box icon
 4. Select a file containing a list of genes.  (e.g. test/test_data/test_gene_list.txt)
 5. confirm the gene search is updated and the genes are shown in the search box
 (note that there is a separate ticket to handle the issue of multiple genes overflowing the search box)

### DIFF
--- a/app/javascript/components/explore/StudyGeneField.js
+++ b/app/javascript/components/explore/StudyGeneField.js
@@ -29,10 +29,7 @@ export default function StudyGeneField({ genes, searchGenes, allGenes }) {
 
   let enteredGeneArray = []
   if (genes) {
-    enteredGeneArray = genes.map(geneName => ({
-      label: geneName,
-      value: geneName
-    }))
+    enteredGeneArray = getOptionsFromGenes(genes)
   }
 
   /** the search control tracks two state variables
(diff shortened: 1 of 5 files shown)


app/javascript/components/explore/StudyGeneField.js
```diff
@@ -80,48 +77,60 @@ export default function StudyGeneField({ genes, searchGenes, allGenes }) {
     }
   }
 
+  /** handles a user selecting a gene list file to use */
+  function readGeneListFile(file) {
+    const fileReader = new FileReader()
+    fileReader.onloadend = () => {
+      const newGenes = fileReader.result.trim().split(/[\s,]+/)
```
bistline, line 84: I just tested, so I know it works without it, but I'm surprised we don't need to call `split()` with `\n` here.

bistline (approved):

knapii-developments (approved):

app/javascript/components/explore/StudyGeneField.js
devonbush: `\s` gets whitespace characters, so it includes  `\n`

PR #910, opened 2021-03-05
Fixing parse launching logic for expression matrices (SCP-3150, SCP-3151)

bistline:
This update addresses issues related to parsing expression matrix files:

SCP-3150: When attempting to sync an MTX file, `FileParseService` will erroneously attempt to push the MTX file back to the bucket when in fact it is already there.  This update will skip this step unless the file is local to the VM, which does not happen during a sync.  This prevents unnecessary admin error emails from being generated that potentially could cause issues if acted upon, as the parse job does initiate.

SCP-3151: When determining whether an expression matrix (MTX or dense) can launch an ingest job, if an upstream matrix (i.e. created before the one we are currently checking) is still marked as "parsing", it will defer launching the job until that matrix has either validated (by which there are cells & at least one gene present), or has finished parsing altogether.  If that upstream matrix never initiated a parse job due to an uncaught exception, all expression matrix files will queue indefinitely for that study until the blocking matrix is either removed, parsed, or marked as "parsed".  This update adds a maximum queueing period of 24 hours, after which a "stuck parsing" matrix will no longer prevent newer matrices from launching ingest runs.

This PR satisfies SCP-3150 and SCP-3151.

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -58,7 +58,7 @@ class FileParseService
                               persist_on_fail: persist_on_fail)
           job.delay.push_remote_and_launch_ingest
         else
-          study.delay.send_to_firecloud(study_file)
+          study.delay.send_to_firecloud(study_file) if study_file.is_local?
           return self.missing_bundled_file(study_file)
         end
       when /10X/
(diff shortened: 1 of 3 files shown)


devonbush (approved):

bistline:
Same failed test for character encoding, false negative.

knapii-developments (approved):

PR #911, opened 2021-03-05
Adding mixpanel to plotly legend clicks

devonbush:
To test:
 1. go to explore tab
 2. click and/or double-click a legend entry of a scatter plot
 3. confirm mixpanel event with appropriate name is sent


bistline (approved):

bistline (approved):
Non-blocking comment:

This works fine and can be merged.  However, I'm noting that in Mixpanel, it does not seem possible to group all actions together.  Rather, each individual "tag" is its own island.  You can look at `click:link` or `click:button`, but not `click`.

I'm not sure if this is something that can be accounted for in Mixpanel with more configuration, but it does seem like an oversight in how we categorize events.

bistline:
Test failure is a false negative (related to file I/O issues in test)

knapii-developments (approved):

PR #915, opened 2021-03-08
Subsampling in explore tab no longer defaults to 'all cells' (SCP-3144)

devonbush:
This now defaults to the minimum subsample -- is that right?  Adding @jlchang  to the review to confirm the default and also the tooltip content (shown below)

Also added a tooltip based on our existing one 
image


bistline:
@devon The default is actually 10K cells if the cluster has more than 10K, otherwise it is 'All Cells'.

devonbush:
Just updated to match @bistline 's comment

devonbush:
test failures are unrelated

bistline (approved):

jlchang (approved):
Thanks to Devon for the walkthrough!

PR #917, opened 2021-03-10
turning on caching for the explore controller API (SCP-3144)

devonbush:


devonbush:
ah, good call.  Let's table this for now, since it's just a performance optimization, and not needed for loading the default study cluster viz

PR #918, opened 2021-03-10
Restoring gene lists in React explore tab (SCP-3110)

bistline:
This update restores the feature for visualizing gene lists, which are study-owner provided "precomputed analyses" that act like expression matrices, where the genes and samples (e.g. columns) are all predefined, and the user cannot load other genes or annotations.  These are visualized as heatmaps.

Sample visualization:

Screen Shot 2021-03-10 at 12 52 12 PM

In addition, this update also adds some additional visualization endpoints to the API, and some missing endpoints to the Swagger documentation.  There are also some test fixes that were discovered due to an update to `yarn` and `node` that caused compilation/test errors.

To test, you will need to first parse a gene list file for a study.  There are samples in `test/test_data` called `marker_{n}_gene_list.txt`.  Additionally, you will need to set the `FeatureFlag` for `react_explore` to `true` for your user account.  After that is completed:

1. Navigate to the study overview page with the gene list
2. Select a gene list from the "Gene Lists" select menu (under subsampling)

This PR satisfies SCP-3110


app/controllers/api/v1/api_docs_controller.rb
```diff
@@ -113,6 +113,7 @@ class ApiDocsController < ActionController::Base
           Api::V1::SearchController,
           Api::V1::Visualization::ClustersController,
           Api::V1::Visualization::AnnotationsController,
+          Api::V1::Visualization::ExpressionController,
```
devonbush, line 116: previously, we had excluded this from api docs since these endpoints were expected to be volatile and change a lot as, e.g. we make performance improvements, and so we didn't want people scripting against it.  
If we're adding it back, we should add a comment in the docs about that volatility

app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -19,6 +19,81 @@ class ExpressionController < ApiBaseController
         # by the SCP UI.
         # We agreed that there would be no swagger docs for this endpoint, as it is not intended
```
devonbush, line 20: see comment above about lack of swagger docs

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -46,13 +46,8 @@ export default function ExploreDisplayTabs(
...
 ) {
   const [, setRenderForcer] = useState({})
-  const isMultiGene = exploreParams.genes.length > 1
-  const isGene = exploreParams.genes.length > 0
   const plotContainerClass = 'explore-plot-tab-content'
-  const hasSpatialGroups = exploreInfo && exploreInfo.spatialGroups.length > 0
-  const hasGenomeFiles = exploreInfo && exploreInfo.bamBundleList.length > 0
-  const enabledTabs = getEnabledTabs(isGene, isMultiGene, hasSpatialGroups, !!exploreParams.consensus, hasGenomeFiles)
-
+  const {enabledTabs, isGeneList, isGene, isMultiGene} = getEnabledTabs(exploreInfo, exploreParams)
```
devonbush, line 50: oh yeah, that looks much nicer :)

app/javascript/components/explore/ExploreView.js
```diff
@@ -27,7 +28,7 @@ function RoutableExploreTab({ studyAccession }) {
   const [exploreInfo, setExploreInfo] = useState(null)
   // tracks whetehr the view options controls are open or closed
   const [showViewOptionsControls, setShowViewOptionsControls] = useState(true)
-  // whether the user is in lass-select mode for selecting points for an annotation
+  // whether the user is in lasso-select mode for selecting points for an annotation
```
devonbush, line 31: heh

app/javascript/components/explore/ExploreView.js
```diff
@@ -82,19 +83,27 @@ function RoutableExploreTab({ studyAccession }) {
...
+
     if (newParams.cluster && !newParams.spatialGroups) {
       newParams.spatialGroups = getDefaultSpatialGroupsForCluster(newParams.cluster, exploreInfo.spatialGroups)
     }
     // if the user updates any cluster params, store all of them in the URL so we don't end up with
     // broken urls in the event of a default cluster/annotation changes
-    const updateParams = {}
+    // also, unset any gene lists as we're about to re-render the explore tab and having gene list selected will show
+    // the wrong tabs
+    const updateParams = {geneList: ''}
```
devonbush, line 94: nice

test/js/explore/explore-display-tabs.test.js
```diff
@@ -0,0 +1,281 @@
...
+
+jest.mock('components/visualization/RelatedGenesIdeogram', () => {
+  return {
+    Ideogram: jest.fn(() => mockPromise)
+  }
+})
+
+import { getEnabledTabs } from 'components/explore/ExploreDisplayTabs'
+
+// function to wrap test assertions
```
devonbush, line 16: You don't need to do property-by-property comparison -- you can just do
`expect(expectedResults).toEqual(getEnabledTabs(...))`
and that's probably even better because it will fail if there are stray unexpected properties

test/js/explore/explore-display-tabs.test.js
```diff
@@ -0,0 +1,281 @@
...
+
+    const expectedResults = {
+      enabledTabs: ['scatter', 'distribution', 'dotplot'],
+      isGeneList: false,
+      isGene: true,
+      isMultiGene: true
+    }
+
+    assertTabConfig(exploreInfo, exploreParams, expectedResults)
+  })
```
devonbush, line 280: great tests!

devonbush (changes requested):
overall great stuff.  just a couple things on the swagger docs and test styling

app/controllers/api/v1/api_docs_controller.rb
bistline: I mainly added this for debugging purposes while I was doing this work as it made it easier for me to iterate.  I'm happy to add the comment though, unless we really feel that these shouldn't be exposed in Swagger.

test/js/explore/explore-display-tabs.test.js
bistline: Ah - you're right.  That will actually test the interface better.

bistline: 1926bcf

app/controllers/api/v1/api_docs_controller.rb
bistline: 1926bcf

app/javascript/components/visualization/controls/GeneListSelector.js
```diff
@@ -0,0 +1,48 @@
...
+const noneSelected = "None selected..."
+
+/** takes the server response and returns subsample options suitable for react-select */
+function getGeneListOptions(studyGeneLists) {
+  let geneListObjects = [{label: noneSelected, value: ''}]
+  geneListObjects = geneListObjects.concat(studyGeneLists.map(geneList => {
+    return { label: geneList, value: geneList }
+  }))
+  return geneListObjects
+}
```
dmohs, line 16: Consider:
```javascript
function getGeneListOptions(studyGeneLists) {
  const lAndV = x => ({label: x, value: x})
  return [{label: noneSelected, value: ''}].concat(studyGeneLists.map(lAndV))
}
```

bistline: Ooh... fancy.  Functional programming to the rescue...

bistline: 75a566f

jlchang (approved):
Checked functionality in Devon's dev environment - looks great!

app/controllers/api/v1/api_docs_controller.rb
devonbush: works for me

devonbush (approved):

PR #920, opened 2021-03-11
Fixing heatmap layout/fit options (SCP-3145, SCP-3146)

bistline:
This update fixes both the general layout of the Morpheus heatmap visualization (now fits inside bounds of parent div), and restores the functionality of the heatmap "Fit options" select menu.  This update only applies to the React-based Explore tab.

To test (must have `react_explore` FeatureFlag set to `true` for your account):
1. Perform multi-gene search in a study
2. Click 'Heatmap' tab
3. Use the "Fit options" dropdown to change heatmap layout (selecting "None" will reset the zoom to defaults)

Examples:

Default zoom:
heatmap_default

Column fit (condenses all columns into viewport):
heatmap_col_fit

This PR satisfies SCP-3145 & SCP-3146.

### DIFF
--- a/app/javascript/components/visualization/DotPlot.js
+++ b/app/javascript/components/visualization/DotPlot.js
@@ -157,7 +157,7 @@ export function morpheusTabManager($target) {
     setTabTitle: () => {},
     setActiveTab: () => {},
     getWidth: () => $target.actual('width'),
-    getHeight: () => $target.height(),
+    getHeight: () => $target.actual('height'),
     getTabCount: () => 1
   }
 }
--- a/app/javascript/components/visualization/DotPlot.js
+++ b/app/javascript/components/visualization/DotPlot.js
@@ -27,7 +27,7 @@ export const dotPlotColorScheme = {
   */
 export default function DotPlot({
   studyAccession, genes=[], cluster, annotation={},
-  subsample, annotationValues, dimensions
+  subsample, annotationValues
 }) {
   const [graphId] = useState(_uniqueId('dotplot-'))
   const expressionValuesURL = getExpressionHeatmapURL({ studyAccession, genes, cluster })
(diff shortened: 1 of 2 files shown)


app/javascript/components/visualization/DotPlot.js
```diff
@@ -165,7 +156,7 @@ export function morpheusTabManager($target, dimensionsFn) {
     },
     setTabTitle: () => {},
     setActiveTab: () => {},
-    getWidth: () => {return dimensionsFn().width},
+    getWidth: () => $target.actual('width'),
```
devonbush, line 159: should we also use .actual for the line below out of caution?

app/javascript/components/visualization/Heatmap.js
```diff
@@ -81,11 +75,15 @@ export default function Heatmap({
   useUpdateEffect(() => {
     if (morpheusHeatmap.current && morpheusHeatmap.current.fitToWindow) {
       const fit = heatmapFit
-      morpheusHeatmap.current.fitToWindow({
-        fitRows: fit === 'rows' || fit === 'both',
-        fitColumns: fit === 'cols' || fit === 'both',
-        repaint: true
-      })
+      if (fit === '') {
+        morpheusHeatmap.current.resetZoom()
```
devonbush, line 79: nice!  I probably spent 30mins digging through the morpheus code looking for something like this resetZoom function before I gave up

devonbush (approved):
good stuff, just one question

app/javascript/components/visualization/DotPlot.js
bistline: Yes, in fact you are correct.  Calling `actual(dimension)` will get the correct value regardless of whether the element is hidden or not.

app/javascript/components/visualization/Heatmap.js
bistline: Yeah it took a while to find it, and I only did by clicking through the toolbar.

jlchang (approved):
Did a functional review - goods great!

PR #921, opened 2021-03-11
Fixing auto-resize of violin plots (SCP-3157)

devonbush:
This also switches from useLayoutEffect to useEffect.  useLayoutEffect does not have a pause between when the event occurs and the redraw happens, so the UI is frozen during that time.  useEffect allows for a pause so the page can return to interactivity, at the cost of being a little uglier (there can be a noticeable delay between when the user finishes resizing, and when the graph finishes resizing).  I think this is the right tradeoff for our users -- better to have the page be more responsive at the cost of a delayed resize.

This also contains a trivial improvement to downloadProvider so that it doesn't need to reload the download size unless the results have actually changed.

to test:
0. enable react_explore feature flag
1. Open a study and do a gene search
2. go to the distribution tab and observe the violin plot
3. resize your browser window, and observe the violin plot resizes to match


bistline (approved):

jlchang (approved):
Looked great in functional review!

PR #922, opened 2021-03-16
Adding inferCNV/Ideogram visualization to React explore tab (SCP-3113)

bistline:
This updates restores the existing inferCNV/Ideogram visualization to the new React explore tab.  This only applies to studies that have previously run `inferCNV` in their study, and have synced outputs to add the Ideogram Annotations files to their study.  Note: it is possible to upload results to your study bucket and sync the results file by selecting the Ideogram Annotations file type and specifying the correct taxon/assembly.  This visualization can load independent of any clustering/annotations/expression data loaded into a study, and as such does not require any of the other file types to be present to enable this tab (which is now separate from the `genome` tab, which loads IGV only now).

Preview, using sample annotations taken from the [Demo: inferCNV oligodendroglioma](https://singlecell.broadinstitute.org/single_cell/study/SCP384/demo-infercnv-oligodendroglioma#study-visualize) study on production:

infercnv_ideogram

To test (once your user account has the `FeatureFlag` for `react_explore` enabled): 
1. Download the `ideogram_exp_means__Observations--Sample--group--cluster.json` file from [Demo: inferCNV oligodendroglioma](https://singlecell.broadinstitute.org/single_cell/study/SCP384/demo-infercnv-oligodendroglioma#study-visualize)
2. Place the file in a bucket for any study connected to your instance of SCP
3. Go to the "Sync Study" page for that study, and specify the following values for `ideogram_exp_means__Observations--Sample--group--cluster.json` and save:
    1. File type: `Ideogram Annotations`
    2. Species: `human`
    3. Assembly: `GRCh37` **!!Important or Ideogram won't render!!**
4. Click "View Study" at the top of the page
5. Click the "Genome (inferCNV)" tab and verify the output looks like the above screenshot

This PR satisfies SCP-3113

PR #923, opened 2021-03-16
improving error handling for explore tab graphs (SCP-3160)

devonbush:
Ooof, this was tougher than expected...  I created a standardized way of indicating errors in visualizations (see screenshot below), and extended it to our four main visualization types.  This also does some overriding of morpheus error handling so we can show a better error message when the user does a gene search with no results (previously a dialog with 'Error opening https:...' would show, which implied a connection problem, rather than an empty search)

This also ran `yarn lint --fix` to fix some outstanding code format issues

To test:

1. Load a study, with the react_explore feature flag on

2. Do a single gene search for a gene that doesn’t exist (e.g. ‘xxxxxx’).  confirm error message is display, similar to screenshot below

3. Do a multi-gene search where all genes don’t exist (e.g. ‘ddddddd’, ‘dfsadfasdf’).  confirm error message is displayed similar to below.

image

### DIFF
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -105,7 +105,7 @@ module Api
           begin
             viz_data = self.class.get_cluster_viz_data(@study, cluster, params)
           rescue ArgumentError => e
-            render json: {error: e.message}, status: 422 and return
+            render json: {error: e.message}, status: 404 and return
           end
           render json: viz_data
         end
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -102,9 +102,10 @@ module Api
               render json: {error: "No cluster named #{cluster_name} could be found"}, status: 404 and return
             end
           end
-          viz_data = self.class.get_cluster_viz_data(@study, cluster, params)
-          if viz_data.nil?
-            render json: {error: 'Annotation could not be found'}, status: 404 and return
+          begin
+            viz_data = self.class.get_cluster_viz_data(@study, cluster, params)
+          rescue ArgumentError => e
+            render json: {error: e.message}, status: 422 and return
           end
           render json: viz_data
         end
@@ -118,7 +119,7 @@ module Api
                                                        annot_type: annot_params[:type],
                                                        annot_scope: annot_params[:scope])
           if !annotation
-            return nil
+            raise ArgumentError, "Annotation #{annot_params[:annot_name]} could not be found"
           end
 
           subsample = get_selected_subsample_threshold(url_params[:subsample], cluster)
@@ -141,6 +142,10 @@ module Api
               range = ClusterVizService.set_range(cluster, coordinates.values)
             end
           else
+            if genes.count == 0
+              # all searched genes do not exist in this study
+              raise ArgumentError, "No genes in this study matched your search"
+            end
             # For single-gene view of Explore tab (or collapsed multi-gene)
             is_collapsed_view = genes.length > 1 && consensus.present?
             y_axis_title = ExpressionVizService.load_expression_axis_title(study)
--- a/app/controllers/api/v1/visualization/expression_controller.rb
+++ b/app/controllers/api/v1/visualization/expression_controller.rb
@@ -119,7 +119,11 @@ module Api
           subsample = ClustersController.get_selected_subsample_threshold(params[:subsample], cluster)
           genes = RequestUtils.get_genes_from_param(@study, params[:genes])
           if genes.empty?
-            render json: {error: 'You must supply at least one gene'}, status: 422
+            if params[:genes].empty?
+              render json: {error: 'You must supply at least one gene'}, status: 422
+            else
+              render json: {error: 'No genes in this study matched your search'}, status: 422
+            end
           else
             render_data = ExpressionVizService.get_global_expression_render_data(
               study: @study,
--- a/app/controllers/api/v1/visualization/expression_controller.rb
+++ b/app/controllers/api/v1/visualization/expression_controller.rb
@@ -120,9 +120,9 @@ module Api
           genes = RequestUtils.get_genes_from_param(@study, params[:genes])
           if genes.empty?
             if params[:genes].empty?
-              render json: {error: 'You must supply at least one gene'}, status: 422
+              render json: {error: 'You must supply at least one gene'}, status: 400
             else
-              render json: {error: 'No genes in this study matched your search'}, status: 422
+              render json: {error: 'No genes in this study matched your search'}, status: 404
             end
           else
             render_data = ExpressionVizService.get_global_expression_render_data(
--- a/app/javascript/components/explore/ExploreDisplayTabs.js
+++ b/app/javascript/components/explore/ExploreDisplayTabs.js
@@ -47,7 +47,7 @@ export default function ExploreDisplayTabs(
 ) {
   const [, setRenderForcer] = useState({})
   const plotContainerClass = 'explore-plot-tab-content'
-  const {enabledTabs, isGeneList, isGene, isMultiGene} = getEnabledTabs(exploreInfo, exploreParams)
+  const { enabledTabs, isGeneList, isGene, isMultiGene } = getEnabledTabs(exploreInfo, exploreParams)
 
   // exploreParams object without genes specified, to pass to cluster comparison plots
   const referencePlotDataParams = _clone(exploreParams)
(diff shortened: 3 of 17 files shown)


app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -102,9 +102,10 @@ def show
               render json: {error: "No cluster named #{cluster_name} could be found"}, status: 404 and return
             end
           end
-          viz_data = self.class.get_cluster_viz_data(@study, cluster, params)
-          if viz_data.nil?
-            render json: {error: 'Annotation could not be found'}, status: 404 and return
+          begin
+            viz_data = self.class.get_cluster_viz_data(@study, cluster, params)
+          rescue ArgumentError => e
+            render json: {error: e.message}, status: 422 and return
```
bistline, line 108: (puts on HTTP status code hat) `422` is `Unprocessable Entity`, and deals specifically the case when the request had the correct syntax/instructions, but the server could not process them.  I've really only ever seen it applied to a `POST` or `PUT` when saving data, and the parameters supplied cause a validation error, in which case the server responds `422` because the request could not be completed as outlined.

In this particular case, we're trying to find some visualization data via a `GET`, but `ClustersController.get_cluster_viz_data` doesn't find what we asked for.  I would posit then that the correct response code would be `404` or `400`, though it would depend on what the request parameters actually were.  If we were missing a required parameter, `400` would be more correct, but if the sum of all params equates to an empty response, then `404`.

app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -119,7 +119,11 @@ def render_violin
           subsample = ClustersController.get_selected_subsample_threshold(params[:subsample], cluster)
           genes = RequestUtils.get_genes_from_param(@study, params[:genes])
           if genes.empty?
-            render json: {error: 'You must supply at least one gene'}, status: 422
+            if params[:genes].empty?
+              render json: {error: 'You must supply at least one gene'}, status: 422
```
bistline, line 123: This to me a is `400` - we're missing a required parameter.

app/controllers/api/v1/visualization/expression_controller.rb
```diff
@@ -119,7 +119,11 @@ def render_violin
...
+            else
+              render json: {error: 'No genes in this study matched your search'}, status: 422
```
bistline, line 125: Again, I'd opt for `404` here.

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -442,34 +442,32 @@ export default function ExploreDisplayTabs(
...
-    if (isGene) {
-      if (isMultiGene) {
-        if (isConsensus) {
-          enabledTabs = ['scatter', 'distribution', 'dotplot']
-        } else if (hasSpatialGroups) {
-          enabledTabs = ['spatial', 'dotplot', 'heatmap']
-        } else {
-          enabledTabs = ['dotplot', 'heatmap']
-        }
+  } else if (isGene) {
```
bistline, line 454: Probably not in the context of this PR, but I really think we should refactor this method for readability/maintainability.  In context with #922 we now have 8 different tabs that can be rendered at any given time, and the cyclomatic complexity of this is such I had no idea what the logic was going to be here without a serious 10 minutes of inspection and copying snippets out to a text document to assemble what it "would" be after the merge.  This clearly isn't going to scale well, so I think it's time for a refactor.

app/javascript/components/visualization/Heatmap.js
```diff
@@ -98,17 +107,22 @@ export default function Heatmap({
...
       { cluster &&
         <div id={graphId} className="heatmap-graph" style={{ minWidth: '80vw' }}></div> }
       { !cluster && <FontAwesomeIcon icon={faDna} className="gene-load-spinner"/> }
     </div>
   )
 }
 
+
+const SafeHeatmap = withErrorBoundary(Heatmap)
+export default SafeHeatmap
```
bistline, line 120: Interesting - I'd love a dive on `withErrorBoundary` at some point to better understand how it works.

app/javascript/lib/error-message.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+/** handler for morpheus errors that catches the error from an empty gene search to
+    replace it with a more useful display and error message. */
+export function morpheusErrorHandler($target, setShowError, setErrorContent) {
+  return err => {
+    // this is a brittle check, but morpheus doesn't give us any visibility into what error occured,
+    // other than that there was an error processing the results returned
+    if (err[0].includes('genes') && err[0].includes('opening')) {
```
bistline, line 48: I would say that perhaps we should talk to Josh about this - there may be something we're not doing in terms of setup that can assist with this.

bistline (changes requested):
Some quibbles about HTTP status codes, and a question regarding `getEnabledTabs`.  Otherwise looks good - I can tell this wasn't simple.

bistline (approved):

app/javascript/components/visualization/DotPlot.js
```diff
@@ -60,20 +68,25 @@ export default function DotPlot({
...
-        <DotPlotLegend/>
+        { !showError && <DotPlotLegend/> }
       </> }
       { !cluster && <FontAwesomeIcon icon={faDna} className="gene-load-spinner"/> }
     </div>
   )
 }
 
+const SafeDotPlot = withErrorBoundary(DotPlot)
+export default SafeDotPlot
```
eweitz, line 83: Blocker: this naming pattern will confuse maintainers.  

The default export is `SafeDotPlot`, but it's imported as `DotPlot` -- the same name as the "unsafe" function in this module.  To avoid confusion over similar names, please align the imported and exported names -- e.g. name the unwrapped dot plot function "UnsafeDotPlot", then:

```suggestion
const DotPlot = withErrorBoundary(UnsafeDotPlot)
export default DotPlot
```

And similarly for the other visualization components.

This would also make it obvious when using the unsafe version of the function.

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -59,26 +62,28 @@ export default function StudyViolinPlot({
...
-    renderViolinPlot(graphElementId, results, {
-      plotType: distributionPlotToUse,
-      showPoints: distributionPoints
-    })
-    if (setAnnotationList) {
-      setAnnotationList(results.annotation_list)
-    }
-    if (updateDistributionPlot) {
-      updateDistributionPlot(distributionPlotToUse)
+    if (checkScpApiResponse(results, () => Plotly.purge(graphElementId), setShowError, setErrorContent)) {
```
eweitz, line 65: For readability, unpack this condition as in `ScatterPlot.js`.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -33,30 +35,32 @@ export const defaultScatterColor = 'Reds'
...
-    layout.width = width
-    layout.height = height
-    formatMarkerColors(clusterResponse.data, clusterResponse.annotParams.type, clusterResponse.gene)
-    formatHoverLabels(clusterResponse.data, clusterResponse.annotParams.type, clusterResponse.gene)
-    const dataScatterColor = processTraceScatterColor(clusterResponse.data, scatterColor)
-    Plotly.newPlot(graphElementId, clusterResponse.data, layout)
-
-    if (dataScatterColor !== scatterColor) {
-      updateScatterColor(dataScatterColor)
+    if (checkScpApiResponse(clusterResponse, () => Plotly.purge(graphElementId), setShowError, setErrorContent)) {
```
eweitz, line 48: For readability, unpack this condition:

```suggestion
   const apiOK = checkScpApiResponse(clusterResponse, () => Plotly.purge(graphElementId), setShowError, setErrorContent)
    if (apiOK) {
```

app/javascript/lib/error-message.js
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from 'react'
+
+
+/** Hook for enabling the conditional display an error message
+  * in the event of an error
+  * the returned ErrorComponent can be placed in the component's markup, and will be
+  * displayed with the content from setErrorContent if setShowError is called with true
+  */
```
eweitz, line 8: Less runny, and more conventional form:

```suggestion
/** 
  * Hook for enabling the conditional display an error message.
  * In the event of an error, the returned ErrorComponent can be placed in the component's markup, 
  * and will be displayed with the content from setErrorContent if setShowError is called with true.
  */
```

eweitz (changes requested):
I'm a fan of moving towards more inline error UI, as done here!  They're sometimes trickier (as seen in demo), but overall better UI.

Two blockers:

* See comment beginning "Blocker: this naming pattern will confuse maintainers."

* This degrades error observability relative to PR #885.  On staging, do a gene search for "asdfasdf".  Note how an `error-modal` event is logged to Bard / Mixpanel.  No analogous client-side error logging is done here.  Even if the event names differ a bit, we should preserve similar functionality.

eweitz (approved):

PR #925, opened 2021-03-17
adding some missing analytics events

devonbush:
This cleans up a couple places in the explore tab where we weren't tracking events well

To test:
1. open a study with the react_explore feature flag on
2. Open the chrome debugger with the network tab
3. type/select a gene in the search box
4. confirm an event gets posted to bard with name 'input:change' and text `study-gene-search:add`
5. Hit enter or click the search bar. 
6. confirm an event gets posted to bard with name 'click:button' and text 'gene-search-submit'
7. use the 'x' or backspace to remove a gene entry from the search bar
8. confirm an event gets posted to bard with name 'input:change' and text `study-gene-search:remove`

PR #928, opened 2021-03-17
Gene search analytics and UX fixes (SCP-3147)

devonbush:
This has a quick-and-dirty fix for handling too many genes in the bar
image
The scrollbar isn't great, but it's better than the previous behavior of overflowing into the graphs (and arguably no worse than the current too-small text field)
image

This also improves handling of copy-pasted gene lists into the bar, as well as giving even more detail on events

To test:
1. open a study with react_explore feature flag enabled
2. paste a list of genes into the search bar (e.g. "foo bar baz") and then hit enter/tab/space
3. Observe the genes get divided appropriately
4. Add a lot of genes, and observe that a vertical scrollbar appears to scroll, rather than overflow

### DIFF
--- a/app/javascript/components/explore/StudyGeneField.js
+++ b/app/javascript/components/explore/StudyGeneField.js
@@ -102,9 +102,10 @@ export default function StudyGeneField({ genes, searchGenes, allGenes }) {
         actionName = 'remove'
         geneDiff = _differenceBy(geneArray, newArray, 'value')
       }
-      log('input:change', {
-        text: `study-gene-search:${actionName}`,
-        changedGenes: geneDiff.map(item => item.value).join(',')
+      log('change:multiselect', {
+        text: geneDiff.map(item => item.value).join(','),
+        action: actionName,
+        type: 'gene'
       })
     } catch (err) {
       // no-op, we just don't want logging fails to break the application
--- a/app/javascript/components/explore/StudyGeneField.js
+++ b/app/javascript/components/explore/StudyGeneField.js
@@ -105,7 +105,8 @@ export default function StudyGeneField({ genes, searchGenes, allGenes }) {
       log('change:multiselect', {
         text: geneDiff.map(item => item.value).join(','),
         action: actionName,
-        type: 'gene'
+        type: 'gene',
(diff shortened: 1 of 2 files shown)


devonbush:
This is now updated to reflect Eric's comments in the (now closed) PR https://github.com/broadinstitute/single_cell_portal_core/pull/925

eweitz (changes requested):
Thanks for including screenshots -- very helpful.  And kudos on the analytics refinement.  I think it nicely addresses this use case.

Overflowing via scrollbar is better than graphically overflowing the container, but the scrollbar will definitely frustrate users.  [One in five gene searches involve >= 10 genes](https://docs.google.com/document/d/1FBQpUBJ0H51dNCeBUM-qN3fY1AM4nleI8HFdfHoVDC8/edit#heading=h.pypfnu913fxu), so I think a reasonable solution needs to be not only not-broken (as you've achieved here) but also not frustrating.

The [React Select docs](https://react-select.com/home) give an example of nicely handling overflow.  

Please implement that here, or open a ticket for next sprint to polish multiselect overflow UX.

Also, SCP-3147 notes "For spatial, we should implement paging controls".  Is a separate ticket needed for that?

bistline (changes requested):
I agree with Eric that the vertical scrollbar will be both difficult to see and use.  I think we should investigate a different option.  The analytics updates are good to go.

devonbush:
https://broadworkbench.atlassian.net/browse/SCP-3169 covers UX improvement for this feature -- quick-and-dirty-but-instrumented was the goal here

eweitz (approved):
Thanks for creating the follow-up!  This works for now.

We'll also want to fix the backwards-incompatibility of these gene search analytics, e.g. so it uses an event named `search` with properties like `context`, `scope`, `numGenes`.  There are also existing analytics on how users interact with gene search -- e.g. whether it's triggered via click on the icon, "enter" key, or via related genes ideogram.

Assuming that analytics issue gets fixed here or ticketed for next sprint prep, I approve.

bistline (approved):

devonbush:
test failures unrelated

PR #929, opened 2021-03-18
adding additional migrate to remove model instances of flag (SCP-3161)

devonbush:
@bistline had actually already written the tool for doing this exact operation, but I forgot to use it in my prior migration script

To test:
1. checkout development, and run `rails db:migrate`
2. Update a user in your DB to have the spatial_transcriptomics feature flag set.
3. confirm that if you go to admin config and try to edit the user roles for that user, the page errors
4. checkout this branch
5. run rails db:migrate
6. confirm you can now edit the user roles for the given user


eweitz (approved):
For auditability, please append Jira ticket to PR title.

bistline (approved):

PR #930, opened 2021-03-18
Fix perfTime analytics for Study Overview (SCP-3042)

eweitz:
This fixes and enriches low-level client-side analytics for plot performance in Study Overview:

* `plot:scatter`
* `plot:violin`
* `plot:box`
* `plot:dotplot`
* `plot:heatmap`

These correspond to low-level analytics diagrammed in #716.  Analytics for "composite" time-to-interactive logging will be addressed in SCP-2907.  These changes provide granular baseline analytics to assess scalability improvements.  The other issue provide higher-level, summary baseline data for the same.

To test:
* Load Study Overview
* Verify that a `plot:scatter` event is logged to Bard / Mixpanel
* Verify that the logged event includes `perfTime`, `perfTime:backend`, and `perfTime:frontend` properties, and that the latter two have values smaller than the first.

Here's an example of what's logged:
Screen Shot 2021-03-18 at 2 00 02 PM

This satisfies SCP-3042.

### DIFF
--- a/app/javascript/components/visualization/DotPlot.js
+++ b/app/javascript/components/visualization/DotPlot.js
@@ -7,6 +7,8 @@ import { log } from 'lib/metrics-api'
 import { getColorBrewerColor } from 'lib/plot'
 import DotPlotLegend from './DotPlotLegend'
 import { getAnnotationCellValuesURL, getExpressionHeatmapURL } from 'lib/scp-api'
+import useErrorMessage, { morpheusErrorHandler } from 'lib/error-message'
+import { withErrorBoundary } from 'lib/ErrorBoundary'
 
 export const dotPlotColorScheme = {
   // Blue, purple, red.  These red and blue hues are accessible, per WCAG.
@@ -25,11 +27,12 @@ export const dotPlotColorScheme = {
   * @param consensus {string} for multi-gene expression plots
   * @param dimensions {obj} object with height and width, to instruct plotly how large to render itself
   */
-export default function DotPlot({
+function RawDotPlot({
   studyAccession, genes=[], cluster, annotation={},
   subsample, annotationValues
 }) {
   const [graphId] = useState(_uniqueId('dotplot-'))
+  const { ErrorComponent, showError, setShowError, setErrorContent } = useErrorMessage()
   const expressionValuesURL = getExpressionHeatmapURL({ studyAccession, genes, cluster })
   const annotationCellValuesURL = getAnnotationCellValuesURL({
     studyAccession,
@@ -45,12 +48,15 @@ export default function DotPlot({
       performance.mark(`perfTimeStart-${graphId}`)
 
       log('dot-plot:initialize')
+      setShowError(false)
       renderDotPlot({
         target: `#${graphId}`,
         expressionValuesURL,
         annotationCellValuesURL,
         annotationName: annotation.name,
-        annotationValues
+        annotationValues,
+        setErrorContent,
+        setShowError
       })
     }
   }, [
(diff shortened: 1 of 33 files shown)


app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -45,24 +45,55 @@ function RawScatterPlot({
...
+      const dataScatterColor = processTraceScatterColor(scatter.data, scatterColor)
+
+      const start = `perfTimeFrontendStart-${graphElementId}`
+      performance.mark(start)
+
+      Plotly.newPlot(graphElementId, scatter.data, layout)
+
+      const end = `perfTimeFrontendEnd-${graphElementId}`
+      performance.mark(end)
+      const perfTimeFrontend = performance.measure(graphElementId, start, end).duration
```
devonbush, line 72: I think moving these performance-related lines after newPlot into a helper logPerformance() method would improve the readability of this handleResponse.

app/javascript/components/visualization/StudyViolinPlot.js
```diff
@@ -72,10 +73,28 @@ function RawStudyViolinPlot({
...
+      const start = `perfTimeFrontendStart-${graphElementId}`
+      performance.mark(start)
+
       renderViolinPlot(graphElementId, results, {
         plotType: distributionPlotToUse,
         showPoints: distributionPoints
       })
+
+      const end = `perfTimeFrontendEnd-${graphElementId}`
+      performance.mark(end)
```
devonbush, line 86: same about mark(end), and perhaps moving these lines after into a logPerformance method so that the loadData() method reads cleaner

devonbush (approved):
Good stuff, just some non-blocking style suggestions

devonbush (changes requested):
I led you astray, sorry!!!    performance.measure does return a value in chrome, but it doesn't return anything in Firefox.  So this code will break on FF.  So I'd go back to the performance.now() approach you had before, because it isn't worth the added complexity of going through the getEntriesByType() rigamarole

bistline (approved):

app/javascript/components/visualization/DotPlot.js
```diff
@@ -116,7 +116,10 @@ function renderDotPlot({
     },
     focus: null,
     tabManager: morpheusTabManager($target),
-    tools
+    tools,
+    loadedCallback() {
+      logMorpheusPerfTime(target, 'dotplot')
+    }
```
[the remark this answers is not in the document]
eweitz: Done in a2b57b3.

app/javascript/components/visualization/DotPlot.js
```diff
@@ -175,3 +178,14 @@ export function morpheusTabManager($target) {
     getTabCount: () => 1
   }
 }
+
+/** Log performance timing for Morpheus dot plots and heatmaps */
+export function logMorpheusPerfTime(target, plotType) {
+  const graphId = target.slice(1) // e.g. #dotplot-1 -> dotplot-1
+  const start = `perfTimeStart-${graphId}`
+  const end = `perfTimeEnd-${graphId}`
+  performance.mark(end)
```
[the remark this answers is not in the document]
eweitz: Good point!  Refined in a517ef9.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -45,24 +45,55 @@ function RawScatterPlot({
...
+      formatHoverLabels(scatter.data, scatter.annotParams.type, scatter.gene)
+      const dataScatterColor = processTraceScatterColor(scatter.data, scatterColor)
+
+      const start = `perfTimeFrontendStart-${graphElementId}`
+      performance.mark(start)
+
+      Plotly.newPlot(graphElementId, scatter.data, layout)
+
+      const end = `perfTimeFrontendEnd-${graphElementId}`
+      performance.mark(end)
```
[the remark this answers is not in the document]
eweitz: Refined in a517ef9.

app/javascript/components/visualization/ScatterPlot.js
eweitz: Agreed, done in a517ef9.

app/javascript/components/visualization/StudyViolinPlot.js
eweitz: Done in a517ef9.

eweitz (commented):
Thanks for the thorough testing across browsers, Devon.

> So I'd go back to the performance.now() approach you had before

Done in a517ef9, expect for the new `logMorpheusPerfTime`.  That retains `performance.measure`, as it seems cleaner than passing a `perfTimeFrontendStart` parameter or some such around.

devonbush (approved):

PR #932, opened 2021-03-19
fixing subsample selector dropdown (SCP-3182)

devonbush:
SCP-3182.  The bug was caused by the selector getting the subsample value as a string "1000" from the url parameter, but the options from the server were integers.  I've taken the simple approach of having it convert everything to strings, since it's always going to get sent to the server as a string anyway, and we have to represent 'all' as a string. 

To test:
1. open a study with at least 1000 cells in the react_explore tab
2. change the subsample to an option other than All cells
3. confirm the visualization updates correctly, and the new option is indicated with dark-blue as the selected option in the dropdown
image

### DIFF
--- a/app/javascript/components/visualization/controls/SubsampleSelector.js
+++ b/app/javascript/components/visualization/controls/SubsampleSelector.js
@@ -16,7 +16,8 @@ function getSubsampleOptions(annotationList, clusterName) {
       clusterSubsamples = []
     }
     subsampleOptions = subsampleOptions.concat(clusterSubsamples.map(num => {
-      return { label: `${num}`, value: num }
+      // convert everything to strings to make the comparisons easier
+      return { label: `${num}`, value: `${num}` }
     }))
   }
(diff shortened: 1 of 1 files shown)


bistline (approved):

app/javascript/components/visualization/controls/SubsampleSelector.js
```diff
@@ -16,7 +16,8 @@ function getSubsampleOptions(annotationList, clusterName) {
       clusterSubsamples = []
     }
     subsampleOptions = subsampleOptions.concat(clusterSubsamples.map(num => {
-      return { label: `${num}`, value: num }
+      // convert everything to strings to make the comparisons easier
+      return { label: `${num}`, value: `${num}` }
```
eweitz, line 20: Since "All" isn't a number, consider using `val` (or something that doesn't entail everything's a number) instead of `num`.

eweitz (approved):

PR #934, opened 2021-03-19
Providing fallback "Unspecified" label for missing group-based annotations (SCP-3070)

bistline:
Since having blank values for group-based annotations is allowed by `scp-ingest-pipeline`, we need to account for this on the front end when rendering any visualization.  Blank strings/null values will cause any number of components to break or have a degraded UX.  This update sanitizes all annotation-based arrays to replace these empty values with a default value of `--Unspecified--`.  

This value was chosen for several reasons:
1. It conveys that the label value was not set by the study owner
1. The `--` characters reduce the risk of this value colliding with a study owner-specified label
1. It works in both Plotly & Morpheus (characters like `<>` do not)

This will work with both normal and array-based metadata, as well as cluster annotation columns.

Sample visualizations using this value:

Scatter:
missing_cluster
Violin:
missing_violin
Dotplot:
missing_dotplot
Heatmap:
missing_heatmap

This PR satisfies SCP-3070.

### DIFF
--- a/app/lib/expression_viz_service.rb
+++ b/app/lib/expression_viz_service.rb
@@ -74,6 +74,7 @@ class ExpressionVizService
       )
       cells.each_with_index do |cell, index|
         values[annotations[index]][:y] << gene['scores'][cell].to_f.round(4)
+        values[annotations[index]][:cells] << cell
       end
     elsif annotation[:scope] == 'user'
       # for user annotations, we have to load by id as names may not be unique to clusters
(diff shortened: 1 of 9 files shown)


bistline:
Test failures unrelated, will address in https://broadworkbench.atlassian.net/browse/SCP-3168

app/lib/expression_viz_service.rb
```diff
@@ -393,7 +411,8 @@ def self.load_gene_set_expression_data_arrays(study, genes, cluster, annotation,
     expression
   end
 
-  # method to initialize containers for plotly by annotation values
+  # method to initialize con
+  # tainers for plotly by annotation values
```
eweitz, line 415:
```suggestion
  # method to initialize containers for plotly by annotation values
```

devonbush (approved):
nice

bistline:
Test failures unrelated

jlchang (approved):
Noting missing labels as "--Unspecified--" is a reasonable default for all the reasons outlined in the PR summary. Many thanks for the thoughtful implementation.

PR #936, opened 2021-03-19
react gene search instrumentation (SCP-3174)

devonbush:
SCP-3174.  This preserves compaatibility with how we've been tracking study gene searches in the past.  @eweitz  let's chat today about ideogram to make sure that's also preserved

To test:
1. Load a study in with react_explore flag on
2. do a gene search
3. confirm via network tab that a 'search' event of type 'gene' is fired, with trigger 'submit'
4. Select 'copy link' from the UX
5. paste the link into a new tab
6. confirm that the new tab also fires a 'search' event of type 'gene', with trigger 'url'


bistline (approved):

eweitz (approved):
Code looks good!

PR #941, opened 2021-03-24
Add informational tooltip for "Scaled mean expression" in dot plot (SCP-3193)

eweitz:
This adds a small tooltip beside the "Scaled mean expression" legend entry in dot plots.  At least two users have been confused by the label.  This tooltip adds the explanation we've given them, now placed in a context where it will more easily help other users.

Clicks on such help icons are also now logged to Mixpanel, so we can assess their usage.

Default icon:
Dot_plot_information_icon_SCP_2021-03-24

Active icon:
Dot_plot_information_icon_active_SCP_2021-03-24

This satisfies SCP-3193.

### DIFF
--- a/app/javascript/components/visualization/DotPlotLegend.js
+++ b/app/javascript/components/visualization/DotPlotLegend.js
@@ -5,7 +5,7 @@ import { faInfoCircle } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import _uniqueId from 'lodash/uniqueId'
 
-const consensusPopover = (
+const scaledPopover = (
   <Popover id="scaled-mean-expression-helptext">
     Scaling is relative to the each gene's expression for all cells in each
     annotation selection, i.e. cells associated with each column label in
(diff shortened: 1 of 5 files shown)


bistline (approved):

app/javascript/components/visualization/DotPlotLegend.js
```diff
@@ -1,7 +1,19 @@
 import React from 'react'
 import { dotPlotColorScheme } from './DotPlot'
+import { Popover, OverlayTrigger } from 'react-bootstrap'
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import _uniqueId from 'lodash/uniqueId'
 
+const scaledPopover = (
+  <Popover id="scaled-mean-expression-helptext">
+    Scaling is relative to the each gene's expression for all cells in each
```
jlchang, line 10:
```suggestion
    Scaling is relative to each gene's expression across all cells in a given
```

jlchang (changes requested):
I've used the word "across" (because "for" could imply for each individual gene expression value being scaled) but if we want to get more technical, we should verify that we're using the calculated mean and then replace "across" with "the calculated mean expression for"

jlchang (approved):

devonbush (approved):
Nice, I appreciate the use of popover here instead of just data-toggle tooltip/title -- let you make the UI very elegant

devonbush (changes requested):
Please add a data-analytics-name to the icon so that we can track clicks on it appropriately, otherwise looks good

eweitz:
The [failed build](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/221220685) is a false-positive, related to an issue with `mimemagic` that Jon noted at standup today.  `yarn test` passes locally.

devonbush (approved):

PR #938, opened 2021-03-22
Improving test reliability/stability (SCP-3168, SCP-3138)

bistline:
Attempting to improve both the reliability and stability of test runs in Travis.  Changes so far:

1. Using block form of `File.open()` where appropriate to ensure no dangling file handlers are left open
2. Use `String#encode!` instead of `String#encode` to force return
3. Invoke tests with `bin/rails test` instead of `bundle exec bin/rake test`
4. Force CI runs in Travis to use full VM config (this is the default, but ensuring it doesn't hurt)

5 consecutive (discounting failures due to `mimemagic` issues, which marked some commits as "failed") clean builds were observed.   While this is not completely indicative that all issues have been permanently resolved, given the previous failure rate of ~85%, this seems like a good enough first start.  Additional enhancements/individual test fixes may be necessary in the future.

This PR satisfies SCP-3168 and SCP-3138.

PR #942, opened 2021-03-24
Log client IP address in nginx/Rails instead of load balancer (SCP-3194)

bistline:
This update changes our nginx configurations to correctly log the client IP address to nginx/Rails logs, rather than the IP address of our GCP load balancers.  This is critical for security/compliance concerns as it will allow us to identify requests by remote IP address.  This is handled in nginx with the Real IP module, and in Rails via the new custom middleware `XForwardedForLogger`.  In both cases, the entire header value will be logged, which will contain each IP address associated with the request, from the client machine, up to (and including) the load balancer.  In local development, this exits without performing any logging as X-Forwarded-For is not being set (although this can be tested via `curl` or an application like Postman).

To test (via `curl`):

1. Boot portal as normal on this branch
2. In a separate terminal, run `curl -k -H "X-Forwarded-For: 123.456.789.0" "https://localhost:3000/single_cell" `
3. Note the following log statement in portal logs: `Forwarded IPs: 123.456.789.0 for GET /single_cell`

In addition, this update pulls in the latest `ubuntu` security patches for our managed base image, and also updates the gem `mimemagic` to a newer release to address the recent issue with `v0.3.5` being removed from rubygems.

This PR satisfies SCP-3194


lib/x_forwarded_for_logger.rb
```diff
@@ -0,0 +1,27 @@
+# x_forwarded_for_logger.rb
+#
+# logs entire X-Forwarded-For header on requests, if present.  This is because Rails only returns the last IP address
+# reported in various headers as it considers this to be the most authoritative, and least likely to be spoofed.  Since
+# GCP reports the load balancer IP last, this is what always shows up in the logs.  This class will log the entire
+# header value to the application log before logging the request as norma, which will mirror the behavior of nginx when
```
eweitz, line 6:
```suggestion
# header value to the application log before logging the request as normal, which will mirror the behavior of nginx when
```

eweitz (approved):
Non-blocking suggestion:

> this can be tested via curl

Instructions for that in the PR description would be a nice touch!

bistline:
> Non-blocking suggestion:
> 
> > this can be tested via curl
> 
> Instructions for that in the PR description would be a nice touch!

Great suggestion - I'll add that.

lib/x_forwarded_for_logger.rb
bistline: b88ad4f

devonbush (changes requested):
Code looks good.  Can you just add a comment to this PR summarizing how to test this in staging (since it's GCP-related, I'm assuming localhost testing will not be representative) ?   

Never mind, just reloaded the page and saw Eric's comment and your update -- thanks!

devonbush (approved):

bistline:
Failure is related to SCP-3138, being addressed in #938

PR #945, opened 2021-03-25
Refine Mixpanel click logging for radio inputs (SCP-3081)

eweitz:
This provides useful Mixpanel logging for clicks on radio-type inputs, as seen in e.g. expression matrix upload UI.

Previously, such clicks were logged but had essentially worthless analytics: `label` was empty and no specific information was provided beyond "there was a radio button click on this page".  Now, we record the `id`, `input-name`, and `value` of the selected radio button option.  

That allows us to analyze things like "What do users do after clicking 'Yes' for 'Is this a raw counts matrix'?"

Due to the complicated nature of radio inputs and how we use them, the raw HTML attribute values are logged rather than their human-friendly labels.  So radio button analysis will need a bit of effort, but is now at least possible -- and the instrumentation is easy to maintain.

To test:
* Go to the Upload Wizard for a study.
* Open Network panel in DevTools, filter to "bard method:POST".
* In the "Expression" tab, click a radio button.
* Verify that an event is sent to Bard with non-empty values for `id`, `input-name`, and `value`.

This is needed for SCP-3081.

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -163,17 +163,24 @@ function getLabelsForElement(element) {
  * Log click on input by type, e.g. text, number, checkbox
  */
 function logClickInput(target) {
-  const domLabels = getLabelsForElement(target)
+  let props
+  if (target.type === 'radio') {
+    const id = 'id' in target ? target.id : ''
+    const inputName = target.name
+    const value = target.value
+    props = { id, 'input-name': inputName, value }
+  } else {
+    const domLabels = getLabelsForElement(target)
 
-  // User-facing label
-  const label = domLabels.length > 0 ? getNameForClickTarget(domLabels[0]) : ''
+    // User-facing label
+    const label = domLabels.length > 0 ? getNameForClickTarget(domLabels[0]) : ''
 
-  const props = { label }
+    props = { label }
 
-  if (target.type === 'submit') {
-    props.text = target.value
+    if (target.type === 'submit') {
+      props.text = target.value
+    }
   }
-
   const element = `input-${target.type}`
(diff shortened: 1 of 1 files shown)


bistline (approved):

app/javascript/lib/metrics-api.js
```diff
@@ -163,17 +163,24 @@ function getLabelsForElement(element) {
  * Log click on input by type, e.g. text, number, checkbox
  */
 function logClickInput(target) {
-  const domLabels = getLabelsForElement(target)
+  let props
+  if (target.type === 'radio') {
+    const id = 'id' in target ? target.id : ''
```
devonbush, line 168: I assume you tested this and so it's there for a reason, but I thought (element).id already returned empty string if no id was specified for the element

devonbush (approved):
Looks good, one html-ese question

app/javascript/lib/metrics-api.js
eweitz: Good catch, refined in f86f76a.  I used this construct yesterday for something else, then removed it, then reflexively used it here.

eweitz:
The [build failure](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/221293524#L3275) is a false positive:

```
Finished in 1274.627726s, 0.2016 runs/s, 0.9281 assertions/s.
  1) Failure:
FileParseServiceTest#test_should_parse_gene_list_file [/home/app/webapp/test/integration/lib/file_parse_service_test.rb:146]:
Expected false to be truthy.
257 runs, 1183 assertions, 1 failures, 0 errors, 1 skips
```

No Rails code was changed, so the test failure is not caused by this change.

PR #946, opened 2021-03-26
Fix occasional infinite loop in related genes ideogram (SCP-3203)

eweitz:
This fixes a bug seen in local development, seemingly caused by a race condition, in which searching via related genes ideogram causes that feature to infinitely loop.  The loop searches the clicked gene, then the previous gene, and repeats _ad infinitum_.

To test:
1. Go to a study in your local SCP with several real genes in SCP React Explore UI
2. Search a gene via the search box
3. Click a different gene in related genes ideogram
4. Wait a few seconds after step 3 completes, and verify that gene from step 2 is not searched again

This satisfies SCP-3203.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.js
+++ b/app/javascript/components/explore/ExploreDisplayTabs.js
@@ -56,9 +56,9 @@ export default function ExploreDisplayTabs(
   referencePlotDataParams.genes = []
 
   /** helper function so that StudyGeneField doesn't have to see the full exploreParams object */
-  function searchGenes(genes, logProps, wasUserSpecified=true) {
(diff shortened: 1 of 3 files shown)


bistline (approved):

devonbush (approved):
Looks good -- @eweitz  can you test locally to confirm the infinite loop problem is solved?

eweitz:
Per chat, the bug re-emerges using 0828710 with a local copy of [SCP41](https://single-cell-staging.broadinstitute.org/single_cell/study/SCP41).

eweitz:
I'll close this PR, as neither of our commits fixes it and we've learned more about the underlying problem.

PR #947, opened 2021-03-26
fix infinite gene search (SCP-3203)

devonbush:
This PR has some general cleanup, what really fixed the bug was the change to `updateExploreParams` that rebuilds the current params from the URL.  I still don't fully understand the root cause of this, as somehow the location of the Reach router was getting updated via a means that wasn't the `navigate` call in updateExploreParams (which is the only navigate call in the Explore component tree).  But this fix is a good idea generally, and fixes the infinite loop, so I think it will be good to have while the underlying cause is still a mystery.   

To test:
0. open a large-ish study with react_explore on
1. Do a gene search, and then VERY quickly delete the gene and type another gene and hit enter to trigger another search (you need to trigger the next search before the previous one finishes)
2. Observe that an infinite toggle between the two genes does NOT occur

### DIFF
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -238,8 +238,7 @@ class ClusterVizService
       annotation[:values] = AnnotationVizService.sanitize_values_array(metadata_obj.values, annotation[:type])
     end
     annotation_array = AnnotationVizService.sanitize_values_array(annotation_array, annotation[:type])
-    Rails.logger.info "annotation array: #{annotation_array}"
-    Rails.logger.info "annotation[:values]: #{annotation[:values]}"
+
     coordinates = {}
     if annotation[:type] == 'numeric'
       text_array = []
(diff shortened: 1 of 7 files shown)


eweitz (approved):
Looks good, verified locally!

app/lib/cluster_viz_service.rb
```diff
@@ -238,8 +238,7 @@ def self.load_cluster_group_data_array_points(study, cluster, annotation, subsam
       annotation[:values] = AnnotationVizService.sanitize_values_array(metadata_obj.values, annotation[:type])
     end
     annotation_array = AnnotationVizService.sanitize_values_array(annotation_array, annotation[:type])
-    Rails.logger.info "annotation array: #{annotation_array}"
```
bistline, line 241: There is also a debug logging statement that needs to be removed on line 213

bistline (changes requested):
This looks good - my only blocking request is that there is one more debug logger statement in `app/lib/cluster_viz_service.rb` that needs to be removed.  

Modulo that, this all looks good, and interestingly this update fixes the issue with very slow autocomplete performance that I had noticed previously in the React explore tab.  It seems as though it was directly related to this issue.

PR #948, opened 2021-03-29
Changing cluster_group.points to field, adding setter (SCP-3104)

bistline:
Currently, when attempting to determine the number of points in a given `ClusterGroup`, calling the method `points` will pull out all of the X coordinates in the database and calculate the length of the array.  While this method ensures that it always returns the correct value, this is unnecessary as the number of points does not change dynamically.  The only way the number of points can change is if the clustering file is re-ingested, in which case the `ClusterGroup` object will have been destroyed.  This update changes `points` from a method to a static field, and adds the method `set_point_count!` to set the number of points after a successful ingest.  

As a part of the associated ticket SCP-3104, benchmarking was performed before & after the refactoring to determine how this will impact the performance of loading clustering data.  The median wall (elapsed real) times in seconds for calling both `cluster.points` and `ClustersController.get_cluster_viz_data` w/ 1.3M cell data are shown below.

BEFORE refactoring:

- `cluster.points` : 1.883
- `ClustersController.get_cluster_viz_data` : 24.293

AFTER refactoring

- `cluster.points` : 0.000027
- `ClustersController.get_cluster_viz_data` : 22.535

As shown, this reduces the elapsed time for calling `points` to essentially 0.  Since `points` can be called up to 4 times during a normal visualization request (in `ClustersController.get_cluster_viz_data`, `Api::V1::Visualization::ExploreController#show`, and in both `ClusterVizService.subsampling_options` and `ClusterVizService.default_subsampling`), this can potentially shave many seconds off of the entire visualization request.

This PR satisfies SCP-3104.

_Note: originally this ticket also looked to make other refinements to loading clustering data by skipping assignment of Z axis or coordinate label information if not present, but these refinements proved to have no impact at all on performance and were summarily dropped._

### DIFF
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -12,6 +12,7 @@ class ClusterGroup
   field :cluster_type, type: String
   field :cell_annotations, type: Array
   field :domain_ranges, type: Hash
+  field :points, type: Integer
   # subsampling flags
   # :subsampled => whether subsampling has completed
   # :is_subsampling => whether subsampling has been initiated
@@ -83,11 +84,6 @@ class ClusterGroup
         data_array.values
       end
     end
-  end
-
-  # return number of points in cluster_group, use x axis as all cluster_groups must have either x or y
-  def points
-    self.concatenate_data_arrays('x', 'coordinates').count
   end
 
   def is_3d?
@@ -354,6 +350,12 @@ class ClusterGroup
       # check if there are any data arrays belonging to this cluster that have a subsample threshold & annotation
       !self.find_subsampled_data_arrays.any?
     end
+  end
+
+  # set the point count for a cluster_group
+  def set_point_count
+    points = self.concatenate_data_arrays('x', 'coordinates').count
+    self.update!(points: points)
   end
 
   ##
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -352,10 +352,11 @@ class ClusterGroup
     end
   end
 
-  # set the point count for a cluster_group
-  def set_point_count
+  # set the point count for a cluster_group and return value
+  def set_point_count!
     points = self.concatenate_data_arrays('x', 'coordinates').count
     self.update!(points: points)
+    self.points
   end
 
   ##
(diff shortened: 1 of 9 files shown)


app/models/ingest_job.rb
```diff
@@ -389,6 +390,17 @@ def set_study_default_options
...
 
+  # set the point count on a cluster group after successful ingest
+  #
+  # * *yields*
+  #   - sets the :points attribute on a ClusterGroup
+  def set_cluster_point_count
+    cluster_group = ClusterGroup.find_by(study_id: self.study.id, study_file_id: self.study_file.id)
+    Rails.logger.info "Setting point count on #{cluster_group.name}:#{cluster_group.id}"
+    cluster_group.set_point_count!
+    Rails.logger.info "Point count on #{cluster_group.name}:#{cluster_group.id} set to #{cluster_group.points}"
```
devonbush, line 401: Do we need both of these statements?  I think we just need the second one -- since set_point_count throws exception on error, we'll know if it fails.

app/models/cluster_group.rb
```diff
@@ -356,12 +352,25 @@ def can_subsample?
...
+  end
+
   ##
   #
   # CLASS INSTANCE METHODS
   #
   ##
 
+  def self.set_all_point_counts!
+    self.all.each do |cluster|
```
devonbush, line 369: does this need to be batched?  I haven't tried ClusterGroup.all in prod

devonbush (approved):
Looks good.  One question which I assume you've already thought through, and one log style question which is definitely non-blocking

app/models/ingest_job.rb
bistline: That's a great point - I can change that.

app/models/cluster_group.rb
bistline: There's only ~1500 or so in prod, so I'm not worried about that part.  But since it needs to be called on each cluster individually I couldn't think of any other way to do this.  Plus, I wanted to be able to call the single method in the migration and do it in the background so that it doesn't slow down the deployment or clog the job queue with 1500 jobs.  If you have a suggestion for a cleaner way of doing this, I'm all ears.

eweitz (approved):
Nice optimization here.

PR #949, opened 2021-03-29
Fixing stability issues with gene list parsing test (SCP-3206)

bistline:
The  test `FileParseServiceTest#test_should_parse_gene_list_file` has periodically failed w/o any specific errors/issues being reported.  Inspection of Travis logs showed that immediately after launching, the test lost the reference to the gene list file:
```
After 0 seconds, Test Gene List is parsing
After 10 seconds,  is unparsed
After 20 seconds,  is unparsed
After 30 seconds,  is unparsed
After 40 seconds,  is unparsed
After 50 seconds,  is unparsed
```

While the cause is not definitively known, local testing has revealed multiple failed calls to GCS to attempt to push the file to a non-existent bucket.  This could be contributing to the anomalous failures, although this has not been directly proven.

This update moves this test into the `ParseUtilsTest` suite which calls the local parser `initialize_precomputed_scores` directly, rather than through the `FireParseService` class.  This is done for several reasons:

1. For gene list files, `ParseUtils.initialize_precomputed_scores` is the only file-type specific line of code executed
2. `FireParseService.run_parse_job` has adequate coverage through other unit/integration tests, so moving this test out should not decrease coverage significantly
3. Calling parser method directly removes the need to poll for job completion, and allows mocking of calls to GCS to avoid errors (these calls cannot be mocked in DelayedJob as the process is detached from test execution)

The result is a test that executes ~10x faster than before, and hopefully addresses any future stability issues with this test.  This update also fixes a broken reference to gene list files when calling `UploadCleanupJob` that appears to be the result of a copy/paste error.

This PR satisfies SCP-3206.


devonbush (approved):

eweitz (approved):

bistline:
Failures are unrelated to this change, but given recent PRs for improving test stability I'm going to leave this open to see if I can determine what's happening with builds.

bistline:
The [build](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/221680749) for fe8ad6d succeeded, as did the previous.  Since 89f3446 was essentially a no-op commit to attempt force a build due to webhooks not triggering, it was cancelled once builds started running again in Travis.

PR #950, opened 2021-03-29
cleanup explore tab code

devonbush:
This fixes a bug where, depending on ordering of requests, we might not respect study default color profile.  It also streamlines and clarifies the application of study default data to the explore params, reducing the number of rerenders.

To test:
0. open a study with a numeric annotation in the explore tab with react_explore, and do not manually specify a color profile
1. change the study default color profile
2. go back to the explore tab, and confirm the new color profile is applied

### DIFF
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -90,7 +90,8 @@ module Api
             # the react refactor is no longer feature-flagged
             spatialGroupNames: spatial_group_options.map { |opt| opt[:name] },
             spatialGroups: spatial_group_options,
-            clusterPointAlpha: @study.default_cluster_point_alpha
+            clusterPointAlpha: @study.default_cluster_point_alpha,
+            defaultColorProfile: @study.default_color_profile
           }
 
           render json: explore_props
--- a/test/api/visualization/clusters_controller_test.rb
+++ b/test/api/visualization/clusters_controller_test.rb
@@ -24,7 +24,8 @@ class ClustersControllerTest < ActionDispatch::IntegrationTest
                                                      y: [7, 5, 3],
                                                      cells: ['A', 'B', 'C']
                                                   },
-                                                  annotation_input: [{name: 'foo', type: 'group', values: ['bar', 'bar', 'baz']}])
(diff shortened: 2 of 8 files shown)


test/api/visualization/clusters_controller_test.rb
```diff
@@ -75,6 +76,17 @@ class ClustersControllerTest < ActionDispatch::IntegrationTest
     assert_equal [1, 4], bar_data['x']
   end
 
+  test 'show should handle numeric clusters' do
+    sign_in_and_update @user
+    execute_http_request(:get, api_v1_study_cluster_path(@basic_study, 'clusterA.txt', {annotation_name: 'intensity', annotation_scope: 'cluster', annotation_type: 'numeric'}))
+    assert_equal 'Reds', json['data'][0]['marker']['colorscale']
+    @basic_study.update!({default_options: {color_profile: 'Portland'}})
+    # since we're going to send a duplicate request, clear the API cache so we can get fresh results
+    Rails.cache.clear
```
bistline, line 85: Non-blocking: rather than the brute-force `Rails.cache.clear`, you can call `CacheRemovalJob.new(@basic_study.accession).perform` to clear the visualization caches for only this study.

bistline (approved):

app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -90,7 +90,8 @@ def show
             # the react refactor is no longer feature-flagged
             spatialGroupNames: spatial_group_options.map { |opt| opt[:name] },
             spatialGroups: spatial_group_options,
-            clusterPointAlpha: @study.default_cluster_point_alpha
+            clusterPointAlpha: @study.default_cluster_point_alpha,
+            defaultColorProfile: @study.default_color_profile
```
eweitz, line 94: For consistency, this should be `colorProfile` or `clusterPointAlpha` should be `defaultClusterPointAlpha`.

```suggestion
            colorProfile: @study.default_color_profile
```

eweitz (changes requested):
With a small tweak for internal consistency, I think this will be good to go.

devonbush:
test fail is unrelated

PR #954, opened 2021-03-31
redrawing plots with expired webgl contexts

devonbush:
Here's what's happening for this bug
1. we render a large number of plots for a gene search w/ spatial data, which uses most of the webgl contexts
2. we do another gene search, which causes all the expression plots to be discarded, and then new ones put in their place, while the cluster plots remain.
3. Chrome hasn't yet garbage collected the webgl contexts of the discarded plots, but needs new contexts to render the new plots.
4. So chrome just expires the oldest webGl context (which is usually the cluster views, since those were rendered first)

There doesn't seem to be any way to manually tell Chrome "garbage collect *this* webgl context instead".  So for now, we'll solve this problem by using plotly.react, which reuses any existing webgl context in the given div, but is otherwise equivalent to newPlot.

To test:
0. load the study "chicken raw counts" with react_explore on
1. do a gene search, and select all the spatial groups to view
2. change the gene and execute another search
3. change the gene again to execute another search
4. observe the while the cluster plot may go away for a few seconds, it returns eventually

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -62,7 +62,7 @@ function RawScatterPlot({
 
       const perfTimeFrontendStart = performance.now()
 
-      Plotly.newPlot(graphElementId, scatter.data, layout)
+      Plotly.react(graphElementId, scatter.data, layout)
 
       const perfTimeFrontend = performance.now() - perfTimeFrontendStart
 
@@ -134,14 +134,16 @@ function RawScatterPlot({
     }
   }, [dimensions.width, dimensions.height])
 
+
   useEffect(() => {
     $(`#${graphElementId}`).on('plotly_selected', plotPointsSelected)
     $(`#${graphElementId}`).on('plotly_legendclick', logLegendClick)
     $(`#${graphElementId}`).on('plotly_legenddoubleclick', logLegendDoubleClick)
     return () => {
-      $(`#${graphElementId}`).off('plotly_selected', plotPointsSelected)
-      $(`#${graphElementId}`).off('plotly_legendclick', logLegendClick)
-      $(`#${graphElementId}`).off('plotly_legenddoubleclick', logLegendDoubleClick)
+      $(`#${graphElementId}`).off('plotly_selected')
+      $(`#${graphElementId}`).off('plotly_legendclick')
+      $(`#${graphElementId}`).off('plotly_legenddoubleclick')
+      Plotly.purge(graphElementId)
     }
   }, [])
 


app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -62,7 +62,7 @@ function RawScatterPlot({
 
       const perfTimeFrontendStart = performance.now()
 
-      Plotly.newPlot(graphElementId, scatter.data, layout)
+      Plotly.react(graphElementId, scatter.data, layout)
```
eweitz, line 65: I'm glad we found this, and also wonder when `Plotly.newPlot` would ever be preferred over `Plotly.react`.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -134,14 +134,16 @@ function RawScatterPlot({
...
     $(`#${graphElementId}`).on('plotly_selected', plotPointsSelected)
     $(`#${graphElementId}`).on('plotly_legendclick', logLegendClick)
     $(`#${graphElementId}`).on('plotly_legenddoubleclick', logLegendDoubleClick)
     return () => {
-      $(`#${graphElementId}`).off('plotly_selected', plotPointsSelected)
-      $(`#${graphElementId}`).off('plotly_legendclick', logLegendClick)
-      $(`#${graphElementId}`).off('plotly_legenddoubleclick', logLegendDoubleClick)
+      $(`#${graphElementId}`).off('plotly_selected')
+      $(`#${graphElementId}`).off('plotly_legendclick')
+      $(`#${graphElementId}`).off('plotly_legenddoubleclick')
```
eweitz, line 145: This would be cleaner with e.g.: 

```
const $graph = $(`#${graphElementId}`)
```

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -134,14 +134,16 @@ function RawScatterPlot({
...
+      Plotly.purge(graphElementId)
```
eweitz, line 146:
```suggestion
      Plotly.purge(graphElementId) // Needed for handling expired WebGL contexts
```

Or a better explanation for maintainers.

eweitz (approved):
Good stuff.  I suggested a few maintainability tweaks, nothing blocking.

app/javascript/components/visualization/ScatterPlot.js
bistline: https://plotly.com/javascript/plotlyjs-function-reference/#plotlyreact

If I read this correctly, I agree with you - `react` seems as though it would be preferable in all circumstances.  There must be some subtlety that is not reflected in the documentation where `newPlot` would be preferable.

bistline (approved):

app/javascript/components/visualization/ScatterPlot.js
devonbush: The purge actually doesn't have any impact on the webgl context (those get cleared up whenever Chrome feels like once the element has been removed) -- this was just added as general good hygiene in case there are other plotly resources associated with the graph

PR #955, opened 2021-03-31
adding annotated scatter tab for numeric annotation gene searches (SCP-3209)

devonbush:
Restores annotated scatter plot tab functionality for numeric annotations on gene searches

image

To test:
0: open a study with react_explore on (e.g. the human lymph node synthetic study)
1: Do a gene search
2: switch the annotation to a numeric facet
3: observe that the annotated scatter plot tab appears with the appropriate plot

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.js
+++ b/app/javascript/components/explore/ExploreDisplayTabs.js
@@ -17,6 +17,7 @@ import useResizeEffect from 'hooks/useResizeEffect'
 const tabList = [
   { key: 'cluster', label: 'Cluster' },
   { key: 'scatter', label: 'Scatter' },
+  { key: 'annotatedScatter', label: 'Annotated Scatter' },
   { key: 'distribution', label: 'Distribution' },
   { key: 'dotplot', label: 'Dot Plot' },
   { key: 'heatmap', label: 'Heatmap' },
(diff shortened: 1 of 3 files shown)


eweitz (approved):

jlchang (approved):
Thanks for restoring annotated scatter!

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -441,7 +459,12 @@ export function getEnabledTabs(exploreInfo, exploreParams) {
         enabledTabs = ['dotplot', 'heatmap']
       }
     } else {
-      enabledTabs = ['scatter', 'distribution']
+      if (exploreParams.annotation.type === 'numeric') {
```
bistline, line 462: I'm going to renew my request that we look at this method and try to figure out a way to simplify the signature.  We're up to 9 different plot/tab types, and this method keeps getting more byzantine as we move forward.

bistline (approved):
Glad we caught this.  I'm making the case again that we should at least ticket the work to simplify `getEnabledTabs()` - the more tabs that we add to Explore, the more difficult it is to parse the logic of this method.

app/javascript/components/explore/ExploreDisplayTabs.js
devonbush: agree that we haven't found the optimal form yet.

devonbush: agree that we haven't found the optimal form for this yet.

PR #956, opened 2021-03-31
Source SECRET_KEY_BASE from vault (SCP-3155)

bistline:
Currently, every time the portal is re-deployed, we generate a new value for `SECRET_KEY_BASE` at runtime and pass this to the Docker container.  This value is used as the [input value](https://api.rubyonrails.org/v5.2.1/classes/Rails/Application.html#method-i-secret_key_base) to the application's secret generator, for things such as secure cookies, and also as the salt for encrypting user's OAuth refresh tokens.  While this is very secure - the value is generated at runtime and never persisted anywhere in code/configuration - it requires us to clear all saved user tokens every time the portal re-deploys.  This leads to a degraded UX that causes all sessions to be invalidated, and causes many `422: InvalidAuthenticityToken` errors as the CSRF token in client browsers cannot be decoded anymore. 

This update replaces the runtime generation for `SECRET_KEY_BASE` and instead sources a static value from `vault`.  This will require a migration to clear all currently saved tokens as `SECRET_KEY_BASE` is changing, but this will be the final time it changes.  All subsequent releases/deployments should no longer sign out users or clear cached tokens (although tokens will still be deleted on sign out as per our Privacy Policy).

To test:
1. Pull this branch 
1. Run `rails_local_setup.rb` to pull all secrets from `vault`
1. Run `bin/rails db:migrate` to unset all stored user tokens
1. Boot server as normal, and sign in a user
1. Stop the server, and re-run `rails_local_setup.rb`
1. Restart the server, and refresh the page.  You should still be signed in, and there should not have been any `OpenSSL` errors when decrypting stored tokens in the database

This PR satisfies SCP-3155


bin/boot_docker
```diff
@@ -39,13 +39,12 @@ $0 [OPTION]
 EOF
 )
 
-# defaults, note there is no default for PROD_DB_PASSWORD, SENDGRID_USERNAME or SENDGRID_PASSWORD
+# defaults, note there is no default for PROD_DB_PASSWORD, SECRET_KEY_BASE, SENDGRID_USERNAME or SENDGRID_PASSWORD
```
devonbush, line 42: Why do we not want to have a default here?  the prod password and sendgrids make sense, but it seems like having a default secret key base if one can't be found in the environment is a reasonable thing

devonbush (approved):
Looks good, just one question about defaults that I don't have any strong opinion on, just curious

eweitz (approved):
Looks good.

To test:
* Sign into staging
* Click "Build now" in https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-development-to-staging/
* After build succeeds, refresh page on staging
* Verify that you don't need to sign in again

Above assumes deployment succeeds within session timeout limit, i.e. 30 minutes.  If it's possible to test this in development, please list those steps!

bin/boot_docker
bistline: We can't a default value because it would have to be represented in code as an explicit string - we can't randomly generate one because the whole point is that we're reusing the same value now, and randomly generating one would cause errors up and down the stack.  Even a default value would do the same thing - if you didn't have it set, and then the default got used, any user auth tokens stored in the DB would be encrypted using that value.  Then if you _did_ set the value correctly from vault, it would cause the aforementioned decryption errors.  Storing a default a value as a string in source would also violate several security practices (and `git-secrets` would disallow the commit as well).

bistline:
> Looks good.
> 
> To test:
> 
> * Sign into staging
> * Click "Build now" in https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-development-to-staging/
> * After build succeeds, refresh page on staging
> * Verify that you don't need to sign in again
> 
> Above assumes deployment succeeds within session timeout limit, i.e. 30 minutes. If it's possible to test this in development, please list those steps!

I've updated the PR description with local testing steps - thanks for the reminder!

PR #957, opened 2021-03-31
Show stack trace for template errors when developing (SCP-3211)

eweitz:
This fixes an observability bug in which stack traces for template errors were inaccessible in development.

To test:
* In your local SCP, insert `<% if @foo.bar < 5 %>test<% end %>` atop `app/views/site/study.html.erb`
* Go to a Study Overview page
* Verify that page contains `NoMethodError in Site#study` error message and stack trace
* Change `development` to `test` in `if !Rails.env.development?` in `app/controllers/application_controller.rb`
* Refresh Study Overview page
* Verify that page contains simple "We're sorry, but something went wrong." message

This satisfies SCP-3211.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,5 @@
 
+
 class ApplicationController < ActionController::Base
 
   ###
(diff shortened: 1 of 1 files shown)


bistline (approved):
Looks good.  Note that this change does not cover the API, but it appears that we do get stack traces in the API already as we're manually logging those here: https://github.com/broadinstitute/single_cell_portal_core/blob/ew-dev-template-stack/app/controllers/api/v1/api_base_controller.rb#L26

devonbush (approved):
Interesting, I had assumed you'd fix this by adding a 
```
if Rails.env.development?  
   Rails.logger.error(exception.backtrace.join("\n"))
end
```
after the ErrorTracker lines.  I kind of like how the way you've done it preserves more of the 'natural' way Rails reports errors in dev, but I am a little worried that this reduces the amount of code surface area we get to see in development.  (but that's probably not a big deal since those trackers were already set up to branch on dev/not-dev).  Anyway, I could see this going either way, so I thought I'd mention it

PR #958, opened 2021-04-01
Refactoring explore tab logic (SCP-3167)

devonbush:
This implements the suggestions from Jerome and the last couple of demos.
  1.  It consolidates the former 'cluster' 'scatter' and 'spatial' plots into a  single "scatter" tab.  This is both good from a UX perspective, and very important from a performance perspective, as it makes it much easier to avoid replots of cluster plots when going back and forth between views
  2. It adds a 'loading' tab for while we're waiting for the exploreInfo to load
  3. It hides controls in the 'view options' that are not relevant to the currently shown tab
  4. It moves the view options controls to be part of the tab with the white background
image

For the code, it simplifies ExploreView -- that component is now just responsible for fetching the explore data nad managing defaults.  ExploreDisplayTabs picks up the logic that used to be in ExploreView.  ScatterTab now has the logic for rendering all the different scatter plot configurations.  This logic has also been moved outside of JSX and into pure javascript so it's easier to follow.  A bunch of logic has been added to manage webgl contexts in a more sophisticated way than we were before.

To test:
 1.  Load the 'chicken raw counts' study with react_explore turned on
 2. Observe the 'loading' tab briefly appears before the cluster.
 3. Observe that once the cluster plot shows, the tab is labeled 'scatter', and observe that the distribution and heatmap controls are hidden
 4. Search for gene 'foo'
 5. Observe the tab is still called 'scatter', and that the cluster graph did not need to re-render
 6. go to the distribution tab, and observe that the 'scatter' controls are replaced with 'distribution' controls
 7. go back to the scatter tab, and select two spatial groups
 8. Observe the spatial plots appear, and the existing scatter plots do not need to be redrawn
 9. Search for the gene 'bar'
 10. Observe the gene expression plots are updated, while the scatter plots persist
 11. Continue searching for genes and toggling spatial groups, and observe that no plots go blank due to missing webGL contexts


app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -42,13 +52,14 @@ const ideogramHeight = 140
...
  */
-export default function ExploreDisplayTabs(
-  {
-    studyAccession, exploreInfo, exploreParams, controlExploreParams, updateExploreParams,
-    isCellSelecting, plotPointsSelected, showViewOptionsControls
-  }
-) {
+export default function ExploreDisplayTabs({ studyAccession, exploreInfo, exploreParams, updateExploreParams, exploreParamsWithDefaults }) {
   const [, setRenderForcer] = useState({})
+    // tracks whetehr the view options controls are open or closed
```
bistline, line 57:
```suggestion
    // tracks whether the view options controls are open or closed
```

app/javascript/components/explore/ScatterTab.js
```diff
@@ -0,0 +1,161 @@
...
+          scatterParams.push({ ...exploreParams, genes: [] })
+        }
+      })
+    }
+  } else if (isGene) {
+    if (isSpatial) {
+      // for single-gene spatial, show the gene expression and reference cluster, then the expresion
+      // overlaid over the spatial plot, then the spatial-cluster plot
+      scatterParams.push({ ...exploreParams })
+      scatterParams.push({ ...exploreParams, genes: [] })
```
bistline, line 102: I _think_ I understand why the multiple assignment is happening here, including the empty `genes`, but some more explanation would be helpful (either in person or as comments).

bistline (approved):
I'm fuzzy on the logic in `ScatterTab`, but I tested and it works as advertised.  I'd love a bit of a rundown on some of `getScatterParams` in `ScatterTab` and why it's doing some of the things it is in terms of assignments.  It feels like there's some logic that isn't being fully explained/outlined.

app/javascript/components/explore/ScatterTab.js
```diff
@@ -0,0 +1,161 @@
...
+      <div className="panel">
+        <span>Due to browser limitations, only 8 plots can be shown at one time.  Deselect some spatial groups</span>
+      </div>
+    }
+  </div>
+}
+
+/** returns an array of params objects suitable for passing into ScatterPlot components
+ * (one for each plot).  Also returns layout variables
+ * This Handles 6 permutations: (spatial / noSpatial) X (no / single / multigene)
```
eweitz, line 66:
```suggestion
 * This handles 6 permutations: (spatial / noSpatial) X (no / single / multigene)
```

app/javascript/components/explore/ScatterTab.js
```diff
@@ -0,0 +1,161 @@
...
+      exploreParams.genes.forEach((gene, index) => {
+        scatterParams.push({ ...exploreParams, genes:  [exploreParams.genes[index]] })
+        if (index === 0) {
+          scatterParams.push({ ...exploreParams, genes: [] })
+        }
+      })
+    }
+  } else if (isGene) {
+    if (isSpatial) {
+      // for single-gene spatial, show the gene expression and reference cluster, then the expresion
```
eweitz, line 99:
```suggestion
      // for single-gene spatial, show the gene expression and reference cluster, then the expression
```

eweitz (approved):

app/javascript/components/explore/ScatterTab.js
eweitz: Having it in comments would help!  The rationale isn't clear to me.

devonbush: Added some comments here and above to hopefully make it clearer.  `{...foo}` is essentially ES6 for `clone(foo)`. And since ExploreTab is passing around param objects and modified param objects across a whole bunch of separate components, I wanted to always use clones to make sure we were never accidentally modifying the original exploreParams object

PR #961, opened 2021-04-02
Remove X Forwarded For logging/configurations (SCP-3234)

bistline:
As discovered on 2021.04.02, our recent configuration/logging updates to log actual client IP addresses introduced an error where users in the most recent version of Chrome could not load SCP, and would instead see a `403` error w/o any explanation.  After much debugging/trial and error, it was determined that the combination of the following two factors caused this issue:

1. In `webapp.conf`, the setting `real_ip_header X-Forwarded-For;` for [`ngx_http_realip_module`](http://nginx.org/en/docs/http/ngx_http_realip_module.html)
2. Chrome sending the `sec-ch-ua` header with `"` and `=` characters in it

While there were no actual errors displayed anywhere, according to [this article](https://www.chromium.org/updates/ua-ch):

> UA-CH is an additive feature, which adds two new request headers that are sent by default: `sec-ch-ua` and `sec-ch-ua-mobile`. Those request headers are based off of Structured Field Values, an emerging standard related to HTTP header values. They contain characters that, though permitted in the HTTP specification, weren’t previously common in request headers, such as double-quotes (“), equal signs (=), forward-slashes (/), and question marks (?). Some Web-Application-Firewall (WAF) software, as well as backend security measures, may mis-categorize those new characters as “suspicious”, and as such, block those requests.

This is in fact exactly what happened - our nginx webserver responded with `403`.  It is unclear why the `ngx_http_realip_module` configuration was the tipping point, but subsequent testing proved this to be the case.

This update removes the `XForwardedForLogger` middleware that attempted to log this header, and all [`ngx_http_realip_module`](http://nginx.org/en/docs/http/ngx_http_realip_module.html) configurations.  In their place is the new `RealIpLogger` which logs the `X-Forwarded-For` header (which is still being set in `nginx`), but performs this task as a controller concern, instead of a middleware.  Also, by calling `ensure => yield` in the concern, we force the Rails app to render a response even if an exception was thrown when attempting to log the IP address.

A test deployment was run on staging and proved that this fix addresses the issue, as it was present on both production and staging.

To test logging locally (the `403` error cannot be reproduced locally as we don't have load balancers):

2. In another terminal, run the following curl command: 
`curl -Ik 'https://localhost:3000/single_cell' --header "X-Forwarded-For: 123.456.789.0"`
3. Back in the portal terminal window, note the log message 
`Forwarded client IP: 123.456.789.0 for GET /single_cell`


bistline:
The configuration change does allow connections to staging w/o `403` errors, and extra IP addresses are being logged to `log/nginx/access.log`, but `staging.log` is not picking up the header.  It may not be getting set by `nginx`.

app/controllers/concerns/real_ip_logger.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  included do
+    around_action :log_real_ip
+  end
+
+  def log_real_ip
+    real_ip = request.headers['HTTP_X_REAL_IP']
+    Rails.logger.info "Real client IP: #{real_ip} for #{request.method} #{request.fullpath}" if real_ip.present?
+  ensure
+    yield # regardless of any issues/errors, request will always execute and render a response
```
devonbush, line 18: nice

devonbush (approved):
This looks good.  Should we also (in either this or a separate PR) set the baseimage back to 1.1.2?

bistline:
> This looks good. Should we also (in either this or a separate PR) set the baseimage back to 1.1.2?

It is still set to `1.1.2` on `development`, so any merges to `master` will preserve that.

eweitz (approved):

eweitz (commented):
I assume the following triggers no 403 error:

```
curl -v 'https://singlecell.broadinstitute.org/single_cell' -H 'sec-ch-ua: "foo" baz="bar"'
```

That's the minimal test case of the original problem that prompted this work.

bistline:
> I assume the following triggers no 403 error:
> 
> ```
> curl -v 'https://singlecell-staging.broadinstitute.org/single_cell' -H 'sec-ch-ua: "foo" baz="bar"'
> ```
> 
> That's the minimal test case of the original problem that prompted this work.

I don't think the last update has been deployed to staging - I can trigger one now and validate that this is the case.

bistline:
@eweitz this works now:
```
curl -I 'https://singlecell-staging.broadinstitute.org/single_cell' -H 'sec-ch-ua: "foo" baz="bar"' | grep status
```

Returns `status: 200 OK`

PR #968, opened 2021-04-05
Adding external resource link to branding groups (SCP-3225)

devonbush:
SCP-3225  This changes the Help menu heading to "Help & Resources" for everyone (which makes sense, since we have resources like the privacy policy under it).  And then for branding groups with the config, it will show an external link at the top of the menu.  I think that makes the most sense, since it's certainly more likely that users are interested in that than the privacy policy, but we can discuss whether it makes sense to use this opportunity to reorder that menu generally.

image

image

To test:
  0. Open a branding group locally (use SyntheticBrandingGroupPopulator to create if you don't have any)
  1. edit the config to give it an external link and description
  2. confirm that the menu option then appears under the "help & resources" menu, and is clickable
  3. confirm that menu option does not appear outside of that branding group

As a side note, I'm a huge fan of OKRs that can be accomplished in less than an hour of coding :)

### DIFF
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -81,6 +81,6 @@ class BrandingGroupsController < ApplicationController
   # Never trust parameters from the scary internet, only allow the white list through.
   def branding_group_params
     params.require(:branding_group).permit(:name, :tag_line, :background_color, :font_family, :font_color, :user_id,
-                                           :splash_image, :banner_image, :footer_image)
+                                           :splash_image, :banner_image, :footer_image, :external_link_url, :external_link_description)
   end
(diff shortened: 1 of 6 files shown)


eweitz (approved):

jlchang (approved):

bistline (approved):
This is all fine, although I worry that the added link will be hidden in that dropdown menu and not terribly discoverable.  But if that's the agreed upon proposal then this looks good to me.

devonbush:
I agree with Jon's comment.  I'm going to merge this so we've got it, and then in demo tomorrow I'll present some options around moving the legal stuff to a footer so this menu is less cluttered

PR #969, opened 2021-04-06
Adding GenericProfiler for method profiling (SCP-3179)

bistline:
This update adds the `GenericProfiler` class to aid in profiling various aspects of SCP performance.  This class can be used to profile any class or instance method, be it SCP-defined or native Rails/Ruby.  All profiling results will be written to both an HTML report showing elapsed wall time for the entire stack (with percentages), and a [flamegraph](https://github.com/brendangregg/FlameGraph) SVG showing the same data, but in an interactive form.  Both can be viewed in any browser.  Results are stored locally in the `tmp/profiling` directory, with each profiling run in a separate folder.  In addition to the reports, an arguments text file will also be written out to for provenance purposes show the exact values/objects passed to the profiled method(s).

The associated ticket with this PR dealt with obtaining benchmarking of the call stack for normal clustering/expression visualization endpoints.  The document with those findings can be found [here](https://docs.google.com/document/d/1oSJklX9bsO4AGEwHwvZurZqnVTd6kWNz4OiZ4Zu2GtE/edit#).  In addition, profiling reports/images used in creating the above document can be found [here](https://drive.google.com/drive/u/0/folders/1ruMFvPrD7naftmv-9dH-kuxpSojaVb5L) (as zip archives, to be downloaded and opened locally in a browser).

This PR (and associated document/reports) satisfy SCP-3179.


eweitz (approved):
Code and document look good!  I made a few comments and suggestions in the latter.

devonbush (approved):
I like this a lot.  I'd lean towards putting it in `lib` instead of `app/lib` since it's a utility that's only used outside manually, but I don't feel strongly about that.

bistline:
> I like this a lot. I'd lean towards putting it in `lib` instead of `app/lib` since it's a utility that's only used outside manually, but I don't feel strongly about that.

That's a good point, I'll go ahead and move it.

PR #972, opened 2021-04-09
Remove extended OAuth scopes, use Terra pet service accounts (SCP-2318, SCP-3204, SCP-3205, SCP-3250, SCP-3259)

bistline:
Previously, when users authenticate into SCP, we asked for two "extended" OAuth scopes: devstorage.read_only, and cloud-billing.readonly.  The former was to allow streaming of files directly to the browser in private studies (for IGV, Ideogram), and the latter to retrieve the list of a user's GCP billing accounts when they are managing Terra billing projects through the "My Billing Projects" section of SCP.

Terra has since stopped requesting extended scopes from users, as some users are unwilling or unable to grant them, and this effectively prevents them from using Terra.  Similar issues have arisen lately with SCP users who object to (or cannot grant) these scope requests.

This update removes both extended scopes from the normal SCP login process, and replaces them with the following functionality:
- **devstorage.read_only**: SCP now leverages Terra pet service accounts, which are Terra-managed service accounts that can be used to grant access tokens w/ extended scopes.  Now, when a user attempts to stream a file directly to the browser in a private study they have access to, SCP issues a token using this pet service account, rather than directly using the user's access token, as it no longer has the requisite scope. For users who have completed their Terra profile, this flow will work.  Users who have not registered with Terra must first complete their Terra registration, as this request will result in a `401` from SAM.
- **cloud-billing.readonly**: If a user clicks the link for "My Billing Projects", they are redirected to a page informing them of the extended scope requirement, and asks them to re-authenticate and grant this scope.  They then have the choice of agreeing or declining at this point.  Users who are unable/unwilling to grant this scope cannot manage billing projects through SCP, but they are no longer blocked from signing into SCP altogether.

This update also displays warnings for users who have not completed their Terra registration when attempting to load IGV or Ideogram in private studies.  Furthermore, as a part of SCP-3259, users can no longer update their Terra profile through SCP until they complete their Terra registration and accept their terms of service.

To test:
1. Pull this branch and sign out of any active sessions
2. Sign in as normal with an account that has registered for Terra
3. Note that no extra consent is required
4. Click the "My Billing Projects" link from the top-righthand nav
5. Note the landing page explaining the extra scope requirements
6. Click 'Sign in and Grant Access'
7. Note the cloud billing scope request
8. Consent to the request, note that the billing projects page loads

If you have any studies with BAM files or Ideogram annotations, to test this functionality:
1. Follow steps 1-3 from above
2. Load the study w/ IGV or Ideogram files
3. Note that the visualization loads
4. Sign out
5. Sign in with an account that has not registered for Terra
6. Load the previous study
7. Note the warning dialog prompting to complete your Terra registration

This PR satisfies SCP-2318, SCP-3204, SCP-3205, SCP-3250, and SCP-3259.


app/javascript/lib/study-overview/terra-profile-warning.js
```diff
@@ -0,0 +1,10 @@
+export const profileWarning = `
+  <div class="container">
+    <p class="bg-danger">
+      The visualization above is attempting to fetch data from this study's private bucket, but it cannot issue an 
+      access token for the request as you have not completed your Terra profile. This is required in order to set your 
+      access permissions on the bucket. Please visit <a href="https://app.terra.bio" target="_blank">Terra</a> and 
+      select 'Sign in with Google' from the top left menu in order to complete your registration, and then reload this page.
```
eweitz, line 7: The help article on this would be more tailored:

```suggestion
      access permissions on the bucket. Please <a href="https://support.terra.bio/hc/en-us/articles/360028235911-How-to-register-for-a-Terra-account" target="_blank">register for Terra</a>, and then reload this page.
```

app/models/fire_cloud_client.rb
```diff
@@ -1332,6 +1336,61 @@ def delete_user_from_billing_project(project_id, role, email)
...
+
+  # issue an access_token for a user's pet service account in the requested project
+  #
+  # * *params*
+  #   - +project_name+ (String) => Name of a FireCloud billing project in which pet service account resides
+  #
+  # * *returns*
+  #   - +String+ pet service account OAuth2 access_token
+  def get_pet_service_account_token(project_name)
+    path = "https://sam.dsde-prod.broadinstitute.org/api/google/v1/user/petServiceAccount/#{project_name}/token"
```
eweitz, line 1374: I recall that [a Bard problem we had last year](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365389125200?thread_ts=1600365317.124600&cid=CBEHTH601) was caused by the (opaque-to-SCP) use of different SAMs for each environment -- i.e. development, staging, and production).

That makes me wary of hard-coding one SAM for all environments.  I'm not confident it'll be problematic, but if we encounter strange auth issues, it might be worth keeping this in mind.

app/models/fire_cloud_client.rb
```diff
@@ -1332,6 +1336,61 @@ def delete_user_from_billing_project(project_id, role, email)
...
+  # response from this API call can be passed to FireCloudClient.new(user, project_name, service_account_contents)
+  # to create an instance of FireCloudClient that is able to call GCS methods as the user in the request project
+  #
+  # * *params*
+  #   - +project_name+ (String) => Name of a FireCloud billing project in which pet service account resides
+  #
+  # * *returns*
+  #   - +Hash+ parsed contents of pet service account JSON keyfile
+  def get_pet_service_account_key(project_name)
+    path = "https://sam.dsde-prod.broadinstitute.org/api/google/v1/user/petServiceAccount/#{project_name}/key"
```
eweitz, line 1390: To avoid needing coordinated updates, it's worth abstracting "https://sam.dsde-prod.broadinstitute.org/api/google/v1/user/petServiceAccount", e.g. to an instance variable, and using that here and in `get_pet_service_account_token`.

```suggestion
    path = "#{@pet_sa_url}/#{project_name}/key"
```

app/views/profiles/show.html.erb
```diff
@@ -74,7 +74,24 @@
...
+        <% if current_user.registered_for_firecloud %>
+          <%= render partial: 'user_firecloud_profile' %>
+        <% else %>
+          <div class="container">
+            <div class="bs-callout bs-callout-danger">
+              <h4>Please complete your Terra registration</h4>
+              <p>
+                You may not update your Terra profile until you have registered with Terra and accepted the terms of service.
+                Please visit <%= link_to 'Terra', 'https://app.terra.bio', target: :_blank %>, select 'Sign in with Google'
+                from the top-lefthand nav menu, and complete the sign in and registration process.
```
eweitz, line 86: See other comment beginning "The help article on this would be more tailored".

app/javascript/components/explore/GenomeView.js
```diff
@@ -32,6 +33,11 @@ function GenomeView({ studyAccession, bamFileName, isVisible, updateExploreParam
   // re-render IGV any time the listing of bamFiles changes
   useEffect(() => {
     if (bamFileList && bamFileList.bamAndBaiFiles.length && isVisible) {
+      // show profile warning from non-existent token due to incomplete Terra registration
+      if (window.accessToken === '') {
+        $('#' + igvContainerId).parent().append(profileWarning)
```
eweitz, line 38: There's probably a more React-idiomatic way to do this, but I'm fine with it as reads clearIy with no obvious functional problem.

eweitz (approved):
Looking good!  Thanks for the thorough description.  I noted two trivial blockers -- for a small abstraction and a more tailored help link -- and made a few tangential comments.

Also, I think five tickets in one PR is a new record!  (Having just opened a separate large PR, I sympathize with the difficulty of breaking changes into smaller efficiently-reviewed chunks.)

app/models/fire_cloud_client.rb
devonbush, line 1374: I think part of this path should be a constant that lives in a config file.

app/models/user.rb
```diff
@@ -33,7 +33,7 @@ def owned_by(user)
   # :confirmable, :lockable, :timeoutable and :omniauthable,
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable, :timeoutable,
-         :omniauthable, :omniauth_providers => [:google_oauth2]
+         :omniauthable, :omniauth_providers => [:google, :google_billing]
```
devonbush, line 36: can you explain why we use 'google' instead of google_oauth2 now?

devonbush (changes requested):
Great stuff -- a big leap forward in user-friendliness.  Biggest concern is that I suspect we want appSec signoff on this.  And then just moving some stuff to react from jquery.

app/controllers/profiles_controller.rb
```diff
@@ -126,6 +124,12 @@ def check_profile_access
     end
   end
 
+  def check_firecloud_registration
+    unless current_user.registered_for_firecloud
+      redirect_to view_profile_path(current_user.id), alert: 'You may not update your Terra profile until you have registered with Terra.' and return
```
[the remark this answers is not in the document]
bistline: Great suggestion

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -26,4 +26,25 @@ def google_oauth2
...
+
+  def google_billing
+    # You need to implement the method below in your model (e.g. app/models/user.rb)
+    @user = User.from_omniauth(request.env["omniauth.auth"])
+
+    if @user.persisted?
+      @user.update(authentication_token: Devise.friendly_token(32))
+      @user.generate_access_token
+      # update a user's FireCloud status
+      @user.delay.update_firecloud_status
```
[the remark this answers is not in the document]
bistline: Fair point, although I think that deserves it's own PR as it touches a ton of code.  My one caveat to that is that while the service/application is now called Terra, the orchestration API is still api.firecloud.org, and many of their internal admin accounts still reference that domain.

app/javascript/components/explore/GenomeView.js
```diff
@@ -32,6 +33,11 @@ function GenomeView({ studyAccession, bamFileName, isVisible, updateExploreParam
   // re-render IGV any time the listing of bamFiles changes
   useEffect(() => {
     if (bamFileList && bamFileList.bamAndBaiFiles.length && isVisible) {
+      // show profile warning from non-existent token due to incomplete Terra registration
+      if (window.accessToken === '') {
```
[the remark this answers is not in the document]
bistline: Gotcha - can do.

app/models/user.rb
bistline: If you look at this: https://github.com/broadinstitute/single_cell_portal_core/blob/jb-pet-service-accounts/config/initializers/devise.rb#L257

What we're doing is configuring two different Omniauth providers, both of which are `google_oauth2` (via the `strategy_class: OmniAuth::Strategies::GoogleOauth2` declaration).  But by using these two different "names", we can have different scope requests that map to discrete callback endpoints.  You can think of it like a `scope` or `namespace` declaration in your `routes.rb` file - it's just a way for Devise & Omniauth to know which provider/strategy to load for each auth endpoint.

This is the only way that you can have separate scope lists unless you do some really complicated dynamic configs via a setup callback endpoint or a fancy setup proc like [this](https://gist.github.com/jiggneshhgohel/4a79aa26cb628533fe132295cffc45b2).  That seemed like a higher risk than this approach entailed.

app/models/fire_cloud_client.rb
bistline: I can do it like `BASE_URL` [here](https://github.com/broadinstitute/single_cell_portal_core/blob/jb-pet-service-accounts/app/models/fire_cloud_client.rb#L18) and call it `SAM_BASE_URL` or `SAM_SERVICE_URL`.  That would at least be internally consistent.

app/models/user.rb
devonbush: gotcha, I knew we needed the new  name to handle the billing-included scopes, wasn't sure why we had to rename the 'default' scope to 'google' from google_oauth2.  thanks for the explanation

app/models/fire_cloud_client.rb
bistline: We would actually encounter auth issues if we used environment-specific domains.  We _always_ point at production for FireCloud API calls (e.g. api.firecloud.org), and if we pointed at a dev/test/alpha SAM instance, the two would not have the same list of users and associated data.  

FWIW, I talked with Kristian about this in the past - specifically should we point at dev in dev, alpha in staging, etc - and he emphatically said no.

app/controllers/profiles_controller.rb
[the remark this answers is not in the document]
bistline: f3422a0

app/javascript/lib/study-overview/terra-profile-warning.js
bistline: f3422a0

app/models/fire_cloud_client.rb
bistline: f3422a0

app/views/profiles/show.html.erb
bistline: f3422a0

app/javascript/components/explore/GenomeView.js
bistline: 79723b0

app/javascript/components/explore/GenomeView.js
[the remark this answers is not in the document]
bistline: 79723b0

app/javascript/components/visualization/InferCNVIdeogram.js
```diff
@@ -267,28 +268,18 @@ function removeIdeogram() {
...
 /** Initialize ideogram to visualize genomic heatmap from inferCNV  */
 function initializeIdeogram(url, organism, assembly, domTarget, showViewOptionsControls) {
   if (typeof window.inferCNVIdeogram !== 'undefined') {
     delete window.inferCNVIdeogram
     removeIdeogram()
   }
 
-  const chrHeight = getChrHeight(showViewOptionsControls)
+  if (window.accessToken === '') {
+    $('#' + domTarget).parent().append(profileWarning);
```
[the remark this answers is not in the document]
bistline: 79723b0

app/models/fire_cloud_client.rb
```diff
@@ -1332,6 +1336,61 @@ def delete_user_from_billing_project(project_id, role, email)
...
+  # create a new instance of FireCloudClient, but use a pet service account keyfile
+  # will only work with users that have been registered in Terra
+  #
+  # * *params*
+  #   - +user+: (User) => User object from which access tokens are generated
+  #   - +project+: (String) => Default GCP Project to use (can be overridden by other parameters)
+  #
+  # * *return*
+  #   - +FireCloudClient+ instance, or nil if user has not registered with Terra
+  def self.new_with_pet_sa(user, project)
```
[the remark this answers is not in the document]
bistline: f3422a0

app/models/fire_cloud_client.rb
bistline: f3422a0

app/javascript/providers/UserProvider.js
```diff
@@ -13,11 +13,26 @@ export function getAccessToken() {
...
+ * from ApplicationHelper#get_read_access_token and stored in the window.SCP object
+ */
+export function getReadAccessToken() {
+  return ('SCP' in window) ? window.SCP.readOnlyToken : 'test'
+}
+
 /** returns true if the current user is logged in */
 export function isUserLoggedIn() {
   return !!getAccessToken()
 }
```
devonbush, line 31: add a /**  comment here

devonbush (approved):
nice, just missing one comment on a function

app/javascript/providers/UserProvider.js
bistline: b6c7e56

app/models/fire_cloud_client.rb
eweitz: Helpful detail, thank you.  I'm no longer wary.  Given we use api.firecloud.org in all environments, it makes sense to use sam.dsde-prod there too.

Notably, that differs from how we use Bard.  [SCP uses environment-specific Bards](https://github.com/broadinstitute/single_cell_portal_core/search?q=bard_host_url) to keep production analytics accurate, and [Bard uses environment-specific SAMs](https://github.com/DataBiosphere/bard/search?q=sam).  

This means SCP implementers will get auth errors for Bard when signed in on development or staging, unless they've gone out of their way to [register in that environment's specific Terra](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365389125200?thread_ts=1600365317.124600&cid=CBEHTH601).  But those are upstream analytics infrastructure issues.  They slightly overlap matters here, and have workarounds, and aren't problematic here.

app/controllers/users/omniauth_callbacks_controller.rb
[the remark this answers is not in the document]
eweitz: Agreed.  Ideally we would suggest to relevant upstream teams that those domains, etc. get renamed to Terra equivalents like api.terra.bio, then (after renames are done) update our code.

It might be worth raising if we're ever asked how upstream FireCloud / Terra APIs could be refined -- a future opportunity rather than an action item.

eweitz (approved):

app/models/fire_cloud_client.rb
bistline: I agree that the current setup is not optimal given the mixed implementations for environments, but merging development and production analytics ones seems like an even larger issue as opposed to this one.

bistline:
Build failures are unrelated and tracked in #982 and SCP-3248.

bistline:
AppSec found an issue with auth scopes - they are apparently editable on the outgoing redirect to Google in the account chooser dialog.  We need to add a validation on the backend that ensures the scopes granted are those that are configured for that provider.

devonbush (approved):
nice work on the scope validation

app/javascript/lib/study-overview/terra-profile-warning.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+export const profileWarning =
+  <div className="container">
+    <p className="bg-danger">
+      The visualization above is attempting to fetch data from this study's private bucket, but it cannot issue an
+      access token for the request as you have not completed your Terra profile. This is required in order to set your
+      access permissions on the bucket. Please&nbsp;
+      <a href="https://support.terra.bio/hc/en-us/articles/360028235911-How-to-register-for-a-Terra-account"
```
eweitz, line 9: At demo, Tim or Amelia suggested less depth in these messages, which I agree would help.  Perhaps:

```suggestion
      This visualization is trying to fetch data from the study's private bucket, but it cannot as you have not completed your Terra profile. Please&nbsp;
      <a href="https://support.terra.bio/hc/en-us/articles/360028235911-How-to-register-for-a-Terra-account"
```

app/javascript/providers/UserProvider.js
```diff
@@ -13,11 +13,39 @@ export function getAccessToken() {
   return ('SCP' in window) ? window.SCP.userAccessToken : 'test'
 }
 
+/**
+ * Returns the current read-access token for the user
+ * This token is specifically for genome visualizations like IGV and Ideogram
+ * as they both require streaming a GCS object directly to the client
+ * This token needs the devstorage.readonly OAuth scope, which is returned
+ * from ApplicationHelper#get_read_access_token and stored in the window.SCP object
```
eweitz, line 21: Great comment.

app/javascript/providers/UserProvider.js
```diff
@@ -13,11 +13,39 @@ export function getAccessToken() {
   return ('SCP' in window) ? window.SCP.userAccessToken : 'test'
 }
 
+/**
+ * Returns the current read-access token for the user
+ * This token is specifically for genome visualizations like IGV and Ideogram
+ * as they both require streaming a GCS object directly to the client
+ * This token needs the devstorage.readonly OAuth scope, which is returned
```
eweitz, line 20: Consistent naming:

```suggestion
 * Returns the current read-only access token for the user
 * This token is specifically for genome visualizations like IGV and Ideogram
 * as they both require streaming a GCS object directly to the client
 * This token needs the devstorage.read_only OAuth scope, which is returned
```

app/javascript/providers/UserProvider.js
```diff
@@ -13,11 +13,39 @@ export function getAccessToken() {
...
+ * from ApplicationHelper#get_read_access_token and stored in the window.SCP object
+ */
+export function getReadAccessToken() {
+  return ('SCP' in window) ? window.SCP.readOnlyToken : 'test'
+}
```
eweitz, line 25: Let's either call the function `getReadOnlyToken`, or call the thing it gets `readAccessToken`, for consistency.  The backend calls it "read access token" instead of "read only token", so I slightly prefer the former.

```suggestion
export function getReadAccessToken() {
  return ('SCP' in window) ? window.SCP.readAccessToken : 'test'
}
```

(This would be a global find-replace, of course.)

app/models/fire_cloud_client.rb
```diff
@@ -1332,6 +1337,61 @@ def delete_user_from_billing_project(project_id, role, email)
...
+
+  # issue an access_token for a user's pet service account in the requested project
+  #
+  # * *params*
+  #   - +project_name+ (String) => Name of a FireCloud billing project in which pet service account resides
+  #
+  # * *returns*
+  #   - +String+ pet service account OAuth2 access_token
+  def get_pet_service_account_token(project_name)
+    path = BASE_SAM_SERVICE_URL + "/api/google/v1/user/petServiceAccount/#{project_name}/token"
```
eweitz, line 1375: Nice refinement.

app/views/billing_projects/access_request.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    and grant access to Single Cell Portal to be able to read this data.</big>
+  </p>
+  <p><big>
+    <strong>Please note:</strong> <u>Single Cell Portal will not be granted access to make billing requests on your behalf,
+    or to manage your billing accounts in any way.</u>  This is simply a request to see a list of billing account names.
+  </big>
+  </p>
+  <p><big>
+    If you wish to continue, please click the 'Sign In and Grant Access' button below and complete the sign-in process,
+    or you may decline and return to the home page.</big>
```
eweitz, line 15: At demo, Tim or Amelia suggested less depth in these messages, which I agree would help. Perhaps:

```suggestion
    To manage your Terra billing projects, Single Cell Portal needs to know your Google
    Cloud Platform billing account names.  This is mediated by Google, so please re-authenticate
    and grant that access to Single Cell Portal.</big>
  </p>
  <p><big>
    <u>Single Cell Portal will not be granted access to make billing requests for you,
    nor to manage your billing accounts.</u>  This is simply a request to see billing account names.
  </big>
  </p>
  <p><big>
    To continue, please click the "Sign In and Grant Access" button below and complete signing in,
    or you may decline and return to the home page.</big>
```

app/views/profiles/show.html.erb
```diff
@@ -74,7 +74,24 @@
...
-        <%= render partial: 'user_firecloud_profile' %>
+        <% if current_user.registered_for_firecloud %>
+          <%= render partial: 'user_firecloud_profile' %>
+        <% else %>
+          <div class="container">
+            <div class="bs-callout bs-callout-danger">
+              <h4>Please complete your Terra registration</h4>
+              <p>
+                You may not update your Terra profile until you have registered with Terra and accepted the terms of service.
+                Please visit <%= link_to 'Terra', 'https://app.terra.bio', target: :_blank %>, select 'Sign in with Google'
```
eweitz, line 85:
```suggestion
                Please <%= link_to 'visit Terra', 'https://app.terra.bio', target: :_blank %>, select 'Sign in with Google'
```

eweitz (approved):
I suggested a variety of refinements, the most important of which I think is simpler and shorter copy for our user-facing messages.  None of my them are major technical issues or functional, so they don't block.

Kudos on reworking SCP's guts to make our UX much more delightful.

bistline:
latest build failure is a problem with Travis not being able to install a gem - not related to any regressions.

PR #974, opened 2021-04-11
removing faceted search feature flag code (SCP-3227)

devonbush:
SCP-3227. 

 Code is ready for review.  Actual merge will depend on PM signoff of https://docs.google.com/document/d/1kXnwyHtC__Tl15N7W4-c1F3zTlGiF1RGEBu_ATGWOHw/edit# (still in progress).  

Note that this removes the feature flag code, but does NOT remove the feature flag.  This is so we can revert this change if necessary, and not have lost the data on which users opted in.  Also, note that we have to remove the code, so this change will only be revertable via a release.  Just changing the feature flag to default-true wouldn't actually test what we want, since users would still have the option to opt-out, which we don't want.

This also updates the faceted search banner display as outlined in the doc. above.

To test:
0. go to home page
1. confirm faceted search appears, and clicking on the "?" icon does NOT give an opt-out option -- it just gives the information
2. confirm the faceted search appears regardless of your current feature flag status.

### DIFF
--- a/app/javascript/components/search/controls/SearchPanel.js
+++ b/app/javascript/components/search/controls/SearchPanel.js
@@ -8,35 +8,14 @@ import FacetsPanel from './FacetsPanel'
 import DownloadButton from './DownloadButton'
 import DownloadProvider from 'providers/DownloadProvider'
 import { StudySearchContext } from 'providers/StudySearchProvider'
-import { UserContext } from 'providers/UserProvider'
 import { SearchSelectionContext } from 'providers/SearchSelectionProvider'
 
-/** render the legacy 'popular' and 'recent' search buttons */
-function CommonSearchButtons() {
-  const searchState = useContext(StudySearchContext)
-  /** update the search params */
-  function handleClick(ordering) {
-    searchState.updateSearch({ order: ordering })
-  }
-  return (
-    <>
-      <span className="facet">
-        <a onClick={() => handleClick('popular')}>Most Popular</a>
-      </span>
-      <span className="facet">
-        <a onClick={() => handleClick('recent')}>Most Recent</a>
-      </span>
-    </>
-  )
-}
 
 const helpModalContent = (<div>
   <h4 className="text-center">Advanced Search</h4><br/>
-  Single Cell Portal supports searching on specific facets of studies by ontology classifications.
-  <br/><br/>
-   For example, you can search on studies that
+  Single Cell Portal supports searching studies by ontology classifications. For example, you can search on studies that
   have <b>species</b> of <b>&quot;Homo sapiens&quot;</b> or have an <b>organ</b> of <b>&quot;brain&quot;</b>.
-  <br/>
+  <br/><br/>
     Currently, about <b>70 out of ~300</b> public studies in SCP provide this metadata information.
   <br/><br/>
   For more detailed information, visit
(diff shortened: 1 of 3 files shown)


devonbush:
Test failure is unrelated, and can't be replicated locally -- trying a rerun

app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -109,69 +50,31 @@ export default function SearchPanel({
...
-          <button className="btn btn-md btn-primary" onClick={() => {
-            setAdvancedSearchEnabled(true, setShowSearchOptInModal)
-          }}>Yes, show advanced search</button>
-          <button className="btn btn-md" onClick={() => {
-            setAdvancedSearchEnabled(false, setShowSearchOptInModal)
-          }}>Cancel</button>
-        </Modal.Footer>
-      </Modal>
+      <FacetsPanel/>
+      <a className="action advanced-opts"
```
eweitz, line 62: If I'd noticed it before, I'd have suggested using `id="search-help"` here, because `advanced-opts` concerns only one DOM element and that other value seems clearer.

But I didn't, and avoiding unnecessary analytics patching seems better than ironing out this nit!

app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -109,69 +50,31 @@ export default function SearchPanel({
...
           { helpModalContent }
         </Modal.Body>
         <Modal.Footer>
-          <button className="btn btn-md btn-primary" onClick={() => {
-            setAdvancedSearchEnabled(true, setShowSearchHelpModal)
-          }}>Yes, use advanced search</button>
-          <button className="btn btn-md" onClick={() => {
-            setAdvancedSearchEnabled(false, setShowSearchHelpModal)
-          }}>Go back to legacy search</button>
+          <button className="btn btn-md btn-primary" onClick={closeSearchModal}>Ok</button>
```
eweitz, line 77: Maybe of interest:

The SCP UI uses both "OK" and "Ok".  (Same with [Terra](https://github.com/DataBiosphere/terra-ui/search?q=OK).)  Chrome, macOS, and [Unicode](https://www.compart.com/en/unicode/U+1F197) all use "OK", but "Ok" is also commonly seen in various products.  I'd opt for "OK", but would be fine with "Ok", as long as we're consistent.

app/javascript/components/search/results/ResultsPanel.js
```diff
@@ -86,10 +86,11 @@ const FacetResultsFooter = ({ studySearchState }) => {
           <FontAwesomeIcon icon={faInfoCircle} className="fa-lg fa-fw icon-left"/>
         </div>
         <div className="">
-          <p>Our advanced search is metadata-powered.  By selecting filters, your search <b>targets only studies that use ontology terms</b> in their metadata file.
-          Currently, about 20% of public studies supply that metadata.</p>
-          Learn more about our search capability on our <a href="https://github.com/broadinstitute/single_cell_portal/wiki/Search-Studies" target="_blank" rel="noreferrer">wiki</a>.<br/>
-          Study authors looking to make their studies more accessible can read our <a href="https://github.com/broadinstitute/single_cell_portal/wiki/Metadata-File#Metadata-powered-Advanced-Search" target="_blank" rel="noreferrer">metadata guide</a>.
+          <p>By selecting filters, your search targets <b>only studies that use ontology terms</b> in their metadata file.
+          <br/>
+          Currently, about 20% of public studies supply such metadata.</p>
```
eweitz, line 91: I've heard "70 of 300" (23%), but I'm not sure when that was queried.  If it was a few months ago, I'd expect it to be 25-30% now, and worth updating given some hesitance about the low figure.

app/javascript/components/search/results/ResultsPanel.js
```diff
@@ -86,10 +86,11 @@ const FacetResultsFooter = ({ studySearchState }) => {
           <FontAwesomeIcon icon={faInfoCircle} className="fa-lg fa-fw icon-left"/>
         </div>
         <div className="">
-          <p>Our advanced search is metadata-powered.  By selecting filters, your search <b>targets only studies that use ontology terms</b> in their metadata file.
-          Currently, about 20% of public studies supply that metadata.</p>
-          Learn more about our search capability on our <a href="https://github.com/broadinstitute/single_cell_portal/wiki/Search-Studies" target="_blank" rel="noreferrer">wiki</a>.<br/>
-          Study authors looking to make their studies more accessible can read our <a href="https://github.com/broadinstitute/single_cell_portal/wiki/Metadata-File#Metadata-powered-Advanced-Search" target="_blank" rel="noreferrer">metadata guide</a>.
+          <p>By selecting filters, your search targets <b>only studies that use ontology terms</b> in their metadata file.
```
eweitz, line 89: Nice trim.

eweitz (approved):
Looks good!  I note some nits, none blocking.

app/javascript/components/search/results/ResultsPanel.js
bistline: I'm not crazy about having `70 out of ~300` hard-coded in the UI.  This will only get more and more stale the longer it is up.  As of 10AM on 4.12.2021, it is 72 out of 335, but by the time this is released, that will likely be higher for both.

I think we should either update these numbers dynamically (like the home banner does), or simply give an approximate percentage range, like 20-30%, or "roughly one quarter", or "only a subset of public studies".

bistline (approved):
This works as advertised, so it's fine to merge - I just have qualms about the stats shown in the popup.

devonbush:
Closing as Amelia has decided more UX refinement is needed before making faceted search GA

PR #975, opened 2021-04-11
improving logging of file downloads (SCP-3227)

devonbush:
* Adds some server-side logging to bulk download requests so that we'll be able to look at the number of files/studies being downloaded.  
* Improves the click tracking of users clicking on the 'download' links to individual files in the UI
* refines our tab-click detection

Please suggest better event names!

TO TEST:
 1. click on some tabs in the UI, confirm they are still logged as tab clicks
 2. perform a bulk download, confirm that a mixpanel event gets logged with name `file-download:curl-config` and `num_studies` and `num_files` properties
 3. download a single file from study overview, confirm a `file-download:study-single` event gets fired (with an `external` suffix if it's a genomic file hosted externally)

### DIFF
--- a/app/lib/bulk_download_service.rb
+++ b/app/lib/bulk_download_service.rb
@@ -48,6 +48,12 @@ class BulkDownloadService
       manifest_config += "output=\"#{study.accession}/file_supplemental_info.tsv\""
       curl_configs << manifest_config
     end
+    byebug
+    MetricsService.log('file-download:curl-config', {
+      num_files: study_files.count,
+      num_studies: studies.count,
+      study_accesions: studies.map(&:accession)
+    }, user)
     curl_configs.join("\n\n")
   end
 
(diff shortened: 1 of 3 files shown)


app/lib/bulk_download_service.rb
```diff
@@ -48,6 +48,12 @@ def self.generate_curl_configuration(study_files:,
       manifest_config += "output=\"#{study.accession}/file_supplemental_info.tsv\""
       curl_configs << manifest_config
     end
+    byebug
+    MetricsService.log('file-download:curl-config', {
+      num_files: study_files.count,
+      num_studies: studies.count,
+      study_accesions: studies.map(&:accession)
```
eweitz, line 55: Please update to use camelCase, which Bard uses for property names like these:

```suggestion
      numFiles: study_files.count,
      numStudies: studies.count,
      studyAccesions: studies.map(&:accession)
```

(The [lexicon](https://mixpanel.com/project/2085496/view/19055/app/lexicon#transformations/event-properties) shows some exceptions and variations, but camel case is the most common for app-defined properties.)

eweitz (changes requested):
Nice observability enhancement here!  With a small tweak, this should be good to merge.

bistline (approved):
Looks good modulo Eric's request.

eweitz (approved):

PR #976, opened 2021-04-12
removing 'url_safe_token' functionality (SCP-2955)

devonbush:
SCP-2955.  Now that we implemented 'cookie allowed API routes', url_safe_token is no longer needed.  url_safe _token was implemented to support passing an API url that morpheus could use, but since morpheus is always invoked on the SCP page itself, the cookie will be passed along, and we can authenticate with that.  

TO TEST:
 0. Sign in, go to a private study, do a multi-gene search
 1. confirm dotplot and heatmap load successfully.
 2. look in the console, and find the URL corresponding to the 'cell_values' request.  e.g. https://localhost:3000/single_cell/api/v1/studies/SCP4/annotations/biosample_id/cell_values?annotation_scope=study&annotation_type=group&cluster=X%20cluster
 3. sign out
 4. try to load that URL directly, confirm it fails

### DIFF
--- a/app/controllers/api/v1/concerns/authenticator.rb
+++ b/app/controllers/api/v1/concerns/authenticator.rb
@@ -11,15 +11,14 @@ module Api
           {controller: 'search', action: 'bulk_download'},
           {controller: 'studies', action: 'generate_manifest'}
         ]
-        URL_SAFE_TOKEN_ALLOWED_ACTIONS = [
-          {controller: 'expression', action: 'show'},
-          {controller: 'annotations', action: 'cell_values'}
-        ]
         # these are API actions that we allow logged-in site users to access using
         # regular rails session validation & csrf protection, which can be useful
         # to save from having to put a token in a URL
         COOKIE_ALLOWED_ACTIONS = [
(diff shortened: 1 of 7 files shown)


bistline (changes requested):
Looks good, and I confirm it works.  The following code should also be removed:

- https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/user.rb#L63
- https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/controllers/users/omniauth_callbacks_controller.rb#L14

Also, the test `test/api/visualization/expression_controller_test.rb#'methods should check view permissions'` lacks a positive assertion (i.e. 200 response check).  The corresponding `annotations_controller_test.rb` does have this assertion, so we should do it on the expression side as well.

devonbush:
I like both those ideas.  I'll add the positive test.  the user.authentication_token code shows up in a lot of places, though, so I'll reserve that for a separate ticket and PR (also it may overlap with SCP-3134)

eweitz (approved):

bistline (approved):

PR #977, opened 2021-04-12
removing byebug (SCP-3229)

devonbush:
SCP-3229 - byebug got added when I was testing, and left in


bistline (approved):

eweitz (approved):
I wonder how feasible it'd be to catch this via linter.

PR #980, opened 2021-04-13
Adding default value for ClusterGroup#points (SCP-3239)

bistline:
This update adds a default value of `0` for calling `points` on an instance of `ClusterGroup`.  This is needed as there is a potential race condition when a cluster file has finished ingesting (and done writing data to MongoDB), but `IngestJob#poll_for_completion` has yet to pick up the finished job and run all of the success callbacks to set the various attributes for the study/cluster (one of which is `cluster.set_point_count!`).  During this window, if a user navigates to the study overview page and attempts to load that cluster, it will throw an error as the subsampling options cannot be rendered due to comparing a `nil` object to an integer.

The default value of `0` will allow rendering the cluster at the "All Cells" level, which at that point is the only data that will be present.  The standard ingest flow will then either continue (if subsampling is required) or exit as normal.

To Test: this unfortunately cannot be tested easily through normal portal usage, as it would require precise timing to pick up the moment when the cluster is present but `set_point_count!` has not yet been called.  Furthermore, calling `cluster.unset(:points)` or `cluster.remove_attribute(:points)` will not cause the default value to be returned, and instead it is simply set to `nil`, which causes the same error as before.

The functionality can be tested in a limited fashion with the following snippet (run in the rails console):

```
new_cluster = ClusterGroup.new
ClusterVizService.subsampling_options(new_cluster)
ClusterVizService.default_subsampling(new_cluster)
```
and should return:
```
=> #<ClusterGroup _id: 6076120ae24139008ee99d11, name: nil, cluster_type: nil, cell_annotations: nil, domain_ranges: nil, points: 1, subsampled: false, is_subsampling: false, study_id: nil, study_file_id: nil>
=> []
=> nil
```

This PR satisfies SCP-3239.

### DIFF
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -12,7 +12,7 @@ class ClusterGroup
   field :cluster_type, type: String
   field :cell_annotations, type: Array
   field :domain_ranges, type: Hash
-  field :points, type: Integer
+  field :points, type: Integer, default: 1
   # subsampling flags
   # :subsampled => whether subsampling has completed
   # :is_subsampling => whether subsampling has been initiated
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -12,7 +12,7 @@ class ClusterGroup
   field :cluster_type, type: String
   field :cell_annotations, type: Array
   field :domain_ranges, type: Hash
-  field :points, type: Integer, default: 1
+  field :points, type: Integer, default: 0
   # subsampling flags
   # :subsampled => whether subsampling has completed
   # :is_subsampling => whether subsampling has been initiated
--- a/test/integration/lib/cluster_viz_service_test.rb
+++ b/test/integration/lib/cluster_viz_service_test.rb
@@ -248,6 +248,7 @@ class ClusterVizServiceTest < ActiveSupport::TestCase
   # before cluster_group.set_point_count! is called at the end of successful ingest
   test 'should fallback to default points for new cluster' do
     new_cluster = @study.cluster_groups.build(name: 'New Cluster', cluster_type: '2d')
+    assert_equal 0, new_cluster.points
     assert_empty ClusterVizService.subsampling_options(new_cluster)
     assert_nil ClusterVizService.default_subsampling(new_cluster)
   end
--- a/test/integration/lib/cluster_viz_service_test.rb
+++ b/test/integration/lib/cluster_viz_service_test.rb
@@ -243,4 +243,12 @@ class ClusterVizServiceTest < ActiveSupport::TestCase
     computed_aspect = ClusterVizService.compute_aspect_ratios(cluster.domain_ranges)
     assert_equal expected_aspect, computed_aspect
   end
+
+  # ensure default value for cluster_group.points will prevent NoMethodError when getting subsampling options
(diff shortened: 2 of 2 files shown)


devonbush (changes requested):
This seems pretty brittle--using 1 to indicate there is an unspecified, but positive number of cells, and having that value be the default?  Where in the ingest process do we actually create the cluster group?  Is there a time when the ClusterGroup is created but before any DataArrays have been ingested, when having "1" as the number of cells will be incorrect?

I could see either
(a) changing the default to zero, and just accepting that the cluster won't be visualizable until  set_study_state_after_ingest runs (that's only going to add a minute or so of delay, right?)
(b) having ingest pipeline set the points on the cluster group during ingest -- since it's doing the ingest, I think it's reasonable to have it own setting the total number of points.  (with the drawback of ingest pipeline having to know a bit more about our portal_core's data model)

There are probably other options, and maybe I'm missing something about how/when cluster groups get created.  But a default of 1 seems like it will cause some downstream headaches

bistline:
@devon to answer your questions:
> Where in the ingest process do we actually create the cluster group? 

The `ClusterGroup` model is created in [transform()](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/clusters.py#L98), after the file format has been validated.  After the document is saved, [load()](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/ingest_pipeline.py#L188) is called which then ingests the `DataArrays`.

> Is there a time when the ClusterGroup is created but before any DataArrays have been ingested, when having "1" as the number of cells will be incorrect?

The value for `points` is not needed in scp-ingest-pipeline until the cluster is subsampled, which happens in a subsequent ingest run, at which point the value is already set, so not having it set here isn't an issue.  The data is only read portal-side when determining subsampling thresholds (both in rendering scatter plots, and in creating user annotations)

> (a) changing the default to zero, and just accepting that the cluster won't be visualizable until set_study_state_after_ingest runs (that's only going to add a minute or so of delay, right?)

The cluster will be visualizable the moment that it is inserted into MongoDB, which happens at the beginning of the ingest run after the file is validated.  This update is directly related to preventing errors in rendering the React explore tab - if the default point count is not set, the whole Explore Tab UI breaks and cannot recover - in fact it doesn't even render.  With this update in place, even though the new cluster that doesn't have any `DataArrays` yet will not render (and you see the DNA spinner stay on the page), the UI is still responsive, and the user can select a different cluster, despite the JS error in the console.  That being said, having a default value of `0` does actually make more sense than `1` from a "logical" standpoint.

> (b) having ingest pipeline set the points on the cluster group during ingest -- since it's doing the ingest, I think it's reasonable to have it own setting the total number of points. (with the drawback of ingest pipeline having to know a bit more about our portal_core's data model)

We could do this, and I in fact thought about it during the refactor for #948.  But we already have convenience methods in SCP for concatenating arrays and counting, and it was simply easier to do it portal-side.  We also do a lot of "state setting" at the end of a successful ingest run, so there was already a framework in place to do this.

devonbush (approved):

eweitz (approved):

bistline:
Test failures are addressed in recently closed PRs.

PR #984, opened 2021-04-14
Further GenericProfiler test fixes (SCP-3248)

bistline:
This update contains further updates to address CI instability for `GenericProfilerTest` runs beyond what was done in #971.  Since all of the errors were instances of `Errno::ENOENT`, `GenericProfiler` will now explicitly catch this error when calling `write_report` in an effort to reduce false negatives during CI, and adds a test that directly invokes the error to prove that it has been addressed as far as the scope of this ticket is concerned.  

It should be noted that while the underlying cause of `Errno::ENOENT: No such file or directory - getcwd` errors during test runs has not been completely identified, efforts through #938, #949 and #971 have drastically reduced the incidence of this particular error.

This PR further supports SCP-3248.

### DIFF
--- a/lib/generic_profiler.rb
+++ b/lib/generic_profiler.rb
@@ -73,11 +73,15 @@ class GenericProfiler
   # * *yields*
   #   - (File) => report file of requested type at specified path
   def self.write_report(report_path, printer)
+    report = File.new(report_path, 'w+')
     begin
-      report = File.new(report_path, 'w+')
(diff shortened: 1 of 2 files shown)


eweitz (approved):

bistline:
Remaining test failures are tracked and addressed in #982.

devonbush (approved):

PR #985, opened 2021-04-15
adding checksum validation to codecov uploader (SCP-3275)

devonbush:
SCP-3275.  Adapted from https://docs.codecov.io/docs/about-the-codecov-bash-uploader.  Tested locally to confirm the diff routine actually exits if the script has been modified (by modifying the script by hand after downloading.

TO TEST:
1. confirm CI still runs properly
2. confirm that you can run lines 140-147 locally
3. confirm that if you run lines 140-143, then manually alter the 'codecov' script file, line 146 exits with error

### DIFF
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -138,7 +138,14 @@ if [[ $code -ne 0 ]]; then
 fi
 if [[ "$CODECOV_TOKEN" != "" ]] && [[ "$CI" == "true" ]]; then
   echo "uploading all coverage data to codecov"
-  bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
+  # code below adapted from https://docs.codecov.io/docs/about-the-codecov-bash-uploader
+  curl -s https://codecov.io/bash > codecov
+  CODECOV_VERSION=$(grep 'VERSION=\".*\"' codecov | cut -d'"' -f2);
+  for i in 1 256 512
+  do
+    diff <(shasum -a $i codecov) <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/$CODECOV_VERSION/SHA${i}SUM)  || { echo "CodeCov checksum validation failed!" >&2; exit 1; }
+  done
+  bash <codecov -t $CODECOV_TOKEN
 fi
 
 clean_up
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -138,14 +138,7 @@ if [[ $code -ne 0 ]]; then
 fi
 if [[ "$CODECOV_TOKEN" != "" ]] && [[ "$CI" == "true" ]]; then
   echo "uploading all coverage data to codecov"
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Looks good -- nice swift mitigation here.  My local run of your test instructions passed: it threw a true positive, i.e. the change I manually introduced was detected and flagged.

Non-blocking suggestion:
* Extract this code (lines 141 - 148) into a separate script, e.g. `codecov_with_validation.sh`.
* Update this script to call that script.
* Have AppSec circulate the new separate script, and perhaps maintain it.

dinvlad (approved):

bistline (approved):

dinvlad:
Great idea! Just created one here: https://github.com/broadinstitute/codecov-bash-uploader/tree/main

bin/run_tests.sh
```diff
@@ -138,7 +138,14 @@ if [[ $code -ne 0 ]]; then
...
   echo "uploading all coverage data to codecov"
-  bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
+  # code below adapted from https://docs.codecov.io/docs/about-the-codecov-bash-uploader
+  curl -s https://codecov.io/bash > codecov
+  CODECOV_VERSION=$(grep 'VERSION=\".*\"' codecov | cut -d'"' -f2);
+  for i in 1 256 512
+  do
+    diff <(shasum -a $i codecov) <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/$CODECOV_VERSION/SHA${i}SUM)  || { echo "CodeCov checksum validation failed!" >&2; exit 1; }
+  done
+  bash <codecov -t $CODECOV_TOKEN
```
eweitz, line 148: Denis centralized your script; let's use it from there.

```suggestion
  bash <(curl -s https://raw.githubusercontent.com/broadinstitute/codecov-bash-uploader/main/codecov-validated.bash)
```

eweitz (changes requested):

bin/run_tests.sh
dinvlad: Sorry, just renamed it to `codecov-verified.bash` 😄

devonbush:
I'm curious to see if this run will pass -- do we not need to explicitly pass in the CODECOV_TOKEN to the codecov script?

bistline:
> I'm curious to see if this run will pass -- do we not need to explicitly pass in the CODECOV_TOKEN to the codecov script?

The `CODECOV_TOKEN` is already in the environment - I'm not sure that we need to explicitly pass it.  I did it just because I was having issues originally getting it to work and thought I needed to do this, but I'm not sure we do.

eweitz (approved):

dinvlad:
Yep, in our script I preserved the semantics of the original script - if `${CODECOV_TOKEN}` is present then it will be used, otherwise the token can be specified with `-t <token>` if needed. I've just updated the script to pass all args transparently.

PR #986, opened 2021-04-15
Removing references to User#authentication_token and other unused code (SCP-3262)

bistline:
Now that Morpheus supports cookie-based authentication, we can now retire the `authentication_token` field from the `User` model, and any references to it.  In addition, this update removes some other recently-obsoleted code that are no longer used.

This also fixes to bugs:
1. An error with the `remove_gene_study_filter_flag` migration that prevented it from running
2. loading Gene Lists in private studies 

This PR satisfies SCP-3262


devonbush (approved):
nice cleanup!

eweitz (approved):

PR #988, opened 2021-04-16
Removing reference to removed field :authentication_token (SCP-3262)

bistline:
This is a bugfix to support SCP-3250 and SCP-3262.  Currently sign in using the `google_billing` provider to add the `cloud-billing.readonly` OAuth scope is broken due to a reference to the recently removed `User` field `:authentication_token`.  

This also fixes an issue in Travis builds due to a problem with the latest release of the `travis` gem not being able to install as a standalone due to a circular dependency error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ arch: amd64
 rvm:
   - 2.6.6
 install:
-  - gem install travis
+  - gem install puma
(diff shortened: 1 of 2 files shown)


eweitz (approved):

devonbush (approved):

PR #991, opened 2021-04-19
removing covid-specific react component (SCP-3263)

devonbush:
SCP-3263. This makes Covid search use the same component as the home page which enables gene search for the covid space.  This removes the covid-specific react component--there are still a couple of hardcoded covid-specific if/elses, but this will be much easier to deal with as we start working on XDSS and making faceted search GA.  

Note that this doesn't address the already-noted concern that StudySearchProvider and GeneSearchProvider share ~80% of their code.   That should be addressed when we look at Jerome's latest proposal for faceted search GA, since that adds a third tab and likely will require redesign of both providers.

image

TO TEST:
0. load the home page, confirm study and gene search load the right studies.
1. Go to covid19 space, confirm study and gene search load the covid-19 specific studies (note that you need to define a Preset Search with a name of 'covid19' in your local instance for this to work)
2. while in the covid19 space, reload the page and confirm the page still loads as expected


eweitz (approved):
Looks good!

bistline (approved):

PR #998, opened 2021-04-21
Updating API caching to hash genes query string re: % encoding errors (SCP-3214)

bistline:
For the last several months, we have noticed a (very) frequently [recurring error](https://sentry.io/organizations/broad-institute/issues/2178215554/?project=1424198&query=is%3Aunresolved+ArgumentError&statsPeriod=14d) in Sentry where attempts to purge entries from the visualization cache fails silently in the background.  While we have been unable to reproduce the error locally, the main issue appears to be that comma-delimited lists of gene names occasionally get corrupted when being transformed into a file path to store responses in the visualization cache. The URL-encoded form of a comma (`%2C`) ends up getting transformed into `%2/C`, or something similar that is not a valid URL-encoded character.  Since all cache expiration is based on regular expression matching, this invalid encoding breaks all `CacheRemovalJob` calls, and ends up preventing caches from being expired when users update files.  Once an invalid cache entry is written, it **_breaks all cache clearing for the entire portal_** until the entire cache is expired on either a deployment, or an administrator running `Rails.cache.clear` in the production console.

This update changes `ApiCaching#get_cache_key` to hash the value of the `genes` query string parameter when computing the cache entry's filename. This will result in more consistently-sized filenames on cache entries by using a hex digest value that maps to the sorted & hashed list of gene names from the request.  

It is worth noting that this feature was actually in place before when visualization requests were handled by `app/controllers/site_controller.rb`, and has now been ported over to `api_caching.rb`.

### Testing

Unfortunately, we have been unable to recreate the conditions where this error occurs.  However, it is possible to test that this new feature works and cache clearing is not impacted.  To test this:

1. Pull this branch, and ensure caching is enabled locally by making sure that the file `tmp/caching-dev.txt` is present.
2. Boot portal and load the `Human HIV and Tuberculosis, blood study`
3. In a separate terminal, enter the rails console and run `Rails.cache.clear` to clear out any existing caches
4. Back in the portal, run a gene search for `APOE` and `AGPAT2`
5. Once the visualization loads, look in the `development.log` file and note the following lines:
```
Writing to API cache: _single_cell_api_v1_studies_SCP18_expression_heatmap_cluster_cluster_tsv_genes_e03d70d640ca9a2f81a83b5de74f3382c241e606f2e2214c8cd74c31ebbb9f50_data_type_heatmap
...
Reading from API cache: _single_cell_api_v1_studies_SCP18_expression_heatmap_cluster_cluster_tsv_genes_e03d70d640ca9a2f81a83b5de74f3382c241e606f2e2214c8cd74c31ebbb9f50_data_type_heatmap
```
6. Back in the rails console, and run the following commands to ensure cache clearing works:
```
study = Study.find_by(name: 'Human HIV and Tuberculosis, blood study')
CacheRemovalJob.new(study.accession).perform
=> true
```
7. Note back in `development.log` the following statements:
```
2021-04-21 14:33:08 -0400: Deleting caches for SCP18
2021-04-21 14:33:08 -0400: caches successfully cleared
```

This PR satisfies SCP-3214.

### DIFF
--- a/app/controllers/api/v1/concerns/api_caching.rb
+++ b/app/controllers/api/v1/concerns/api_caching.rb
@@ -9,7 +9,7 @@ module Api
         CACHE_PATH_BLACKLIST = %w(controller action format study_id)
 
         # character regex to convert into underscores (_) for cache path setting
-        PATH_REGEX =/(\/|%2C|%2F|%20|\?|&|=|\.)/
+        PATH_REGEX =/(\/|%2C|%2F|%20|\?|&|=|\.|,|\s)/
 
         # check Rails cache for JSON response based off url/params
         # cache expiration is still handled by CacheRemovalJob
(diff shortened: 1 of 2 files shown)


eweitz (approved):
It's too bad this hasn't been reproducible.  (Cache invalidation is a hard problem indeed!)  But the hypothesized cause and proposed fix seem plausible, and given the silent-but-high impact and tests above, this seems worth trying.

devonbush (approved):
nice approach

PR #1000, opened 2021-04-22
fixing plotly usage in create annotation (SCP-3298)

devonbush:
SCP-3298. 
This also adds a basic check to confirm that the component can render itself without error.  A full behavioral check is well beyond the capabilities of our current test framework.

TO TEST:
 0. sing in and open a study
 1.  click 'create annotations'
 2.  observe the create annotations controls render


eweitz (approved):
Nice quick fix.  

The test seems independent of the bug and slightly extraneous, but (as you note) the pertinent test isn't feasible to automate.

bistline (approved):
I agree about the test - all I was looking for was an assertion that we could render the component, and this fulfills that.

PR #1004, opened 2021-04-23
Refactoring `CellMetadatum#cell_annotations` for performance

bistline:
This refactoring has three key improvements:

1. Fully leverages the unique `linear_data_arrays_index` on the `DataArray` collection
2. Uses `pluck` and `reduce` to concatenate values rather than `each` and `+=`
3. Avoids calling `any?` on query cursor to prevent running count operation (~2s)

When [profiled](https://docs.google.com/spreadsheets/d/1WHJzz7vL5XDhyKuLp16Cx10_BhhzIJyD28VyLeqny3U/edit#gid=0) with `GenericProfiler`, this showed a 26% improvement over the original implementation.

### DIFF
--- a/app/models/cell_metadatum.rb
+++ b/app/models/cell_metadatum.rb
@@ -55,7 +55,7 @@ class CellMetadatum
     query = {name: array_name, array_type: array_type, linear_data_type: 'CellMetadatum', linear_data_id: self.id,
              study_id: self.study_id, study_file_id: self.study_file_id, cluster_group_id: nil,
              cluster_name: self.study_file.upload_file_name, subsample_threshold: nil, subsample_annotation: nil}
-    DataArray.where(query).order(:array_index => 'asc').pluck(:values).reduce(&:+)
+    DataArray.where(query).order(:array_index => 'asc').pluck(:values).reduce([], :+)
   end
 
   # create dropdown menu value for annotation select
--- a/app/models/cell_metadatum.rb
+++ b/app/models/cell_metadatum.rb
@@ -43,7 +43,7 @@ class CellMetadatum
 
   # concatenate all the necessary data_array objects and construct a hash of cell names => expression values
   def cell_annotations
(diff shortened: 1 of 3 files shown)


app/models/cell_metadatum.rb
```diff
@@ -43,23 +43,26 @@ class CellMetadatum
...
     Hash[cells.zip(annot_values)]
   end
 
-  # concatenate data arrays of a given name/type in order
+  def all_study_cells
+    query = {name: 'All Cells', array_type: 'cells', linear_data_type: 'Study', linear_data_id: self.study_id,
+             cluster_name: self.study_file.upload_file_name, study_id: self.study_id, study_file_id: self.study_file_id,
+             cluster_group_id: nil, subsample_annotation: nil, subsample_threshold: nil}
+    DataArray.where(query).order(:array_index => 'asc').pluck(:values).reduce([], :+)
+  end
```
devonbush, line 59: Let's talk through this Monday -- I like that we're able to utilize more of the index fields.  I wonder how much of this we can port over to the method in `study.rb`. Theoretically, this is doing exactly the same thing, and study annotations are always defined in the same study_file (the metadata file), so we should be able to get the performance gains you've gained here, but keep the all_cells_array method in study.rb where it makes more sense

devonbush (changes requested):
Looks great -- 20% improvement is nothing to sneeze at!  Just some questions about whether we can get these performance gains while keeping the all_cells_array in its more natural location

app/models/cell_metadatum.rb
bistline: Happy to talk about it - but what I had noticed was that part of the slowdown was using the association (i.e. `self.study.all_cells_array`) to call the method.

devonbush:
I think we can just close this -- it'll get picked up when I rebase the long-lived branch on development

PR #1005, opened 2021-04-26
Log web vitals and client hardware metrics to Mixpanel (SCP-3299)

eweitz:
This logs metrics on page-load performance -- known as [web vitals](https://web.dev/vitals/) -- and client hardware to Mixpanel via Bard.

Good web vitals improve user experience and search engine ranking.  They help assess impact of simple-to-implement-and-hard-to-quantify optimizations, like changing compression from gzip to brotli.  They are measured for all page views.

Tracked web vitals:
* TTFB: time to first byte, the earliest page load event
* FCP: first contentful paint, when users gets first visual feedback
* LCP: largest contentful paint, a core measure of perceived load time
* FID: first input delay, when user can first interact (e.g. click)
* CLS: cumulative layout shift, measures visual stability

LCP, FID, and CLS are core web vitals.  TTFB and FCP are useful supplements.

Since users' machines and connection speeds impact performance, this also adds tracking for client hardware:
* Memory (GB RAM)
* CPU (# CPUs)
* Internet connection: 
  * Type: effective connection type, e.g. "4g"
  * RTT: round trip time, i.e. latency / ping (ms)
  * Downlink: download bandwidth in megabits per second (Mbps)

Client hardware is also tracked in any event that tracks `perfTime` -- e.g. fetching scatter plot data.  This could help, later perhaps, tailor how much data we send based on the client's connection speed.

Here's what it looks like in Mixpanel:
Web_vitals_and_client_hardware_properties_in_Mixpanel_SCP_2021-04-26

To test:
* Open a page in SCP
* Go to another page
* (Note: cancelled web request is OK.  [`sendBeacon`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) would be ideal, but complicated for Bard to handle.)
* Go to [Live View in Mixpanel dev](https://mixpanel.com/report/2085496/view/19055/live)
* Search "web-vitals", verify that properties marked in screenshot appear

This satisfies SCP-3299.


app/javascript/lib/metrics-web-vitals.js
```diff
@@ -0,0 +1,141 @@
...
+
+    result = Object.assign(result, hardwareStats)
+
+    isSent = true
+
+    if (typeof navigator === 'undefined') {return}
+    return log('web-vitals', result, true)
+  }
+
+
```
bistline, line 80: A comment would be helpful here.

app/javascript/lib/metrics-web-vitals.js
```diff
@@ -0,0 +1,141 @@
...
+  /** @param {Metric} metric */
+  const report = metric => {
+    if (!isCalled) {isCalled = true}
+    result = { ...result, ...mapMetric(metric, result) }
+  }
+
+  // should be the last call to capture latest CLS
+  setTimeout(() => {
+    // Safari does not fire "visibilitychange" on the tab close
+    // So we have 2 options: loose Safari data, or loose LCP/CLS that depends on "visibilitychange" logic.
```
bistline, line 95:
```suggestion
    // So we have 2 options: lose Safari data, or loose LCP/CLS that depends on "visibilitychange" logic.
```

bistline (approved):
Looks good - minor non-blocking comments suggestions.

app/javascript/lib/metrics-web-vitals.js
eweitz: Done in a739a52.

app/javascript/lib/metrics-web-vitals.js
```diff
@@ -0,0 +1,141 @@
...
+ *  - TTFB: time to first byte, a very early page load event
+ *  - FCP: first contentful paint, when users gets first visual feedback
+ *  - LCP: largest contentful paint, a middle measure perceived load time
+ *  - FID: first input delay, when user can first interact (e.g. click)
+ *  - CLS: cumulative layout shift, measures visual stability
+ *  - Hardware stats on connection speed, # CPUs, memory
+ *
+ * LCP, FID, and CLS are the core web vitals.
+ * TTFB and FCP are useful supplements.
+ */
```
devonbush, line 132: do we have any sense for how long this call takes?  Right now we're blocking page load on this call completing, and I'm not sure that's what we want to do.  for performance, and error-safety, it might be better to execute this in a callback and/or try/catch

devonbush: Reading this PR more, it looks like this doesn't actually perform the logging, it just sets it up?  so maybe 'initializeWebVitalsLogging' is a more descriptive name?

devonbush (changes requested):
Cool stuff, just some code structure/style questions to sort out

app/javascript/lib/metrics-web-vitals.js
eweitz: Yup, the setup call is non-blocking.  The log request happens when the page is *un*loaded, and also doesn't block.

This theoretically means that `web-vitals` log requests could get dropped.  The new-ish `sendBeacon` JS API is guarantees such requests wouldn't drop -- it was specifically designed to account for analytics use cases like this.   However, Bard requires an `Authorization: Bearer` HTTP request header for registered users, and `sendBeacon` doesn't support such non-standard request headers.  

So until Bard is enhanced to authenticate user identities some other way, this approach is the next best thing.

app/javascript/lib/metrics-api.js
```diff
@@ -46,6 +48,10 @@ if (env != 'test') {
   registeredForTerra = window.SCP.registeredForTerra
 }
```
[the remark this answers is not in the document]
eweitz: Done in b80cace.

app/javascript/lib/scp-api-metrics.js
```diff
@@ -129,6 +129,34 @@ export function logSearch(type, searchParams, perfTime) {
   )
 }
 
+/** Logs scatter plot metrics */
+export function logScatterPlot(perfTimes, plotData) {
```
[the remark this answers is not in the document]
eweitz: Done in 72468be.

Passing `perfTime` parameters like this was me anticipating a small refactor, which I had overlooked following through on.  I just did in 3834001, removing some error-prone duplication.

app/javascript/lib/metrics-web-vitals.js
eweitz: `b80cace` uses clearer function names.

app/javascript/lib/metrics-api.js
```diff
@@ -46,6 +48,14 @@ if (env != 'test') {
...
 
+/** Initializes logging for web vitals, if appropriate */
+export function setupWebVitalsLogging() {
+  console.log('in setup')
```
devonbush, line 53: extra console.log

app/javascript/lib/metrics-web-vitals.js
devonbush: cool

app/javascript/lib/metrics-api.js
```diff
@@ -46,6 +48,14 @@ if (env != 'test') {
   registeredForTerra = window.SCP.registeredForTerra
 }
 
+/** Initializes logging for web vitals, if appropriate */
+export function setupWebVitalsLogging() {
```
devonbush, line 52: maybe this should just be called 'initializeLogging', and then any other stuff that we need to initialize goes here too.  that also will reduce confusion about setupWebVitalsLogging vs. initializeWebVitalsLogging?

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -64,27 +64,10 @@ function RawScatterPlot({
...
-        'numAnnotSelections': scatter.annotParams.values.length,
-        'annotName': scatter.annotParams.name,
-        'annotType': scatter.annotParams.type,
-        'annotScope': scatter.annotParams.scope
-      }
-
-      log('plot:scatter', perfLogProps)
+      logScatterPlot(
+        { scatter, genes, width, height },
+        { perfTime, perfTimeFrontendStart }
```
devonbush, line 69: I still think this is a cleaner if you do 
```const perfTimeFrontend = performance.now() - perfTimeFrontendStart 
logScatterPlot(..., {perfTimeBackend, perfTimeFrontend})```

devonbush (changes requested):
console.log should definitely get cleaned, and then a couple of non-blocking suggestions

app/javascript/lib/metrics-api.js
eweitz: Fixed in 7551e40.

app/javascript/lib/metrics-api.js
eweitz: I refined it to `setupPageTransitionLog`, to clarify it while keeping it narrow in purpose.

devonbush (approved):

app/javascript/components/visualization/ScatterPlot.js
eweitz: That would give cleaner parameters, but add duplication (`perfTimeFrontend ` would also need to be calculated above `logViolinPlot`) and decrease coherence (the less `perfTime` calculation needed directly in app code, the better).

I'm preparing `perfTime` refinements for another PR, which I think will resolve both our concerns.

PR #1006, opened 2021-04-26
Optimizing performance of loading annotation/cell name arrays (SCP-3302)

bistline:
This update optimizes performance of loading annotation/cell name arrays in the `Study` and `CellMetadatum` models.  The substantive changes are:

1. Leverage the entire `linear_data_arrays_index` in the `DataArray` model
2. Avoid calling `any?` after creating query cursor as this performs a document count - checking for the existence of a parsed metadata file achieves the same goal in a fraction of the time (~2s savings)
3. Use `pluck` and `reduce` instead of `each` and `+=` when concatenating values (20% faster)

When profiling in development, this showed an overall increase of 26% in load performance compared to the original implementation.

This PR satisfies SCP-3302.

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -1250,7 +1250,7 @@ class Study
   # return an array of all single cell names in study, will check for master list of cells or concatenate all
   # cell lists from individual expression matrices
   def all_cells_array
-    if study.metadata_file&.parsed? # nil-safed via &
+    if self.metadata_file&.parsed? # nil-safed via &
       query = {name: 'All Cells', array_type: 'cells', linear_data_type: 'Study', linear_data_id: self.id,
                cluster_name: self.metadata_file.upload_file_name, study_id: self.id, study_file_id: self.metadata_file.id,
                cluster_group_id: nil, subsample_annotation: nil, subsample_threshold: nil}
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -1250,9 +1250,11 @@ class Study
   # return an array of all single cell names in study, will check for master list of cells or concatenate all
   # cell lists from individual expression matrices
   def all_cells_array
-    arrays = DataArray.where(study_id: self.id, linear_data_type: 'Study', linear_data_id: self.id, name: 'All Cells').order_by(&:array_index)
-    if arrays.any?
-      arrays.pluck(:values).reduce(&:+)
+    if self.metadata_file.present? && self.metadata_file.parsed?
+      query = {name: 'All Cells', array_type: 'cells', linear_data_type: 'Study', linear_data_id: self.id,
+               cluster_name: self.metadata_file.upload_file_name, study_id: self.id, study_file_id: self.metadata_file.id,
+               cluster_group_id: nil, subsample_annotation: nil, subsample_threshold: nil}
+      DataArray.where(query).order(:array_index => 'asc').pluck(:values).reduce(&:+)
     else
       self.all_expression_matrix_cells
     end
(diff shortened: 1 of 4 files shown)


bistline:
Comparison of staging vs. production using the same cluster/annotation (SCP383, 10x mouse data, all cells, uncached viz request TTFB):

production: 46.9s
after: 37.6s

Roughly 20% faster - so this improvement appears to hold from development to a production-like environment.

app/models/study.rb
```diff
@@ -1250,9 +1250,11 @@ def all_cells
   # return an array of all single cell names in study, will check for master list of cells or concatenate all
   # cell lists from individual expression matrices
   def all_cells_array
-    arrays = DataArray.where(study_id: self.id, linear_data_type: 'Study', linear_data_id: self.id, name: 'All Cells').order_by(&:array_index)
-    if arrays.any?
-      arrays.pluck(:values).reduce(&:+)
+    if self.metadata_file.present? && self.metadata_file.parsed?
```
eweitz, line 1253: Since we're optimizing, it seems we could shave off more time by:

```suggestion
    if self.metadata_file.parsed?
```

(`self.metadata_file.present? == false and self.metadata_file.parsed? == true` seems impossible.)

eweitz (approved):
Great value here.  Below, I speculate about a further optimization.

eweitz:
> staging: 37.6s
production: 46.9s
> Roughly 20% faster - so this improvement appears to hold from development to a production-like environment.

@bistline, out of curiosity, how many seconds is this metric on staging without this optimization?  It seems plausible that it could be faster than production, too.

app/models/study.rb
bistline: That is correct, but if there is no metadata file present in the study, then `self.metadata_file.parsed?` results in a `NoMethodError`.  The `self.metadata_file.present?` call first is to simply avoid that.

bistline:
> > staging: 37.6s
> > production: 46.9s
> > Roughly 20% faster - so this improvement appears to hold from development to a production-like environment.
> 
> @bistline, out of curiosity, how many seconds is this metric on staging without this optimization? It seems plausible that it could be faster than production, too.

You are correct, as I just realized that this branch isn't even _on_ staging right now.  So the `37.6` time is pre-optimization.  Once this is merged and deployed, then I can get the real number and compare it against the time of `37.6s`.

app/models/study.rb
devonbush: This seems like a good place for nil-safe function calling: `self.metadata_file&.parsed?`

devonbush (changes requested):
I looove performance improvements like this that don't actually increase the complexity of what we have to do, just do things faster.  Nice work!  Just one minor update

bistline:
> I looove performance improvements like this that don't actually increase the complexity of what we have to do, just do things faster. Nice work! Just one minor update

85ba79f

app/models/study.rb
```diff
@@ -1250,9 +1250,11 @@ def all_cells
   # return an array of all single cell names in study, will check for master list of cells or concatenate all
   # cell lists from individual expression matrices
   def all_cells_array
-    arrays = DataArray.where(study_id: self.id, linear_data_type: 'Study', linear_data_id: self.id, name: 'All Cells').order_by(&:array_index)
-    if arrays.any?
-      arrays.pluck(:values).reduce(&:+)
+    if !!study.metadata_file&.parsed? # nil-safed via &, !! converts to boolean: nil|false == false
```
devonbush, line 1253: I think the `!!` is confusing here.  It's not necessary because ruby will treat already nil as false for `if` purposes.  Typically you only need the `!!` if you're printing out values or passing them to another function that might differentiate between actual `false` and false-y values.

devonbush (changes requested):

app/models/study.rb
bistline: >  the !! wouldn't help either, because it would make it true

That is not correct - `!!nil` equates to `false`, as seen here:

```
irb(main):001:0> !!nil
=> false
irb(main):002:0> !nil
=> true
irb(main):003:0> !!false
=> false
irb(main):004:0> !!true
=> true
```

If it didn't, then `nil` being treated like `false` wouldn't be the expected behaviour.  That being said, I take the point about the rest of it - since this is the default behavior for `nil`, the double-bang isn't necessary here.

bistline: 2a29c4b

devonbush (approved):

PR #1016, opened 2021-04-30
Rescue RuntimeError in token_for_storage_object to prevent downstream UI errors (SCP-3327)

bistline:
When generating a pet service account token to stream a GCS object back to the client, if an error is thrown during the subsequent FireCloud API call, instead of a token being returned, and the contents of the error message get returned to the UI as the actual token value.  This causes a cascade of errors that will break the UI until the user refreshes the page.

This update explicitly rescues the `RuntimeError` that is raised from all FireCloud API errors, and now returns `nil`, which will then be handled downstream by the UI via `UserProvider` and the two genome visualization components that use this token (Ideogram and IGV).

TO TEST:
1. [Sign in with registeredinscpbutnotterra@gmail.com](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365349124700?thread_ts=1600365317.124600&cid=CBEHTH601) (or any Google account that is not registered for Terra)
2. Run:
```
rails console
user = User.find_by(email: 'registeredinscpbutnotterra@gmail.com')
user.update(registered_for_firecloud: true)
```
3. Call `user.token_for_storage_object` and note the return value of `nil`, and no error is raised (would have been a `401`).
4. Cleanup: `user.update(registered_for_firecloud: false)`

This PR satisfies SCP-3327.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -214,8 +214,14 @@ class User
   # will return nil if user is not registered for Terra as API call would return 401
   def token_for_storage_object(project=FireCloudClient::PORTAL_NAMESPACE)
     if self.refresh_token.present? && self.registered_for_firecloud
-      client = FireCloudClient.new(self, project)
-      client.get_pet_service_account_token(project)
+      begin
+        client = FireCloudClient.new(self, project)
+        client.get_pet_service_account_token(project)
+      rescue RuntimeError => e
+        # returning nil here will be caught at the UI level and show an error message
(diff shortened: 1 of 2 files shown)


devonbush (approved):

eweitz (approved):
Looks good.  I verified locally, and refined the test instructions with more explicit procedural commands.

bistline:
Thanks @eweitz - I couldn't remember if we had shared Google accounts for this.

PR #1020, opened 2021-05-03
optimized cluster endpoint (SCP-3286, SCP-3285, SCP-3284)

devonbush:
This refactors the return values of clusters_controller.rb to be more plain-data, and moves as much visualization logic to the frontend as possible.  This also adds a plotDataCache, to cache coordinate data and cell names across plots. 

TO TEST:
This really implicates all of the scatter plots on the explore tab, so I recommend going through https://docs.google.com/document/d/1CfpDVfmHJPvHISt-e1v90grKu9LrIrtEClxlRQJtFDs/edit and just clicking around a lot, as there could be edge cases I haven't handled around quickly toggling genes/annotations in a certain way


devonbush:
closing as I found bugs which turned out to need a large refactor

PR #1021, opened 2021-05-03
Add Zendesk email to user-facing error modals (SCP-3321, SCP-3332)

bistline:
This update adds the `scp-support@broadinstitute.zendesk.com` email to all user-facing error modal messages.  Note: this update only covers modals where an error has been handled, and the page is redirected back to another action and the `alert` property has been set on the `redirect_to` call in the non-API portion of SCP.  This covers the vast majority of known error modes, but it should be noted that there may be others that are not addressed by this change.  It does cover the main error modes of users attempting to access resources (studies, files, profiles, etc.) they do not have permission to view, or other portions of SCP that require admin access.

Preview:
scp-error-modal

In addition, this will log the `copy` and `contextmenu` events to Mixpanel when a user copies the support email address, either by using the context menu on the anchor tag, or manually using `cmd + C`.

TO TEST:
1. [Sign in with registeredinscpandterra@gmail.com](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365349124700?thread_ts=1600365317.124600&cid=CBEHTH601), or any other non-admin Google account
2. Go to https://localhost:3000/single_cell/admin
3. Note the `scp-support@broadinstitute.zendesk.com` address in the error modal, and that the modal does not close until manually cleared.
4. Open the "Network" tab in Chrome Dev Tools
5. Right click on the support email and select "copy email address"
6. Note in the Network tab the `copy` event posted to Mixpanel with the support email as the `text` property

This PR satisfies SCP-3321 & SCP-3332.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,11 @@ class ApplicationController < ActionController::Base
 
   include RealIpLogger
 
+  # Error modal contact message
+  ZENDESK_CONTACT = "If this error persists and you require assistance, please contact support at " \
+                    "<a href='mailto:scp-support@broadinstitute.zendesk.com' id='zendesk-alert-email'>" \
+                    "scp-support@broadinstitute.zendesk.com</a>."
+
   ###
   #
   # These are methods that are not specific to any one controller and are inherited into all
@@ -71,14 +76,14 @@ class ApplicationController < ActionController::Base
   # auth action for portal admins
   def authenticate_admin
     unless current_user.admin?
-      redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]), alert: 'You do not have permission to access that page.' and return
+      redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]), alert: "You do not have permission to access that page.  #{ZENDESK_CONTACT}" and return
     end
   end
 
   # auth action for portal reporters
   def authenticate_reporter
     unless current_user.acts_like_reporter?
-      redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]), alert: 'You do not have permission to access that page.' and return
+      redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]), alert: "You do not have permission to access that page.  #{ZENDESK_CONTACT}" and return
     end
   end
 
@@ -97,7 +102,7 @@ class ApplicationController < ActionController::Base
     redirect = request.referrer.nil? ? site_path : request.referrer
     access = AdminConfiguration.firecloud_access_enabled?
     if !access
-      redirect_to merge_default_redirect_params(redirect, scpbr: params[:scpbr]), alert: "Study access has been temporarily disabled by the site adminsitrator.  Please contact #{view_context.mail_to('scp-support@broadinstitute.zendesk.com')} if you require assistance." and return
+      redirect_to merge_default_redirect_params(redirect, scpbr: params[:scpbr]), alert: "Study access has been temporarily disabled by the site adminsitrator.  #{ZENDESK_CONTACT}" and return
     end
   end
 
@@ -202,7 +207,7 @@ class ApplicationController < ActionController::Base
 
     if redirect_parameters.any?
       redirect_to merge_default_redirect_params(redirect_parameters[:url], scpbr: params[:scpbr]),
-                  alert: redirect_messages.dig(redirect_parameters[:message_key]) and return
+                  alert: redirect_messages.dig(redirect_parameters[:message_key]) + "  #{ZENDESK_CONTACT}" and return
     end
 
     # next check if downloads have been disabled by administrator, this will abort the download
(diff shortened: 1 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -3,6 +3,11 @@ class ApplicationController < ActionController::Base
 
   include RealIpLogger
 
+  # Error modal contact message
+  ZENDESK_CONTACT = "If this error persists and you require assistance, please contact support at " \
+                    "<a href='mailto:scp-support@broadinstitute.zendesk.com' id='zendesk-alert-email'>" \
+                    "scp-support@broadinstitute.zendesk.com</a>."
+
```
devonbush, line 10: I think we should add a whitespace: no-wrap  css class to this so that the email address never line-wraps

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,11 @@ class ApplicationController < ActionController::Base
 
   include RealIpLogger
 
+  # Error modal contact message
+  ZENDESK_CONTACT = "If this error persists and you require assistance, please contact support at " \
```
devonbush, line 7: IMO the 'and you require assistance' is redundant, but that's a good topic for demo

devonbush: Also, I'd name this constant SCP_SUPPORT_EMAIL_MESSAGE, since it's not just the contact, and who knows when zendesk will get replaced by <<new DSP tech>>

app/javascript/packs/application.js
```diff
@@ -38,6 +38,10 @@ document.addEventListener('DOMContentLoaded', () => {
 
   $(document).on('click', 'body', logClick)
   $(document).on('change', 'select', logMenuChange)
+  $(document).on('copy', 'body', logCopy)
+  // contextmenu event is to handle when users use context menu "copy email address" instead of cmd+C copy event as
+  // this does not emit the copy event
+  $(document).on('contextmenu', '#zendesk-alert-email', logCopy)
```
devonbush, line 44: instead of hardcoding this to the id, I'd create a new functional css class like "log-on-context-copy" or somesuch.  That way we don't have to update this application.js file each time we want to do something like this

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,11 @@ class ApplicationController < ActionController::Base
...
+                    "<a href='mailto:scp-support@broadinstitute.zendesk.com' id='zendesk-alert-email'>" \
```
devonbush, line 8: this should probably be a class, not an id, since we could want to have multiple alerts on a given page

devonbush: also, I could see giving this element a data-analytics-name of "SUPPORT_EMAIL" just so it's easier to find in mixpanel, but I defer to Eric on naming questions

devonbush (changes requested):
Good stuff.  A few polishing suggestions -- the ones I feel most strongly about are not allowing line breaks on the email address, and not hardcoding specific ids inside application.js (other than the root id)

app/controllers/application_controller.rb
eweitz: > a data-analytics-name of "SUPPORT_EMAIL"

Let's use kebab-case, per [convention](https://github.com/broadinstitute/single_cell_portal_core/search?q=data-analytics-name%3D%22).

app/controllers/application_controller.rb
eweitz: Also, a few line breaks between the error message and support email message might help.

app/javascript/packs/application.js
eweitz: Let's log all `contextmenu` events (on `body`, not any more particular selector) in a `logContextMenu` function like `logClick` or `logCopy`.

That would solve two things:
1. Opening the context menu does not necessarily entail selecting "Copy email address" (even on `mailto` links).
2. Additional, not-yet-known use cases for wanting to know when users open their context menu.

Regarding (1), users can open context menu then select "Copy", which emits a `copy` event, and the current instrumentation double counts copies in that.  Logging `contextmenu` separately would us have refined downstream handling for that in Mixpanel queries.

eweitz (changes requested):
This is coming along well!

Modulo Devon's, my only blocking concern is with `contextmenu` event logging.

app/javascript/packs/application.js
bistline: > Regarding (1), users can open context menu then select "Copy", which emits a copy event, and the current instrumentation double counts copies in that case.

My testing has shown this does not in fact happen - selecting the "Copy" option from a context menu does not emit a copy event.  The JS copy event is only emitted by `cmd + C`.  I think it has something to do with the `contextmenu` event taking precedence during the right-click interaction.

app/controllers/application_controller.rb
bistline: Can do.

bistline: 390a54e

app/controllers/application_controller.rb
bistline: 390a54e

app/javascript/packs/application.js
bistline: 390a54e

app/controllers/application_controller.rb
bistline: 390a54e

devonbush (approved):

app/javascript/packs/application.js
eweitz: Hmm, I still see `copy` events logged to Mixpanel when selecting the "Copy" option from a context menu.  But my blocking concern is resolved, because I verified your `contextmenu` handler change addresses my main point!

(If you confirm that "Copy" via context menu _doesn't_ log, then that strikes me as a bug, but I can't reproduce it.)

eweitz (approved):

app/javascript/packs/application.js
bistline: Ah - I see what is happening.  There is a distinction between selecting "Copy" and "Copy email address".  The former emits the `copy` event, and the latter does not.  I was always using "Copy email address".  I don't believe that there is a way for us to prevent this from happening as this is all native browser behavior.

PR #1022, opened 2021-05-03
relaxing template rules to prevent eslint breakage (SCP-3336)

devonbush:
See https://github.com/babel/babel-eslint/issues/681.  This sidesteps a eslint/babel version issue that has broken linting in sublime and VSCode from time to time.  Since we don't care that much about formatting rules for inside of backticks, better to avoid this issue than have it keep resurfacing

TO TEST:
 0. open DotPlot.js in Sublime/VSCode
 1. confirm syntax highlighting works

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,6 +23,7 @@ module.exports = {
         'ga': 'readonly',
         'bamAndBaiFiles': 'readonly'
     },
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Looks good, verified locally.

bistline (approved):
Nice to see Codecov is working again as well...

PR #1024, opened 2021-05-05
Log network and detailed client performance metadata, refine related metrics (SCP-3238)

eweitz:
This enhances performance timing metrics for SCP API calls to log finer-grained `perfTime` metadata.  It also refines tracking of coarser-grained `perfTime` metadata.  These changes will help systematically assess how much various processing segments contribute to overall performance, and thus better quantify the impact of optimizations, e.g with scatter plots.

Below are the `perfTime` segments we'll log to Mixpanel with these changes.

### Total timing
* _Full_: API call start to plot render end
* _Legacy_: API call start to end time plus plot time.  Inadequate because it excludes some client timing segments ("parse" and "other"; see below), and assumes the logged event is a plot.  This is a [forward-compatible](https://en.wikipedia.org/wiki/Forward_compatibility) name for old `perfTime` values, if/when they are replaced with preferable `perfTime:full` values.
* _`perfTime`_: Currently same as `perfTime:legacy`, for continuity.  Enables historical analysis without patching analytics queries in Mixpanel.

### Server timing
* _Backend_: Time for server to process request.  Previously included transfer time, which was a poor fit as that is in the client's domain.  (In an upcoming PR, coarse segments of backend processing time will be reported via the [Server Timing API](https://medium.com/bbc-design-engineering/server-timing-in-the-wild-bfb34816322e).)

### Client timing
* _Frontend_: API call response start to plot render end
* _Frontend, transfer_: Response start to response end; time to download entire response from server to client
* _Frontend, plot_: Start to end of plot render call.  (Only logged for plots.)
* _Frontend, parse_: Start to end of call to parse response body (currently only JSON)
* _Frontend, other_: Time for otherwise-unaccounted for client processing segments

### Data
* _Compressed size_: Encoded response body size
* _Uncompressed size_: Decoded response body size
* _Compression bytes diff_: Absolute amount compressed
* _Compression ratio_: Relative amount compressed

All time values are in milliseconds and all data values (except compression ratio) are in bytes.

Here's what the logged values look like in Mixpanel:
Network_performance_metadata_and_refined_perfTime_properties_SCP_2021-05-05

To test:
1. Go to a study with visualizations
2. Go to [Live View in Mixpanel dev](https://mixpanel.com/report/2085496/view/19055/live)
3. Search "plot:scatter", expand row
4. Confirm `perfTime` properties include those shown in above screenshot, e.g. `perfTime:frontend:transfer` and `perfTime:frontend:parse`.

This satisfies SCP-3238.


bistline (approved):
Really great - looking forward to having this data available.

app/javascript/lib/metrics-perf.js
```diff
@@ -138,3 +138,93 @@ export function setupWebVitalsLog() {
...
+  let compressionRatio = uncompressedSize / compressedSize
+  // Round to 2 digits, e.g. "3.44".  Number.EPSILON ensures numbers like 1.005 round correctly.
+  compressionRatio = Math.round((compressionRatio + Number.EPSILON) * 100) / 100
+  perfProps['perfTime:data:compression-ratio'] = compressionRatio // Relative amount compressed
+
+  // Accounts for `null`, '', 'non-empty string', etc.
+  const errorKeys = Object.keys(rawPerfProps).filter(k => isNaN(parseFloat(rawPerfProps[k])))
+  if (errorKeys.length > 0) {
+    const specifics = errorKeys.map(k => `${k}: ${rawPerfProps[k]}\n`)
+    throw Error(
```
devonbush, line 214: I think this Error throwing should just be a log('metrics-error', ...) or somesuch, followed by a no-op where it just doesn't try to add the perf numbers.   I don't want the UX thread to break due to a logging irregularity

devonbush (changes requested):
Looks good, nice addition of a lot of detail while keeping a relatively clean API into the main code.  Just one blocker on the error handling

app/javascript/lib/metrics-perf.js
eweitz: Good ideas, done!

app/javascript/lib/metrics-perf.js
```diff
@@ -138,3 +138,97 @@ export function setupWebVitalsLog() {
...
+  }
+  const perfProps = roundValues(Object.assign({}, rawPerfProps))
+
+  let compressionRatio = uncompressedSize / compressedSize
+  // Round to 2 digits, e.g. "3.44".  Number.EPSILON ensures numbers like 1.005 round correctly.
+  compressionRatio = Math.round((compressionRatio + Number.EPSILON) * 100) / 100
+  perfProps['perfTime:data:compression-ratio'] = compressionRatio // Relative amount compressed
+
+  // Accounts for `null`, '', 'non-empty string', etc.
+  const errorKeys = Object.keys(rawPerfProps).filter(k => isNaN(parseFloat(rawPerfProps[k])))
```
devonbush, line 212: As discussed, move this to above line 204, so it's closer to where the possible error occurred.

devonbush (approved):
Looks good once you make the change we agreed on

app/javascript/lib/metrics-perf.js
eweitz: Done in f1f273a.

PR #1027, opened 2021-05-07
hiding scatter color control for non-numeric annotations

devonbush:
This hides the color selector if it is not applicable to the current displayed plots (if it is just a cluster plot of a non-numeric annotation).  The value is still preserved though, so if the user is toggling back and forth, they don't have to reenter their choice

TO TEST:
1. open study "Male mouse brain"
2. observe that "Scatter" section with "Continuous color scale" dropdown menu is NOT shown when the default cluster/annotation are shown
3.  switch to the 'intensity' annotation -- observe the scatter color picker is restored
4.  search a single gene, e.g. "Apoe".  Observe the scatter color picker is not displayed on the annotated scatter, but is displayed on the scatter tab


bistline (approved):

eweitz (approved):
Looks good!  I verified locally, and refined test instructions.

PR #1028, opened 2021-05-11
Ease discovery by adding labels in related genes ideogram (SCP-3007)

eweitz:
This makes it easier to discover genes related to a gene of interest, by labeling them in related genes ideogram.

Previously, users had to hover over related genes to see their name and other information.  Now, the names are usually shown by default.  This lowers the interaction cost for discovery, while keeping the engaging biological visualization.  It also fulfills a user request made when earlier demonstrating the component.

Previously, related genes ideogram looked like this:
Related_genes_ideogram_with_no_labels_2021-05-11

Now it looks like this:
Related_genes_ideogram_with_labels_2021-05-11

Per another user request at a demo of a prototype of this enhancement, the active feature has some highlighting around the label and triangle -- a UI detail influenced by Google Maps:
Tooltip_in_related_genes_ideogram_with_labels_2021-05-11

Sometimes, many features appear, and labeling them all would cause overlaps that greatly degrade readability.  This is avoided by omitting certain feature labels when they would overlap.

To test:
* Go to a study that has visualizations and a typical number of genes (not a synthetic study)
* Search a gene
* Verify label for searched gene and related genes
* Hover over a related gene
* Verify that active feature label is highlighted

This satisfies SCP-3007.

### DIFF
--- a/app/javascript/components/visualization/RelatedGenesIdeogram.js
+++ b/app/javascript/components/visualization/RelatedGenesIdeogram.js
@@ -147,7 +147,6 @@ export default function RelatedGenesIdeogram({
       chrHeight: height - verticalPad,
       chrLabelSize: 12,
       annotationHeight: 7,
-      showAnnotLabels: false,
       onClickAnnot,
       onPlotRelatedGenes,
       onWillShowAnnotTooltip,
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "camelcase-keys": "^6.1.2",
     "core-js": "^3.6.4",
-    "ideogram": "1.29.0",
+    "ideogram": "1.30.0",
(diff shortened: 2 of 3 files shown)


bistline (approved):
Very slick.

devonbush (approved):
Code looks simple enough.  I have some concerns that the text display is too 'loud' with colors.  Note how, for example, on google search, 'related searches' is actually one of the least colorful parts of the page.   
image

but that's an issue more for Jerome/Jean than me.  I'm good to merge this once they sign off.

eweitz:
Jean and Jerome approved, so this is good to merge per Devon's rubric.

The consensus at demo today seemed to be that the visual noise mitigation techniques are currently decent and (corroborating Devon's point) worth refining, but not concerning enough to outweigh the added value of this more glanceable UI.  Another key point others noted was that releasing this now gives us a much better opportunity for learning from users.

PR #1033, opened 2021-05-12
Change references to master/slave/black/white to more appropriate terms (SCP-3345)

bistline:
In line with the industry-wide move away from computing terms that have a racial context, this update renames methods/variables that contain references to the above terms with one of the following replacements:

* `master`: `main`, or `parent`
* `slave`: `worker`, or `child`
* `black` (used to denote denial/restriction, rather than the actual color): `deny`
* `white` (used to denote allowed/accepted, rather than the actual color): `allow`, or `permit`

There is an associated ticket (SCP-3346) to rename the `master` branch in all SCP Github repositories to `main`.  This work impacts all of our automation for CI/deployments, so is not included in this update.

This PR satisfies SCP-3345.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,10 +106,10 @@ to log in with their Google accounts.  To do so:
   * Save the client id
   * You will also want to create a second OAuth Client ID to use in local development, using `localhost` as the hostname
 
-* **Whitelisting your OAuth Audience**
-  * Once you have exported your OAuth credentials, you will need to have your client id whitelisted to allow it to make
+* **Authorizing your OAuth Audience**
+  * Once you have exported your OAuth credentials, you will need to have your client id authorized to allow it to make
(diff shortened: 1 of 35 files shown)


devonbush (approved):
Nice!  thanks for making the effort on this.  I'm approving, although I'd recommend we don't merge this until after the release tomorrow, since it touches a bunch of files and is non-urgent.  (That said, it's the release engineer's call

eweitz (approved):

bistline:
> Nice! thanks for making the effort on this. I'm approving, although I'd recommend we don't merge this until after the release tomorrow, since it touches a bunch of files and is non-urgent. (That said, it's the release engineer's call

I've already decided that I want to hold this out of tomorrow's release.  I expect there to be regressions given the amount of refactoring that happened, so I want to see what build says.  Even if it's green, I still want to poke around a little more.

PR #1035, opened 2021-05-13
Add feature flag for cross dataset search

eweitz:
This enables easy development of cross dataset search, in parallel with existing code, by adding a feature flag.

To test:
1. Do migration: run `ruby rails_local_setup.rb`, then `source config/secrets/.source_env.bash`, then `rails db:migrate`
2. Load home page, verify tab says "Search Studies"
3. Set feature flag for yourself: go to https://localhost:3000/single_cell/admin#users, select your account, then set "cross_dataset_search" to "True".
4. Load home page, verify tab says "Search Studies (XDS feature flag)"

This satisfies SCP-3280.

### DIFF
--- a/app/views/site/index.html.erb
+++ b/app/views/site/index.html.erb
@@ -29,7 +29,11 @@
         Ensure any updates to this ID are reflected in the ReactDOM.render call
         in app/javascript/packs/application.js
       #%>
-      <div id="home-page-content"></div>
+      <% if User.feature_flag_for_instance(current_user, 'cross_dataset_search') && @selected_branding_group.nil? %>
+        <div id="home-page-content-xds"></div>
+      <% else %>
+        <div id="home-page-content"></div>
+      <% end %>
     </div>
   </div>
 </div>
--- a/app/javascript/components_xds/HomePageContent.js
+++ b/app/javascript/components_xds/HomePageContent.js
@@ -1,0 +1,89 @@
+
+import React, { useContext } from 'react'
+import { Router, Link, useLocation } from '@reach/router'
+
+import GeneSearchView from 'components/search/genes/GeneSearchView'
+import GeneSearchProvider from 'providers/GeneSearchProvider'
+import SearchPanel from 'components/search/controls/SearchPanel'
+import ResultsPanel from 'components/search/results/ResultsPanel'
+import StudyDetails from 'components/search/results/Study'
+import StudySearchProvider, { StudySearchContext } from 'providers/StudySearchProvider'
+import SearchFacetProvider from 'providers/SearchFacetProvider'
+import UserProvider from 'providers/UserProvider'
+import ErrorBoundary from 'lib/ErrorBoundary'
+import * as queryString from 'query-string'
+
+/** include search controls and results */
+export function StudySearchView({ advancedSearchDefault }) {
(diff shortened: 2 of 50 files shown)


bistline (approved):

devonbush (changes requested):
the 'components' directory is conventionally used to hold all components.  So I would make an 'xds_search' directory inside the components folder, rather than a new top-level folder.  It might be worth then moving the current HomePageContent.js file inside the 'search' folder, so it's clearer what belongs to what.  Alternatively, I could see making an 'xds' folder inside the 'components/search' directory.  Either way makes sense

app/views/site/index.html.erb
```diff
@@ -29,7 +29,11 @@
         Ensure any updates to this ID are reflected in the ReactDOM.render call
         in app/javascript/packs/application.js
       #%>
-      <div id="home-page-content"></div>
+      <% if User.feature_flag_for_instance(current_user, 'cross_dataset_search') && @selected_branding_group.nil? %>
```
devonbush, line 32: I think this is already showing we might need two feature flags -- one for XDS UX, and one for XDS backend.  Because we'll want the new XDS UX to work with branded spaces -- we just don't want it to then allow actual XDS search.

app/views/site/index.html.erb
```diff
@@ -29,7 +29,11 @@
...
+        <div id="home-page-content-xds"></div>
+      <% else %>
+        <div id="home-page-content"></div>
```
devonbush, line 35: For the frontend flag, I'd prefer the feature flag branch to be controlled from JS, rather than rails.  This both lets us have less logic inside rails templates, and fewer 'magic strings' that have to be shared between rails and our JS.   So I'd create a new, trivial component that just handles reading the feature flag and rendering the appropriate search content.  (then you may want to rename the existing HomePageContent.js file to 'ClassicSearch.js' or somesuch, so that the new, trivial, component can just be called "HomePageContent"

devonbush (changes requested):

app/views/site/index.html.erb
eweitz: I've split the one XDS flag into frontend and backend flags in bc3e5d3.

app/views/site/covid19.html.erb
```diff
@@ -36,7 +36,7 @@
         <a href="#" id="covid-contact-link">contact us</a>.
        </span>
       <div class="tab-content top-pad">
-        <div id="home-page-content"></div>
+        <div id="home-view"></div>
```
devonbush, line 39: Is this supposed to have been changed? I don't see a corresponding hook for this id anywhere in the PR

devonbush (changes requested):
I wonder if some changes got indavertently left out of this PR -- there isn't a migration script here anymore?  
JS code looks good, with the exception of what might be a hanging change to an id

app/views/site/covid19.html.erb
eweitz: Good catch -- I had changed this in both indexes, decided against it, and rolled back one index but not the other.

eweitz (commented):
The moved migrations did indeed get inadvertently omitted.  They're here now!

devonbush (approved):

PR #1036, opened 2021-05-13
Partial frontend caching of cluster responses (SCP-3284, SCP-3285, SCP-3286, SCP-2957)

devonbush:
SCP-3284, SCP-3285, SCP-3286, SCP-2957

Provides a transparent caching mechanism for calls to fetchCluster.  For each call, it attempts to cache three separate things from the response:
    cellsAndCoords: the coordinates and cell names of the cluster
    annotations: an array of annotations
    expression: an array of expression values
  
Prior to a new call, it checks the cache if any of those are available for the given params.  If so, it removes those fields from the list of fields to request from the server.  Once the response from the server comes back, it handles merging in all the cached fields into the response.

See the top of `plotDataCache.js` for full description.  

TO TEST:  This is pretty far-ranging in its impact.  So use https://docs.google.com/document/d/1CfpDVfmHJPvHISt-e1v90grKu9LrIrtEClxlRQJtFDs/edit, and make special effort to mix in very fast clicks (e.g. very quickly changing genes/annotations before some graphs finish loading)

TO TEST SCP-2957:  Load a study that has all zeros for a given gene expression (e.g. Male mouse brain, gene 'Adcy5').  Confirm that the coloring of the expression graph is no longer misleading.


app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -125,6 +155,13 @@ def self.get_cluster_viz_data(study, cluster, url_params)
           subsample = get_selected_subsample_threshold(url_params[:subsample], cluster)
           consensus = url_params[:consensus].blank? ? nil : url_params[:consensus]
 
+          default_data_fields = ['coordinates', 'expression', 'annotation', 'cells']
```
bistline, line 158: This seems like it should be a constant for the controller as I expect it shouldn't change: `DEFAULT_DATA_FIELDS`

app/javascript/components/explore/plotDataCache.js
```diff
@@ -0,0 +1,278 @@
...
+  /** get the data for a given study/cluster */
+  cache._putEntry = (accession, clusterName, subsample, entry) => {
+    const key = getKey(accession, clusterName, subsample)
+    cache.entries[key] = entry
+  }
+
+  /** based on cache contents and desired return values, returns an array suitable for the 'fields' argument
+    * to fetchCluster in scp-api, and an array of promises for any in-flight requests relating to the needed data
+     */
+  cache._getFieldsToRequest = ({
```
bistline, line 226: The structure of this method seems overly complicated... I'm wondering if it can be split up into some smaller helper methods to reduce the cyclomatic complexity.  I'm having a hard time reading it with all of the nested if/else logic.

app/javascript/components/explore/plotDataCache.js
```diff
@@ -0,0 +1,278 @@
+import { fetchCluster } from 'lib/scp-api'
```
bistline, line 1: Since there's a _lot_ going on here, it would be great if each method had standard docstrings with params/returns/etc.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -172,96 +191,204 @@ const ScatterPlot = withErrorBoundary(RawScatterPlot)
...
+    return acc
+  }, {})
+}
+
+/** sorts the scatter plot legend alphabetically
+  this is super-hacky in that it relies a lot on plotly internal classnames and styling implementation
+  it should only be used as a last resort if https://github.com/plotly/plotly.js/pull/5591
+  is not included in Plotly soon
+ */
+function sortLegend({ graphElementId, annotValues, isAnnotatedScatter, annotType, gene }) {
```
bistline, line 289: Since the legend is auto-generated, and the sort order is determined by the input order of the traces (for group-based annotations), couldn't we just do a lexical sort of the traces by their name once we have all the data assembled?

app/lib/cluster_viz_service.rb
```diff
@@ -174,47 +174,64 @@ def self.compute_aspect_ratios(range)
...
-      range[:z] = absolute_range
-    end
-    range
-  end
-
   # generic method to populate data structure to render a cluster scatter plot
   # uses cluster_group model and loads annotation for both group & numeric plots
   # data values are pulled from associated data_array entries for each axis and annotation/text value
-  def self.load_cluster_group_data_array_points(study, cluster, annotation, subsample_threshold=nil, colorscale=nil)
+  def self.load_cluster_group_data_array_points(study, cluster, annotation, subsample_threshold=nil, include_coords: true, include_cells: true, include_annotations: true)
```
bistline, line 180: This is starting to venture into the "too many parameters" range, even though they're keyword-based.  7 params seems like a lot.

app/lib/cluster_viz_service.rb
```diff
@@ -174,47 +174,64 @@ def self.compute_aspect_ratios(range)
...
+    x_array = []
+    y_array = []
+    z_array = []
+    cells = []
+    annotation_array = []
+
+    if annotation[:scope] == 'user'
+      user_annotation = UserAnnotation.find(annotation[:id])
+      subsample_annotation = user_annotation.formatted_annotation_identifier
+      if include_coords
```
bistline, line 193: I think this can all be simplified:
```
data_source = cluster
if annotation[:scope] == 'user'
  user_annotation = UserAnnotation.find(annotation[:id])
  subsample_annotation = user_annotation.formatted_annotation_identifier
  data_source = user_annotation
end
if include_coords
  x_array = data_source.concatenate_user_data_arrays('x', 'coordinates', subsample_threshold, subsample_annotation)
  y_array = data_source.concatenate_user_data_arrays('y', 'coordinates', subsample_threshold, subsample_annotation)
  z_array = data_source.concatenate_user_data_arrays('z', 'coordinates', subsample_threshold, subsample_annotation)
end
if include_cells || annotation[:scope] == 'study'
  # for study annotations, we have to grab the cluster cells to match with the study-wide annotation
  cells = data_source.concatenate_data_arrays('text', 'cells', subsample_threshold, subsample_annotation)
end
```

app/lib/cluster_viz_service.rb
```diff
@@ -225,106 +242,13 @@ def self.load_cluster_group_data_array_points(study, cluster, annotation, subsam
...
-      x_array = user_annotation.concatenate_user_data_arrays('x', 'coordinates', subsample_threshold, subsample_annotation)
-      y_array = user_annotation.concatenate_user_data_arrays('y', 'coordinates', subsample_threshold, subsample_annotation)
-      z_array = user_annotation.concatenate_user_data_arrays('z', 'coordinates', subsample_threshold, subsample_annotation)
-      cells = user_annotation.concatenate_user_data_arrays('text', 'cells', subsample_threshold, subsample_annotation)
     else
       # for study-wide annotations, load from study_metadata values instead of cluster-specific annotations
       metadata_obj = study.cell_metadata.by_name_and_type(annotation[:name], annotation[:type])
       annotation_hash = metadata_obj.cell_annotations
-      annotation[:values] = AnnotationVizService.sanitize_values_array(metadata_obj.values, annotation[:type])
+      annotation_array = cells.map { |cell| annotation_hash[cell] }
```
bistline, line 249: We still need to call `AnnotationVizService.sanitize_values_array` here so that we don't end up with blank values (that method interpolates `--Unspecified--` in for blank entries)

app/lib/expression_viz_service.rb
```diff
@@ -107,121 +107,35 @@ def self.load_expression_boxplot_data_array_scores(study, gene, cluster, annotat
...
-    annotation_array = AnnotationVizService.sanitize_values_array(annotation_array, annotation[:type])
-    values = {}
-    values[:all] = {x: [], y: [], cells: [], annotations: [], text: [], marker: {size: study.default_cluster_point_size,
-      line: { color: 'rgb(40,40,40)', width: study.show_cluster_point_borders? ? 0.5 : 0}}}
-    if annotation[:scope] == 'cluster' || annotation[:scope] == 'user'
-      annotation_array.each_with_index do |annot, index|
-        annotation_value = annot
-        cell_name = cells[index]
-        expression_value = gene['scores'][cell_name].to_f.round(4)
+    annotation_array = ClusterVizService.get_annotation_values_array(study, cluster, annotation, cells, subsample_annotation, subsample_threshold)
```
bistline, line 115: Same concern as above with not sanitizing this array.

bistline (commented):
Great start!  Really liking where this is going.  I've put in some comments for areas that I think can be improved.

app/javascript/components/visualization/ScatterPlot.js
devonbush: I wish -- that's what we used to do.  Now we only send a single trace to plotly, and plotly's groupBy transformation handles splitting it -- so we don't have access to the subsequent generated traces.  So plotly determines the order based on the order the annotations are encountered in the *individual cells*, which would be extremely difficult to reorder.

app/javascript/components/explore/plotDataCache.js
devonbush: I added a bunch more function-level comments.  I'm hesitant to go full docstring, because there are a *lot* of small methods that all pass around the same params, and to me, the readability suffers if  `@param cluster (string) - the cluster name` gets repeated everywhere. But I'll defer to you and Eric on what you think

app/lib/cluster_viz_service.rb
devonbush: NICE!

app/lib/cluster_viz_service.rb
devonbush: yes, that gets taken care of just before the return

app/lib/expression_viz_service.rb
devonbush: The sanitization happens inside that utility function

app/lib/cluster_viz_service.rb
devonbush: Yeah, not really sure what to do about it -- I could condense the include_* params into a single include object, but I think that makes it less clear.

eweitz (changes requested):
Exciting stuff!  This seems like a huge optimization for single-gene search in Explore, and certain other use cases.

To help soundly quantify impact, we'll need to carefully code around `perfTime` logging as explained below.  I've prefixed those issues with "Blocker" for convenience.

Beyond those blocking functional problems, I suggest a bevy of non-blocking refinements for variable names, style, etc.

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -13,6 +14,7 @@ import CreateAnnotation from 'components/visualization/controls/CreateAnnotation
 import PlotDisplayControls from 'components/visualization/PlotDisplayControls'
 import GeneListSelector from 'components/visualization/controls/GeneListSelector'
 import InferCNVIdeogramSelector from 'components/visualization/controls/InferCNVIdeogramSelector'
+import { newCache } from './plotDataCache'
```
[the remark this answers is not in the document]
devonbush: I've changed to `createCache` to make it more verby

app/javascript/components/explore/plotDataCache.js
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+      apiCallPromise = Promise.resolve([
+        {
+          genes,
+          consensus,
+          cluster,
+          subsample,
+          annotParams: annotation,
+          data: {},
+          isPureCache: true // set a flag indicating that no fresh request to the server was needed
```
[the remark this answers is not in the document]
eweitz: With this name, I foresee maintainers likely wondering (and falsely assuming):

* Which cache, [server](https://github.com/broadinstitute/single_cell_portal_core/pull/1029)?  Nope, client.  (This is less concerning if we use a different prop name, like `cache: "client/server/none"`, for logging how/whether calls are cached.)

* Is the value giving me all the data that's in cache?  Instead, it's a Boolean.

app/javascript/components/explore/plotDataCache.js
```diff
@@ -0,0 +1,302 @@
...
+      Object.assign(scatter, entry.clusterProps)
+      return clusterProps
+    }
+  }
+}
+
+/**
+ * Get a fresh, empty cache.
+ */
+export function newCache() {
```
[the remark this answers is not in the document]
eweitz: They note [classes can save memory better than objects](https://gist.github.com/indiesquidge/f8c486795d7dd455c0327ce7e0aa8c16#pitfalls-of-prototypal-inheritance).  Is that relevant here?

In any case, that gist is a great read -- thanks for sending it my way.  This approach works.

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -181,28 +216,28 @@ def self.get_cluster_viz_data(study, cluster, url_params)
...
             titles: titles,
-            ranges: range,
             aspects: aspect
           }
 
           coordinate_labels = ClusterVizService.load_cluster_group_coordinate_labels(cluster)
           {
             "data": plot_data,
+            "pointSize": study.default_cluster_point_size,
+            "clusterSpecifiedRanges": cluster.domain_ranges,
```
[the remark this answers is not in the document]
eweitz: Hmm, the discerning principle still seems murky -- it seems applicable to several other properties here that aren't prefixed with `userSpecified`.

app/lib/cluster_viz_service.rb
```diff
@@ -174,47 +174,59 @@ def self.compute_aspect_ratios(range)
...
+    y_array = []
+    z_array = []
+    cells = []
+    annotation_array = []
+
+    data_source = cluster
+    if annotation[:scope] == 'user'
+      user_annotation = UserAnnotation.find(annotation[:id])
+      subsample_annotation = user_annotation.formatted_annotation_identifier
+      data_source =user_annotation
```
bistline, line 194: I just tested this and realized it doesn't work, because the method in `UserAnnotation` is actually called `concatenate_user_data_arrays`, not `concatenate_data_arrays`.  So this either should revert back to what it was before, or that method in `UserAnnotation` should be renamed.  Either solution would be fine.

app/lib/cluster_viz_service.rb
bistline: Ah, I see it now.  Small quibble - we're reassigning the variable to itself after sanitizing it.  It doesn't really make a difference from a functional standpoint, but I think it's safer to rename the sanitized array something else.

test/js/explore/plot-data-cache.test.js
```diff
@@ -0,0 +1,329 @@
...
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('Plot data cache', () => {
+  it('caches a single cluster call for use when annotation and gene are changed', async () => {
+    const cache = createCache()
+    const apiFetch = jest.spyOn(ScpApi, 'fetchCluster')
+    // pass in a clone of the response since it may get modified by the cache operations
+    apiFetch.mockImplementation(() => Promise.resolve([_cloneDeep(FETCH_CLUSTER_RESPONSE), 230]))
```
bistline, line 85: What does the `230` map to?  Is that the `perfTime` calculation?

bistline (changes requested):
This looks good, but there is the blocking bug with my suggestion for cleaning up `ClusterVizService.load_cluster_group_data_array_points`.  Also, I have noticed an issue when testing locally.  If you load a cluster/annotation that has blank values (either from an unmapped cell, or a blank entry in a metadata file), then the scatter plot looks off:

Screen Shot 2021-05-18 at 10 17 35 AM

As you can see, the points aren't the same size as all the rest (from the study default options), and the DNA spinner never goes away.  Definitely a corner case, as we don't have that many of these in the portal, but something that should be looked at and addressed.

app/javascript/components/explore/plotDataCache.js
```diff
@@ -0,0 +1,302 @@
...
+          consensus,
+          cluster,
+          subsample,
+          annotParams: annotation,
+          data: {},
+          isPureCache: true // set a flag indicating that no fresh request to the server was needed
+        }, {
+          url: 'cache',
+          legacyBackend: 0, // a backend time of zero since this request will never go to the server
+          parse: 0
```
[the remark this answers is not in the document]
eweitz: Thanks for the insightful chat.  In summary:

1.  We agree that omitting props for perfTime segments that don't happen can be accounted for, by filtering analytics queries with logged props like `cache: "client/server/none"`.  That approach makes analysis harder (but not impossible) and simplifies instrumentation, so it's net reasonable.

2.  We'll need to be careful that omitted props don't skew summary metrics, e.g. definitely not `perfTimeFull`, and ideally not `perfTime:frontend` or `perfTime:backend`.

I'll re-review once those takeaways are implemented here or ticketed for this sprint.

eweitz (commented):
Thanks for the insightful chat on [zero vs. undefined](https://github.com/broadinstitute/single_cell_portal_core/pull/1036#discussion_r633822328). In summary:

We agree that omitting props for `perfTime` segments that don't happen can be accounted for when analyzing within a segment (e.g. "How has transfer performance changed over time?") and across segments ("What's the principal bottleneck in performance?"), by filtering analytics queries with logged props like `cache: "client/server/none"`.  That approach makes analysis harder (but not impossible) and simplifies instrumentation, so it's net reasonable.

We'll need to be careful that omitted props don't skew summary metrics, e.g. definitely not `perfTime:full`, and ideally not `perfTime:frontend` or `perfTime:backend`.  For example, we'll want to ensure that when transfers don't happen (due to caching), we properly account for that in `perfTime:full` and ideally e.g. `perfTime:frontend`.

I'll re-review once those takeaways are implemented here or ticketed for this sprint.

test/js/explore/plot-data-cache.test.js
devonbush: yeah, so for purposes of this test it just needs to be any number/object

app/lib/cluster_viz_service.rb
devonbush: Good point, since this is at the end of the function, we actually don't need a variable for it at all

app/lib/cluster_viz_service.rb
devonbush: good catch -- I've just renamed it to concatenate_data_arrays for consistency

devonbush:
@bistline  I can't reproduce the missing point issue -- I can render clusters with 'unspecified' cells ok.  I wonder if there's something specific about the structure in the cluster you're using that brings out the bug.  Let's screenshare tomorrow -- since the spinner is still going, chances are good there'll be an informative error in the JS console

bistline (approved):

eweitz (approved):

PR #1037, opened 2021-05-17
fixing display of coordinate labels in 3d plots (SCP-3363)

devonbush:
TO TEST:
1. load a study with a 3d plot and coordinate labels -- e.g. the now-updated human raw counts study (note you'll have to repopulate it so you get the coordinate labels added here)
2. Confirm the coordinate labels appear


bistline (approved):
Glad it was so straightforward!

eweitz (approved):

PR #1044, opened 2021-05-20
Terra Data Repo API client scaffold (SCP-3353)

bistline:
This update adds a minimal client for [Terra Data Repo API](https://jade-terra.datarepo-prod.broadinstitute.org/swagger-ui.html) bindings.  This will be used as the scaffold on which all future "cross-dataset search" work can be done.

The client is structured very similarly to all of the other external API client classes: `FireCloudClient` and `PapiClient`.  Currently, the client has methods to retrieve datasets and snapshots, as well as abstract methods for making requests and processing responses.

This PR satisfies parts of SCP-3353.

### DIFF
--- a/app/models/data_repo_client.rb
+++ b/app/models/data_repo_client.rb
@@ -66,11 +66,13 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   #
   # * *return*
   #   - +DateTime+ timestamp of new access token expiration
-  def refresh_access_token
+  def refresh_access_token!
+    Rails.logger.info "DataRepoClient token expired, refreshing access token"
     new_token = self.class.generate_access_token(self.service_account_credentials)
     new_expiry = Time.zone.now + new_token['expires_in']
     self.access_token = new_token
     self.expires_at = new_expiry
+    new_token
   end
 
   # check if an access_token is expired
@@ -86,7 +88,7 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   # * *return*
   #   - +Hash+ of access token
   def valid_access_token
-    self.access_token_expired? ? self.refresh_access_token : self.access_token
+    self.access_token_expired? ? self.refresh_access_token! : self.access_token
   end
 
   # get issuer of storage credentials
@@ -105,14 +107,10 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   def process_api_request(http_method, path, payload: nil, retry_count: 0)
     # Log API call for auditing/tracking purposes
     Rails.logger.info "Terra Data Repo API request (#{http_method.to_s.upcase}) #{path}"
-    # check for token expiry first before executing
-    if self.access_token_expired?
-      Rails.logger.info "DataRepoClient token expired, refreshing access token"
-      self.refresh_access_token
-    end
+
     # set default headers, appending application identifier including hostname for disambiguation
     headers = {
-      'Authorization' => "Bearer #{self.access_token['access_token']}",
+      'Authorization' => "Bearer #{self.valid_access_token['access_token']}",
       'Accept' => 'application/json',
       'x-app-id' => "single-cell-portal",
       'x-domain-id' => "#{ENV['HOSTNAME']}"
--- a/app/models/data_repo_client.rb
+++ b/app/models/data_repo_client.rb
@@ -2,22 +2,14 @@
 # API client bindings for retrieving information about Terra Data Repo datasets/snapshots/schemas from their API
 ##
 class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_at, :service_account_credentials)
-  extend AccessTokenGenerator # class method for generating OAuth2 access tokens
-  include GoogleServiceClient # helper module for refreshing access tokens, accessing GCS storage driver attributes
-  include ApiHelpers # process external API requests/responses
+  extend ServiceAccountManager
+  include GoogleServiceClient
+  include ApiHelpers
 
-  # path to read-only service account JSON keyfile
-  SERVICE_ACCOUNT_KEY = !ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'].blank? ? (ENV['NOT_DOCKERIZED'] ? ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'] : File.absolute_path(ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'])) : ''
   # Google authentication scopes necessary for querying TDR API
   GOOGLE_SCOPES = %w(openid email profile)
-  # GCP Compute project
-  COMPUTE_PROJECT = ENV['GOOGLE_CLOUD_PROJECT'].blank? ? '' : ENV['GOOGLE_CLOUD_PROJECT']
-  # Base API URL to request against
+   # Base API URL to request against
   BASE_URL = Rails.application.config.tdr_api_base_url
-  # constant used for retry loops in process_firecloud_request and execute_gcloud_method
-  MAX_RETRY_COUNT = 5
-  # constant used for incremental backoffs on retries (in seconds); ignored when running unit/integration test suite
-  RETRY_INTERVAL = Rails.env.test? ? 0 : 15
 
   # control variables
   SORT_DIRECTIONS = %w(asc desc).freeze
@@ -30,16 +22,16 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   ##
 
   # initialize is called after instantiating with DataRepoClient.new
-  # will set the access token, TDR base api url root and GCP storage driver instance
+  # will set the access token, TDR base API URL  root and GCP storage driver instance
   #
   # * *params*
   #   - +service_account_key+: (String, Pathname) => Path to service account JSON keyfile
   # * *return*
   #   - +DataRepoClient+ object
-  def initialize(service_account=SERVICE_ACCOUNT_KEY)
+  def initialize(service_account=self.class.get_primary_keyfile)
     # GCS storage driver attributes
     storage_attr = {
-      project: COMPUTE_PROJECT,
+      project: self.class.compute_project,
       timeout: 3600,
       keyfile: service_account
     }
@@ -78,11 +70,11 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
       context = " encountered when requesting '#{path}', attempt ##{current_retry}"
       log_message = "#{e.message}: #{e.http_body}; #{context}"
       Rails.logger.error log_message
-      retry_time = retry_count * RETRY_INTERVAL
+      retry_time = retry_count * ApiHelpers::RETRY_INTERVAL
       sleep(retry_time)
       # only retry if status code indicates a possible temporary error, and we are under the retry limit and
       # not calling a method that is blocked from retries
-      if should_retry?(e.http_code) && retry_count < MAX_RETRY_COUNT
+      if should_retry?(e.http_code) && retry_count < ApiHelpers::MAX_RETRY_COUNT
         process_api_request(http_method, path, payload: payload, retry_count: current_retry)
       else
         # we have reached our retry limit or the response code indicates we should not retry
--- a/app/models/data_repo_client.rb
+++ b/app/models/data_repo_client.rb
@@ -2,6 +2,10 @@
 # API client bindings for retrieving information about Terra Data Repo datasets/snapshots/schemas from their API
 ##
 class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_at, :service_account_credentials)
+  extend AccessTokenGenerator # class method for generating OAuth2 access tokens
+  include GoogleServiceClient # helper module for refreshing access tokens, accessing GCS storage driver attributes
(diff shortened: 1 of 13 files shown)


app/models/data_repo_client.rb
```diff
@@ -0,0 +1,337 @@
+##
+# API client bindings for retrieving information about Terra Data Repo datasets/snapshots/schemas from their API
+##
+class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_at, :service_account_credentials)
+  # path to read-only service account JSON keyfile
+  SERVICE_ACCOUNT_KEY = !ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'].blank? ? (ENV['NOT_DOCKERIZED'] ? ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'] : File.absolute_path(ENV['READ_ONLY_SERVICE_ACCOUNT_KEY'])) : ''
```
eweitz, line 6: This code is the same for `fire_cloud_client.rb` and `papi_client.rb`, so let's put it in a superclass.

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,337 @@
...
+  # Google authentication scopes necessary for querying TDR API
+  GOOGLE_SCOPES = %w(openid email profile)
+  # GCP Compute project
+  COMPUTE_PROJECT = ENV['GOOGLE_CLOUD_PROJECT'].blank? ? '' : ENV['GOOGLE_CLOUD_PROJECT']
+  # Base API URL to request against
+  BASE_URL = Rails.application.config.tdr_api_base_url
+  # constant used for retry loops in process_firecloud_request and execute_gcloud_method
+  MAX_RETRY_COUNT = 5
+  # constant used for incremental backoffs on retries (in seconds); ignored when running unit/integration test suite
+  RETRY_INTERVAL = Rails.env.test? ? 0 : 15
```
eweitz, line 16: This code is the same for `fire_cloud_client.rb`; let's put it in a superclass.

eweitz (changes requested):
Functionality here looks good!  With a few rather similar classes and lots of exact overlap, it's worth abstracting some code to superclasses (or another type of shared module) for maintainability.

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,337 @@
...
+  def self.generate_access_token(service_account)
+    # if no keyfile present, use environment variables
+    creds_attr = {scope: GOOGLE_SCOPES}
+    if !service_account.blank?
+      creds_attr.merge!(json_key_io: File.open(service_account))
+    end
+    creds = Google::Auth::ServiceAccountCredentials.make_creds(creds_attr)
+    token = creds.fetch_access_token!
+    token
+  end
```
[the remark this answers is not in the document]
bistline: Yeah, I had considered that, but it seemed like maybe more than I wanted to bite off.  But we're now at the rule of 3 (TDR, FireCloud, and PAPI clients) so it's probably time to do that refactoring.

lib/api_helpers.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  # handle a RestClient::Response object
+  #
+  # * *params*
+  #   - +response+ (String) => an RestClient response object
+  #
+  # * *return*
+  #   - +Hash+ if response body is JSON, or +String+ of original body
+  def handle_response(response)
```
devonbush, line 48: `parse_reponse_if_json` might be a more descriptive name for this function

lib/api_helpers.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  # URI-encode parameters for use in API requests
+  #
+  # * *params*
+  #   - +parameter+ (String) => Parameter to encode
+  #
+  # * *returns*
+  #   - +String+ => URI-encoded parameter
+  def uri_encode(parameter)
```
devonbush, line 98: let's drop this method -- it'll be clearer to call URI.escape directly in the code above

lib/api_helpers.rb
```diff
@@ -0,0 +1,101 @@
...
+##
+module ApiHelpers
+  # check if OK response code was found
+  #
+  # * *params*
+  #   - +code+ (Integer) => integer HTTP response code
+  #
+  # * *return*
+  #   - +Boolean+ of whether or not response is a known 'OK' response
+  def ok?(code)
```
devonbush, line 12: that's really weird that there's not a built-in Rails method for this, because it seems so useful, but a few minutes of googling didn't turn up anything.

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,207 @@
...
+    self.expires_at = Time.zone.now + self.access_token['expires_in']
+    self.api_root = BASE_URL
+  end
+
+  ##
+  # Abstract request handlers
+  ##
+
+  # submit a request to TDR API
+  def process_api_request(http_method, path, payload: nil, retry_count: 0)
```
devonbush, line 59: this is a semantic nitpick, but `execute_api_request` reads better to me.  'process' sounds like we're just processing an already-sent request

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,207 @@
...
+
+  # generic validator for an argument to a method
+  #
+  # * *params*
+  #   - +name+ (String, Symbol) => argument name
+  #   - +value+ (Various) => Value for argument
+  #   - +control+ (Array) => Array of acceptable values
+  #
+  # * *raises*
+  #   - (ArgumentError) => if value is not a member of control
```
devonbush, line 195: this seems generally useful.  I like that you've got it here now, and we can keep an eye out for other uses where we might want to move it into a shared library

lib/access_token_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+  #
+  # * *params*
+  #   - +service_account+ (Pathname) => path to service account JSON keyfile
+  #
+  # * *returns*
+  #   - (Hash) => OAuth2 access token hash, with the following attributes
+  #     - +access_token+ (String) => OAuth2 access token
+  #     - +expires_in+ (Integer) => duration of token, in seconds
+  #     - +token_type+ (String) => type of access token (e.g. 'Bearer')
+  def generate_access_token(service_account)
```
devonbush, line 12: I'd make this a class method `self.generate_access_token`, and have it take the scopes as an argument.  that makes it clearer what this method depends on

lib/google_service_client.rb
```diff
@@ -0,0 +1,91 @@
...
+module GoogleServiceClient
+  ##
+  # OAuth token methods
+  ##
+
+  # refresh access_token when expired and stores back in instance of parent class
+  #
+  # * *return*
+  #   - +DateTime+ timestamp of new access token expiration
+  def refresh_access_token!
```
devonbush, line 13: Same, I'd make this a class method.  (and you can always get the callee by introspection if you want it for logging purposes.)

devonbush (changes requested):
Great stuff, just some polishing on where the code lives.  I think you did a great job of scoping this ticket well

lib/access_token_generator.rb
bistline: 5eb5494 - this is renamed to `ServiceAccountManager`, and since it is only ever used via `extend` in other classes, this means that it is in fact a class method.  

This is the only way to get this to work as using `include` provides access as instance (not class) methods, and you can't mix the two without lots of chicanery that breaks other things.

lib/google_service_client.rb
bistline: I don't actually want this to be a class method as the classes that this gets included in are all singletons. 
 `FireCloudClient`, `PapiClient`, and `DataRepoClient` each are instantiated as class variables for `ApplicationController` - we do this specifically so that everyone uses the same instance, and share the same access token.  Therefore, when I call `ApplicationController.firecloud_client.refresh_access_token!`, it updates the access token _in place_ for that shared instance of `FireCloudClient`.  If we don't do this, eventually we'd run afoul of the Google quota on generating access tokens, and everything would break until the lock expires.

devonbush: Exactly, that's why this should be a class method (self.refresh_access_token). If this GoogleServiceClient is included in 3 other classes, having this be an instance method implies that each of them have their own token, and we have 3 separate Google tokens.  Which isn't what we want, right?

devonbush: Essentially, I was expecting this class to be the 'Google access token manager', and other classes call it when they need a token.  And so I wouldn't necessarily include in other services at all -- it would be a singleton class of its own.

lib/access_token_generator.rb
devonbush: I think I'm fuzzy on the difference between this and the google_service_client.rb. Should this be moved into that file so we centralize google token creation/refresh?

devonbush (changes requested):
Let's chat tomorrow about this.  I think we have different visions for how the refactored services interact.  And yours might be just as good/better than mine, so I want to hear it before I make a bunch more PR comments that may be pushing in a different direction

bistline:
Great comments, everyone - I've addressed them in 5eb5494 and c26fc0b.

> Let's chat tomorrow about this. I think we have different visions for how the refactored services interact. And yours might be just as good/better than mine, so I want to hear it before I make a bunch more PR comments that may be pushing in a different direction

Yeah, let's talk.  My vision for this altered once I started coding it up, and the difference between `include` and `extend` and how that treats instance vs. class methods is what informed this approach.

bistline:
Last build failure was a red herring due to a momentary blip on the FireCloud API (we got a 503 at one point).

app/models/data_repo_client.rb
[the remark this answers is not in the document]
bistline: 5eb5494 & c26fc0b

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,337 @@
...
+  # URI-encode parameters for use in API requests
+  #
+  # * *params*
+  #   - +parameter+ (String) => Parameter to encode
+  #
+  # * *returns*
+  #   - +String+ => URI-encoded parameter
+  def uri_encode(parameter)
+    URI.escape(parameter)
+  end
```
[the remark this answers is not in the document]
bistline: 5eb5494 & c26fc0b

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,199 @@
...
+  #
+  # * *returns*
+  #   - (Array<Hash>) => Array of snapshot JSON atributes
+  def snapshots(datasetIds: [], direction: 'desc', filter: nil, limit: 100, offset: 0, sort: 'name')
+    validate_argument(:direction, direction, SORT_DIRECTIONS)
+    validate_argument(:sort, sort, SORT_OPTIONS)
+    query_opts = merge_query_options(
+      {
+        datasetIds: datasetIds.join(','), direction: direction, filter: filter,
+        limit: limit, offset: offset, sort: sort
```
eweitz, line 158: Usually I want a Ruby language feature in JavaScript, but here it's vice versa.  [Object initializer shorthand](https://stackoverflow.com/q/37365394) would be useful, but TIL it isn't available in Ruby.

lib/service_account_manager.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  # return the GCP project this instance is running in
+  def compute_project
+    ENV['GOOGLE_CLOUD_PROJECT']
+  end
+
+  # resolve filepath to primary service account (e.g. project owner) keyfile
+  def get_primary_keyfile
+    ENV['NOT_DOCKERIZED'] ? ENV['SERVICE_ACCOUNT_KEY']: File.absolute_path(ENV['SERVICE_ACCOUNT_KEY'])
```
eweitz, line 33: Conventional ternary spacing:

```suggestion
    ENV['NOT_DOCKERIZED'] ? ENV['SERVICE_ACCOUNT_KEY'] : File.absolute_path(ENV['SERVICE_ACCOUNT_KEY'])
```

eweitz (approved):
Much cleaner!  Thanks for going the extra mile for maintainability.

I noted a few additional nits, no blockers.

app/models/data_repo_client.rb
bistline: TOTALLY agree.  I've thought that many times myself.

app/models/data_repo_client.rb
```diff
@@ -0,0 +1,199 @@
...
+  # fetch a single dataset
+  #
+  # * *params*
+  #   - +dataset_id+ (String: UUID) => Dataset UUID
+  #   - +include+ (Array) => Dataset fields to return from DATASET_INCLUDE_FIELDS (default: SCHEMA,PROFILE,DATA_PROJECT,STORAGE)
+  #
+  # * *returns*
+  #   - (Hash) => Hash of dataset JSON attributes
+  def dataset(dataset_id, include: ALL_DATASET_FIELDS)
+    path = api_root + "/api/repository/v1/datasets/#{dataset_id}"
```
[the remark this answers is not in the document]
bistline: We wouldn't be able to do that because of the DRS endpoints: https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService

Those will be needed to get information about the location of files.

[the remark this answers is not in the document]
eweitz: Oof, nice catch on something of an API gotcha!

devonbush (approved):
Looks good, modulo the method naming suggestions which are optional

PR #1048, opened 2021-05-25
restoring sorting to violin plots (SCP-3380)

devonbush:
To TEST:
1. load a study with expression data
2. do a search for a gene
3. confirm violin plot annotations are alphabetically sorted
4. change annotation
5. confirm violin plot annotations are still sorted


jlchang (approved):
Thanks for restoring the sorting!

eweitz (approved):

bistline (approved):

PR #1052, opened 2021-05-27
Adding correlation plots for two-gene searches (SCP-2348)

devonbush:
SCP-2348.  As shown in demo, this shows a correlation plot whenever a user searches for 2 genes, in addition to the existing dotPlot and Heatmap.  The title of the plot includes the correlation coefficient.  This uses the same functions for loading data from Mongo as the current expression plots, and so I would expect performance to be about the same as a cluster plot, since it has to load 2 genes, but no cluster coordinates.

Gene_correlation_plot_1_2021-06-02

Gene_correlation_plot_2_2021-06-02

TO TEST:
1. load a study with expression data
2. do a search for two genes which exist in the study
3. confirm a 'Correlation' tab appears, with the genes as the x and y axis.


app/lib/expression_viz_service.rb
```diff
@@ -137,6 +137,44 @@ def self.load_expression_data_array_points(study, gene, cluster, annotation, sub
...
+    viz_data[:x] = gene0_expression
+    viz_data[:y] = gene1_expression
+    viz_data[:pearsonsR] = pearsons_r(gene0_expression, gene1_expression).to_f.round(4)
+    viz_data
+  end
+
+  # adapted from https://blog.chrislowis.co.uk/2008/11/24/ruby-gsl-pearson.html
+  # note that the age of the blog means his performance numbers of the ruby version are off
+  # in local testing, this can compute correlation for 1MM cells in 0.5 seconds
+  def self.pearsons_r(x,y)
```
bistline, line 155: Style: I get this was copied from the above source, but it needs to be reformatted for readability and general Ruby style.  The variables are camel-case, and the spacing is all off in places.

app/lib/expression_viz_service.rb
```diff
@@ -137,6 +137,44 @@ def self.load_expression_data_array_points(study, gene, cluster, annotation, sub
...
+    n = x.length
+
+    sumx = x.inject(0) { |r,i| r + i }
```
bistline, line 158:
```suggestion
    sumx = x.reduce(0, :+)
```

bistline (approved):
Looks good, modulo some style nits about `pearsons_r`

app/lib/expression_viz_service.rb
eweitz, line 155: Having a hand-transcribed algorithm like this is pretty neat, but also makes me wary.  Ideally, we'd use a Ruby statistics library that's been tested, gets bug reports if something is wrong, etc.

Are any of these suitable?

* https://github.com/sciruby/statsample
* https://github.com/clbustos/statsample
* https://github.com/alfonsojimenez/pearson

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -285,6 +286,23 @@ export default function ExploreDisplayTabs({
...
+                  isCorrelatedScatter={true}
+                  dimensions={getPlotDimensions({
+                    numColumns: 1,
+                    numRows: 1,
+                    hasTitle: true,
+                    showRelatedGenesIdeogram: false
+                  })}
+                  isCellSelecting={isCellSelecting}
+                  plotPointsSelected={plotPointsSelected}
+                />
```
eweitz, line 303: All [examples of gene correlation plots](https://www.google.com/search?q=gene+correlation+plot) I've seen -- except for a sketch from the original SCP feature request in https://github.com/broadinstitute/single_cell_portal/issues/38 -- have numerical tick marks / gridline labels for the X and Y axes, but your helpful screenshot lacks them.

I suspect we need tick marks for scientific quality here, but I defer to @jlchang on their priority -- e.g. blocker for MVP, ticket-worthy for next sprint, or a later opportunity for improvement.

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -202,6 +209,17 @@ def self.get_cluster_viz_data(study, cluster, url_params)
...
               }
+            elsif is_correlated_scatter
+              if genes.count != 2
+                raise ArgumentError, "Correlated scatter plots require specifying 2 genes"
+              end
+              plot_data = ExpressionVizService.load_correlated_data_array_scatter(
+                  study, genes, cluster, annotation, subsample)
+              gene_names = genes.map {|g| g['name']}
+              titles = {
+                x: "#{gene_names[0]} #{titles[:magnitude]}",
```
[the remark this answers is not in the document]
devonbush: works for me

app/lib/expression_viz_service.rb
devonbush: heh, yeah, I cleaned it up a little bit, but you're right it needs more work

app/javascript/lib/scp-api-metrics.js
```diff
@@ -145,16 +145,18 @@ export function logScatterPlot(
   perfTimes
 ) {
   const props = {
-    'numPoints': scatter.numPoints, // How many cells are we plotting?
+    numPoints: scatter.numPoints, // How many cells are we plotting?
     genes,
-    'gene': scatter.gene,
-    'is3D': scatter.is3D,
+    gene: scatter.genes,
```
eweitz, line 150: This seems like breaking change in analytics.  Previously, `gene` was a string.  Now it's an array.  I can't think of any analytics reports that currently use `gene`, but this will prevent analyzing that property it before our launch of this feature.  That said, I don't think it's a big deal, and I'd have done the same.

Two trivial change requests: 
* Make this `genes` like `getLogPlotProps`
* Add `numGenes` since it's trivial to implement, infeasible to derive in Mixpanel, and foreseeable we could support scatter plots with more genes

```suggestion
    genes: scatter.genes,
    numGenes: scatter.genes.length,
```

eweitz (changes requested):
Stellar work!  This will clearly be useful.

I note two blockers below -- a likely scientific error with _R^2_ vs. _r_, and an instrumentation inconsistency.  

Not necessarily blockers: please get Jean's opinion on the priority of tick marks for the graph, and I wonder if a statistics library would be better than our hand-transcribed correlation algorithm.  Finally, given this is a significant new visual feature, it's probably also worth getting Jerome's UX / design input, if you haven't already.

app/javascript/components/explore/ExploreDisplayTabs.js
jlchang: Yes we should have tick marks but I don't consider the lack to be blocking as long as we intend to add them in a timely manner.

app/javascript/components/visualization/PlotTitle.js
```diff
@@ -19,5 +21,8 @@ export default function PlotTitle({ cluster, annotation, genes, consensus, subsa
   return <h5 className="plot-title">
     <span className="cluster-title">{contentString} </span>
     <span className="detail"> {detailString}</span>
+    { isCorrelatedScatter &&
+      <span> &nbsp; R<sup>2</sup>={pearsonsR}</span>
```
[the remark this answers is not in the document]
jlchang: There is a question of whether Pearson's correlation is the most appropriate calculation to make. [correlatePairs](https://rdrr.io/bioc/scran/man/correlatePairs.html) from scran uses Spearman's rank correlation. [Seurat](https://github.com/satijalab/seurat/blob/af2925ce71f51d47cce8e45d94b638ca434a4d33/R/visualization.R#L6898) seems to use Pearsons (the default for "cor") - so there's varying usage and precedent for using Pearson's.

This [explaination](https://towardsdatascience.com/clearly-explained-pearson-v-s-spearman-correlation-coefficient-ada2f473b8) of the difference between the two is compelling that Pearsons will only detect linear relationships well while spearman can detect non-linear but monotonically increasing/decreasing relationships. Since biological relationships aren't reliably linear in nature, Spearman's seems like a better correlation metric to use.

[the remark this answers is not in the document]
eweitz: Starting with Pearson correlation is reasonable, I think.

But as shown in the original feature request for this ([here](https://github.com/broadinstitute/single_cell_portal/issues/38#issuecomment-436822747)), we may well want _both_ Pearson's _and_ Spearman's (and p-value / adjusted p-value).  This strengthens the case for using a Ruby statistics library instead of us writing the algorithms, but maybe using such a library can weight until we add those other common statistics for gene correlation plots.

What do you think, @jlchang?

app/lib/expression_viz_service.rb
devonbush: I totally concur that this should be in a gem if possible.  The problem was my 15 minutes of googling turned up nothing.   I'd looked at the statsample gems, and none had recent updates (e.g. sciruby statsample was last updated in 2017), and  both have complex dependencies on native C extensions that suggested it would not be worth bringing in for a single function.  I hadn't seen the alfonsojimenez/pearson gem.  But as best as I can tell that's not maintained or widely used.

It boggles my mind that there isn't a standard, up-to-date Ruby stats library.  I guess all the cool kids really do use Python

app/javascript/lib/scp-api-metrics.js
devonbush: technically, this isn't a breaking change -- this is a fixing change. :-)  The renaming of gene => genes happened in an unrelated earlier, already merged PR -- scatter.gene is always undefined now.  We're already storing a 'genes' param (line 149 above), so we can't call this 'genes' too.   Maybe call this one 'receivedGenes' -- maybe rename them both to receivedGenes and requestedGenes now since we haven't used either in a report yet?

app/javascript/components/visualization/PlotTitle.js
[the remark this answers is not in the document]
devonbush: I've got a meeting with Sarah to talk about other visualizations, I can also ask her

app/javascript/components/explore/ExploreDisplayTabs.js
devonbush: cool, I can't imagine adding tickmarks in plotly will be too hard

app/javascript/lib/scp-api-metrics.js
eweitz: Ah, thanks for pointing out that `genes` is already logged.  I slightly prefer to keep property names as close as possible across interfaces (so e.g. `genes` instead of `requestedGenes`), but I don't object if you think it's worthwhile.

For easier visual grouping -- Mixpanel sorts properties alphabetically -- let's order property name parts as <subject><modifier> (e.g. `genesRequested`).  We use this order in e.g. #1056, with `numPointsPlotted` instead of `numPlottedPoints`.

I'd also opt for `genesPlotted` (and `numGenesPlotted`) over `genesReceived`, as it seems more descriptive.

app/javascript/components/visualization/PlotTitle.js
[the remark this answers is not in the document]
eweitz: Good stuff.  For clarity, let's distinguish the two issues discussed in this thread:

1.  Assertion that writing Pearson correlation coefficient as "R^2" is incorrect, and should be "r".

2. Question about the most appropriate type of correlation for this scenario.

app/lib/expression_viz_service.rb
devonbush: Update -- after chatting with Sarah, she suggested using both spearman and Pearson, which means we should definitely start using a gem, and we can deal with the added install complexity

app/javascript/components/explore/ExploreDisplayTabs.js
devonbush: added

app/lib/expression_viz_service.rb
devonbush: double update, the gem is performance prohibitive.  So we'll try client side, and start with just Spearman, since that's the most 'correct'

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -63,21 +68,27 @@ function RawScatterPlot({
...
       { scatter, genes, width: dimensions.width, height: dimensions.height },
       perfTimes
     )
+
+    if (isCorrelatedScatter) {
+      // compute correlation stats, but do it after the graph render so it doesn't delay
+      // the visualizations or impact logging
+      const spearmanRho = new SpearmanRho(scatter.data.x, scatter.data.y)
+      spearmanRho.calc().then(value => setSpearman(value))
+    }
```
eweitz, line 91: Let's swap these -- the time needed to show this important statistic seems relevant to the plot.

```suggestion
    if (isCorrelatedScatter) {
      // compute correlation stats, but do it after the graph render so it doesn't delay
      // the visualizations or impact logging
      const spearmanRho = new SpearmanRho(scatter.data.x, scatter.data.y)
      spearmanRho.calc().then(value => setSpearman(value))
    }

    logScatterPlot(
      { scatter, genes, width: dimensions.width, height: dimensions.height },
      perfTimes
    )
```

Including this segment now lets us still analyze just front-end performance of Plotly rendering (via `perfTime:frontend:plot`, which isn't affected by this swap), while _also_ capturing this segment in `perfTime:frontend:other`, `perfTime:frontend`, and `perfTime:full`.  Including this segment now also avoids a discontinuity in those metrics for gene correlation plots in the likely case we want a named segment to track time to compute correlation.

eweitz (changes requested):
Looking good.  With two small bits of polish -- for performance analytics and user-facing notation -- I think this will be good to go.

app/javascript/components/visualization/ScatterPlot.js
devonbush: Sorry, the code comment is misleading.  The stats calculation is done asynchronously, so the change you suggest wouldn't impact logging except trivially.  I'm not sure I like waiting to log the plot until after the stats resolve, since a lot can happen, and I don't want to lose logging in the event the stats process errors or the user navigates away before they render.  I'll update the comment, and would suggest that we hold off on instrumenting the performance of these stats until we get a better sense of their usage and whether we'll be adding others.  (FWIW, the computations take ~1sec for 1MM cells in the console)

app/javascript/components/visualization/PlotTitle.js
```diff
@@ -1,23 +1,36 @@
...
+    detailContent = `subsample[${subsample}]`
   }
   return <h5 className="plot-title">
-    <span className="cluster-title">{contentString} </span>
-    <span className="detail"> {detailString}</span>
+    <span className="cluster-title">{content} </span>
+    <span className="detail"> {detailContent} </span>
+    { isCorrelatedScatter && !!correlations.spearman &&
+      <span className="correlation">
+        Spearman rho={ Math.round(correlations.spearman * 1000) / 1000}
```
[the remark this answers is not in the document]
devonbush: Yup, google image search confirms you are right that most preserve spaces

app/javascript/components/visualization/ScatterPlot.js
eweitz: Ah, right, it'd need to be e.g.:

```
if (isCorrelatedScatter) {
  // compute correlation stats, but do it after the graph render so it doesn't delay
  // the visualizations or impact logging
  const spearmanRho = new SpearmanRho(scatter.data.x, scatter.data.y)
  spearmanRho.calc().then(value => {
    setSpearman(value)
    logScatterPlot({ scatter, genes, width: dimensions.width, height: dimensions.height }, perfTimes)
  })
} else {
    logScatterPlot({ scatter, genes, width: dimensions.width, height: dimensions.height }, perfTimes)
  }
}
```

The benefits strike me as greater than the possible costs you describe, but I don't consider it blocking.  I appreciate the rationale and point-measurement data!

eweitz (approved):

PR #1059, opened 2021-06-02
Adding DNS rebinding/host header injection protection (SCP-3373)

bistline:
This update adds DNS rebinding/host header injection protection through the `Rails.application.config.hosts` setting, which was [added](https://guides.rubyonrails.org/configuring.html#configuring-middleware) in Rails 6.  This will validate any incoming request to ensure that the `Host` header is not being spoofed that would cause the request to be redirected away from SCP to another site, potentially controlled a malicious actor.

This update corrects the issues identified in #1050 that caused Google-based load balancer health checks to be blocked, which then caused the load balancer to take the instance offline.  The specific fix is to allow connections from a variety of IP ranges as detailed [here](https://cloud.google.com/load-balancing/docs/health-checks#firewall_rules) in Google's documentation.  In addition to external IP ranges, connections on the private internal network (`10.128.0.0/24`) are also allowed, as this is the value set for the `Host` header on health checks.  Normal user-based traffic is covered by the actual hostname of the deployed instance (`PROD_HOSTNAME`).

TO TEST:

Testing this is somewhat convoluted as we cannot use a tool like Burp to intercept traffic and add headers.  However, this can be tested to an extent in `curl`:

1. Pull this branch and boot portal as normal and sign in with an account
2. In `development.log`, copy the quoted request path for the OAuth callback, which will look like the following in the log:
```
D, [2021-05-26T11:08:50.044257 #30071] DEBUG -- omniauth: (google) Request phase initiated.
Started GET "/single_cell/users/auth/google/callback?state=f53303c40ad362fc49d876c468d289d42e3ab36bf01eac59&code=4/0AY0e-g5sg9Ej1hR3CpaDyRNjFt0oOchV9CvQV2hTaYRuqzEWPI1SU0DksJA07qXGEKEArw&scope=email%20profile%20https://www.googleapis.com/auth/userinfo.email%20https://www.googleapis.com/auth/userinfo.profile%20openid&authuser=0&hd=broadinstitute.org&prompt=consent" for 127.0.0.1 at 2021-05-26 11:08:52 -0400
D, [2021-05-26T11:08:52.453752 #30071] DEBUG -- omniauth: (google) Callback phase initiated.
```
3. In a new terminal window, execute the following `curl` call:
```
curl -I --location --request GET 'https://locahost:3000/[ ... request path from above ... ]' --header 'Host: malicious.com'
```
4. Validate the response is `403`:
```
HTTP/1.1 403 Forbidden
Content-Type: text/html; charset=UTF-8
Strict-Transport-Security: max-age=15768000; includeSubdomains; preload
...
```
_Note: as this is handled at the middleware level, the bad request will not show up in `development.log`._

This PR satisfies SCP-3373.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,6 +42,7 @@ module SingleCellPortal
     EXTENDED_GOOGLE_SCOPES = BASIC_GOOGLE_SCOPES.dup + %w(cloud-billing.readonly)
 
     # DNS rebinding/host header injection protection
+    # CIDR ip ranges from https://cloud.google.com/load-balancing/docs/health-checks#firewall_rules
     config.hosts = [
       ENV['PROD_HOSTNAME'], # hostname for server
       IPAddr.new('10.128.0.0/24'), # connection on private network for LB health check
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,5 +40,15 @@ module SingleCellPortal
     # extended scopes add cloud-billing.readonly which requires user consent
     # these are only requested when users attempt to visit the "My Billing Projects" page
     EXTENDED_GOOGLE_SCOPES = BASIC_GOOGLE_SCOPES.dup + %w(cloud-billing.readonly)
+
+    # DNS rebinding/host header injection protection
+    config.hosts = [
+      ENV['PROD_HOSTNAME'], # hostname for server
+      IPAddr.new('10.128.0.0/24'), # connection on private network for LB health check
+      IPAddr.new('35.191.0.0/16'), # external IP range for LB health check
+      IPAddr.new('130.211.0.0/22'), # external IP range for LB health check
+      IPAddr.new('209.85.152.0/22'), # external IP range for network LB
+      IPAddr.new('209.85.204.0/22') # external IP range for network LB
+    ]
   end
 end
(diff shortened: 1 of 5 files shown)


config/application.rb
```diff
@@ -40,5 +40,15 @@ class Application < Rails::Application
...
     EXTENDED_GOOGLE_SCOPES = BASIC_GOOGLE_SCOPES.dup + %w(cloud-billing.readonly)
+
+    # DNS rebinding/host header injection protection
+    config.hosts = [
+      ENV['PROD_HOSTNAME'], # hostname for server
+      IPAddr.new('10.128.0.0/24'), # connection on private network for LB health check
+      IPAddr.new('35.191.0.0/16'), # external IP range for LB health check
+      IPAddr.new('130.211.0.0/22'), # external IP range for LB health check
+      IPAddr.new('209.85.152.0/22'), # external IP range for network LB
+      IPAddr.new('209.85.204.0/22') # external IP range for network LB
```
eweitz, line 51: I'll quickly forget where these come from.  For maintainability, could you either link to [your PR here](https://github.com/broadinstitute/single_cell_portal_core/pull/1059) (it is superb documentation) or to https://cloud.google.com/load-balancing/docs/health-checks#firewall_rules?

eweitz (approved):
Looks good!  I suggested a small maintainability improvement.

Given AppSec found this bug, is it feasible to have them verify the fix before it's deployed to production?

config/application.rb
devonbush: concur

devonbush (approved):

config/application.rb
bistline: Can do.

bistline:
> Looks good! I suggested a small maintainability improvement.
> 
> Given AppSec found this bug, is it feasible to have them verify the fix before it's deployed to production?

Yes - since this is being done at their request, once it is deployed to staging I will have them re-run their scans to confirm this has fixed the problem.  I expect those to pass without issue.

PR #1060, opened 2021-06-03
Adding test for special characters in filenames (SCP-3364)

bistline:
This update adds a test to validate that file uploads with special characters in the filenames (e.g. parentheses) still work.  This was found as an issue during the [Rails 6.1/Carrierwave](#1030) migration, and while it was addressed, a test was not added to directly validate this moving forward.  This test validates that all non-word characters are replaced with underscores (`_`).

TO TEST:
1. Boot portal as normal on this branch
2. In a study, navigate to the file upload wizard and select the 'Miscellaneous' tab
3. Upload the test file `test/test_data/README (1).txt`
4. Note that the file uploads w/o issue and is renamed to `README__1_.txt`

This PR satisfies SCP-3364.


eweitz (approved):

test/test_data/README (1).txt
```diff
@@ -0,0 +1 @@
+This is an example readme file
```
devonbush, line 1: I think both this file and the README in this directory should be renamed to something like "mock_study_doc_upload.txt" and "mock_study_doc_upload2.txt", so that they are not confused with READMEs for the code.  

Also, does running this test delete the file?  if so, the test should copy the file to a temp directory, and upload that, so that it can be run without modifying files under source control, and run successfully more than once. See, e.g. lines 30-33 of SyntheticStudyPopulator.rb

devonbush (changes requested):
Thanks for following up and writing this test.  A couple changes, one of which I think is needed for test rerun reliability

test/test_data/README (1).txt
bistline: Good points.

bistline: Actually, for the 2nd point - we patched the uploader for study files so that in the test environment, it performs a copy rather than a move.  See [here](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/uploaders/upload_uploader.rb#L29) and [here](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/uploaders/upload_uploader.rb#L33).  So we shouldn't have to worry about that in tests.

devonbush: ah, thanks for reminding me, good point

devonbush (approved):

PR #1063, opened 2021-06-04
Document observability (SCP-3399)

eweitz:
This documents observability in SCP.  It covers the tools SCP uses for insight on bugs and product strategy, including a dive into concepts we use to guide performance optimization efforts.

You can view the formatted document by clicking "..." at top right in the diff for `observability.md`, then "View file".  As discussed in standup, defaulting such non-sensitive technical docs to our public repo seems best for review, long-term discoverability, and maintainability.  Once merged, I'll also link it from other places.

This satisfies SCP-3399.


bistline (approved):

ehanna4 (approved):
LGTM Walked through this doc with Eric & Mohs in a meeting and found it well organized and helpful. Thanks for putting this together!

jlchang (approved):

devonbush (approved):
I like the doc a lot.  I'm not as sure about the /docs location for it.  If we're going to create that folder, we should probably move our other docs there, (e.g. readme, non_containerized_dev_readme, etc..), and replace our current readme with a pointer there.  But this is fine to start with, we can see how it goes

PR #1066, opened 2021-06-09
Bulk download selection modal (SCP-3356, SCP-2333)

devonbush:
SCP-3356 SCP-2333
image
image

This does not yet handle TDR search results, but the APIs should be relatively straightforward to update once the structure of those results and the ways to get to the file paths become clearer.  To that end, I've taken the bulk download APIs from the search controller and moved them into their own controller

On the frontend, I've removed the DownloadProvider, consistent with our restriction of contexts/providers to only globally-applicable stuff (e.g. user info).  I've also started using react-testing-library.  I agree with Eric that the experience is superior to enzyme, so I think we should use it instead of enzyme going forward.

TO TEST:
  0. sign in and do a search
  1. click 'Download'
  2. Observe that a modal appears
  3. select/deselect studies/files/categories, and observe the total size changes
  4. hit 'next' while at least one file is selected
  5. paste the code that appears into your terminal, and confirm the download works


app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -0,0 +1,299 @@
...
+          study_accession = valid_accessions.first
+          directories = self.class.find_matching_directories(params[:directory], study_accession)
+          directory_files = ::BulkDownloadService.get_requested_directory_files(directories)
+        else
+          directories = []
+          directory_files = []
+        end
+
+        files_requested = nil
+        byebug
```
bistline, line 209: Debug call

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -0,0 +1,299 @@
...
+        # reference BulkDownloadService as ::BulkDownloadService to avoid NameError when resolving reference
+        if params[:file_ids]
+          files_requested = StudyFile.where(:id.in => file_ids)
+        else
+          files_requested = ::BulkDownloadService.get_requested_files(file_types: sanitized_file_types,
+                                                                      study_accessions: valid_accessions)
+        end
+
+        # determine quota impact & update user's download quota
+        # will throw a RuntimeError if the download exceeds the user's daily quota
```
bistline, line 220: I know that's what the current functionality is (and I wrote it), but for the life of me I can't figure out _why_ we decided that throwing a `RuntimeError` was the right call here...

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -0,0 +1,117 @@
...
+            <div className="col-md-4">
+              <h3 className={stepNum === 1 ? '' : 'greyed'}>
+                <span className="badge">1</span>
+                &nbsp; Select the files
+              </h3>
+            </div>
+            <div className="col-md-4">
+              <h3 className={stepNum === 2 ? '' : 'greyed'}>
+                <span className="badge">2</span>
+                &nbsp; Get terminal link
```
eweitz, line 69: Calling this a link is a bit confusing.  Users expect to click on links and go somewhere.  And while the command includes a link, but that's more of a technical detail.

```suggestion
                &nbsp; Get terminal command
```

I could also see labeling this "Get download command" or simply "Download".

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -0,0 +1,117 @@
...
+    animation={false}
+    bsSize='large'>
+    <Modal.Body>
+      <div className="download-modal">
+        <div>
+          <div className="download-steps-header row">
+            <div className="col-md-4">
+              <h3 className={stepNum === 1 ? '' : 'greyed'}>
+                <span className="badge">1</span>
+                &nbsp; Select the files
```
eweitz, line 63: Let's omit articles in call-to-action labels, [per convention](https://github.com/broadinstitute/single_cell_portal_core/pull/1064#discussion_r647410393).

```suggestion
                &nbsp; Select files
```

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -0,0 +1,117 @@
...
+            </div>
+            <div className="col-md-4">
+              <h3 className={stepNum === 2 ? '' : 'greyed'}>
+                <span className="badge">2</span>
+                &nbsp; Get terminal link
+              </h3>
+            </div>
+          </div>
+          <div className="greyed">
+            Files are downloaded via the command line. Once you confirm your files selection, you will get a link
```
eweitz, line 74:
```suggestion
            Files are downloaded via the command line. Once you confirm your files selection, you will get a command
```

app/lib/request_utils.rb
```diff
@@ -79,6 +79,18 @@ def self.sanitize_search_terms(terms)
     SANITIZER.sanitize(inputs).encode!(Encoding.find('ASCII-8BIT'), invalid: :replace, undef: :replace)
   end
 
+  # takes a comma-delimited string of ids (e.g. StudyFile ids) and returns an array of ids
+  # raises Argument error if any of the strings are not valid ids
+  def self.validate_id_list(id_list_string)
```
bistline, line 84: This looks like it works fine, but there's an easier way to check IDs.  Since what we're actually trying to validate is "does this string represent a `StudyFile` id", and all instances of such are `BSON::ObjectId` objects, we can simply cast them:
```
ids = id_list_string.split(',').map(&:strip) # we need to strip any whitespace or it can error out
ids.map {|id| BSON::ObjectId.from_string(id) }
ids
```
If any id is not valid, it will raise an `BSON::ObjectId::Invalid` error.

app/lib/bulk_download_service.rb
```diff
@@ -160,21 +160,47 @@ def self.get_requested_directory_files(directories)
...
+  #   -+study+ (Study) => The study to retrieve download preview information for
+  #
+  # * *returns*
+  #   - (Hash) => Hash of download information about the study, including an array of the study files
+  def self.study_download_info(study)
+    {
+      name: study.name,
+      accession: study.accession,
+      description: study.description,
+      study_files: study.study_files.map { |study_file| BulkDownloadService.study_file_download_info(study_file) }
```
bistline, line 182: I had thought that we were only doing the 3 'main' types for bulk downloads.  Are we changing to "all files" by default now?  We may run into issues w/ studies containing sequence data.  Just a thought.

eweitz (changes requested):
Great stuff!  This looks almost ready to launch as an MVP.

The only things I consider blocking are two small, unambiguous bugs:
* Error message referring to ObjectId values are UUIDs
* Lack of space between unit and number in file sizes

I also suggest a few non-blocking, subjective refinements in user-facing copy, code style, etc.

Thanks as usual for the helpful screenshots.  I think it's also worth investigating a way to make the modal height adjust to the height of its content (to some reasonable maximum).  This would fix e.g. the awkwardly large empty space between the terminal command and the "DONE" button.

Finally, if "other" file types -- manifests / supplementary info files, gene lists, annotation labels, BAMs, etc. -- won't be downloadable through this UI, is it worth briefly noting that in the UI?

bistline (changes requested):
Overall this looks good - I found one stray `byebug` call, and I suggested an enhancement for checking file IDs, but I found a blocker: if you only perform a faceted search, the "Download" button does not enable.

Screen Shot 2021-06-09 at 10 26 45 AM

It apparently only turns on once you have performed a keyword search.  

I also noticed that when I signed in after loading this branch, the UI had "forgotten" that I turned on advanced search a while ago in my local instance.  There appears to be something amiss with honoring that selection.

app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -0,0 +1,255 @@
...
+                </label>
+              </td>
+              <td width="40%">
+                Study name
+              </td>
+              { COLUMN_ORDER.map(colType => {
+                return <td key={colType}>
+                  <label>
+                    <input type="checkbox"
+                      data-analytics-name="download-modal-checkbox"
```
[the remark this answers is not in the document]
devonbush: confirmed, the label is recorded as the 'label' property

app/lib/request_utils.rb
devonbush: nice, agree that's much cleaner

app/lib/bulk_download_service.rb
devonbush: Good question, Eric and I just chatted about this.  I think the right approach is to have the backend only worry about auth. and quota, and otherwise just do whatever the frontend asks.  So it will give the frontend a list of all available files, and the frontend can handle deciding what to ask for.  

 This will make it easier to update the UX to support additional ways of downloading.  I did update the actual download method so that if you just provide accessions, without supplying any file types, it defaults to only giving you the main 3.

app/javascript/styles/_search.scss
```diff
@@ -335,3 +335,37 @@ nav.search-links, #search-panel {
...
+  vertical-align: inherit;
+  border-radius: 2em;
+}
+
+.download-steps-header .greyed {
+	color: #888;
+	.badge {
+	 	background-color: #888;
+	}
+}
```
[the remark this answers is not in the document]
devonbush: Yeah, somehow Sublime stopped replacing tabs with spaces in scss files.  (sigh)

app/lib/request_utils.rb
bistline: Come to think of it, we actually don't even need the `ids` at the end since the output of `map()` is the same thing and Ruby will auto-return that...

app/controllers/api/v1/api_base_controller.rb
```diff
@@ -26,9 +26,11 @@ class ApiBaseController < ActionController::API
         ErrorTracker.report_exception(exception, current_api_user, params)
         logger.error ([exception.message] + exception.backtrace).join($/)
         if Rails.env.production?
-          render json: {error: "An unexpected error has occurred"}, status: 500
+          render json: {error: "An unexpected error has occurred"}, status: 500 and return
```
bistline, line 29: I think we actually proved that these don't need to be here, but they certainly don't hurt anything.

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -0,0 +1,308 @@
...
+        end_time = Time.zone.now
+        runtime = TimeDifference.between(start_time, end_time).humanize
+        logger.info "Curl configs generated for studies #{valid_accessions}, #{files_requested.size + directory_files.size} total files"
+        logger.info "Total time in generating curl configuration: #{runtime}"
+        send_data @configuration, type: 'text/plain', filename: 'cfg.txt'
+      end
+
+      private
+
+      def check_accession_permissions(valid_accessions)
```
bistline, line 258: I think it would be better/more consistent if this were a class method, and we passed in any other parameters we need (like `current_api_user`).  This would make it more consistent with the structure of all the other private methods in this controller.

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -0,0 +1,308 @@
...
+        if valid_accessions.blank?
+          render json: {error: 'Invalid request parameters; study accessions not found'}, status: 400 and return
```
bistline, line 260: This `render` call here will raise the same `DoubleRenderError` we were seeing before.

bistline (changes requested):
Nice work, and kudos for finally figuring out the `DoubleRenderError` we were seeing.  The `check_accession_permissions` still could potentially throw that, which is the only blocking change I can see.  Everything else looks good to me.

eweitz (approved):
Looks good, modulo Jon's note on `DoubleRenderError`!

app/controllers/api/v1/api_base_controller.rb
devonbush: concur

app/controllers/api/v1/bulk_download_controller.rb
devonbush: concur

app/controllers/api/v1/bulk_download_controller.rb
devonbush: ah, good catch, thanks

bistline (approved):

devonbush:
test failures are due to TDR upstream change, and unrelated to this PR

PR #1069, opened 2021-06-10
Extending DataRepoClient functionality (SCP-3354, SCP-3398)

bistline:
This update adds more functionality to `DataRepoClient`, including the ability to query snapshot indexes (dev only), and resolving references to files in Terra Data Repo (TDR) using DRS ids.  Currently these IDs are not available in search results, but once [this ticket](https://broadworkbench.atlassian.net/browse/DR-1870) is implemented, those should be available in row-level results.

This also adds `FacetNameConverter`, which is a stopgap solution for converting from SCP metadata convention names to HCA metadata short names (not fully namespaced, as this is not how the data is currently indexed).   This is _only_ intended for current "proof of concept" work, and will not be the eventual implementation for handling name conversions across schemas.

For now, searching is behind the `cross_dataset_search_backend` feature flag, and results are printed to the log only, and are not rendered in the UI.  As this is still all at the PoC level, there is some necessary "hand waving" in terms of terms/values being hard-coded, for now.  Searches are currently only implemented at the string-matching level.  Still to be implemented are numeric range queries, and handling conversion of time units.

TO TEST:
2. Enable the `cross_dataset_search_backend` feature flag in the "Admin => User Roles" panel for your user account
3. Perform a faceted search in the front end, using `"species: Homo sapiens"` and `"organism_age: 46-72"`
4. Note results in the log, detailing the project that was found, including the facet matches:
```
Found 1 results in Terra Data Repo
{"PulmonaryFibrosisGSE135893"=>{:name=>"Single-cell RNA-sequencing reveals profibrotic roles of distinct epithelial and mesenchymal lineages in pulmonary fibrosis", :description=>"Pulmonary fibrosis (PF) is a form of chronic lung disease characterized by progressive destruction of normal alveolar gas-exchange surfaces and accumulation of extracellular matrix (ECM). In order to comprehensively define the cell types, mechanisms and mediators driving ECM deposition and fibrotic remodeling in lungs with pulmonary fibrosis, we performed single-cell RNA-sequencing (scRNA-seq) of single-cell suspensions generated from non-fibrotic control and PF lungs. Analysis of over 114,000 cells from 20 PF and 10 control lungs identified 31 distinct cell types. We identified multiple distinct lineages directly contribute to ECM expansion, including a novel HAS1hi fibroblast subtype and a previously undescribed KRT5-/KRT17+, collagen and ECM-producing epithelial cell population that was highly enriched in PF lungs. Together these data provide high-resolution insights into the basic mechanisms of pulmonary fibrosis, and indicate a direct profibrotic role of the lung epithelium in PF pathogenesis. Overall design: We performed single-cell RNA-sequencing (scRNA-seq) of single-cell suspensions generated from non-fibrotic control and pulmonary fibrosis (PF) lungs", :matches=>[{"genus_species"=>"Homo sapiens"}, {"organism_age"=>"46-72"}]}}
```

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -178,7 +178,7 @@ module Api
           if @preset_search.accession_list.any?
             sort_type = :accession_list if params[:terms].blank?
             @accession_list = @preset_search.accession_list
-            Rails.logger.info "Scoping search results to accessions from preset search: #{@preset_search.name}: #{@accession_list}"
+            logger.info "Scoping search results to accessions from preset search: #{@preset_search.name}: #{@accession_list}"
             @viewable = @viewable.where(:accession.in => @accession_list)
           end
         end
@@ -291,8 +291,16 @@ module Api
           end
         end
 
+        # perform TDR search, if enabled & user performed faceted search
+        if @facets.any? && User.feature_flag_for_instance(current_api_user, 'cross_dataset_search_backend')
+          @trd_results = self.class.get_tdr_results(@facets)
+          # just log for now
+          logger.info "Found #{@trd_results.keys.size} results in Terra Data Repo"
+          logger.info pp @trd_results
+        end
+
         @matching_accessions = @studies.map(&:accession)
-        Rails.logger.info "Final list of matching studies: #{@matching_accessions}"
+        logger.info "Final list of matching studies: #{@matching_accessions}"
         @results = @studies.paginate(page: params[:page], per_page: Study.per_page)
         render json: search_results_obj, status: 200
       end
@@ -925,6 +933,48 @@ module Api
         else
           return matching_facet[:filters].detect { |filter| filter[:id] == search_result[result_key] }
         end
+      end
+
+      # execute a search in TDR and get back normalized results
+      def self.get_tdr_results(selected_facets)
+        query_json = ApplicationController.data_repo_client.generate_query_from_facets(selected_facets)
+        raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
+        results = {}
+        raw_tdr_results.dig('result').each do |result_row|
+          short_name = result_row.dig('project_short_name')
+          results[short_name] ||= {
+            name: result_row.dig('project_title'),
+            description: result_row.dig('project_description'),
+            matches: []
+          }
+          # determine facet filter matches
+          selected_facets.each do |facet|
+            matches = get_facet_match_for_tdr_result(facet, result_row)
+            matches.each do |col_name, matched_val|
+              entry = {col_name => matched_val}
+              results[short_name][:matches] << entry unless results[short_name][:matches].include?(entry)
+            end
+          end
+        end
+        results
+      end
+
+      # determine facet matches for an individual result row from TDR
+      def self.get_facet_match_for_tdr_result(facet, result_row)
+        tdr_name = FacetNameConverter.to_hca(facet[:id])
+        if facet[:filters].is_a? Hash
+          # this is a numeric facet, so convert to range for match
+          # TODO: determine correct unit/datatype and convert
+          filter_value = "#{facet.dig(:filters, :min).to_i}-#{facet.dig(:filters, :max).to_i}"
+          matches = result_row.each_pair.select {|col, val| col == tdr_name && val == filter_value}
+        else
+          matches = []
+          facet[:filters].each do |filter|
+            # look for match on the column name, and see which filter value also matched (ontology label or id)
+            matches << result_row.each_pair.select {|col, val| col == tdr_name && (val == filter[:name] || filter[:id])}.flatten
+          end
+        end
+        matches
       end
     end
   end
--- a/app/models/data_repo_client.rb
+++ b/app/models/data_repo_client.rb
@@ -12,6 +12,8 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   BASE_URL = Rails.application.config.tdr_api_base_url.freeze
   # Hostname of repo, needed for parsing out DRS identifiers
   REPOSITORY_HOSTNAME = URI(BASE_URL).host.freeze
+  # prefix used to identify valid DRS ids
+  DRS_PREFIX = "drs://#{REPOSITORY_HOSTNAME}/".freeze
 
   # control variables
   SORT_DIRECTIONS = %w(asc desc).freeze
@@ -338,8 +340,8 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   # * *raises*
   #   - (ArgumentError) => if drs_id is not formatted correctly
   def parse_drs_id(drs_id)
-    raise ArgumentError.new("#{drs_id} is not a valid DRS id") unless drs_id.starts_with?("drs://#{REPOSITORY_HOSTNAME}/")
-    drs_id.split("drs://#{REPOSITORY_HOSTNAME}/").last
+    raise ArgumentError.new("#{drs_id} is not a valid DRS id") unless drs_id.starts_with?(DRS_PREFIX)
+    drs_id.split(DRS_PREFIX).last
   end
 
   ##
--- a/app/models/data_repo_client.rb
+++ b/app/models/data_repo_client.rb
@@ -7,7 +7,7 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   include ApiHelpers
 
   # Google authentication scopes necessary for querying TDR API
-  GOOGLE_SCOPES = %w(openid email profile)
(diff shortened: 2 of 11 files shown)


app/models/data_repo_client.rb
```diff
@@ -92,6 +97,47 @@ def process_api_request(http_method, path, payload: nil, retry_count: 0)
...
+      'Accept' => 'application/json',
+      'Content-Type' => 'application/json',
+      'x-app-id' => "single-cell-portal",
+      'x-domain-id' => "#{ENV['HOSTNAME']}"
+    }
+    begin
+      response = RestClient::Request.execute(method: :get, url: path, headers: headers)
+      JSON.parse(response.body)
+    rescue RestClient::ExceptionWithResponse => e
+      Rails.logger.error "Terra Data Repo status error: #{e.message}"
```
eweitz, line 123:
```suggestion
      logger.error "Terra Data Repo status error: #{e.message}"
```

You refactored `Rails.logger` to `logger` in `search_controller.rb `, might as well do so in this file too.

app/models/data_repo_client.rb
```diff
@@ -169,11 +215,140 @@ def snapshots(datasetIds: [], direction: 'desc', filter: nil, limit: 100, offset
...
+  ##
+
+  # get a file using a DRS identifier
+  #
+  # * *params*
+  #   - +drs_id+ (String) => A DRS idenfifier, formatted as drs://[repository_hostname]/v1_[collection_id]_[file_id], where
+  #                          [repository_hostname] is a domain name, e.g. jade.datarepo-dev.broadinstitute.org,
+  #                          [collection_id] is a UUID of a collection, and
+  #                          [file_id] is a UUID of a file object, e.g:
+  #                          drs://jade.datarepo-dev.broadinstitute.org/v1_fcaee9e0-4745-4006-9718-7d048a846d96_5009a1f8-c2ee-4ceb-8ddb-40c3ddee5472
```
eweitz, line 303: Great docs.

lib/facet_name_converter.rb
```diff
@@ -0,0 +1,35 @@
...
+    donor_id: 'donor_id',
+    disease: 'disease',
+    library_preparation_protocol: 'library_construction_method',
+    organ: 'organ',
+    organism_age: 'organism_age',
+    sex: 'sex',
+    species: 'genus_species',
+    study_name: 'project_title',
+    study_description: 'project_description'
+  }
```
eweitz, line 19: What's the use case for storing equivalent mappings?

Naively, storing only non-equivalent mappings (e.g. `species: 'genus_species'`) seems more maintainable.

eweitz (approved):
Looks good!  I noted a handful of maintainability issues, but nothing non-trivial; no blockers.

app/models/data_repo_client.rb
```diff
@@ -56,6 +60,7 @@ def process_api_request(http_method, path, payload: nil, retry_count: 0)
     headers = {
       'Authorization' => "Bearer #{self.valid_access_token['access_token']}",
       'Accept' => 'application/json',
+      'Content-Type' => 'application/json',
```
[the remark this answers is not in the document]
bistline: It is safe to omit on `GET` requests, but not `POST`.  We get a `415 Unsupported Media Type` without it.

app/models/data_repo_client.rb
```diff
@@ -92,6 +97,47 @@ def process_api_request(http_method, path, payload: nil, retry_count: 0)
...
+  def api_status
+    path = self.api_root + '/status'
+    # make sure access token is still valid
+    headers = {
+      'Authorization' => "Bearer #{self.valid_access_token['access_token']}",
+      'Accept' => 'application/json',
+      'Content-Type' => 'application/json',
+      'x-app-id' => "single-cell-portal",
+      'x-domain-id' => "#{ENV['HOSTNAME']}"
+    }
```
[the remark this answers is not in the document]
bistline: Probably, but not the `Authorization: Bearer` token.  We need to check that on every request.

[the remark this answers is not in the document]
eweitz: Maybe via a new method `get_default_headers()`?

app/models/data_repo_client.rb
bistline: It actually is not - any controller automatically sets `logger` to `Rails.logger`, but other classes do not.

lib/facet_name_converter.rb
bistline: This is really just super quick-and-dirty, and I did the equivalent mappings just to ensure a return for a "known good".  I'm open to any suggestions on how to do this more robustly, in the absence of an ontology server to do conversions for us.

app/models/data_repo_client.rb
eweitz: Oof, I'll try to remember that.  Thanks for the note!

app/models/data_repo_client.rb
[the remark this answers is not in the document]
eweitz: Thanks for the detail.  To clarify, my previous comment was merely noting it'd be nice if this weren't required, i.e. if it were made unnecessary by an upstream change in Data Repo.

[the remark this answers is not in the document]
bistline: Ah - gotcha.  I thought you were making a suggestion.  I completely agree with the sentiment.

app/models/data_repo_client.rb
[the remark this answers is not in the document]
bistline: That's a great idea - this is actually shared across two clients, so we can move that into `ApiHelpers` so it covers `DataRepoClient` and `FireCloudClient`.

app/controllers/api/v1/search_controller.rb
```diff
@@ -178,7 +178,7 @@ def index
           if @preset_search.accession_list.any?
             sort_type = :accession_list if params[:terms].blank?
             @accession_list = @preset_search.accession_list
-            Rails.logger.info "Scoping search results to accessions from preset search: #{@preset_search.name}: #{@accession_list}"
+            logger.info "Scoping search results to accessions from preset search: #{@preset_search.name}: #{@accession_list}"
```
devonbush, line 181: interesting, I didn't realize that could be omitted

app/models/data_repo_client.rb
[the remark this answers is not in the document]
bistline: 0a45463 and 83e4cda

app/controllers/api/v1/search_controller.rb
```diff
@@ -926,6 +934,48 @@ def self.match_results_by_filter(search_result:, result_key:, facets:)
...
+      # execute a search in TDR and get back normalized results
+      def self.get_tdr_results(selected_facets)
+        query_json = ApplicationController.data_repo_client.generate_query_from_facets(selected_facets)
+        raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
+        results = {}
+        raw_tdr_results.dig('result').each do |result_row|
+          short_name = result_row.dig('project_short_name')
+          results[short_name] ||= {
+            name: result_row.dig('project_title'),
+            description: result_row.dig('project_description'),
```
devonbush, line 947: great use of 'dig' in this method

app/models/data_repo_client.rb
```diff
@@ -7,9 +7,13 @@ class DataRepoClient < Struct.new(:access_token, :api_root, :storage, :expires_a
   include ApiHelpers
 
   # Google authentication scopes necessary for querying TDR API
-  GOOGLE_SCOPES = %w(openid email profile)
+  GOOGLE_SCOPES = %w(openid email profile https://www.googleapis.com/auth/devstorage.read_only)
```
devonbush, line 10: how does this impact the scopes work from earlier this quarter?  Does this mean we will have to go back to prompting people again?

app/models/data_repo_client.rb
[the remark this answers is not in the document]
devonbush: It looks like you made the default headers in api_helpers, but then forgot to use it here?

devonbush (changes requested):
Looks really good--excited that this TDR integration is coming together at long last!  One question about scopes, and then one place where I think you just forgot to use a helper method you added

app/models/data_repo_client.rb
bistline: No - this is the read-only service account we're generating the token for, so we can add/remove scopes as we see fit.  Since we don't plan on using people's credentials for accessing TDR, it doesn't require the usual OAuth "3 legged" authentication/confirmation flow.  In the future, if we did want to use user credentials, we can use the same [SAM endpoint](https://sam.dsde-prod.broadinstitute.org/#/Google/getArbitraryPetServiceAccountToken) for generating pet service account tokens with the `devstorage.read_only` scope, and would still not need to prompt users.

app/models/data_repo_client.rb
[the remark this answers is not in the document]
bistline: Ah yes - good catch.  I'll update that.

devonbush (approved):

PR #1071, opened 2021-06-11
fixing inclusion/exclusion of annotation values in expression endpoint (SCP-3423)

devonbush:
SCP-3423  Certain patterns of gene searches meant that the cache would request only the expression and annotation values.  The API endpoint would then only return the expression values, causing the error.

TO TEST:
1. do a single-gene search in.a study
2. add a gene, and search again
3.  remove the first gene from the list, then search again
4. confirm the plots render correctly


eweitz (approved):
Nice quick fix.  This had reproduced for me in almost all cases of:

* Search 1 gene
* Search > 1 gene
* Search 1 gene

bistline (approved):

PR #1072, opened 2021-06-12
Adds auth error messages to scp-api.js calls

devonbush:
Previously, authentication errors were not handled in a uniform way, and many resulted in unhelpful errors displayed to users. This change pops a single, standard modal any time an ajax call from scp-api returns a 401/403

image

This modal dedupes these errors, so only a single modal & message will be displayed if there are multiple simultaneous errors (which is common for the explore tab, as multiple plots will all have requests that fail). 

TO TEST:
 1.  edit the first line of the `show` method in  `clusters_controller.rb` to be  `render json: {error: 'Unauthorized'}, status: 403 and return`
  2. load a study 
  3.  confirm a modal pops up as above


app/javascript/lib/scp-api.js
```diff
@@ -5,7 +5,7 @@
  *
  * API docs: https://singlecell.broadinstitute.org/single_cell/api
  */
-
+import React from 'react'
```
eweitz, line 8:
```suggestion
```

Is this left over from before modularizing `MessageModal`?

app/javascript/lib/scp-api.js
```diff
@@ -592,6 +594,15 @@ export default async function scpApi(
     } else {
       return [response, perfTimes, true]
     }
+  } else {
+    if (response.status === 401 || response.status === 403) {
+      showMessage(
+        <div>
+          Authentication failed<br/>
+          Your session may have timed out. Please sign in again<br/><br/>
+        </div>, 'api-auth-failure'
```
eweitz, line 603: Let's instrument this in a way that enables more analytics use cases, e.g.:

* Find all auth errors thrown by Rails non-API endpoints (currently logged via `error-modal`) and Rails API endpoints (logged via `message-modal`)
* Find all API errors, e.g. to make a graph of different errors
* Find errors by HTTP status code.  Requires less familiarity with our app-specific keys, useful if quickly comparing errors logged in Mixpanel to errors logged in e.g. Sentry or Stackdriver.

Likely not exactly this, but maybe something along the lines of:
```suggestion
        </div>, 'api-auth-failure', {
          source: 'api', // Did the server error come from the API controllers, other controllers, or elsewhere?
          url, // This has proven very useful in other Mixpanel use cases
          isOk: false, // Applies to both server and client errors
          messageType: 'client-error', // Applies for all 400-status errors
          statusCode: response.status // The specific standard, invariant error code
        }
```

eweitz (changes requested):
Nice UX refinement for errors!

I note some trivial issues, and where some small changes would likely get us more versatile and consistent logging that I think we need.

app/javascript/lib/scp-api.js
devonbush: we need this to be able to pass JSX into showMessage

bistline (approved):
No issues aside from the maintainability/analytics concerns Eric raised.

eweitz (approved):

PR #1074, opened 2021-06-15
Honor user-provided domain ranges on 3d clusters (SCP-3440)

bistline:
This update fixes a bug where study owner-provided domain ranges (i.e. x, y, and z axis bounds) were no longer being set in visualization requests, and produced an error in rendering when trying to compute the aspect ratio for 3d clusters.  These ranges are now included in all visualization requests for scatter plots (under the `axes` key).

MANUAL TESTING
As there are no 3d clusters in the synthetic studies, you will need to ingest a 3d cluster using [this cluster file](https://github.com/broadinstitute/single_cell_portal_core/blob/development/test/test_data/cluster_example.txt) from the `test/test_data` directory.  It can be added to any existing study and visualized by itself as it contains cluster-specific annotations.  When the file is uploaded, please use the following values for the various axis min/max settings:
```
x_axis_min: -65
x_axis_max: 65
y_axis_min: -55
y_axis_max: 55
z_axis_min: -75
z_axis_max: 75
```

Once this file has been ingested:
1. Load the selected study/cluster, and select the `Category` annotation
2. Verify that the cluster has a non-cube aspect ratio and looks like the following:
Screen Shot 2021-06-15 at 2 59 18 PM

This PR satisfies SCP-3440.

### DIFF
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -233,8 +233,7 @@ module Api
 
           axes_full = {
             titles: titles,
-            aspects: aspect,
-            ranges: cluster.domain_ranges
+            aspects: aspect
           }
 
           coordinate_labels = ClusterVizService.load_cluster_group_coordinate_labels(cluster)
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -228,12 +228,13 @@ module Api
           end
 
           if cluster.is_3d? && cluster.has_range?
-            aspect = ClusterVizService.compute_aspect_ratios(range)
+            aspect = ClusterVizService.compute_aspect_ratios(cluster.domain_ranges)
           end
 
           axes_full = {
             titles: titles,
-            aspects: aspect
+            aspects: aspect,
+            ranges: cluster.domain_ranges
           }
 
           coordinate_labels = ClusterVizService.load_cluster_group_coordinate_labels(cluster)
--- a/test/factories/cluster_group.rb
+++ b/test/factories/cluster_group.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   # creates a ClusterGroup object
   factory :cluster_group do
-    cluster_type { '2d' }
+    cluster_type { }
     cell_annotations { [] }
     name { study_file.name }
     study { study_file.study }
--- a/test/factories/cluster_group.rb
+++ b/test/factories/cluster_group.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   # creates a ClusterGroup object
   factory :cluster_group do
-    cluster_type { }
+    cluster_type { '2d' }
     cell_annotations { [] }
     name { study_file.name }
     study { study_file.study }
--- a/test/factories/study_file.rb
+++ b/test/factories/study_file.rb
@@ -48,7 +48,7 @@ FactoryBot.define do
         cell_input {
           {}
         }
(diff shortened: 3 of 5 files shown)


eweitz (approved):

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -228,12 +228,13 @@ def self.get_cluster_viz_data(study, cluster, url_params)
...
           if cluster.is_3d? && cluster.has_range?
-            aspect = ClusterVizService.compute_aspect_ratios(range)
+            aspect = ClusterVizService.compute_aspect_ratios(cluster.domain_ranges)
           end
 
           axes_full = {
             titles: titles,
-            aspects: aspect
+            aspects: aspect,
+            ranges: cluster.domain_ranges
```
devonbush, line 237: we're already sending the `domain_ranges` in the response as `userSpecifiedRanges`.  I think the bug is in get3DScatterProps() in `ScatterPlot.js`.  There we check for userSpecifiedRanges, but are still trying to read values from axes.ranges.

Alternatively, if you think sending them as part of the axes is clearer (I don't have strong feelings, except I'd prefer that something in the variable name should indicate they were set by the study owner, and aren't just auto-computed), feel free to remove that userSpecifiedRanges from the root object and update the frontend code in ScatterPlot.

test/factories/cluster_group.rb
```diff
@@ -1,7 +1,7 @@
 FactoryBot.define do
   # creates a ClusterGroup object
   factory :cluster_group do
-    cluster_type { '2d' }
+    cluster_type { }
```
devonbush, line 4: remember, that's just saying the default is '2d', so I'd leave this as-is, as I think that's a sensible default

test/factories/study_file.rb
```diff
@@ -48,7 +48,7 @@
         cell_input {
           {}
         }
-        cluster_type { '2d' }
+        cluster_type { cell_input.dig(:z).present? ? '3d' : '2d' }
```
devonbush, line 51: Nice

devonbush (changes requested):
this is some good work on a tricky edge case.  I want to make sure we're not sending the same data twice, but otherwise this looks good, and appreciate the tests.

  FYI: We do have a synthetic study with 3d clusters -- 'human_raw_counts' has a cluster 'raw1_human_5k_cells_80_genes.cluster_3D.txt'.   I could see a separate ticket 'Upgrade testing for plot axes' that would involve updating the synth populator to allow specifying ranges, and then updating a couple of the synth studies to have ranges specified for 2d and 3d plots.  (and maybe even adding a frontend test for checking that we can pass the right stuff to plotly)

app/controllers/api/v1/visualization/clusters_controller.rb
bistline: Ah - thanks for the clarification.  You're right in that it is being shipped twice... I just glossed over the `userSpecifiedRanges`.  I think it's fine to leave it in there since it already exists, and I can adjust to make sure we read/honor those, and update the tests accordingly.

bistline:
> FYI: We do have a synthetic study with 3d clusters -- 'human_raw_counts' has a cluster 'raw1_human_5k_cells_80_genes.cluster_3D.txt'. I could see a separate ticket 'Upgrade testing for plot axes' that would involve updating the synth populator to allow specifying ranges, and then updating a couple of the synth studies to have ranges specified for 2d and 3d plots. (and maybe even adding a frontend test for checking that we can pass the right stuff to plotly)

I'll open that ticket now.

devonbush (approved):

PR #1078, opened 2021-06-17
Adding user-friendly error message to prod environments (SCP-3427)

devonbush:
SCP-3427 This updates the ErrorBoundary code to display a user-friendly error message in production, but a full stack trace in all other environments.  (as a reminder -- ErrorBoundary handles errors that occur during the direct rendering of React components.  Errors that occur during callbacks/async operations cannot be handled by ErrorBoundary)
PRODUCTION VIEW:
image

OTHER ENVIRONMENTS:
image

 `process.env.VIEW_ENV` is introduced as the variable that tracks the environment, this allows it to be easily changed during development workflow.  To see the production view,  just change line 4 of development.js to `process.env.VIEW_ENV = 'production'`, and restart the webpack server.  
 The alternative, using either NODE_ENV or RAILS_ENV directly, would mean that you couldn't test the production views without doing a full production build, which disables the live reload server and breaks the non-dockerized development workflow.  

This change therefore partially addresses SCP-2237, but I will hold off on updating the other environment references in JS until this is confirmed to work in actual production.

TO TEST:
 1. update a component that's within an error boundary to have an obvious JS error in its main render path.  (e.g. change line 45 of ScatterPlot.js to be `const [isLoading, setIsLoading] = foobar(false)`
 2. Load the explore tab for a study
 3. confirm an error message appears with full stack trace
 4. edit `config/webpack/development.js:4` to be `process.env.VIEW_ENV = 'production'`
 5. restart your webpack dev server and reload the page
 6. confirm the user-friendly error message appears with the support email

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -42,7 +42,7 @@ function RawScatterPlot({
   studyAccession, cluster, annotation, subsample, consensus, genes, scatterColor, dimensions,
   isAnnotatedScatter=false, isCorrelatedScatter=false, isCellSelecting=false, plotPointsSelected, dataCache
 }) {
-  const [isLoading, setIsLoading] = duseState(false)
+  const [isLoading, setIsLoading] = useState(false)
   const [spearman, setSpearman] = useState(null)
   const [scatterData, setScatterData] = useState(null)
   const [graphElementId] = useState(_uniqueId('study-scatter-'))
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -42,7 +42,7 @@ function RawScatterPlot({
   studyAccession, cluster, annotation, subsample, consensus, genes, scatterColor, dimensions,
   isAnnotatedScatter=false, isCorrelatedScatter=false, isCellSelecting=false, plotPointsSelected, dataCache
 }) {
-  const [isLoading, setIsLoading] = useState(false)
+  const [isLoading, setIsLoading] = duseState(false)
   const [spearman, setSpearman] = useState(null)
   const [scatterData, setScatterData] = useState(null)
   const [graphElementId] = useState(_uniqueId('study-scatter-'))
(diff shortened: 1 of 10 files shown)


ehanna4:
Is "Staging" intended to be most like "Prod" or "local/dev" ? If the former would it make sense to have the error messaging appear the same in "Staging" and "Prod"?

devonbush:
> Is "Staging" intended to be most like "Prod" or "local/dev" ? If the former would it make sense to have the error messaging appear the same in "Staging" and "Prod"?

Good question -- I put the full error trace in for easier debugging of staging errors, but I think you're right that it should be prod-like.  I'll switch the logic so the full trace only appears in dev

eweitz (changes requested):
Good stuff!  I tested this locally and confirm it works.

I like the use of Node environment, but [other scenarios](https://github.com/broadinstitute/single_cell_portal_core/pull/450#discussion_r392524219) seem like a better use case.  Its use here slightly decreases observability, and I don't think the benefit of removing technical details from the error UI offsets that cost.  Details below!

bistline (approved):
Very helpful enhancement, but I agree with Eric's sentiment that this would also be useful in `production` as well for QA/UL purposes.  I'm approving proactively with the request that we apply this to all environments.

devonbush:
Yup, that's a good point.  I was less thinking of UX, but just the general security principle of 'never expose stack traces to users'.  But since this is all client-side, I don't think the security principle applies

devonbush:
Updated: 
image

eweitz (approved):

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -42,7 +42,7 @@ function RawScatterPlot({
   studyAccession, cluster, annotation, subsample, consensus, genes, scatterColor, dimensions,
   isAnnotatedScatter=false, isCorrelatedScatter=false, isCellSelecting=false, plotPointsSelected, dataCache
 }) {
-  const [isLoading, setIsLoading] = useState(false)
+  const [isLoading, setIsLoading] = duseState(false)
```
ehanna4, line 45: typo?

devonbush: ah, yup, leftover from when I was deliberately generating errors.  thanks!

devonbush:
test failures are caused by upstream TDR change -- unrelated to this work

PR #1081, opened 2021-06-23
Refine gene search autocomplete in Study page (SCP-3432)

eweitz:
This fixes a UI/UX bug with gene search autocomplete suggestion, and refines its other aspects in the Study page.  It shows:

* Exact matches first
* Suggestions (performantly!) even upon entering 1 character
* Prefix matches above other substring matches
* Similar results even if they're not precise substring matches

Old UI/UX bug: exact match not visible by default
Screen Shot 2021-06-23 at 12 57 06 PM

New UI/UX: exact match ranked first
Screen Shot 2021-06-23 at 12 58 54 PM

To test:
* Go to a Study page with many genes
* Search for "nf2"
* Verify that "Nf2" (or "NF2") appears first among autocomplete suggestions

This satisfies SCP-3432.


devonbush (changes requested):
Cool stuff -- big improvement in relevance and that it can appear after just one character.

Must have: This should gracefully handle the case where a study has no genes (currently it errors out)
Probably should have: 
  * Sort the prefix matches lexicographically
  * some indication that the list is truncated at 20, so users

app/javascript/lib/search-utils.js
```diff
@@ -0,0 +1,42 @@
...
+  const text = inputText.toLowerCase()
+
+  const exactMatch = targets.find(gene => gene === inputText)
+
+  // Get genes that start with the input text
+  const prefixMatches =
+    targets.filter(gene => {
+      return gene !== inputText && gene.toLowerCase().startsWith(text)
+    })
+
```
devonbush, line 26: I don't know how expensive the simila operations are, but it might be worth skipping the lines below if we already have >20 prefix matches (which I suspect will often be the case when they've just typed one letter)

app/javascript/lib/search-utils.js
```diff
@@ -0,0 +1,42 @@
...
+        .filter(match => {
+          const target = match.target
+          return target !== inputText && !prefixMatches.includes(target)
+        })
+        .map(match => match.target)
+
+  // Show top 20 matches -- exact match, prefix matches, then similar matches
+  const topMatches = prefixMatches.concat(similarMatches)
+  if (exactMatch) {topMatches.unshift(exactMatch)} // Put any exact match first
+  return topMatches.slice(0, 20)
```
devonbush, line 41: the 20 here should either be a constant at the top of this file, or a param to the method (I'd lean towards the latter)

app/javascript/lib/search-utils.js
eweitz: Done, good idea. It's very fast, but why not optimize?

app/javascript/lib/search-utils.js
eweitz: Makes sense, I've parameterized this.  Per our chat about truncation indicator etc., I've limited it to 8 -- the number of suggestions shown before scrolling is available.

bistline (approved):
Looks good, modulo Devon's suggestions.  I'm somewhat ambivalent about telling the user that there are more results than shown, I feel like this is standard behavior for autocomplete, and most users would be saavy enough to know that as they continue typing, results get more accurate.

eweitz:
> Must have: This should gracefully handle the case where a study has no genes (currently it errors out)

Done.

> Probably should have: Sort the prefix matches lexicographically

Done.

> Maybe should have: some indication that the list is truncated at 20, so users know they're seeing a subset of possible matches

Worked around by limiting list to 8, which avoids scrolling.  All other autocomplete UIs we looked at -- GTEx, gnomAD, NCBI, Google -- did the same.

app/javascript/lib/search-utils.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Returns top matches: exact prefix matches, then similar matches
+ *
+ * @param {String} inputString String typed by user into text input
+ * @param {Array<String>} targets List of strings to match against
+ * @param {Integer} numSuggestions Number of suggestions to show
+ */
+export function getAutocompleteSuggestions(inputText, targets, numSuggestions=8) {
+  // Autocomplete when user starts typing
+  if (!targets || targets.length === 0 || !inputText) {
```
devonbush, line 14: FYI that this is a case where you could just do 
`if (!targets?.length || !inputText) { `

devonbush (approved):

eweitz:
The [build failed](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/230618360#L3095) due to upstream Terra Data Repo changes, unrelated to this PR.

app/javascript/lib/search-utils.js
eweitz: Neat, done in 1261991.

PR #1085, opened 2021-06-24
Data Repo search results integration (SCP-3431)

bistline:
This update provides minimal scaffolding to query for and return search results from Terra Data Repo to the SCP search UI.  The React integration is extremely minimal, and continued in SCP-3420.  Currently, keyword and faceted search are supported.  This update also covers converting from SCP metadata convention names to Terra Interoperability Model names (via `FacetNameConverter`).  Work still left to be done:

* Refining TDR search result weighting/matching for keywords/facets
* Adding API endpoints to include TDR results in bulk download requests
* UX work to show TDR results in bulk download modal

MANUAL TESTING

* Pull this branch and boot as normal
* Set the `cross_dataset_search_backend` feature flag to `true` for your account
* Perform any of the following searches:
  * Keyword search: `pulmonary`
  * Faceted search: `Species: Homo sapiens`
  * Combination of above searches
* Note the result at the end for `Single-cell RNA-sequencing reveals profibrotic roles of distinct epithelial and mesenchymal lineages in pulmonary fibrosis`


app/javascript/components/search/results/DataRepoStudy.js
```diff
@@ -0,0 +1,66 @@
...
+                return filter.name
+              }
+            }).join(',')
+          }
+        </span>)
+    })}
+  </>)
+}
+
+export default function DataRepoStudy({ study }) {
```
ehanna4, line 50: It seems to me there is enough overlap in the `SCPStudy` and `DataRepoStudy` that they could be within one component to reduce some of the duplication, I was making that change in my branch (which is branched off this one) so not blocking but just wanted to mention it.

bistline: I would say that any React refactoring you want to do in your branch should supersede anything here.  These components were simply a _very_ rough way of getting the TDR result into the UI.

app/controllers/api/v1/search_controller.rb
```diff
@@ -291,15 +291,18 @@ def index
...
-          logger.info pp @trd_results
+          logger.info "Found #{@tdr_results.keys.size} results in Terra Data Repo"
+          logger.info pp @tdr_results if @tdr_results.any?
+          @tdr_results.each do |short_name, tdr_result|
+            @studies << tdr_result
+          end
         end
 
-        @matching_accessions = @studies.map(&:accession)
+        @matching_accessions = @studies.map {|study| study.is_a?(Study) ? study.accession : study[:accession]}
```
devonbush, line 305: FYI: this could be replaced with just `@studies.map{ |study| study[:accession] }`  but it might be clearer the way you have it that 'hey, there are two different kinds of things in this list'

app/controllers/api/v1/search_controller.rb
```diff
@@ -664,32 +664,65 @@ def self.match_results_by_filter(search_result:, result_key:, facets:)
...
+        end
+        if terms.present?
+          term_json = ApplicationController.data_repo_client.generate_query_from_keywords(terms)
+        end
+        # now we merge the two queries together to perform a single search request
+        query_json = ApplicationController.data_repo_client.merge_query_json(facet_query: facet_json, term_query: term_json)
+        logger.info "Executing TDR query with: #{query_json}"
+        raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
+        raw_tdr_results['result'].each do |result_row|
+          # get column name mappings for assembling results
```
devonbush, line 680: this section should get pulled out into a helper method `process_tdr_result_row`.  That will help in clarity and testing

app/controllers/api/v1/study_search_results_objects.rb
```diff
@@ -26,39 +24,53 @@ def search_results_obj
...
-          study_obj[:term_matches] = inferred_weight[:terms].keys
-          study_obj[:term_search_weight] = inferred_weight[:total]
-        end
-        if @preset_search.present? && @preset_search.accession_list.include?(study.accession)
-          study_obj[:preset_match] = true
-        end
-        if @gene_results.present?
-          study_obj[:gene_matches] = @gene_results[:genes_by_study][study.id].uniq
-          study_obj[:can_visualize_clusters] = study.can_visualize_clusters?
+        if study.is_a?(Study)
```
devonbush, line 27: I would put each of the two sides of this if statement into helper methods "scp_study_response_obj" and "tdr_study_response_obj", again for clarity and easier testing

app/javascript/components/search/results/DataRepoStudy.js
devonbush: Concur.  My two cents is that I would lean towards keeping these as separate components, since they are likely to be distinct visual entities. And since they share a lot of common attributes, the DataRepoStudy component might import a lot of logic and display from the SCPStudy component.  But that's something the implementor can figure out :)

lib/facet_name_converter.rb
```diff
@@ -1,35 +1,51 @@
 # class for converting from alexandria convention names to HCA or TIM metadata model names (i.e. columns, not individual values)
 # this is currently for PoC work on XDSS - eventually this will be replaced by an onotology server that can handle
 # conversions programmatically
+# entries have a :name (column name in query results for facet matching) and :id (ElasticSearch encoded property name)
 class FacetNameConverter
   # map of alexandria metadata convention names to HCA 'short' names
-  # not fully namespaced as this is what is indexed in TDR
   ALEXANDRIA_TO_HCA = {
```
devonbush, line 7: Curious, do we foresee any use case for this HCA conversion going forward?  Or are we just keeping it while we're in PoC stage in case TDR throws curveballs at us?

devonbush (changes requested):
Nice work!  a few minor suggestions, the only one I feel strongly about is the helper method for process_tdr_result_row

app/controllers/api/v1/search_controller.rb
bistline: Good call.

app/controllers/api/v1/study_search_results_objects.rb
bistline: Yeah, this was super "I just want to get this in the UI I don't care how gross this is" and I never went back to do the actual refactoring.

test/integration/data_repo_client_test.rb
```diff
@@ -18,8 +18,8 @@ class DataRepoClientTest < ActiveSupport::TestCase
 
   # skip a test if the TDR API is not up (since it is their dev instance there is no uptime guarantee)
   def skip_if_api_down
-    if !@data_repo_client.api_available?
-      puts "-- skipping due to TDR API being unavailable --" ; skip
+    unless @data_repo_client.api_available?
```
eweitz, line 21: Ironically, even though it's plain English, I find `unless` harder to understand than `if !`.  I'll adapt!

lib/facet_name_converter.rb
bistline: I think that `ALEXANDRIA_TO_HCA` might get dumped from here as it seems that TDR will be using TIM names for everything.  I left it in for now, but it is somewhat vestigial currently.

test/integration/data_repo_client_test.rb
```diff
@@ -144,68 +144,110 @@ def skip_if_api_down
...
   end
 
   test 'should get file in snapshot' do
     skip_if_api_down
     file_info = @data_repo_client.get_snapshot_file_info(@snapshot_id, @file_id)
     assert file_info.present?
-    assert_equal @file_id, file_info.dig('fileId')
-    assert file_info.dig('size') > 0, 'did not get a valid file size'
+    assert_equal @file_id, file_info['fileId']
+    assert file_info['size'].positive?, 'did not get a valid file size'
```
eweitz, line 157: I find `> 0` easier to understand than `.positive?`.  The latter is idiomatic Ruby, but also _uniquely_ Ruby.

test/integration/data_repo_client_test.rb
```diff
@@ -144,68 +144,110 @@ def skip_if_api_down
...
+    keywords = %w[pulmonary human lung]
+    name_field = FacetNameConverter.convert_to_model(:tim, :study_name, :id)
+    description_field = FacetNameConverter.convert_to_model(:tim, :study_description, :id)
+    expected_query = "(#{name_field}:pulmonary OR #{name_field}:human OR #{name_field}:lung) OR " \
+                     "(#{description_field}:pulmonary OR #{description_field}:human OR #{description_field}:lung)"
+    query_json = @data_repo_client.generate_query_from_keywords(keywords)
+    assert_equal expected_query, query_json.dig(:query_string, :query)
+  end
+
+  test 'should merge query json for facets and keywords' do
```
eweitz, line 190:
```suggestion
  test 'should merge query JSON for facets and keywords' do
```

app/javascript/components/search/results/DataRepoStudy.js
```diff
@@ -0,0 +1,66 @@
...
+  const matchedKeys = Object.keys(matches)
+    .filter(key => key != 'facet_search_weight')
+  return (<>
+    { matchedKeys.map((key, index) => {
+      const helpText = `Metadata match for ${key}`
+      return (
+        <span key={index}
+              className="badge badge-secondary facet-match"
+              data-toggle="tooltip"
+              title={helpText}>
```
eweitz, line 35: SCP JS uses two-space offsets in cases like this.

```suggestion
          className="badge badge-secondary facet-match"
          data-toggle="tooltip"
          title={helpText}>
```

app/javascript/components/search/results/DataRepoStudy.js
```diff
@@ -0,0 +1,66 @@
...
+
+export function shortenDescription(textDescription) {
+  const suffixTag = <span className="detail"> ...(continued)</span>
+  const displayedStudyDescription = { __html: textDescription.slice(0, descriptionCharacterLimit) }
+  if (textDescription.length>descriptionCharacterLimit) {
+    return <>
+      <span className="studyDescription" dangerouslySetInnerHTML={displayedStudyDescription}></span>{suffixTag}
+    </>
+  } else {
+    return <><span className = 'studyDescription' dangerouslySetInnerHTML={displayedStudyDescription}></span></>
```
eweitz, line 16: Blocker: SCP JS always uses kebab-case for attribute values.

```suggestion
      <span className="study-description" dangerouslySetInnerHTML={displayedStudyDescription}></span>{suffixTag}
    </>
  } else {
    return <><span className="study-description" dangerouslySetInnerHTML={displayedStudyDescription}></span></>
```

(It also typically double quotes them with no space around the equals sign.)

app/javascript/components/search/results/DataRepoStudy.js
```diff
@@ -0,0 +1,66 @@
+/* show a search result entry from Data Repo */
```
eweitz, line 1: Starting comments this way gets them picked up in IDEs (e.g., hovering over the module name in an `import`).

```suggestion
/** show a search result entry from Data Repo */
```

eweitz (changes requested):
Looking pretty good!

I noted a few trivial blockers -- all such comments start with "Blocker".  My other suggestions are all a bit more subjective or lower priority.

lib/facet_name_converter.rb
devonbush: cool, yeah, I like leaving it in while everything is still in relative flux

app/javascript/components/search/results/ResultDetail.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable require-jsdoc */
+import React from 'react'
+import DataRepoStudy from "components/search/results/DataRepoStudy";
+import SCPStudy from "components/search/results/SCPStudy";
```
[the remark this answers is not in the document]
bistline: Good to know - that was autocompleted by my IDE but I'll override those.

test/integration/data_repo_client_test.rb
bistline: That was a `rubocop` suggestion which was why I did it.  I tend to agree, but I also wanted to start getting in the practice of accepting the linter's advice...

app/javascript/components/search/results/DataRepoStudy.js
bistline: Interesting... I actually have no idea how that got in there.  Will change it back.

app/javascript/components/search/results/DataRepoStudy.js
bistline: I clearly need to update my IDE with the correct JS linting!  I think it did it before, but since I upgraded my laptop it's apparently applying the wrong rules.

test/integration/data_repo_client_test.rb
eweitz: Given [Rubocop's default here](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericPredicate) impedes readability for both of us, we could change the linting rule by adding:

```
Style/NumericPredicate
  Enabled: false
```

in `rubocop.yml`.  (We occasionally change defaults in our JS lint rules, via `.eslintrc.js`, as well.)

eweitz (approved):

devonbush (approved):

PR #1086, opened 2021-06-25
Add a visual cue that a study is TDR (SCP-3420)

ehanna4:
Branched from Jon's initial work so the bulk of the changes are coming from his branch. Probably should wait to review until his is merged.

Main changes introduced by this branch
Code things: 
1. Consolidation of SCPStudy and TDRStudy components into one StudySearchResult component.
2. Some linting/formatting
3. Added highlighting to HCA/TDR study search results

Feature:
1. Adding a badge to the right of the title with a tool tip to indicate whether a study is from TDR/HCA
Screen Shot 2021-06-30 at 1 05 09 PM

Please note the TDR studies do not currently go anywhere upon clicking them. This is known and awaiting work with the TDR folks so they are not links anymore

Also added the highlighting for text searches:
Screen Shot 2021-06-30 at 1 05 04 PM


devonbush:
I think 'visual cue' is a better title for this PR :)

ehanna4:
> I think 'visual cue' is a better title for this PR :)

oops 😅 🤦‍♀️

devonbush:
Nice, I like the attention to narrow screen resolutions.  I wonder if we can drop the badge for SCP studies, since that's the 'default'.  And that will help the TDR results stand out more as something different.  Code looks reasonable, although I suspect you'll want to consolidate some of the markup that's shared between the TDR and SCP result displays.  Also I'm intrigued by the use of `grid-template-columns` that's a CSS attribute that's new to me, but looks very concise and powerful

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -129,33 +142,57 @@ function facetMatchBadges(study) {
...
-          { facetMatchBadges(study) }
+    return (
+      <>
+        <div key={study.accession}>
+          <label htmlFor={study.name} id="result-title" className={'study-item'}>
+            <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle}/>{inferredBadge}
+            <span className="badge badge-secondary study-type"
+              data-toggle="tooltip"
+              title={'Study comes from Single Cell Portal'}>
+              SCP </span>
```
eweitz, line 167: That a study in SCP comes from SCP ought to be implied, I think.

By analogy, Google search marks results that are not default -- e.g. with a small "PDF" badge.  Like Google not marking results that are web pages with an "HTML" badge, I think not marking SCP study results with "SCP" would improve visual signal-to-noise ratio.


```suggestion
```

(I see Devon beat me to this suggestion, but thought this inline concurrence might help.)

app/controllers/api/v1/study_search_results_objects.rb
```diff
@@ -61,7 +62,7 @@ def study_response_obj(study)
           end
         else
           study_obj = {
-            tdr_result: true,
+            study_source: 'TDR',
```
ehanna4, line 65: switched to a string `study_source` (if anybody has a better name tho totally open) so that in case ever in the future there are other sorts of studies we'd be ready rather than just being TDR specific

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -128,31 +129,50 @@ function facetMatchBadges(study) {
...
+  const studyTitle = highlightText(study.name, termMatches).styledText
+  const studyDescription = formatDescription(study.description, termMatches)
+  const displayStudyTitle = { __html: studyTitle }
 
   return (
     <>
       <div key={study.accession}>
-        <label htmlFor={study.name} id= 'result-title'>
-          <a href={study.study_url} dangerouslySetInnerHTML = {displayStudyTitle}></a>{inferredBadge}
+        <label htmlFor={study.name} id="result-title" className={'study-label'}>
```
ehanna4, line 168: emily - remove curly bracs

ehanna4: emily - go change the "result-title" to a class rather than id

app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -47,7 +47,7 @@ const helpModalContent = (<div>
     target="_blank" rel="noreferrer">wiki
   </a>.  Study authors looking to make their studies more accessible can read our{' '}
```
eweitz, line 48: Your solution for spaces before links in JSX is more elegant than mine, so let's remove my redundant solution:

```suggestion
  </a>.  Study authors looking to make their studies more accessible can read our
```

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -128,31 +129,50 @@ function facetMatchBadges(study) {
...
+          <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} />
```
eweitz, line 169: Blocker: clicking on this link for TDR results does nothing -- it brings the user to the top of the page.  That strikes me as a UX bug.

Unless there are plans/designs on where TDR result links would go, let's not wrap their titles in a link.  This avoids confusing users who click on links and go nowhere.  Getting Jerome's input on this might be worthwhile!

eweitz (changes requested):
This is coming along well!  I tested it locally.

I noted one blocker, a seeming UX bug with non-functional TDR result links.  Beyond that, I suggested a handful of UI and code style refinements.

app/javascript/styles/_resultsPanel.scss
```diff
@@ -126,3 +126,18 @@
...
+  border: 1px solid #4d72aa;
+  background-color: white;
+  font-weight: normal;
+  color: #777;
+  width: 44px;
+  height: 20px
+}
+
+.study-label {
+  grid-template-columns: 6fr 0.25fr;
```
bistline, line 140: This rule is causing issues with the inferred search "text match" label that shows up when we run a search using a facet, but then also attempt to make a text match using the facet label.  Example: running a search on production using `disease: COVID-19` will match some studies because they have "COVID-19" in their title or description.  This displays a badge like this with a tooltip explaining what the match means:

Screen Shot 2021-06-29 at 11 28 46 AM

This new CSS rule will result in those badges looking like this:

Screen Shot 2021-06-29 at 11 28 59 AM

I think we either need a different class for this particular badge, or we need to handle the grid spacing or word wrapping differently.

bistline (changes requested):
Overall this looks good, modulo the blocker on the CSS for the new SCP/TDR badges.

Also, non-blocking: it might be nice to shade the div for TDR matches slightly so that they are more distinctive at a first glance.  The badges are a nice touch, but it is very easy to miss them (especially when they're right-aligned and the study name is short).

app/javascript/components/search/results/StudySearchResult.js
bistline: @eweitz unfortunately there is an issue on the TDR UI side where external hyperlinks do not resolve to an entity in the TDR UI.  Even if we had a valid link (which is unclear how we would generate it from the search results, as there is no reference to the snapshot/dataset), clicking it in a browser just dumps you on the home page, even after you've authenticated.  The TDR team took a look and they are apparently not caching the link origin after they complete the auth challenge.

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -128,31 +129,50 @@ function facetMatchBadges(study) {
   </>)
 }
 
-/* displays a brief summary of a study, with a link to the study page */
-export default function SCPStudy({ study }) {
-  const termMatches = study.term_matches
-  const studyTitle= highlightText(study.name, termMatches).styledText
-  const studyDescription = formatDescription(study.description, termMatches)
-  const displayStudyTitle = { __html: studyTitle }
+// generate a cell count badge for SCP studies
```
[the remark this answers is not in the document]
ehanna4: thanks for all the comment formatting fixes, i will make sure to fix up my editor/lint settings

app/javascript/components/search/results/StudySearchResult.js
eweitz: Thanks for the context!  So, indeed, we should not show TDR titles as links.

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -128,31 +129,50 @@ function facetMatchBadges(study) {
...
-    inferredBadge = <span className="badge soft-badge" data-toggle="tooltip" title={helpText}>text match only</span>
+    return <span className="badge soft-badge" data-toggle="tooltip" title={helpText}>text match only</span>
   }
+}
+
+/* generate a badge to indicate to users the study origin for non-SCP studies */
+function studyTypeBadge(study) {
+  if (study.study_source === 'TDR') {
+    return <span className="badge badge-secondary study-type" data-toggle="tooltip"
+      title={'Study comes from Terra Data Repo'}> TDR </span>
```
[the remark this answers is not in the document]
ehanna4: I updated the tool tip to reference HCA & TDR, please let me know if it looks alright or if you'd still like to see line breaks or other formatting - thanks!

app/javascript/styles/_resultsPanel.scss
ehanna4: I think this is fixed, I also moved the "HCA" badge over so it's not far-right-aligned anymore so didn't change the coloring, please take a look and let me know though if you still think that would be helpful!

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -128,31 +129,52 @@ function facetMatchBadges(study) {
...
+}
 
-  let inferredBadge = <></>
+/** Generate the inferredBadge for SCP studies */
+function inferredBadge(study, termMatches) {
   if (study.inferred_match) {
     const helpText = `${termMatches.join(', ')} was not found in study metadata,
      only in study title or description`
-    inferredBadge = <span className="badge soft-badge" data-toggle="tooltip" title={helpText}>text match only</span>
+    return <span className="badge soft-badge match-badge" data-toggle="tooltip" title={helpText}>text match only</span>
```
devonbush, line 146: do we need the extra match-badge class?  I don't see soft-badge used anywhere else, so maybe just consolidate those styles for now?

devonbush (approved):
Looks good, just one non-blocking question about css classes

app/javascript/components/search/results/StudySearchResult.js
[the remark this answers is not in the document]
eweitz: Looks good, nicely refined!

eweitz (approved):

app/controllers/api/v1/study_search_results_objects.rb
jlchang: Good idea! The latest Emerald Empire update mentioned that they're ingesting LungMap data, which we're told may have cell-level metadata so it'd be good to be prepared for more than one study_source :)

jlchang (approved):
No blocking issues, just a comment on findability.

The cell number and matching facet "bubbles" are all in the 2nd line, left justified. I worry that having the HCA "bubble" come after the end of the Study title will cause it to be easily overlooked... Also, the text for "HCA" seems grey - do we want it more bold so it jumps out a bit more?

ehanna4:
Screen Shot 2021-07-07 at 9 40 44 AM

based on feedback have moved the "TDR" badge down to the line where the cell count and facet match bubbles are. Also wrote out "Human Cell Atlas". Hoping to get final sign off from Jerome before merging for colors.

bistline (approved):

PR #1087, opened 2021-06-25
Allow re-upload of same study file after successful deletion (SCP-3451)

bistline:
A regression was introduced during the Rails 6.1  & `CarrierWave` [upgrade](#1030) where the `StudyFile` attribute `upload_file_name` is not always being successfully cleared when invoking `DeleteQueueJob`, which results in the user not being able to upload a file with the same filename after deleting it.  This is a common workflow when addressing file validation issues.  The error is not completely reproducible - it seems somewhat random as to when it happens - although files uploaded prior to the upgrade seem much more likely to encounter this issue.

This update fixes the problem via the following mechanisms:
1. `queued_for_deletion: false` is now appended to queries where SCP is trying to find a file in a study using `upload_file_name`
2. Validating uniqueness of `upload_file_name` ignores files where `queued_for_deletion: true`
3. `DeleteQueueJob` now automatically deletes `StudyFile` documents once all parsed data has been removed from the database

The third change is the main mechanism for correcting the regression, whereas the first & second changes address potential race conditions while `DeleteQueueJob` runs.

MANUAL TESTING

1. Sign in and create a new study (or use an existing study)
2. Add a `Gene List` file, using `test/test_data/marker_1_gene_list.txt`
3. Add a doc/miscellaneous file, using `test/test_data/mock_study_doc_upload.txt`
4. Delete both files
5. Without refreshing the page, re-upload both files and note that neither throws an error saying the file already exists

This PR satisfies SCP-3451.


devonbush (changes requested):
That's a bummer that this isn't consistently reproducible.  Adding the `:queued_for_deletion` qualifier seems like a reasonable workaround, but it's dangerous in that we can no longer rely on names being unique.  Now we have to make sure we're never querying just by name or else we'll get strange and hard-to-reproduce bugs.  So if at all possible, I'd prefer to fix this by ensuring the name gets updated instead.   But if we're going this route, here's more places I found that might need their query updated with the `queued_for_deletion: false` qualifier.  (and I would lean towards adding it everywhere, even if not strictly necessary, just for good hygiene)

studies_controller.rb:1298

site_controller.rb:1070

cluster_viz_service.rb:16 & 51 (these probably should have had the queued_for_deletion filter even before this change)

synthetic_study_populator.rb:102, 143, 158, 178. (probably not necessary here since we've just recreated the study, but I think it would be good hygiene to add the queued_for_deletion qualifier)

parse_utils.rb:562  ???

study_file_bundle.rb:250

bistline:
> So if at all possible, I'd prefer to fix this by ensuring the name gets updated instead. 

I would vastly prefer that as well, but it seems almost random as to when it does update and when it does not.  It has to do with[ this line](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/delete_queue_job.rb#L106) in `DeleteQueueJob`.  For some reason, this doesn't always update `upload_file_name`, but when it doesn't, no error is emitted.  I have a suspicion that it has to do with the couplet of `update_attributes` and `save!(validate: false)`, but that was a workaround to avoid issues with `expression_file_info` on older expression matrices not deleting at all.

I'll keep trying things to find out how to reliably reproduce it, but in the absence of that, I'm not sure what else we can do.

eweitz (approved):

eweitz (approved):

app/models/delete_queue_job.rb
```diff
@@ -22,6 +22,9 @@ def perform
       file_type = object.file_type
       study = object.study
 
+      # delete embedded expression_file_info to prevent validation errors later, if present
+      object.expression_file_info.delete if object.expression_file_info.present?
+
```
devonbush, line 27: I don't understand why this is necessary.  Is this leftover from troubleshooting?

app/models/delete_queue_job.rb
```diff
@@ -99,12 +102,8 @@ def perform
...
       end
 
-      # queue study file object for deletion, set file_type to DELETE to prevent it from being picked up in any queries
-      # use assign_attributes and save!(validate: false) to avoid any validation issues that might prevent saving
-      # this ensures file is queued for deletion, as dependent records have already been deleted
-      new_name = "DELETE-#{SecureRandom.uuid}"
-      object.assign_attributes({queued_for_deletion: true, upload_file_name: new_name, name: new_name, file_type: 'DELETE'})
-      object.save!(validate: false)
+      # it is safe to call :destroy now as all child documents have been removed
+      object.destroy
```
devonbush, line 106: oh, very cool, so we'll just no longer have the renamed files lingering around for cleanup, we can just delete them straightway.  What was the original reason we weren't deleting them here?

devonbush (changes requested):
I think there's some leftover troubleshooting code, and then one question

app/models/delete_queue_job.rb
bistline: Ah yeah, that's a leftover now.  Since we're not calling `save(validate: false) anymore, this isn't needed.  I was testing to see if preemptively clearing out the `expression_file_info` would help with the update so we could use `update!` instead of `update_attributes` and `save(validate: false)`.

app/models/delete_queue_job.rb
bistline: The original reason was that calling `destroy` on a `StudyFile` would invoke all of the callbacks on child records, and since the `has_many` associations all have the `dependent: :destroy` clause, this could take quite a while. This was why we initially created `DeleteQueueJob` years ago.  

Over time we've gotten more proactive about cleanup up child records before deletion, and now by the time we get to this line, we've already removed all of the dependent documents, so calling `destroy` doesn't take any extra time.  This was probably something we could have changed a while ago but there wasn't any compelling reason to (until now).

bistline:
Closing this PR as the approach of proactively deleting the `StudyFile` is having many unintended consequences and causing more than a few test regressions.  I'm going to circle all the way back on a new branch to see if I can reliably get `upload_file_name` to overwrite.

PR #1088, opened 2021-06-25
Fix missing space before links in advanced search footer (SCP-3463)

eweitz:
This fixes missing spaces, along with some other updates and refinements, in user-facing text related to advanced search.

Before (see footer):
Screen Shot 2021-06-25 at 4 31 05 PM

After:
Screen Shot 2021-06-25 at 5 10 03 PM

This satisfies SCP-3463.

### DIFF
--- a/app/javascript/components/search/controls/SearchPanel.js
+++ b/app/javascript/components/search/controls/SearchPanel.js
@@ -34,17 +34,18 @@ const helpModalContent = (<div>
   Single Cell Portal supports searching on specific facets of studies by ontology classifications.
   <br/><br/>
    For example, you can search on studies that
-  have <b>species</b> of <b>&quot;Homo sapiens&quot;</b> or have an <b>organ</b> of <b>&quot;brain&quot;</b>.
-  <br/>
-    Currently, about <b>70 out of ~300</b> public studies in SCP provide this metadata information.
+  have <b>species</b> of <b>&quot;Homo sapiens&quot;</b> or have an <b>organ</b> of <b>&quot;brain&quot;</b>.{' '}
+  Almost <b>25% (84 of ~350)</b> public studies in SCP provide this metadata information.
+  {/*
+    84 of 353 studies as of 2021-06-22,
+    per https://docs.google.com/spreadsheets/d/1FSpP2XTrG9FqAqD9X-BHxkCZae9vxZA3cQLow8mn-bk
+  */}
   <br/><br/>
   For more detailed information, visit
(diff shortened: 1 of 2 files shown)


app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -34,17 +34,18 @@ const helpModalContent = (<div>
...
+  have <b>species</b> of <b>&quot;Homo sapiens&quot;</b> or have an <b>organ</b> of <b>&quot;brain&quot;</b>.{' '}
+  Almost <b>25% (84 of ~350)</b> public studies in SCP provide this metadata information.
+  {/*
+    84 of 353 studies as of 2021-06-22,
+    per https://docs.google.com/spreadsheets/d/1FSpP2XTrG9FqAqD9X-BHxkCZae9vxZA3cQLow8mn-bk
+  */}
   <br/><br/>
   For more detailed information, visit
-  our
+  our{' '}
```
devonbush, line 45: I'm curious why you use this instead of `&nbsp;`?

app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -34,17 +34,18 @@ const helpModalContent = (<div>
...
   <a href="https://github.com/broadinstitute/single_cell_portal/wiki/Search-Studies"
     target="_blank" rel="noreferrer">wiki
-  </a>.
-  <br/>If you are a study creator and would like to provide that metadata for your study to be searchable,
-  see our
+  </a>.  Study authors looking to make their studies more accessible can read our{' '}
```
devonbush, line 48: same, I'd put &nbsp; here -- I think it reads cleaner

devonbush (approved):
trivial style suggestion.  Thanks for cleaning this up -- in hindsight, I realize that I introduced these bugs when I went through and linted the frontend

app/javascript/components/search/controls/SearchPanel.js
eweitz: I considered `nbsp;`, as it's a classic solution, but saw no good reason to make this space non-breaking.  [Reading around](https://stackoverflow.com/a/41142436/10564415), `{' '}` seems to be the popular recommendation here.

app/javascript/components/search/controls/SearchPanel.js
eweitz: See above!

bistline (approved):

PR #1089, opened 2021-06-28
upgrading plotly version, and using new legendrank feature (SCP-3464)

devonbush:
legendrank sorting, added in plotly 2.1, allows us to avoid manually sorting the SVG of the legend, which led to a lot of subtle bugs. Note that there is not a large difference between plotly 1.x and plotly 2.x -- the v2 got added just because they updated/deprecated some rarely-used features (which we do not use)

MANUAL TEST:
0. load a study, and go to explore tab
1. click around a bunch of different categorized scatter plots
2. Observe that the legend is correct, and correctly sorted (first red, then blue, etc...)

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -70,12 +70,6 @@ function RawScatterPlot({
     const startTime = performance.now()
     Plotly.react(graphElementId, plotlyTraces, layout)
 
-    sortLegend({
-      graphElementId,
-      isAnnotatedScatter,
-      annotType: scatter.annotParams.type,
-      genes: scatter.genes
-    })
     perfTimes.plot = performance.now() - startTime
 
     logScatterPlot(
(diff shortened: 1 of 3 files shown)


bistline (approved):
FYI: Developers may need to run `yarn install` and `Rails.cache.clear` in order for these changes to fully kick in.  Once I did both of those, everything worked fine.

Non-blocking: it would be _awesome_ if we could sort lexically, rather than alphabetically.  Group annotations that are in fact numeric (like the 10x 1.3M mouse study) show the following order:
```
1
10
11
...
```

This has been SCP behavior since day 1, so definitely not a regression.  But it would be great if we could finally correct that oversight.

eweitz (approved):
Looks good, tested locally.

Following Jon's comment, I wonder if there's a way to incorporate [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) (for natural sorting) into `legendrank`.

devonbush:
@eweitz  We're already using localeCompare for the legend sort, with special handling so that '--Unspecified--' shows up last.  This is the traceNameSort function on ScatterPlot.js:318.   Agree with Jon's comment, that sounds good, but is probably complex enough to warrant a separate ticket

PR #1092, opened 2021-06-29
Correctly overwrite upload_file_name when queueing files for deletion (SCP-3451)

bistline:
This update fixes a regression introduced with the Rails 6.1  & `CarrierWave` [upgrade](#1030) where the `StudyFile` attribute `upload_file_name` is not always being successfully cleared when invoking `DeleteQueueJob`, which results in the user not being able to upload a file with the same filename after deleting it.  This is a common workflow when addressing file validation issues.  

The problem appears to be that the passed reference to the `StudyFile` in the `DeleteQueueJob` instance (which is executing in a background process) does not have the ability to overwrite this attribute.  This could be due to how `DelayedJob` marshals/serializes objects to and from YAML, or how `CarrierWave` manages the `upload_file_name` attribute via the `mount_uploader` command (or a combination of the two).  In any event, the solution is to instantiate a new `StudyFile` object and perform the update with that, rather than the existing reference.  

In addition, this update changes how the attributes are overwritten to no longer call `save(validate: false)`.  Since previous validation errors were due to invalid `expression_file_info` embedded documents, removing them prior to saving sidesteps the issue.  The existing test `should allow deletion of legacy expression matrices` in `DeleteQueueJobTest` validates that this functionality has not regressed due to this update.

This PR satisfies SCP-3451.

### DIFF
--- a/app/models/delete_queue_job.rb
+++ b/app/models/delete_queue_job.rb
@@ -22,6 +22,9 @@ class DeleteQueueJob < Struct.new(:object)
       file_type = object.file_type
       study = object.study
 
+      # remove expression_file_info, if present to avoid validation issues later
+      object.expression_file_info.destroy if object.expression_file_info.present?
+
       # now remove all child objects first to free them up to be re-used.
       case file_type
       when 'Cluster'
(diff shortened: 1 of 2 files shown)


app/models/delete_queue_job.rb
```diff
@@ -22,6 +22,9 @@ def perform
       file_type = object.file_type
       study = object.study
 
+      # remove expression_file_info, if present to avoid validation issues later
```
eweitz, line 25: Slightly clearer:

```suggestion
      # remove any expression_file_info to avoid validation issues later
```

eweitz (approved):

app/models/delete_queue_job.rb
devonbush: do we need these lines?  I would just as soon rewrite the update below to use `save(validate: false)` so that we don't have to add corner cases up here every time a model validation changes

devonbush (approved):
I'd rather go back to the unvalidated save than try to sanitize objects beforehand, since the unvalidated save will require less maintenance as file constraints continue to evolve.  But it's not a blocking issue

app/models/delete_queue_job.rb
bistline: You are correct in that we don't explicitly need them, but given that we were already having problems with the documents not saving correctly, and I'm _never_ going to feel comfortable with not enforcing validations on document save, I made the call to change it.  We knew the problem was the invalid `expression_file_info` document, and it seems more "correct" in my brain to address that directly.

Also, regarding future validation changes - we need to do a better job of migrating data and writing tests surrounding it.  I would rather have more tests/coverage around that then expedient solutions like disabling validations.

PR #1094, opened 2021-07-01
Support bulk download for cross dataset search (SCP-3397)

devonbush:
This adds TDR capability to bulk download.   ​A new model, DownloadRequest, has been added to capture details of a download request so that we can simplify the provided curl commands.  the UX handles sending the download request params along with the auth code request, which now also returns a download id.  This also updates a synthetic study to have the keyword 'pulmonary' in it, so that it can be easily queried alongside the test datarepo ingested study.

Note this branch also contains Emily's frontend work, so some merging may be necessary.

image

NOTE: The authCode test was removed from scp-api.test as that API call now includes a body, which means it cannot be mocked as a GET to a file -- the GET is rejected since GETs aren't allowed to have a body


app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -218,23 +255,34 @@ export function getSelectedFileStats(downloadInfo, selectedBoxes, isLoading) {
...
   const fileCount = files.length
-  const fileSize = files.reduce((sum, studyFile) => sum + studyFile.upload_file_size, 0)
+  const fileSize = files.reduce((sum, studyFile) => {
+    return sum + (studyFile.upload_file_size ? studyFile.upload_file_size : 0)
+  }, 0)
   return { fileCount, fileSize }
 }
 
-/** Gets the file ids selected, given downloadInfo and the current selection state
+/** Gets the file ids/urls selected, given downloadInfo and the current selection state
```
eweitz, line 264:
```suggestion
/** Gets the file IDs/URLs selected, given downloadInfo and the current selection state
```

app/models/download_request.rb
```diff
@@ -0,0 +1,17 @@
+class DownloadRequest
+
+  ###
+  #
+  # Stores information about files requested for a download and a corresponding auth. code
+  # Used primarily so that the curl command given to users can be concise and not require excessive escape characters
+  #
+  ###
```
eweitz, line 8: Move class docs above, per https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/Documentation:

```suggestion
# Stores information about files requested for a download and a corresponding auth. code
# Used mainly so the curl command given to users can be concise and not require excessive escape characters
class DownloadRequest
```

test/api/bulk_download_controller_test.rb
```diff
@@ -138,4 +139,36 @@ class BulkDownloadControllerTest < ActionDispatch::IntegrationTest
     excluded_file = study.study_files.by_type('Fastq').first
     refute json.include?(excluded_file.name), 'Bulk download config did not exclude external fastq link'
   end
+
+  test 'bulk download should support a download_id and tdr files' do
```
eweitz, line 143:
```suggestion
  test 'bulk download should support a download_id and TDR files' do
```

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -106,13 +138,29 @@ def summary
         render json: @study_file_info
       end
 
+      def drs_info
```
eweitz, line 141: A comment on why this DRS information is needed, and how it's used, would help here.

eweitz (changes requested):
Looking good!

I noted a few blockers on error message clarity and observability, and various non-blocking readability nits.

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -75,14 +122,31 @@ export default function DownloadSelectionModal({ studyAccessions, show, setShow
...
+          <div>
+            <DownloadSelectionTable
+              isLoading={isLoading}
+              downloadInfo={downloadInfo}
+              selectedBoxes={selectedBoxes}
+              setSelectedBoxes={setSelectedBoxes}
+              columnTypes={SCP_COLUMNS}/>
+            { showTDRSelectionPane &&
+              <div>
+                <h4>HCA studies</h4>
```
eweitz, line 135: Blocker: users often won't intuit the initialism, so let's default to the full name in the UI:

```suggestion
                <h4>Human Cell Atlas studies</h4>
```

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -106,13 +138,29 @@ def summary
...
+        drs_ids = params[:drs_ids]
+        file_info = Parallel.map(drs_ids, in_threads: 100) do |drs_id|
```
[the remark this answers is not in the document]
devonbush: I think this code already does what you want.  It accepts one or multiple ids, it will spawn up to 100 threads (but only 1 if only one id is requested).  And this method is synchronous.  Parallel.map will wait until all threads finish until it returns.  So the client code is really simple.  Jon and I talked about potentially expanding this to be a jobs-based system (right now, for the example with ~400 ids, it takes ~40 seconds for this method to return, so for larger batches we'll need a different approach)

app/javascript/components/search/controls/download/DownloadButton.js
```diff
@@ -5,13 +5,15 @@ import Tooltip from 'react-bootstrap/lib/Tooltip'
...
 
 /**
- * Component for "Download" button and Bulk Download modal.
+ * Component for "Download" button which shows a Bulk Download modal on click.
+ * DemoMode is included which will limit the number of TDR files queried, and make sure
+ * analysis files are seeded as well.  Ordinarily, we wouldn't merge demo code, but this feature may be in demo-only
+  * state for a long time, and need lots of demos
  */
-export default function DownloadButton({ searchResults={} }) {
+export default function DownloadButton({ searchResults={}, isDemoMode=false }) {
```
ehanna4, line 16: If `isDemoMode` is only used for determining which files to show I wonder if it would make sense to call it something more file specific like `shouldShowDemoFiles` ?

app/javascript/components/search/controls/download/DownloadButton.js
```diff
@@ -56,7 +70,26 @@ export default function DownloadButton({ searchResults={} }) {
...
+          setShow={setShowModal}
+          tdrFileInfo={tdrFileInfo}
+          studyAccessions={matchingAccessions}/> }
     </>
   )
 }
+
+/** limit the files returned to 10, and add two fake analysis files
+  */
+function makeDemoAppropriate(fileInfomation) {
```
ehanna4, line 84: Same thought here, wondering if something like `makeFilesDemoAppropriate` would be more precise ?

app/javascript/components/search/controls/download/DownloadButton.js
```diff
@@ -56,7 +70,26 @@ export default function DownloadButton({ searchResults={} }) {
...
+  return fileInfomation.slice(0, 10).concat(FAKE_ANALYSIS_FILES)
+}
+
+/** these are real (sequence file) DRS ids */
+const FAKE_ANALYSIS_FILES = [{
```
ehanna4, line 89: One more NIT on naming, the comment says these are real DRS ids, so calling them `FAKE_ANALYSIS_FILES` feels funny, wondering if maybe `DEMO_ONLY_ANALYSIS_FILES` or something might fit better?

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -1,36 +1,83 @@
...
   const [isLoading, setIsLoading] = useState(true)
+  const [isLoadingTDR, setIsLoadingTDR] = useState(true)
   const [downloadInfo, setDownloadInfo] = useState([])
   const [selectedBoxes, setSelectedBoxes] = useState()
+  const [downloadInfoTDR, setDownloadInfoTDR] = useState([])
+  const [selectedBoxesTDR, setSelectedBoxesTDR] = useState()
   const [stepNum, setStepNum] = useState(1)
 
+  const scpAccessions = studyAccessions.filter(accession => accession.startsWith('SCP'))
+  const tdrAccessions = studyAccessions.filter(accession => !accession.startsWith('SCP'))
```
ehanna4, line 30: Is it safe to assume if not `SCP` will be `TDR`? If the `TDR` accessions start with `TDR` could we look for that instead so if we add more types in the future it's a bit safer?

app/controllers/api/v1/bulk_download_controller.rb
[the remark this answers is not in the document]
bistline: I've logged this as [feature request](https://broadworkbench.atlassian.net/browse/DR-1947) in the Data Repo board.

app/javascript/components/search/controls/download/DownloadButton.js
bistline: I like the idea, but `shouldShowDemoAppropriateFiles` seems overly long.  Maybe `showDemoFiles`?

app/javascript/components/search/controls/download/DownloadButton.js
bistline: Agreed, the variable and the function ideally should align somehow to be more intuitive.

app/javascript/components/search/controls/download/DownloadSelectionModal.js
bistline: `TDR` accessions don't have a naming scheme, only `SCP` ones.  We're shoe-horning in the "project short name" as an accession here just to make the code easier to work with.

app/javascript/components/search/controls/download/DownloadButton.js
ehanna4: ah yes `showDemoFiles` is much better 👍

app/javascript/components/search/controls/download/DownloadSelectionModal.js
ehanna4: oh okay, thanks for the context 👍 this comment can be considered resolved then

app/lib/request_utils.rb
```diff
@@ -82,6 +82,7 @@ def self.sanitize_search_terms(terms)
   # takes a comma-delimited string of ids (e.g. StudyFile ids) and returns an array of ids
   # raises Argument error if any of the strings are not valid ids
   def self.validate_id_list(id_list_string)
+    byebug
```
ehanna4, line 85:
```suggestion
   
```

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -106,13 +138,29 @@ def summary
...
+          begin
+            client = DataRepoClient.new
+            client.get_drs_file_info(drs_id)
+          rescue StandardError => e
+            {
+              drs_id: drs_id,
+              error: e.message
+            }
+          end
```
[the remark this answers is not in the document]
bistline: It actually will have already been reported to Sentry once via [this](https://github.com/broadinstitute/single_cell_portal_core/blob/db-bulk-download-tdr/app/models/data_repo_client.rb#L76), but I can add Mixpanel tracking to this block as well.

app/lib/request_utils.rb
eweitz: Great catch.  

Flagging active `byebug` calls would be a high-value lint rule.  It's easy to automate, often missed by humans, and will crash at runtime.

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -187,19 +242,33 @@ def summary
...
         file_ids = []
-        # sanitize study accessions and file types
-        if params[:file_ids]
+        tdr_files = {}
+
+        # branch based on whether they provided a download_id, file_ids, or accessions
+        if params[:download_id]
+          download_req = DownloadRequest.find(params[:download_id])
+          if !download_req
+            render json: {error: 'Invalid download_id provided'}, status: 400 and return
```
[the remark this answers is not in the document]
bistline: @eweitz that isn't technically true - Calling `Model.find(id)` where `id` is either `nil` or a non-existent ID will both return `nil`.  This is standard for both `ActiveRecord`, and we have configured `Mongoid` to return `nil` rather than throw an error via [this line](https://github.com/broadinstitute/single_cell_portal_core/blob/development/config/mongoid.yml#L118) in our `Mongoid` configuration.  As such I think error message is actually appropriate.

app/lib/request_utils.rb
bistline: This is enabled by default in `rubocop` once we actually turn it on...

app/controllers/api/v1/bulk_download_controller.rb
bistline: I'll actually add it to the swagger docs for consistency since it is a public method.

app/controllers/api/v1/bulk_download_controller.rb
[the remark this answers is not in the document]
eweitz: Ah, right.  I think I somehow conflated `download_req` and `download_id`, and thus misunderstood.

app/controllers/api/v1/search_controller.rb
```diff
@@ -674,14 +674,20 @@ def self.get_tdr_results(selected_facets:, terms:)
         query_json = ApplicationController.data_repo_client.merge_query_json(facet_query: facet_json, term_query: term_json)
         logger.info "Executing TDR query with: #{query_json}"
         raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
+        added_file_ids = {}
         raw_tdr_results['result'].each do |result_row|
-          results = process_tdr_result_row(result_row, results, selected_facets: selected_facets, terms: terms)
+          results = process_tdr_result_row(result_row, results,
```
ehanna4, line 679: Rails/Ruby questions:
1. How come `result_row` didn't have to be passed as `row: result_row`?
2. Why can `results` be passed through with no assignment but the other args with the same name as their parameters need to be assigned? ie `results` vs `selected_facets: selected_facets`

bistline: This is the difference between positional and keyword arguments in Ruby.  Since the private method `process_tdr_result_row` is declared like this:
```
def self.process_tdr_result_row(row, results, selected_facets:, terms:, added_file_ids:)
```

The first two arguments are positional (`row` and `results`), so the order in which they are passed determines their assignment.  The rest are keyword-based, so we need to declare to the method which is which.  I think the confusing part here is that we've mirrored the names of both the local variables and the arguments.  It probably would have been better to use different names for one or the other so that the difference is clearer.

As far as `results` is concerned, it is initialized as an empty hash on the first line of `get_tdr_results`.  After that, we're passing it in to `process_tdr_result_row` and then the rows get merged into that single hash, which is then returned on the last line of the method (hence the assignment of `results = process_tdr_result_row(result_row, results, ...)`.  It all works because of this line:

```
results[short_name] ||= { ... assigning values here }
```

The `||=` operator is Ruby's conditional assignment.  It will assign the following value to the variable, but only if `results[short_name]` evaluates to `nil` or `false`.  Otherwise it will exit silently and leave whatever was already there untouched.  The first time through it initializes the structure, and after that it's a no-op until we get a different value for `short_name`.  This lets us modify it in place without any messy `if/else` logic, or calling `Hash#merge!` repeatedly.

ehanna4: Really helpful info, thank you!

eweitz (approved):

ehanna4 (approved):
LGTM - looked over code (albeit my ROR is still a work in progress) and tested locally

PR #1097, opened 2021-07-09
List non-plottable annotations in Annotation select (SCP-3436)

bistline:
If a user uploads a cluster/metadata file with group-based annotations that are not viewable (i.e. have only 1 unique value, or more than 200), these annotations will not be shown in any select menus dealing with annotations.  This can be confusing to users as it is not immediately obvious why these are omitted, and previously led to errors where no annotations in a study were considered viewable.  This update adds those annotations back in under the "Cannot Display" option group, and does not allow them to be selected when interacting with the Explore tab.  

**NOTE**: As a part of #1093, it is possible for the study owner to set one of these as the default annotation in the case where there are no available valid annotations.  If there are valid annotations, then the first available annotation will be displayed instead, even if the study owner selects an invalid one as the default.

MANUAL TESTING
To test this feature, you will need to add a new cluster file to a study that has only one invalid annotation.  You can use the following data:
```
NAME	X	Y	Z	Category
TYPE	numeric	numeric	numeric	group
CELL_0001	34.472	32.211	60.035	A
CELL_0002	15.975	10.043	21.424	A
CELL_0003	-11.688	-53.645	-58.374	A
CELL_0004	30.04	31.138	33.597	A
CELL_0005	23.862	33.092	26.904	A
CELL_0006	-39.07	-14.64	-44.643	A
CELL_0007	40.039	27.206	55.211	A
CELL_0008	28.755	27.187	34.686	A
CELL_0009	-48.601	-13.512	-51.659	A
CELL_00010	14.653	27.832	28.586	A
CELL_00011	20.603	32.071	45.484	A
CELL_00012	-10.333	-51.733	-26.631	A
CELL_00013	-52.966	-12.484	-60.369	A
CELL_00014	38.513	26.969	63.654	A
CELL_00015	12.838	13.047	17.685	A
```

1. Create a new study, and upload the non-convention metadata file from `test/test_data/metadata_example.txt` (specifying `No` for convention validation).
2. Add a cluster file, using the contents from above.
3. Once both files have ingested, navigate to the explore tab, and note that the `Category` annotation is listed under `Cannot Display (Only one label)` and cannot be selected.
Screen Shot 2021-07-09 at 1 12 20 PM

This PR satisfies SCP-3436.

### DIFF
--- a/app/javascript/components/visualization/controls/AnnotationSelector.js
+++ b/app/javascript/components/visualization/controls/AnnotationSelector.js
@@ -9,7 +9,7 @@ function getAnnotationOptions(annotationList, clusterName) {
   return [{
     label: 'Study Wide',
     options: annotationList.annotations
-      .filter(annot => annot.scope == 'study').map(annot => annotationKeyProperties(annot))
+      .filter(annot => annot.scope === 'study').map(annot => annotationKeyProperties(annot))
   }, {
     label: 'Cluster-Based',
     options: annotationList.annotations
@@ -23,7 +23,7 @@ function getAnnotationOptions(annotationList, clusterName) {
   }, {
     label: 'Cannot Display',
     options: annotationList.annotations
-      .filter(annot => annot.scope == 'invalid' && (annot.cluster_name == clusterName || !annot.cluster_name))
+      .filter(annot => annot.scope === 'invalid' && (annot.cluster_name == clusterName || !annot.cluster_name))
       .map(annot => annotationKeyProperties(annot))
   }]
 }
--- a/app/javascript/components/visualization/controls/AnnotationSelector.js
+++ b/app/javascript/components/visualization/controls/AnnotationSelector.js
@@ -23,7 +23,8 @@ function getAnnotationOptions(annotationList, clusterName) {
   }, {
     label: 'Cannot Display',
     options: annotationList.annotations
-      .filter(annot => annot.scope == 'invalid').map(annot => annotationKeyProperties(annot))
+      .filter(annot => annot.scope == 'invalid' && (annot.cluster_name == clusterName || !annot.cluster_name))
+      .map(annot => annotationKeyProperties(annot))
   }]
 }
 
--- a/app/javascript/components/visualization/controls/AnnotationSelector.js
+++ b/app/javascript/components/visualization/controls/AnnotationSelector.js
@@ -20,6 +20,10 @@ function getAnnotationOptions(annotationList, clusterName) {
     options: annotationList.annotations
       .filter(annot => annot.cluster_name === clusterName && annot.scope === 'user')
       .map(annot => annotationKeyProperties(annot))
+  }, {
+    label: 'Cannot Display',
+    options: annotationList.annotations
+      .filter(annot => annot.scope == 'invalid').map(annot => annotationKeyProperties(annot))
   }]
 }
 
@@ -59,6 +63,7 @@ export default function AnnotationControl({
       <label>Annotation</label>
       <Select options={annotationOptions}
         value={shownAnnotation}
+        isOptionDisabled={annotation => annotation.isDisabled}
         getOptionLabel={annotation => annotation.name}
         getOptionValue={annotation => annotation.scope + annotation.name + annotation.cluster_name}
         onChange={newAnnotation => updateClusterParams({ annotation: newAnnotation })}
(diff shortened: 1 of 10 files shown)


eweitz (approved):
Code and screenshot look good!

Non-blocking suggestion: a very brief note on why the annotation cannot be displayed would help.  If I were a user, it'd be the first thing I'd wonder.

Perhaps something like:

> CANNOT DISPLAY
> $annotationName (Too many labels)

and

> CANNOT DISPLAY
> $annotationName (Only one label)

On the other hand, maybe that adds too much code complexity given the rarity of this issue.  I defer to you!

bistline:
> Code and screenshot look good!
> 
> Non-blocking suggestion: a very brief note on why the annotation cannot be displayed would help. If I were a user, it'd be the first thing I'd wonder.
> 
> Perhaps something like:
> 
> > CANNOT DISPLAY
> > $annotationName (Too many labels)
> 
> and
> 
> > CANNOT DISPLAY
> > $annotationName (Only one label)
> 
> On the other hand, maybe that adds too much code complexity given the rarity of this issue. I defer to you!

Great idea - I don't think that will be hard at all.  We normally return the unique annotation values in that response, but I had omitted them for the `invalid` annotations as they weren't needed.  I could easily add those back in, and then display the message for 1 vs. too many values.

app/javascript/lib/cluster-utils.js
```diff
@@ -42,7 +42,8 @@ export function annotationKeyProperties(annotation) {
     name: annotation.name,
     type: annotation.type,
     scope: annotation.scope,
-    id: annotation.id
+    id: annotation.id,
+    isDisabled: annotation.scope == 'invalid'
```
ehanna4, line 46: Safer to use '===' rather than '==' here

app/javascript/components/visualization/controls/AnnotationSelector.js
```diff
@@ -20,6 +20,11 @@ function getAnnotationOptions(annotationList, clusterName) {
     options: annotationList.annotations
       .filter(annot => annot.cluster_name === clusterName && annot.scope === 'user')
       .map(annot => annotationKeyProperties(annot))
+  }, {
+    label: 'Cannot Display',
+    options: annotationList.annotations
+      .filter(annot => annot.scope == 'invalid' && (annot.cluster_name == clusterName || !annot.cluster_name))
```
ehanna4, line 26: Best to use '===' rather than '==' to be safest

app/models/cell_metadatum.rb
```diff
@@ -61,7 +61,7 @@ def concatenate_data_arrays(array_name, array_type)
 
   # create dropdown menu value for annotation select
   def annotation_select_value
-    "#{self.name}--#{self.annotation_type}--study"
+    "#{self.name}--#{self.annotation_type}--#{can_visualize? ? 'study' : 'invalid'}"
```
ehanna4, line 64: Ruby question - is the first `?` in `can_visualize? ? ...` used to convert `can_visualize` to a boolean?

ehanna4 (commented):
Code looks good (as far as i can tell) other than the couple of comments on the js files. Going to  manually test next

app/models/cell_metadatum.rb
bistline: No, that's just the Ruby convention for naming a method that returns a boolean.  If you want to convert a value to boolean in Ruby, it's the standard preceding double-bang that many languages use, like so:
```
irb(main):001:0> true_val = true
=> true
irb(main):002:0> false_val = false
=> false
irb(main):003:0> nil_val = nil
=> nil
irb(main):004:0> !!true_val
=> true
irb(main):005:0> !!false_val
=> false
irb(main):006:0> !!nil_val
=> false
irb(main):007:0> string_val = 'foo'
=> "foo"
irb(main):008:0> !!string_val
=> true
```

app/javascript/lib/cluster-utils.js
bistline: fb73f5e

app/javascript/components/visualization/controls/AnnotationSelector.js
bistline: fb73f5e

app/models/cell_metadatum.rb
ehanna4: oh interesting, that will take some getting used to!😅 Thanks for the clarification!

ehanna4 (approved):
Followed manual repro steps and saw the disabled annotation as described 👍

PR #1098, opened 2021-07-09
Fixing broken variable references (SCP-3485, SCP-3499)

bistline:
This fixes a pair of regressions that occurred during recent refactoring that left two undefined variables, both of which cause errors that were identified during Sentry triage.  They are merged into one PR for brevity's sake.

### DIFF
--- a/app/controllers/api/v1/user_annotations_controller.rb
+++ b/app/controllers/api/v1/user_annotations_controller.rb
@@ -101,7 +101,7 @@ module Api
           ErrorTracker.report_exception(e, current_user, @study, log_params)
           render json: {error: e.message}, status: 400 # Bad request
         rescue => e
-          ErrorTracker.report_exception(e, current_user, error_context)
+          ErrorTracker.report_exception(e, current_user, @study, log_params)
           message = 'An unexpected error prevented the annotation from being saved: ' + e.message
           render json: {error: message}, status: 500 # Server error
         end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/user_annotations_controller.rb
```diff
@@ -101,7 +101,7 @@ def create
           ErrorTracker.report_exception(e, current_user, @study, log_params)
           render json: {error: e.message}, status: 400 # Bad request
         rescue => e
-          ErrorTracker.report_exception(e, current_user, error_context)
+          ErrorTracker.report_exception(e, current_user, @study, log_params)
```
ehanna4, line 104: This change follows the same convention as the other error report so it seems like an appropriate move 👍 however I have a Rails question:
I thought that instance variables defined with the `@` sign were to be defined in a Controller to then be useable in a View, however I don't see a definition/assignment of `@study` in this Controller class. Could you explain where `@study` gets assigned? Thanks!

bistline: Great question!  This is definitely a Rails-specific thing.  Controllers can have helper methods that fire before the actual requested method is executed via the `before_action` declaration.  These are usually setters, getters, or authentication methods, although you can do anything you want, and any instance variables you define will then be available in the actual method.  In this case, the we're calling the [`set_study`](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/controllers/api/v1/concerns/study_aware.rb#L7) method with this line:

```
include Concerns::StudyAware
... 
before_action :set_study
```
In Ruby, calling `include` will make any instance methods from a module available in the class you're calling `include` from, and `set_study` is defined in `app/controllers/api/v1/concerns/study_aware.rb` as this:

```
def set_study
    head 404 and return
  end
end
```

ehanna4: ah okay great, thanks!

eweitz (approved):

ehanna4 (approved):

PR #1099, opened 2021-07-12
Adding public_filters to SearchFacet for anonymous searches (SCP-3471)

bistline:
Currently, when viewing the SCP home page before logging in, the filter values displayed when clicking on a search facet can contain values that only reside in private studies that the user may not be able to view, once logged in.  This can lead to confusing situations where a user selects one of those filter values, runs a search, and gets no results (due to correct permission filtering on studies).

This update adds the `public_filters` attribute to `SearchFacet`, where only filter values from public studies are shown.  This way, when an anonymous user selects a filter value, they are guaranteed to see at least one result.  Once logged in, the former scenario still applies, but solving that issue is out of scope for this PR.  In addition, this update applies general linting format updates to files updated in this PR.

MANUAL TESTING
Since waiting for an ingest of a convention-compliant metadata file in a private study will take several minutes, the easiest way to test this feature is to manually update the `public_filters` value on a search facet.  Also, you will need to set the default value for the "faceted_search" `FeatureFlag` to `true` in order to have advanced search show up when not logged in.

1. From the project root, after running `rails_local_setup.rb`, run all queued migrations:
```
bin/rails db:migrate
```
2. Enter the rails console, and find the 'faceted_search' `FeatureFlag`:
```
feature_flag = FeatureFlag.find_by(name: 'faceted_search')
```
3. Set the `default_value` to true:
```
feature_flag.update(default_value: true)
```
4. Load the `SearchFacet` for 'species'
```
facet = SearchFacet.find_by(name: 'species')
```
5. Update the `public_filters` to be only the first two filters:
```
facet.update(public_filters: facet.filters.take(2))
```
6. In another terminal window, boot the portal as normal, and click on the "species" facet, confirming there are only 2 values:
Screen Shot 2021-07-12 at 3 14 49 PM
7. Sign in, and confirm that all values for species are now shown:
Screen Shot 2021-07-12 at 3 15 42 PM
8. (Optional): Back in the rails console, reset the 'species' facet:
```
facet.update_filter_values!
```

This PR satisfies SCP-3471

### DIFF
--- a/app/models/search_facet.rb
+++ b/app/models/search_facet.rb
@@ -11,6 +11,7 @@ class SearchFacet
   field :name, type: String
   field :identifier, type: String
   field :filters, type: Array, default: []
+  field :public_filters, type: Array, default: [] # filters for public studies only
   field :is_ontology_based, type: Boolean, default: false
   field :ontology_urls, type: Array, default: []
   field :data_type, type: String
@@ -39,7 +40,8 @@ class SearchFacet
   }.with_indifferent_access.freeze
   TIME_UNITS = TIME_MULTIPLIERS.keys.freeze
 
-  validates_presence_of :name, :identifier, :data_type, :big_query_id_column, :big_query_name_column, :convention_name, :convention_version
+  validates_presence_of :name, :identifier, :data_type, :big_query_id_column, :big_query_name_column, :convention_name,
+                        :convention_version
   validates_uniqueness_of :big_query_id_column, scope: [:convention_name, :convention_version]
   validate :ensure_ontology_url_format, if: proc {|attributes| attributes[:is_ontology_based]}
   before_validation :set_data_type_and_array, on: :create,
@@ -47,7 +49,7 @@ class SearchFacet
   after_create :update_filter_values!
 
   swagger_schema :SearchFacet do
-    key :required, [:name, :identifier, :data_type, :big_query_id_column, :big_query_name_column, :convention_name, :convention_version]
+    key :required, %i[name identifier data_type big_query_id_column big_query_name_column convention_name convention_version]
     key :name, 'SearchFacet'
     property :name do
       key :type, :string
@@ -67,7 +69,7 @@ class SearchFacet
       key :description, 'Array of filter values for facet'
       items type: :object do
         key :title, 'FacetFilter'
-        key :required, [:name, :id]
+        key :required, %i[name id]
         property :name do
           key :type, :string
           key :description, 'Display name of filter'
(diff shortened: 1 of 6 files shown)


app/models/search_facet.rb
```diff
@@ -164,10 +166,10 @@ class SearchFacet
     end
     property :filters do
       key :type, :array
-      key :description, 'Array of filter values for facet'
+      key :description, 'Array of filter values for facet (will default to public_filters if no user is signed in)'
```
eweitz, line 169: Blocker: there seems to be a subtle documentation bug that substantially changes meaning.

```suggestion
      key :description, 'Array of filter values for facet (will default to public_filters if user is not signed in)'
```

app/models/search_facet.rb
```diff
@@ -309,80 +311,96 @@ def convert_time_between_units(base_value:, original_unit:, new_unit:)
...
-    values = self.get_unique_filter_values
-    unless values.empty?
-      if self.is_numeric?
-        self.update(min: values[:MIN], max: values[:MAX])
-      else
-        self.update(filters: values.to_a)
-      end
+    if is_numeric?
+      values = get_unique_filter_values
+      return false if values.empty? # found no results, meaning an error occurred
```
eweitz, line 349: If an error occurred, should it be thrown and logged here?  I'd think so, but might be overlooking something.

eweitz (changes requested):
This will soften a search UX papercut!

I noted a trivial blocker, and suggested a non-blocking maintainability improvement.

Although this fix for unauth'd users doesn't fix how auth'd users see some filters for which no results will be returned, does this work lower effort for fixing that?  Or would fixing that related issue require an orthogonal solution?

app/models/search_facet.rb
bistline: That is handled in `get_unique_filter_values`, which is where the actual call to BigQuery happens.  This is just the setter.

eweitz (approved):

bistline:
> Although this fix for unauth'd users doesn't fix how auth'd users see some filters for which no results will be returned, does this work lower effort for fixing that? Or would fixing that related issue require an orthogonal solution?

Not really... the only way I can see of truly solving this problem completely is maintaining a map of all facet/filter values to study accessions, and then filtering the UI based on which studies the user is allowed to see.  That seems a rather complicated and expensive solution to an admittedly minor UX grumble.

ehanna4 (approved):
Looks good, tested manually following the repro steps provided and compared to dev and this definitely is a nicer UX. Code looks good as far as I can tell. The use of `%w[]` for string arrays is a neat ruby-feature/change

PR #1101, opened 2021-07-14
Avoid long-lived query cursors when caching all study defaults (SCP-3489)

bistline:
Previously, when running `ClusterCacheService.cache_all_defaults`, the entire job was performed inside a single Mongo query cursor (`Study.all`), which works fine in development/staging.  However, when deployed to production, we began seeing instances of `Mongo::Error::OperationFailure` where the query cursor timed out, causing the job to error and stop running.  This has the end effect of not caching any default visualization responses for higher-accession studies, since they are all processed in order of creation.

This update replaces the `Study.all` query with `Study.pluck(:accession)`, and then each call to `cache_study_defaults(study)` is wrapped in a `begin/rescue` block.  This has two main benefits:

1.  `Study.pluck(:accession)` returns an array, rather than a query cursor, so the operation completes w/o leaving a database connection open
2. Any instances of `Mongo::Error::OperationFailure` will be explicitly handled, allowing the job to continue operating even if the error happens again

This does have a minor drawback in that it can potentially take longer, as `Study.all.map` is a single query, as opposed to 1000+ calls to `Study.find_by(accession: accession)`, but since this is a background process, there is no real harm in this.

MANUAL TESTING
Since this error has only ever occurred in production, and relied on the fact that there are approx. 1000 studies, there is no feasible way to replicate it in development/staging.  Still, anyone can verify that the job runs w/o error by issuing the following commands:

1. In a terminal window, run `rails_local_setup.rb` and then start a DelayedJob work with:
```
bin/rails jobs:work
```
2. In a separate terminal window, run `rails_local_setup.rb` and then issue the following command:
```
rails runner "ClusterCacheService.delay(queue: :cache).cache_all_defaults"
```
3. Back in the first terminal window, note that the job starts with a message like:
```
[Worker(host:wm306-020 pid:86370)] Job ClusterCacheService.cache_all_defaults (id=60ef083694ec8f5248ad984d) (queue=cache) RUNNING
```
3. In `development.log` note entries to the effect of `Checking defaults on SCP13 for pre-caching` and subsequent messages detailing what cache paths are written (or skipped if there are no defaults to cache)
4. In the DelayedJob worker window, note the job eventually completes with a message like:
```
[Worker(host:wm306-020 pid:86370)] Job ClusterCacheService.cache_all_defaults (id=60ef083694ec8f5248ad984d) (queue=cache) COMPLETED after 114.9203
[Worker(host:wm306-020 pid:86370)] 1 jobs processed at 0.0087 j/s, 0 failed
```

This PR satisfies SCP-3489.

### DIFF
--- a/app/lib/cluster_cache_service.rb
+++ b/app/lib/cluster_cache_service.rb
@@ -15,8 +15,20 @@ class ClusterCacheService
   end
 
   # pre-cache all default clusters/annotations for every study
+  # use accession list to prevent long-lived query cursors from timing out in Delayed::Job
   def self.cache_all_defaults
-    Study.all.map {|study| cache_study_defaults(study) }
+    accessions = Study.pluck(:accession)
+    accessions.each do |accession|
+      begin
+        study = Study.find_by(accession: accession)
+        cache_study_defaults(study)
+      rescue Mongo::Error::OperationFailure => e
+        ErrorTracker.report_exception(e, nil,
+                                      { study_accession: accession, operation: :cache_study_defaults })
+        Rails.logger.error "Error caching study defaults for #{accession}: #{e.message}"
+        next
+      end
+    end
   end
 
   # pre-cache the default cluster & annotation for a given study


eweitz (approved):

ehanna4 (approved):
Makes sense to me 👍

PR #1102, opened 2021-07-19
Sort the annotations AlphaNumerically [SCP-3466]

ehanna4:
Sort the annotations for scatterplot Alphanumerically.

Remove "legendrank" as that was not actually setting the order in the legend, it is coming from the annotation groups order.
Sorting the groups even with the million cell study added only about 3 ms so it seemed reasonable to continue with this approach.

Before:
Screen Shot 2021-07-19 at 4 10 11 PM
After:
Screen Shot 2021-07-19 at 4 07 03 PM

Another example of how the annotation groups will be sorted now:
Screen Shot 2021-07-19 at 3 11 55 PM

Sorts just numbers logically too:
Screen Shot 2021-07-19 at 4 54 35 PM

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -67,6 +67,9 @@ function RawScatterPlot({
       is3D: scatter.is3D
     })
 
+    // sort the annotations array alphanumerically - this will determine order of legend in graph
+    plotlyTraces[0].transforms[0].groups.sort(sortAlphaNumerically)
+
     const startTime = performance.now()
     Plotly.react(graphElementId, plotlyTraces, layout)
 
@@ -237,7 +240,6 @@ function getPlotlyTraces({
           target: val,
           value: {
             name: `${val} (${traceCounts[val]} points)`,
-            legendrank: index,
             marker: {
               color: getColorBrewerColor(index),
               size: pointSize
@@ -296,6 +298,11 @@ function traceNameSort(a, b) {
   if (a === UNSPECIFIED_ANNOTATION_NAME) {return 1}
   if (b === UNSPECIFIED_ANNOTATION_NAME) {return -1}
   return a.localeCompare(b)
+}
+
+/** sort alphanumerically */
+function sortAlphaNumerically(a, b) {
+  return a.localeCompare(b, 'en', { numeric: true })
 }
 
 /** makes the data trace attributes (cells, trace name) available via hover text */


jlchang (commented):
One major motivation was that Studies with numbers-as-labels (the default for Seurat clustering output) would sort numerically despite looking like a number.

Could you include in the PR description what the label sort order would look like for a study that had Seurat Cluster labels 1 through 12?

If you need a "real data" example, you could use the data in SCP1238 - just keep the study private as the content is unpublished data shared by our U19 collaborators.

ehanna4:
> One major motivation was that Studies with numbers-as-labels (the default for Seurat clustering output) would sort numerically despite looking like a number.
> 
> Could you include in the PR description what the label sort order would look like for a study that had Seurat Cluster labels 1 through 12?
> 
> If you need a "real data" example, you could use the data in SCP1238 - just keep the study private as the content is unpublished data shared by our U19 collaborators.

Will do! Thanks for providing an example to use

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -298,6 +300,11 @@ function traceNameSort(a, b) {
   return a.localeCompare(b)
 }
 
+/** sort alphanumerically */
+function sortAlphaNumerically(a, b) {
```
eweitz, line 304: "[Natural sort](https://en.wikipedia.org/wiki/Natural_sort_order)" is the conventional term for this kind of sort, so preferring that name would help maintainability.

```suggestion
/** Give natural sort order for alphanumeric strings */
function sortNaturally(a, b) {
```

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -67,6 +67,9 @@ function RawScatterPlot({
       is3D: scatter.is3D
     })
 
+    // sort the annotations array alphanumerically - this will determine order of legend in graph
+    plotlyTraces[0].transforms[0].groups.sort(sortAlphaNumerically)
```
eweitz, line 71:
```suggestion
    plotlyTraces[0].transforms[0].groups.sort(sortNaturally)
```

eweitz (approved):
Looks good!  Thanks for the timing data.  I suggested a small maintainability improvement.

bistline (changes requested):
I'm not sure what is causing this, but it appears that we are losing the assignments and the plots aren't showing the correct populations.  I've tried with both strings and numbers as "group" type labels.  Oddly, the colors are correct in terms of the labels, but the points are not.  I tried blowing away the cache, but the problem is persisting.

I've included screenshots from SCP1 (sNuc-seq) and SCP383 (10X 1.3M cell) both locally and on production, for comparison:

localhost, strings
localhost_strings

production, strings
production_strings

localhost, numbers
localhost_numbers

production, numbers
production_numbers

ehanna4:
Closing this PR - hoping to utilize the library Jon pointed out as a better way to do legends and with the hope that it supports longer term goals of better customizing plotly things.

PR #1103, opened 2021-07-20
Enforcing metadata convention rules when syncing files (SCP-3514)

bistline:
Currently, there is a gap in functionality concerning the SCP metadata convention when a user is syncing study files, rather than uploading them.  There is only a yes/no select menu, and all of the acceptance logic displayed in the metadata upload form is omitted.  This can lead to a corner case where a user can select "No" for the metadata convention while syncing, which will then throw an error in production (as it is required), leading to the file being "synced" but not ingested.

This update applies the same UI/UX for the metadata convention to sync forms.  In addition, a server-side validation has been put in place to honor the value of the `convention_required` feature flag.  If set to `true`, it will reject saving any metadata files that do not conform to the metadata convention unless that user (or any other user associated with the current study) has obtained an exemption from SCP admins.  This is handled through a new `associated_users` method on studies that will return an array of all users that have access to the study in question.  

This will be useful as well for upcoming work regarding requiring raw count expression matrices, as that will also attempt to enforce acceptance via a validation which will also be governed by a feature flag.

MANUAL TESTING
To test, you will first need to create a new Terra workspace in your developer-specific project, and upload the non-convention metadata file from `test/test_data/metadata_example.txt` to the workspace bucket.  In addition, you will want to set the default value for `convention_required` to `true`:

```
./rails_local_setup.rb
source config/secrets/.source_env.bash && bin/rails c
irb(main):001:0> flag = FeatureFlag.find_by(name: 'convention_required')
=> #<FeatureFlag _id: 6033f69ee241391a2371c242, name: "convention_required", default_value: false, description: "whether metadata convention compliant files are required">
irb(main):002:0> flag.update(default_value: true)
=> true
```

1. Boot the portal as normal, sign in, and click "Create a study"
2. Set `use_existing_workspace` to `true` and put in the name of the above workspace
3. Click "Create" to go to the sync page
4. For the one unsynced file, select the file type of 'Metadata' and confirm that "No" is greyed out for the `use_metadata_convention` field
Screen Shot 2021-07-20 at 12 30 17 PM
5. Go to the Admin page, load your user account, and set the `convention_required` feature flag to `false`
6. Reload the sync page, and confirm that you can now specify "No" for `use_metadata_convention`
7. Select "No", save the file, and note in `development.log` that an ingest job has launched:
```
Remote found for metadata_example.txt:60f7037794ec8f7f4c8d9b5e, launching Ingest job
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 60ef404d94ec8f6c291f59d8
  - "--study-file-id"
  - 60f7037794ec8f7f4c8d9b5e
  - "--user-metrics-uuid"
  - ab2abaf4-4322-48c2-b886-9bb111bef757
  - ingest_cell_metadata
  - "--cell-metadata-file"
  - gs://fc-19feb94f-9d21-4f93-8aeb-ef0c5e014fbe/metadata_example.txt
  ...

```
8. (Optional) Back in the Rails console, set the `default_value` to `false` for the `convention_required` feature flag.

This PR satisfies SCP-3514.

### DIFF
--- a/app/views/studies/_metadata_file_fields.html.erb
+++ b/app/views/studies/_metadata_file_fields.html.erb
@@ -1,6 +1,46 @@
 <div class="form-group row">
   <div class="col-sm-4">
-    <%= f.label :use_metadata_convention, "Use Metadata Convention?" %> <%= render partial: 'metadata_convention_help_popover', locals: {id: f.object.id.to_s} %><br />
-    <%= f.select :use_metadata_convention, options_for_select([['Yes', true],['No', false]], f.object.use_metadata_convention), {}, class: 'form-control use-metadata-convention', disabled: !f.object.new_record? %>
+    <% convention_required = User.feature_flag_for_instance(current_user, 'convention_required') %>
+    <%= f.label :use_metadata_convention do %>
+      Do you use SCP conventional names for required metadata column headers?
+      <%= render partial: 'metadata_convention_help_popover', locals: {id: f.object.id.to_s} %>
+    <% end %>
+    <br />
+    <%= f.label :use_metadata_convention_true, 'data-analytics-name': 'metadata-convention-optin' do %>
+      <%= f.radio_button :use_metadata_convention, true, checked: true, disabled: !f.object.new_record?%>
+      Yes
+    <% end %>
+    &nbsp;
+    <%= f.label :use_metadata_convention_false, 'data-analytics-name': 'metadata-convention-optout' do %>
+      <% if convention_required %>
+        <a href="#/" style="color:#999" id="convention-decline-label-<%= f.object.id.to_s %>"
+           data-content="Using the convention is now required.  If this presents a particular issue for your study, please contact us at <a href='mailto:scp-support@broadinstitute.zendesk.com'>scp-support@broadinstitute.zendesk.com</a> so we can help or make any updates to the convention."
+           data-toggle="popover"
+           data-analytics-name="convention-decline-label">
+          <%= f.radio_button :use_metadata_convention, false, checked: false, disabled: true %>
+          No
+        </a>
+
+      <% else %>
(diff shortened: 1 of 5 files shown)


app/views/studies/_metadata_file_fields.html.erb
```diff
@@ -1,6 +1,46 @@
...
+    &nbsp;
+    <%= f.label :use_metadata_convention_false, 'data-analytics-name': 'metadata-convention-optout' do %>
+      <% if convention_required %>
+        <a href="#/" style="color:#999" id="convention-decline-label-<%= f.object.id.to_s %>"
+           data-content="Using the convention is now required.  If this presents a particular issue for your study, please contact us at <a href='mailto:scp-support@broadinstitute.zendesk.com'>scp-support@broadinstitute.zendesk.com</a> so we can help or make any updates to the convention."
+           data-toggle="popover"
+           data-analytics-name="convention-decline-label">
+          <%= f.radio_button :use_metadata_convention, false, checked: false, disabled: true %>
+          No
+        </a>
```
eweitz, line 22: Out-of-scope, but just to surface recent team chats: 

It's not ideal to show a disabled selection UI that can only be enabled after emailing.  It works, but it's awkward.  It'd likely be better to not show radio buttons in this case, and instead just show an button like "Using the convention is a problem for me" that shows these instructions on click.

Radio buttons make sense if the user is exempted, but not until then.

eweitz (approved):

app/views/studies/_metadata_file_fields.html.erb
bistline: No argument, but this is duplicating the UX from the upload wizard - the main point (other than fixing the bug) was to homogenize the experience so that any time a user is attempting to ingest a metadata file they are presented with the same controls.

ehanna4 (approved):
LGTM, manually tested with steps provided and saw the appropriate results

PR #1104, opened 2021-07-21
Increase scientific precision by increasing default cluster resolution (SCP-3513)

eweitz:
This increases the scientific precision of our leading visualization by increase its default resolution by 10x.  Specifically, this increases the default cluster subsampling threshold from 10K to 100K cells.  If a cluster has < 100K cells, all its cells are shown.

Performance optimizations done last quarter made it trivial to implement this scientific enhancement.

A disadvantage of subsampling is that the number of cells for each subsampled label is equal, while the underlying data can have labels with very unequal numbers of cells.  So subsampling can greatly skew proportions in underlying data, which is a scientific problem.  

Compare the default 10K-subsampled cluster below to its full resolution cluster, as Jean for our U19 stakeholders.  The relatively small subsampling obscures:

* The thick band of grey cells (i.e. annotation-labeled "3").
* The preponderence of red cells (label “0") at bottom and light pink cells (label 2) at top.

Old, low resolution default:
10K_subsample_old_default_of_spatial_cluster__SCP_2021-07-21

New, high resolution default:
All_cells_for_100K_subsample_new_default_of_spatial_cluster__SCP_2021-07-21

To test:
* Load a study that has a cluster with > 100K cells
* Verify "Subsampling" says "100000", not "10000"
* Switch to "10000"
* Verify it has lower resolution

This satisfies SCP-3513.

### DIFF
--- a/app/javascript/lib/cluster-utils.js
+++ b/app/javascript/lib/cluster-utils.js
@@ -23,12 +23,12 @@ export const UNSPECIFIED_ANNOTATION_NAME = '--Unspecified--'
 export function getDefaultSubsampleForCluster(annotationList, clusterName) {
   const subsampleOptions = annotationList.subsample_thresholds[clusterName]
   if (subsampleOptions?.length) {
-    // find the max subsample less than or equal to 10000
-    const defaultSubsample = Math.max(...subsampleOptions.filter(opt => opt <= 10000))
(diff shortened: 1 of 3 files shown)


bistline (approved):

ehanna4 (approved):
Looks good to me, looked over code change and manually tested using the steps suggested and saw the appropriate changes.

PR #1106, opened 2021-07-22
Adding TDR snapshot IDs configuration for scoping search requests (SCP-3490)

bistline:
Since SCP will be enabling cross dataset search via requests to Terra Data Repo (TDR), and will be doing so by connecting to their development instance, there is the possibility that search requests from production SCP could potentially expose data on the development TDR instance that is not intended for public release.  This is because search requests to TDR will query all available snapshots shared with the SCP service account unless we provide a list of snapshots that we want to query.

This update adds an admin configuration option to append a comma-delimited list of snapshot UUIDs to all TDR queries.  This will limit results to only those snapshots, allowing admins to dynamically control what data is returned to users without having to perform a release.  In addition, this update also adds error handling to TDR requests to avoid breaking search requests entirely in case snapshots are redacted.  It also increases the default number of rows returned by TDR from `1000` to `10000` to avoid having to paginate row-level results.

MANUAL TESTING
Since there is currently only 1 snapshot shared with SCP service accounts, the only way to test this is to put in dummy UUIDs for the configuration value, and check the logs to confirm that the query has been limited to those snapshots.

1. Boot portal as normal, sign in, and go to Admin page
2. Click "New Config Option"
3. Select `TDR Snapshot IDs` as the configuration type, and `string` as the value type
4. Paste the following UUIDs into the `value` box: `b5cabf5b-1202-4017-93c6-9537a773f1a3, 7bc2a4d8-95ac-4f5b-9906-92d1f7175eaa` and click "Save"
5. Go back to the home page, and select "Homo sapiens" from the "species" facet
6. Note that normal SCP results are returned
7. In `development.log`, note the following entry:
```
Scoping TDR query to snapshots: b5cabf5b-1202-4017-93c6-9537a773f1a3, 7bc2a4d8-95ac-4f5b-9906-92d1f7175eaa
```
8. Note the `401 Unauthorized` error (since those snapshots do not exist; this proves the error handling works)
9. Back in the admin section, delete the config option for `TDR Snapshot IDs`

This PR satisfies SCP-3490.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -674,8 +674,9 @@ module Api
           end
           # now we merge the two queries together to perform a single search request
           query_json = ApplicationController.data_repo_client.merge_query_json(facet_query: facet_json, term_query: term_json)
+          logger.info "Executing TDR query with: #{query_json}"
           snapshot_ids = AdminConfiguration.get_tdr_snapshot_ids
-          logger.info "Executing TDR query with: #{query_json} ()"
+          logger.info "Scoping TDR query to snapshots: #{snapshot_ids.join(', ')}" if snapshot_ids.present?
           raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json,
                                                                                           snapshot_ids: snapshot_ids)
           added_file_ids = {}
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -665,22 +665,29 @@ module Api
       # execute a search in TDR and get back normalized results
       def self.get_tdr_results(selected_facets:, terms:)
         results = {}
-        if selected_facets.present?
-          facet_json = ApplicationController.data_repo_client.generate_query_from_facets(selected_facets)
-        end
-        if terms.present?
-          term_json = ApplicationController.data_repo_client.generate_query_from_keywords(terms)
-        end
-        # now we merge the two queries together to perform a single search request
-        query_json = ApplicationController.data_repo_client.merge_query_json(facet_query: facet_json, term_query: term_json)
-        logger.info "Executing TDR query with: #{query_json}"
-        raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
-        added_file_ids = {}
-        raw_tdr_results['result'].each do |result_row|
-          results = process_tdr_result_row(result_row, results,
-            selected_facets: selected_facets,
-            terms: terms,
-            added_file_ids: added_file_ids)
+        begin
+          if selected_facets.present?
+            facet_json = ApplicationController.data_repo_client.generate_query_from_facets(selected_facets)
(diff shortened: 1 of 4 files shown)


app/models/admin_configuration.rb
```diff
@@ -53,7 +53,7 @@ def self.get_ingest_docker_image_config
   # Ingest docker image getter
   def self.get_ingest_docker_image
     ingest_config = self.get_ingest_docker_image_config
-    if Rails.env != 'production' && ingest_config.present?
+    if !Rails.env.production? && ingest_config.present?
```
eweitz, line 56: I find the previous code more readable.

If you agree, could we disable this Rubocop rule?  If not, no worries, I'll adapt!

app/controllers/api/v1/search_controller.rb
```diff
@@ -665,22 +665,30 @@ def self.match_results_by_filter(search_result:, result_key:, facets:)
...
-        raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json)
-        added_file_ids = {}
-        raw_tdr_results['result'].each do |result_row|
-          results = process_tdr_result_row(result_row, results,
-            selected_facets: selected_facets,
-            terms: terms,
-            added_file_ids: added_file_ids)
+        begin
+          if selected_facets.present?
+            facet_json = ApplicationController.data_repo_client.generate_query_from_facets(selected_facets)
```
eweitz, line 670: For succinct readability, could we do something like:

`tdr_client = ApplicationController.data_repo_client` 

atop this controller, then replace references throughout this file accordingly?  E.g.:

```suggestion
            facet_json = tdr_client.generate_query_from_facets(selected_facets)
```

eweitz (approved):
Looks good!  I suggested a few readability refinements.

app/models/admin_configuration.rb
bistline: This is a Rubocop suggestion, and we have mixed usage throughout the portal.  I'm updating them as we find them.  Personally I think it's easier to read, but I'm a Rubyist, so I get your objection.  I also prefer `unless` to a negative `if` as it reads more like language.

app/controllers/api/v1/search_controller.rb
bistline: This would be a departure from how we reference this API client vs. all the other ones (`FireCloudClient`, `BigQueryClient`, and `PapiClient`).  Furthermore, if we assigned it to a local variable like this, it would mean we're reinstantiating the client for every request, which will reissue OAuth tokens every time.  Eventually we'd run afoul of Google's token quotas.  This is the main reason why we consolidated into class variables in `ApplicationController`.

eweitz: Thanks for the reminder!

ehanna4: thanks for the question/context !

ehanna4 (approved):
Tested manually and talked through code w/ Jon and LGTM 👍

PR #1107, opened 2021-07-23
Adding HCA Azul service API client (SCP-3503)

bistline:
This update adds a minimal API client for interacting with the Human Cell Atlas [Azul service](https://service.azul.data.humancellatlas.org).  Currently, the only API endpoint integration is for generating file manifests for available files in Azul.  These manifests can also contain experimental metadata (if requesting the `full` manifest), though mostly they contain information about files from various catalogs/projects.  This will allow SCP to export metadata pertaining to HCA catalogs that exist in TDR as a part of bulk download requests.

MANUAL TESTING
As this is not wired into advanced search or bulk download yet, the only way to test is manually through the Rails console:

1. Pull this branch, run local setup, and enter the Rails console:
```
./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails c
```
2. Set up variables for making Azul requests:
```
pulmonary_project = 'c1a9a93d-d9de-4e65-9619-a9cec1052eaa'
tcell_act_project = '4a95101c-9ffc-4f30-a809-f04518a23803'
catalog = 'dcp7'
azul_client = ApplicationController.hca_azul_client
```
3. Get a link to the standard manifest for the Pulmonary Fibrosis HCA project:
```
pulmonary_manifest = azul_client.get_project_manifest_link(catalog, pulmonary_project)
```
4. Open the `Location` URL in a browser tab to download the manifest
5. Get a link to the `full` manifest for the Human Tissue Tcell Activation project:
```
tcell_manifest = azul_client.get_project_manifest_link(catalog, tcell_act_project, 'full')
```
6. Open the `Location` URL in a browser tab to download the full manifest
7. (Optional) Generate a curl manifest to download all files from the Tcell project:
```
tcell_curl = azul_client.get_project_manifest_link(catalog, tcell_act_project, 'curl')
```
8. (Optional) Copy the `CommandLine` > `bash` curl config into a terminal window to execute file downloads (you may want to quit after a while as there is quite a lot of data).

This PR satisfies SCP-3503.

### DIFF
--- a/app/models/hca_azul_client.rb
+++ b/app/models/hca_azul_client.rb
@@ -1,0 +1,122 @@
+# Query Human Cell Atlas Azul service for metadata associated with both experimental and analysis data
+# No ServiceAccountManager or GoogleServiceClient includes as all requests are unauthenticated for public data
+class HcaAzulClient < Struct.new(:api_root)
+  include ApiHelpers
+
+  GOOGLE_SCOPES = %w[openid email profile].freeze
+  BASE_URL = 'https://service.azul.data.humancellatlas.org'.freeze
+
+  # list of available HCA catalogs
+  HCA_CATALOGS = %w[dcp1 dcp6 dcp7 it1 it6 it7 it0lungmap lungmap].freeze
+
+  # List of accepted formats for manifest files
+  MANIFEST_FORMATS = %w[compact full terra.bdbag terra.pfb curl].freeze
+
+  ##
+  # Constructors & token management methods
+  ##
+
+  # base constructor
+  #
+  # * *return*
+  #   - +HcaAzulClient+ object
+  def initialize
+    self.api_root = BASE_URL
+  end
+
+  ##
+  # Abstract request handlers/heplers
+  ##
+
+  # submit a request to HCA Azul Service API
+  #
+  # * *params*
+  #   - +http_method+ (String, Symbol) => HTTP method, e.g. :get, :post
+  #   - +path+ (String) => Relative URL path for API request being made
+  #   - +payload+ (Hash) => Hash representation of request body
+  #   - +retry_count+ (Integer) => Counter for tracking request retries
+  #
+  # * *returns*
+  #   - (Hash) => Parsed response body, if present
+  #
+  # * *raises*
+  #   - (RestClient::Exception) => if HTTP request fails for any reason
+  def process_api_request(http_method, path, payload: nil, retry_count: 0)
+    # Log API call for auditing/tracking purposes
+    Rails.logger.info "HCA Azul API request (#{http_method.to_s.upcase}) #{path}"
+    # process request
+    begin
+      headers = {
+        'Accept' => 'application/json',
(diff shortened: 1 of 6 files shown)


app/models/hca_azul_client.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+
+    path = self.api_root + "/fetch/manifest/files?catalog=#{catalog}"
+    project_filter = { 'projectId' => { 'is' => [project_id] } }
+    filter_query = format_hash_as_query_string(project_filter)
+    path += "&filters=#{filter_query}&format=#{format}"
+    # since manifest files are generated on-demand, keep making requests until the Status code is 302 (Found)
+    # Status 301 means that the manifest is still being generated; if no manifest is ready after 30s, return anyway
+    time_slept = 0
+    manifest_info = process_api_request(:get, path)
```
eweitz, line 126: I wonder why Azul generates the manifest files on-demand.  Presumably, their upstream components don't frequently change.

eweitz (approved):

app/models/hca_azul_client.rb
bistline: This is because the endpoint we are using (`/fetch/manifest/files`) supports a query interface.  For our usage, we're only asking for project-level information by passing the `{ 'projectId' => { 'is' => [project_id] } }` query, but you can provide more detailed queries to further refine the row-level data returned.  Logically, there's no way that Azul can precompute everything, so manifests are generated on-demand.

You can read more about the particular endpoint here: https://service.azul.data.humancellatlas.org/#/Manifests/get_fetch_manifest_files

eweitz: Makes sense, thanks for the pointer!

ehanna4 (commented):
I've tried following the manual test steps but I get a '500 Internal Server' error when attempting: `pulmonary_manifest = azul_client.get_project_manifest_link(catalog, pulmonary_project)` 

Going through their swagger page I can hit the `/fetch/manifest/files` end point and run the curl command locally to get the files, so I know I have access but it's working when trying to follow the steps provided. Did you run into this at all? Is it likely something I've got configured wrong on my local?

bistline:
> I've tried following the manual test steps but I get a '500 Internal Server' error when attempting: `pulmonary_manifest = azul_client.get_project_manifest_link(catalog, pulmonary_project)`
> 
> Going through their swagger page I can hit the `/fetch/manifest/files` end point and run the curl command locally to get the files, so I know I have access but it's working when trying to follow the steps provided. Did you run into this at all? Is it likely something I've got configured wrong on my local?

I'm getting that now too, so something is up.  I must have broke something in the second commit.

bistline:
@ehanna4 apparently dcp6 is no longer available via their API anymore.  I'm updating everything to reflect that.

ehanna4 (approved):
LGTM - was able to test manually after the latest fixes and worked as expected. Walked through code questions w/ Jon and makes sense to me 👍

PR #1108, opened 2021-07-27
Ignore email case when checking study shares (SCP-3510)

bistline:
When sharing studies with other Google accounts, users are asked to enter and email address and select a permission level.  When that other user receives an email notifying them of the share, they can then authenticate into SCP and view that study.  SCP will compare the email address of the current logged in account against the list of "shared" email accounts, and if there is a match, that user can view the study in question.  However, there is a corner case where the string case of the email is entered differently than what Google declares the actual account email to be (Google does not honor case or punctuation in email addresses).  For instance, sharing a study with `JohnDoe@gmail.com` will deliver an email to `johndoe@gmail.com`, but when `johndoe@gmail.com` signs in, they will not be able to view the study in question because `'JohnDoe@gmail' == 'johndoe@gmail.com'` will evaluate to `false`.

This update addresses the problem twofold: first, when comparing email addresses, all emails are downcased prior to comparison.  Secondly, whenever querying for existing shares, a case-insensitive search is run using the `/#{email}/i` regular expression.  While this is slower, the size of the `StudyShare` collection is small enough that no performance impact will be noticed.

MANUAL TESTING
1. Log into with your Broad institute email, and click 'Create a Study'
2. Give the study a name, and set `Public` to `No`
3. At the bottom, click 'Share Study', and enter a different email address (such as a personal Gmail account), but enter the email address in all capitals (e.g. `JOHNDOE@GMAIL.COM`)
4. Click 'Create Study'
5. From the upload wizard, click the "View Study" button at the top right to go to the study overview page
6. Copy the URL for this study, then sign out of your Broad account and log in with the shared user account
7. Paste the URL into the search bar and note that the study loads without error

This PR satisfies SCP-3510.

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -704,7 +704,7 @@ class Study
       self.where(queued_for_deletion: false)
     else
       studies = self.where(queued_for_deletion: false, user_id: user._id)
-      shares = StudyShare.where(email: user.email, permission: 'Edit').map(&:study).select {|s| !s.queued_for_deletion }
+      shares = StudyShare.where(email: /#{user.email}/i, permission: 'Edit').map(&:study).select {|s| !s.queued_for_deletion }
       [studies + shares].flatten.uniq
     end
   end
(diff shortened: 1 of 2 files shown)


eweitz (approved):

ehanna4 (approved):
LGTM 👍 thanks for fixing this!

PR #1110, opened 2021-07-29
Download HCA Project manifests via bulk download (SCP-3505)

bistline:
This update allows downloading of project manifest files from the HCA Azul service during bulk download requests.  These files manifests contain project-level metadata about all analysis/sequence/supplemental files associated with a given HCA project, including file provenance and access methods.  Since these are all public HCA files, the access URLs contained in the file manifest are all valid, and can be pasted into any browser to download the files directly (in case the user does not already download them through the bulk download UX).

This update also fixes an issue in the `BulkDownloadServiceTest` where the test `should get requested file sizes by query` was not asserting anything other than the expected outputs being declared.  The test has been rewritten to perform a strict equality comparison of method outputs against a known good structure.

MANUAL TESTING

1. Boot the portal and sign in with an account that has the `cross_dataset_search_backend` feature flag to `true`
2. Execute a keyword search for `pulmonary`
3. Once the Pulmonary Fibrosis study loads, click the "Download" button
4. Click "Next" (or optionally, wait for the "Sequence" column to finish updating and select all sequence files to download)
5. Click the copy icon once the `curl` command loads
6. Paste the `curl` command into a terminal window to download the data
7. Once the download completes, open the `PulmonaryFibrosisGSE135893.tsv` file and validate that it has project data in it

This PR satisfies SCP-3505.

### DIFF
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -31,10 +31,10 @@ module Api
             key :required, true
           end
           parameter do
-            key :name, :hca_project_id
-            key :type, :string
+            key :name, :hca_projects
+            key :type, :object
             key :in, :body
-            key :description, 'HCA project UUID, if specified'
+            key :description, 'Hash of HCA project short names to UUIDs, if specified'
             key :required, false
           end
           response 200 do
@@ -67,7 +67,7 @@ module Api
         half_hour = 1800 # seconds
 
         totat = current_api_user.create_totat(half_hour, api_v1_bulk_download_generate_curl_config_path)
-        valid_params = params.permit({file_ids: [], tdr_files: {}, hca_project_id: ''}).to_h
+        valid_params = params.permit({file_ids: [], tdr_files: {}, hca_project_ids: {}}).to_h
 
         # for now, we don't do any permissions validation on the param values -- we'll do that during the actual download, since
         # quota/files/permissions may change between the creation of the download and the actual download.
@@ -75,7 +75,7 @@ module Api
           auth_code: totat[:totat],
           file_ids: valid_params[:file_ids],
           tdr_files: valid_params[:tdr_files],
-          hca_project_id: valid_params[:hca_project_id],
+          hca_projects: valid_params[:hca_projects],
           user_id: current_api_user.id)
         auth_code_response = {
           auth_code: totat[:totat],
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -30,13 +30,6 @@ module Api
             key :description, 'Hash of file arrays to download from TDR, keyed by accession. Each file should specify URL and name'
             key :required, true
           end
-          parameter do
-            key :name, :hca_projects
-            key :type, :object
-            key :in, :body
-            key :description, 'Hash of HCA project short names to UUIDs, if specified'
-            key :required, false
-          end
           response 200 do
             key :description, 'One-time auth code and time interval, in seconds'
             schema do
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -289,9 +289,8 @@ module Api
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
           download_req = DownloadRequest.find(params[:download_id])
-          if !download_req
-            render json: { error: 'Invalid download_id provided' }, status: 400 and return
-          end
+          render json: { error: 'Invalid download_id provided' }, status: 400 and return if download_req.blank?
+
           file_ids = download_req.file_ids
           tdr_files = download_req.tdr_files
         elsif params[:file_ids]
@@ -329,15 +328,8 @@ module Api
           directory_files = []
         end
 
-        files_requested = []
-        # get requested files
-        # reference BulkDownloadService as ::BulkDownloadService to avoid NameError when resolving reference
-        if file_ids.any?
-          files_requested = StudyFile.where(:id.in => file_ids)
-        elsif valid_accessions.any? && sanitized_file_types.present?
-          files_requested = ::BulkDownloadService.get_requested_files(file_types: sanitized_file_types,
-                                                                      study_accessions: valid_accessions)
-        end
+        # get requested files, depending on access method
+        files_requested = self.class.load_study_files(ids: file_ids, accessions: valid_accessions, file_types: sanitized_file_types)
 
         # determine quota impact & update user's download quota
         # will throw a RuntimeError if the download exceeds the user's daily quota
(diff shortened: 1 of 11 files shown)


app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -43,41 +43,51 @@ export default function DownloadSelectionModal({ studyAccessions, tdrFileInfo, s
     setDownloadInfoTDR(tdrFileInfo)
     setIsLoadingTDR(false)
     // pull the drs ids out of each study and put them in a single array
+    // calling filter at end removes metadata file that does not have a DRS id
     const drsIds = tdrFileInfo.map(study => study.studyFiles.map(sfile => sfile.drs_id))
-      .reduce((acc, val) => acc.concat(val), [])
+      .reduce((acc, val) => acc.concat(val), []).filter(e => e != null)
```
eweitz, line 48: Strict equality is conventional:

```suggestion
      .reduce((acc, val) => acc.concat(val), []).filter(e => e !== null)
```

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -43,41 +43,51 @@ export default function DownloadSelectionModal({ studyAccessions, tdrFileInfo, s
...
-            sfile.name = matchFile.name
+          if (sfile.drs_id) {
+            const fileId = sfile.drs_id.split('/').slice(-1)[0]
+            const matchFile = result.find(file => file.id === fileId)
+            if (matchFile) {
+              sfile.upload_file_size = matchFile.size
+              // we add the url and name to each file info so that they can be sent along with the download
+              // request, this allows us to avoid having to query the DRS service again
+              sfile.url = matchFile.accessMethods.find(method => method.type === 'https').access_url.url
+              sfile.name = matchFile.name
```
eweitz, line 63: For readability, I'd change all uses of `sfile` to `studyFile`.  That's how the rest of SCP refers to these.

app/lib/bulk_download_service.rb
```diff
@@ -66,7 +77,7 @@ def self.generate_curl_configuration(study_files:,
     MetricsService.log('file-download:curl-config', {
       numFiles: study_files.count,
       numStudies: studies.count,
-      studyAccesions: studies.map(&:accession),
+      studyAccessions: studies.map(&:accession),
```
eweitz, line 80: Good catch!

Is it trivial to get counts for each study file _type_ -- at least for the big three?  E.g., `numMetadataFiles`, `numClusterFiles`, `numExpressionFiles`.  This is something Leyla was analyzing in the context of single-file downloads, and if it's < 10 minutes to instrument that here it might be worthwhile.  (If not, no worries, I can ticket it for later.)

app/models/hca_azul_client.rb
```diff
@@ -144,4 +175,27 @@ def get_project_manifest_link(catalog, project_id, format = 'compact')
...
+      # fall back to HCA_CATALOGS
+      is_valid = HCA_CATALOGS.include? catalog
+    end
+    raise ArgumentError, "#{catalog} is not a valid catalog: #{HCA_CATALOGS.join(',')}" unless is_valid
+  end
+
+  # validate requested format is valid
+  def validate_manifest_format(format)
+    unless MANIFEST_FORMATS.include?(format)
+      raise ArgumentError, "#{format} is not a valid format: #{MANIFEST_FORMATS.join(',')}"
```
eweitz, line 198: Blocker: Let's also log these to Sentry, at least, if they're not already.

eweitz (changes requested):
Looking pretty good!  I noted one hopefully trivial blocker re observability, and a few non-blocking maintainability suggestions.

app/lib/bulk_download_service.rb
bistline: Yeah that should be very easy - I can make that change.

eweitz (approved):

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -280,30 +282,31 @@ def generate_curl_config
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
           download_req = DownloadRequest.find(params[:download_id])
-          if !download_req
```
ehanna4, line 283: i think using `!download_req` produced slightly different results than `download_req.blank?` will (referencing these docs: https://apidock.com/rails/Object/blank%3F) is this logic shift intended?

app/controllers/api/v1/search_controller.rb
```diff
@@ -688,9 +688,9 @@ def self.match_results_by_filter(search_result:, result_key:, facets:)
         if db_facet.is_numeric?
           match = matching_facet[:filters].dup
           match.delete(:name)
-          return match
+          match
```
ehanna4, line 691: Just to make sure I understand it - having an explicit `return` is not preferred by Ruby right?

bistline: Yes - all Ruby methods implicitly return the last object operated on, so just calling `foo` will return whatever value is assigned to `foo`.  Calling `return` is only required when you need to explicitly halt execution and return a value.

app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -43,41 +43,51 @@ export default function DownloadSelectionModal({ studyAccessions, tdrFileInfo, s
...
-        study.studyFiles.forEach(sfile => {
-          const fileId = sfile.drs_id.split('/').slice(-1)[0]
-          const matchFile = result.find(file => file.id === fileId)
-          if (matchFile) {
-            sfile.upload_file_size = matchFile.size
-            // we add the url and name to each file info so that they can be sent along with the download
-            // request, this allows us to avoid having to query the DRS service again
-            sfile.url = matchFile.accessMethods.find(method => method.type === 'https').access_url.url
-            sfile.name = matchFile.name
+        study.studyFiles.forEach(studyFile => {
```
ehanna4, line 54: i appreciate the rename of `sfile` to `studyFile`😆

ehanna4 (approved):
Looks good! Manually tested and was able to download files however I did see this in the terminal not sure it's a blocker but figured it was worth mentioning
Screen Shot 2021-08-05 at 3 51 32 PM

bistline:
> Looks good! Manually tested and was able to download files however I did see this in the terminal not sure it's a blocker but figured it was worth mentioning
> Screen Shot 2021-08-05 at 3 51 32 PM

I've seen that issue as well, it's part of a lower-level issue with "phantom" empty parameters being added to API requests.  It has to do with how we've configured our [routing](https://github.com/broadinstitute/single_cell_portal_core/blob/development/config/routes.rb#L8), and I've never been able to fully figure out why they're being sent.  Regardless, they're unused and get filtered anyway, so no harm done.  Thanks for checking!

PR #1111, opened 2021-08-02
Add the badges for TDR results for facet matching (SCP-3473)

ehanna4:
This adds the facet matching badge for TDR results in the study search.

To Test locally:
Try searching for a study using the facet "homo sapien" and you should see the TDR result with the badge same as any SCP study 

Search on both keyword and facet matching:
Screen Shot 2021-07-30 at 12 30 49 PM

Hovering tooltip on TDR:
Screen Shot 2021-07-30 at 12 30 56 PM

Just Facet match:
Screen Shot 2021-07-30 at 12 31 08 PM

Just keyword match still works as usual
Screen Shot 2021-07-30 at 12 31 22 PM

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -294,9 +294,9 @@ module Api
         # perform TDR search, if enabled, and there are facets/terms provided by user
         if User.feature_flag_for_instance(current_api_user, 'cross_dataset_search_backend') && (@facets.present? || @term_list.present?)
           @tdr_results = self.class.get_tdr_results(selected_facets: @facets, terms: @term_list)
-          
+
           if @facets.present?
-            simple_tdr_results = self.class.simplify_TDR_facet_search_results(@tdr_results, @facets)
+            simple_tdr_results = self.class.simplify_tdr_facet_search_results(@tdr_results, @facets)
             matched_tdr_studies = self.class.match_studies_by_facet(simple_tdr_results, @facets)
             
             if @studies_by_facet.present?
@@ -309,9 +309,7 @@ module Api
           logger.info "Found #{@tdr_results.keys.size} results in Terra Data Repo"
           logger.info pp @tdr_results if @tdr_results.any?
           @tdr_results.each do |short_name, tdr_result|
-              if tdr_result[:accession] != nil
-                @studies << tdr_result
-              end
+              @studies << tdr_result
           end
         end
 
@@ -636,20 +634,18 @@ module Api
       def self.simplify_tdr_facet_search_results(query_results, search_facets)
         simple_TDR_result = {}
         simple_TDR_results = []
-        query_results.each do |result|
-          unless result[0].nil?
-            result_details = result[1]
-            accession = result_details[:accession]
-            facet_matches = result_details[:facet_matches]
-            simple_TDR_result[:study_accession] = accession
-            if facet_matches.present?
-              facet_matches[0].each_pair { |key, val| 
-              short_name_field = FacetNameConverter.convert_to_scp(key, :name)
-              simple_TDR_result[short_name_field] = val
+        query_results.each_pair do |accession, result|
+          facet_matches = result[:facet_matches]
+          simple_TDR_result[:study_accession] = accession
+          if facet_matches.present?
+            facet_matches.each { |mapping| 
+              mapping.each_pair { |key, val| 
+                short_name_field = FacetNameConverter.convert_to_model(:tim, :alex, key, :name)
+                simple_TDR_result[short_name_field] = val
               }
-            end
+            }
           end
-          simple_TDR_results.append(simple_TDR_result)
+          simple_TDR_results << simple_TDR_result
         end
         simple_TDR_results
       end
@@ -793,9 +789,8 @@ module Api
         else
           matches = []
           facet[:filters].each do |filter|
-            # look for match on the column name, and see which filter value also matched (ontology label or id)
-            matches << result_row.each_pair.select { |col, val| col == tdr_name && (val == filter[:name] || filter[:id]) }
-                                 .flatten
+            matches << result_row.each_pair.select { |col, val| col == tdr_name && (val == filter[:name] || val == filter[:id] ) }
+                                 .flatten                                 
           end
         end
         matches
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -215,7 +215,6 @@ module Api
 
         # only call BigQuery if list of possible studies is larger than 0 and we have matching facets to use
         if @studies.count > 0 && @facets.any?
-          # puts "studies at begining of the facet matching: #{@studies}"
           sort_type = :facet
           @studies_by_facet = {}
           @big_query_search = self.class.generate_bq_query_string(@facets)
@@ -223,7 +222,7 @@ module Api
           query_results = ApplicationController.big_query_client.dataset(CellMetadatum::BIGQUERY_DATASET).query @big_query_search
           job_id = query_results.job_gapi.job_reference.job_id
           @studies_by_facet = self.class.match_studies_by_facet(query_results, @facets)
-
+          # build up map of study matches by facet & filter value (for adding labels in UI)
           # uniquify result list as one study may match multiple facets/filters
           @convention_accessions = query_results.map {|match| match[:study_accession]}.uniq
           logger.info "Found #{@convention_accessions.count} matching studies from BQ job #{job_id}: #{@convention_accessions}"
@@ -319,11 +318,7 @@ module Api
 
         @matching_accessions = @studies.map {|study| study.is_a?(Study) ? study.accession : study[:accession]}
         logger.info "Final list of matching studies: #{@matching_accessions}"
-        logger.info "Final list of matching studies: #{@studies}"
-
         @results = @studies.paginate(page: params[:page], per_page: Study.per_page)
-        logger.info "results: #{@results}"
-        logger.info "search_results_obj: #{search_results_obj}"
         render json: search_results_obj, status: 200
       end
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -294,11 +294,24 @@ def index
         # perform TDR search, if enabled, and there are facets/terms provided by user
         if User.feature_flag_for_instance(current_api_user, 'cross_dataset_search_backend') && (@facets.present? || @term_list.present?)
           @tdr_results = self.class.get_tdr_results(selected_facets: @facets, terms: @term_list)
+          
+          if (@facets.present?)
```
bistline, line 298: Redundant parentheses - Ruby doesn't require them for `if/else` blocks.
```suggestion
          if @facets.present?
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -294,11 +294,24 @@ def index
...
+            simple_tdr_results = self.class.simplify_TDR_facet_search_results(@tdr_results, @facets)
+            matched_tdr_studies = self.class.match_studies_by_facet(simple_tdr_results, @facets)
+            
+            if @studies_by_facet.present?
+              @studies_by_facet = @studies_by_facet.merge(matched_tdr_studies)
```
bistline, line 303: You can call `merge!` here instead to do an in-place update, rather than reassigning to itself:
```suggestion
              @studies_by_facet.merge!(matched_tdr_studies)
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -294,11 +294,24 @@ def index
...
+            else
+              @studies_by_facet = matched_tdr_studies
+            end
+          end
           # just log for now
           logger.info "Found #{@tdr_results.keys.size} results in Terra Data Repo"
           logger.info pp @tdr_results if @tdr_results.any?
           @tdr_results.each do |short_name, tdr_result|
-            @studies << tdr_result
+              if tdr_result[:accession] != nil
```
bistline, line 312: How could `:accession` be `nil` for a TDR result?  Shouldn't we always get back an HCA short name for all TDR results from `process_tdr_result_row`?

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
         matches
       end
 
+      # Simplify TDR results to be mappable for the UI badges for facet search
+      def self.simplify_TDR_facet_search_results(query_results, search_facets)
```
bistline, line 636: Ruby convention is to not use capitals in method names - those are typically reserved for class/module names.
```suggestion
      def self.simplify_tdr_facet_search_results(query_results, search_facets)
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+        simple_TDR_result = {}
+        simple_TDR_results = []
+        query_results.each do |result|
```
bistline, line 639: Since `query_results` is a Hash, we can iterate over the keys & values natively like below.  In this case, the keys are "accessions", or project short names.  This way you don't need to assign `results_details` as it will automatically be `result`:
```suggestion
        query_results.each do |accession, result|
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+          unless result[0].nil?
```
bistline, line 640: I definitely would like to grab a few minutes of your time to look into this more - this shouldn't ever be the case.  If we're getting back results without accessions/short names, that's a problem.

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+            result_details = result[1]
+            accession = result_details[:accession]
+            facet_matches = result_details[:facet_matches]
+            simple_TDR_result[:study_accession] = accession
+            if facet_matches.present?
+              facet_matches[0].each_pair { |key, val|
```
bistline, line 646: Why are we only ever looking at the first entry in `facet_matches`?  Don't we want to iterate over all of them to see what matches?

Also, style nit: when doing a multi-line block in Ruby, don't use the inline form of `{ |foo| }`, but rather the `do` keyword:

```suggestion
              facet_matches[0].each_pair do |key, val| 
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+          unless result[0].nil?
+            result_details = result[1]
+            accession = result_details[:accession]
+            facet_matches = result_details[:facet_matches]
+            simple_TDR_result[:study_accession] = accession
+            if facet_matches.present?
+              facet_matches[0].each_pair { |key, val| 
+              short_name_field = FacetNameConverter.convert_to_scp(key, :name)
+              simple_TDR_result[short_name_field] = val
+              }
```
bistline, line 649: This won't be needed with the `do` form of the block.

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +632,29 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+            end
+          end
+          simple_TDR_results.append(simple_TDR_result)
```
bistline, line 652: Personally, I prefer `<<` over `append` but that's just me.

lib/facet_name_converter.rb
```diff
@@ -48,4 +65,19 @@ def self.convert_to_model(model_name, column_name, property)
...
+
+  # convert from Terra Interoperability Model to SCP metadata names 
+  #
+  # * *params*
+  #   - +column_name+ (String, Symbol) => facet name to convert
+  #   - +property+ (String, Symbol) => property to return (:id or :name only)
+  #
+  # * *returns*
+  #   - (String) => String value of requested column/property
+  def self.convert_to_scp(column_name, property)
```
bistline, line 79: Rather than the new method here (which is totally fine, btw!), I would rather refactor `convert_to_model` so that we can go from any one convention to the other.  Essentially, you'd have `source_model` and `target_model` models (such as `:tim`, `:hca`, and `:scp`), and then depending on the combinations, you'd pick the right constant to convert to/from.

I'm thinking something like this (there might be a better way to assign `mappings`, this is quick & dirty):
```
def self.convert_to_model(source_model, target_model, column_name, property)
  if source_model.to_sym == :tim
    mappings = TIM_TO_ALEXANDRIA
  else
    mappings = target_model.to_sym == :hca ? ALEXANDRIA_TO_HCA : ALEXANDRIA_TO_TIM
  end
  # perform lookup, but fall back to provided column name if no match is found
  mappings[column_name.to_sym]&.dig(property.to_sym) || column_name
end
```

bistline (changes requested):
Overall, pretty good!  Mostly style/linting requests, but I would like to talk about the potential missing accessions to make sure something isn't amiss.

app/controllers/api/v1/search_controller.rb
```diff
@@ -294,11 +294,22 @@ def index
...
+            else
+              @studies_by_facet = matched_tdr_studies
+            end
+          end
           # just log for now
           logger.info "Found #{@tdr_results.keys.size} results in Terra Data Repo"
           logger.info pp @tdr_results if @tdr_results.any?
           @tdr_results.each do |short_name, tdr_result|
-            @studies << tdr_result
+              @studies << tdr_result
```
bistline, line 312: non-blocker: extra whitespace

app/controllers/api/v1/search_controller.rb
```diff
@@ -757,9 +789,8 @@ def self.get_facet_match_for_tdr_result(facet, result_row)
         else
           matches = []
           facet[:filters].each do |filter|
-            # look for match on the column name, and see which filter value also matched (ontology label or id)
-            matches << result_row.each_pair.select { |col, val| col == tdr_name && (val == filter[:name] || filter[:id]) }
-                                 .flatten
+            matches << result_row.each_pair.select { |col, val| col == tdr_name && (val == filter[:name] || val == filter[:id] ) }
```
bistline, line 792: Blocker: this fixes the problem we saw yesterday with bogus matches, however it now throws an error if you search for other facets that don't match on the Pulmonary TDR study:
```
undefined method `to_sym' for nil:NilClass
single_cell_portal_core/lib/facet_name_converter.rb:70:in `convert_to_model'
single_cell_portal_core/app/controllers/api/v1/search_controller.rb:644:in `block (3 levels) in simplify_tdr_facet_search_results'
single_cell_portal_core/app/controllers/api/v1/search_controller.rb:642:in `each_pair'
single_cell_portal_core/app/controllers/api/v1/search_controller.rb:642:in `block (2 levels) in simplify_tdr_facet_search_results'
```
The issue is that we now end up with a "match" that looks like this: `{nil => nil}`, which throws the `nil` error when trying to convert from TIM to Alexandria.  If we add the following line right before the final return of `matches`, it will work, though we still have the problem identified in SCP-3544 where our matching is too lax (out of scope for this):
```
matches.reject! { |key, value| key.blank? || value.blank? }
```

bistline (approved):
Conditionally approving modulo adding the specified fix for `SearchController#get_facet_match_for_tdr_result`.  Otherwise, looks good!

app/controllers/api/v1/search_controller.rb
ehanna4: oh shoot good catch, thanks!

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +630,27 @@ def self.match_studies_by_facet(query_results, search_facets)
         matches
       end
 
+      # Simplify TDR results to be mappable for the UI badges for facet search
```
eweitz, line 633: More conventional phrasing:

```suggestion
      # Simplify TDR results to be mappable for the UI badges for faceted search
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -619,6 +630,27 @@ def self.match_studies_by_facet(query_results, search_facets)
...
+        simple_TDR_results = []
+        query_results.each_pair do |accession, result|
+          facet_matches = result[:facet_matches]
+          simple_TDR_result[:study_accession] = accession
+          if facet_matches.present?
+            facet_matches.each { |mapping| 
+              mapping.each_pair { |key, val| 
+                short_name_field = FacetNameConverter.convert_to_model(:tim, :alex, key, :name)
+                simple_TDR_result[short_name_field] = val
+              }
```
eweitz, line 645: Great readability!

eweitz (approved):

PR #1113, opened 2021-08-04
Fixing query syntax for TDR results (SCP-3544)

bistline:
This update fixes an issue where search results in TDR are far too lax, resulting in partial matches on filter values that give erroneous results.  For instance, searching for `cataract (disease)` would result in matching on either `cataract` or `disease` anywhere in the entry, and not just within the `disease` field.  The issue was due to incorrect formatting of queries in regards Terra Interoperability Model (TIM) names, and also not quoting values correctly.  

TIM names can be passed in human-readable form (e.g. `TerraCore:hasDisease`), but must be enclosed in brackets `[]` to denote the field.  Also, values must be fully quoted to prevent partial matching.  For instance, a query for `HIV infectious disease` would look like this in ElasticSearch DSL:

```
{:query_string=>{:query=>"([TerraCore:hasDisease]:\"MONDO_0005109\") OR ([TerraCore:hasDisease]:\"HIV infectious disease\")"}}
```

This update also simplifies and hardens the `FacetNameConverter` class.  It is now possible to convert from Alexandria to TIM or HCA names, and back.

MANUAL TESTING
1. Boot portal as normal, and sign in with an account that has `cross_dataset_search_backend` feature flag enabled
2. Use facet for `species: Homo sapiens`, and note both SCP & TDR results
3. Search for `Pulmonary`, and note only TDR results
4. Use facet for `disease: HIV infectious disease` and note only SCP results

This PR satisfies SCP-3544.

### DIFF
--- a/lib/facet_name_converter.rb
+++ b/lib/facet_name_converter.rb
@@ -1,72 +1,54 @@
 # class for converting from Alexandria convention names to HCA or TIM metadata model names (i.e. columns, not individual values)
 # this is currently for PoC work on XDSS - eventually this will be replaced by an onotology server that can handle
 # conversions programmatically
-# entries have a :name (column name in query results for facet matching) and :id (ElasticSearch encoded property name)
 class FacetNameConverter
   # map of Alexandria metadata convention names to HCA 'short' names
   ALEXANDRIA_TO_HCA = {
-    biosample_id: { name: 'biosample_id', id: 'biosample_id' },
-    cell_type: { name: 'cell_type', id: 'cell_type' },
-    donor_id: { name: 'donor_id', id: 'donor_id' },
-    disease: { name: 'disease', id: 'disease' },
-    library_preparation_protocol: {name: 'library_construction_method', id: 'library_construction_method' },
-    organ: { name: 'organ', id: 'organ' },
-    organism_age: { name: 'organism_age', id: 'organism_age' },
-    sex: { name: 'sex', id: 'sex' },
-    species: { name: 'genus_species', id: 'genus_species' },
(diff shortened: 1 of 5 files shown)


lib/facet_name_converter.rb
```diff
@@ -1,72 +1,70 @@
...
+    'disease' => 'disease',
+    'library_preparation_protocol' => 'library_construction_method',
+    'organ' => 'organ',
+    'organism_age' => 'organism_age',
+    'sex' => 'sex',
+    'species' => 'genus_species',
+    'study_name' => 'project_title',
+    'study_description' => 'project_description',
+    'accession' => 'project_short_name'
+  }.with_indifferent_access.freeze
```
eweitz, line 22: Nice readability improvement!

eweitz (approved):

lib/facet_name_converter.rb
```diff
@@ -1,72 +1,70 @@
...
-    'TerraCore:hasAnatomicalSite': { name: 'organ'},
-    'organism_age': { name: 'organism_age'},
-    'TerraCore:hasSex': { name: 'sex'},
-    'TerraCore:hasOrganismType': {name: 'species'},
-    'dct:title': { name: 'study_name' },
-    'dct:description': { name: 'study_description'},
-    'rdfs:label': { name: 'accession'}
-  }.freeze
-  
+  # inverted mappings of TIM/HCA to Alexandria
```
ehanna4, line 39: wow so much better!!

ehanna4 (approved):
Woo thanks for fixing this, no more wonky results 👍. Manually tested and verified TDR results actually only come up for appropriate facets

PR #1115, opened 2021-08-05
Replacing `node-sass` with `sass`, JS updates, UserAssetServiceTest fix (SCP-3555, SCP-3569)

bistline:
This update replaces the deprecated `node-sass` package with `sass`, and updates various packages to the next minor/patch release.  This is part of an effort to remediate the [security warning](https://github.com/broadinstitute/single_cell_portal_core/security/dependabot/yarn.lock/tar/open) for the `tar` package, which was pinned to a deprecated version due to `node-sass`, and therefore could not be automatically patched by `dependabot`.

In addition, this updates to `1.3.0` of `scp-rails-baseimage`, which includes an update to `nodejs` v16.  Local testing has shown that no problems with static asset compilation are introduced with the update (currently on node 12 in Dockerized deployments).

This PR also includes a fix for `UserAssetServiceTest#test_should_push_and_pull_assets_from_remote`, which was failing on all builds starting yesterday.  The issue was that the test is not idempotent, and previous failures left the remote asset storage bucket in an inconsistent state, which cause all subsequent test runs to fail.  The issue has been corrected, and local state caching has been added so that developers can run tests locally w/o losing the state of any cached user assets (like for branding groups).

This PR satisfies SCP-3555 and SCP-3569.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,3 +49,6 @@ Style/Documentation:
 
 Style/NumericPredicate:
   Enabled: false
+
+Style/ConditionalAssignment:
+  Enabled: false
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # use KDUX base Rails image, configure only project-specific items here
-FROM gcr.io/broad-singlecellportal-staging/rails-baseimage:1.2.1
+FROM gcr.io/broad-singlecellportal-staging/rails-baseimage:1.3.0
 
 # Set ruby version
 RUN bash -lc 'rvm --default use ruby-2.6.6'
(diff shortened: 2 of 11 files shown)


eweitz (approved):
Thanks for this!  

A while ago I encountered `yarn install` issues with `node-sass` ("gyp: No Xcode or CLT version detected!"), and had to run commands [here](https://github.com/schnerd/d3-scale-cluster/issues/7#issuecomment-550579897).  I recall you did too.  

Such fixes take forever.  I'm glad it seems we'll be free of them from now on.

ehanna4 (approved):
Looks good, tested clicking around the UI locally and everything works as expected and I don't see any new warnings in the console

PR #1116, opened 2021-08-05
Prototype: Validate files in client: reduce time to fail from 2-3 min to < 1 s (SCP-3496, SCP-3498)

eweitz:
This adds a prototype of basic client-side validation for study files.  It has two big benefits:

* Reduce time to fail from 2-3 minutes to less than 1 second
* Report errors within app UI instead of via email

For now, this feature is hidden by default behind the flag `clientside_validation`.  Work remains for additional validation rules, error reporting UI, and analytics.

The implementation uses the `FileReader` API to read a file that the user selects from their local file system.  The file is then validated using JavaScript that closely follows patterns from the Python implementation of validation rules in Ingest Pipeline (entry point e.g. [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/annotations.py#L328)).

To test:
1. Do migration: run `ruby rails_local_setup.rb`, then `source config/secrets/.source_env.bash`, then `rails db:migrate`
2. Sign in
3. Open upper right nav menu -> Admin Controls -> User Roles (https://localhost:3000/single_cell/admin#users)
4. Find yourself, set `clientside_validation` feature flag to "True"
5. Create a new study, e.g. "Client-side validation test"
6. In "Metadata" tab of upload wizard, click "Choose File"
7. Select file at `single_cell_portal_core/test/test_data/metadata_example_bad_TYPE.txt`
8. Verify that you see an alert reporting an error, as shown in brief video below.

https://user-images.githubusercontent.com/1334561/128394912-1213fa03-042c-4d43-bafb-a128da94d506.mov

This satisfies SCP-3496 and SCP-3498.


bistline:
I doubt any of your changes caused this, but it appears the build failed due to a `webpack` issue:
```
Compiling...
Compilation failed:
yarn run v1.22.5
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command "webpack" not found.
```

eweitz:
Good spot, Jon.  Another build failed for the proximate cause, [ultimately caused by](https://app.travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/234623812#L1767):

> error node-sass@6.0.1: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.24.1"
error Found incompatible module.

Serendipitously, I think [your open PR to replace `node-sass`](https://github.com/broadinstitute/single_cell_portal_core/pull/1115) will fix that true positive error.  When that merges into `development`, I'll merge it into this branch.

app/javascript/lib/validation/validate-file.js
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Verify second row starts with TYPE (case-insensitive)
+ */
+function validateTypeKeyword(annotTypes) {
+  const issues = []
+
+  const value = annotTypes[0]
+
+  if (value.toUpperCase() === 'TYPE') {
+    if (value !== 'TYPE') {
```
ehanna4, line 51: What is this trying to check? The comment made me think "type" or "TYPE" would be acceptable but this check makes it seem like only "TYPE" is ok?

eweitz: Great question.  

Like the [reference implementation](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/annotations.py#L258) in Ingest Pipeline, if the provided header is "type", then it passes a case-insensitive match but fails a case-sensitive match and thus gets a warning (see line 53 here).  

However, if the provided header is e.g. "foo", then it doesn't even pass a case-insensitive match, and thus gets an error.

If the provided header is "TYPE", then it passes a case sensitive match, and gets neither a warning nor an error.

bistline (approved):
Great start!  Really looking forward to seeing this refined and released.

ehanna4 (approved):
Looks good to me - tested manually following steps provided and saw warning as expected

PR #1117, opened 2021-08-09
Removing static list of HCA catalogs (SCP-3571)

bistline:
Since the Human Cell Atlas releases new catalogs on a regular basis, maintaining a static, hard-coded list will not scale well as it will require frequent changing.  This update removes the HcaAzulClient::HCA_CATALOGS constant and replaces it with an instance variable that populates on initialization.  This list is then used to validate requests, which will throw an error if a non-public or non-existent catalog is requested.  Also, if the list of catalogs is out of date with the remote, a fallback validation will query the API to see if a requested catalog is valid.

In addition, this update will prevent the test `HcaAzulClientTest#should_get_catalogs` from failing when a new catalog is released, and the previous version is deprecated.

This PR satisfies SCP-3571.


eweitz (approved):

ehanna4 (approved):
Looks good to me 👍

PR #1118, opened 2021-08-09
Legend Fixes SCP-3572 & SCP-3466

ehanna4:
- Sort Legend annotations "Naturally"
- Add space so that legend does not overlap Plotly tool menu
- Add title to legend for gene-gene correlation

I think I've ended up being able to fix this using Plotly legends for now. (I must have missed something small earlier and super overcomplicated things..). 

Note: The colors are slightly different with this implementation as it appears the colors stay in a certain order no matter the annotation order. However they match to the graph and I checked with Jean and it seems to be alright. 

Before:
Screen Shot 2021-08-08 at 3 06 11 PM

After:
Screen Shot 2021-08-09 at 10 46 19 AM

Screen Shot 2021-08-09 at 10 46 31 AM

Overlap Before:
Screen Shot 2021-08-09 at 10 07 37 AM

Overlap Fixed:
Screen Shot 2021-08-09 at 10 19 38 AM

Before Gene-Gene correlation:
Screen Shot 2021-08-09 at 11 50 54 AM

AFter:
Screen Shot 2021-08-09 at 11 51 55 AM

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -295,7 +295,7 @@ function countOccurences(array) {
 function traceNameSort(a, b) {
   if (a === UNSPECIFIED_ANNOTATION_NAME) {return 1}
   if (b === UNSPECIFIED_ANNOTATION_NAME) {return -1}
-  return a.localeCompare(b)
+  return a.localeCompare(b, 'en', { numeric: true, ignorePunctuation: true })
 }
 
 /** makes the data trace attributes (cells, trace name) available via hover text */
@@ -366,7 +366,8 @@ function getPlotlyLayout({ width, height }={}, {
   if (!genes.length && annotParams && annotParams.name) {
     layout.legend = {
       itemsizing: 'constant',
-      title: { text: annotParams.name }
+      title: { text: annotParams.name },
+      y: 0.94
     }
   }
   layout.width = width
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -364,6 +364,7 @@ function getPlotlyLayout({ width, height }={}, {
     Object.assign(layout, props2d)
   }
   if (!genes.length && annotParams && annotParams.name) {
+  if (annotParams && annotParams.name) {
     layout.legend = {
       itemsizing: 'constant',
       title: { text: annotParams.name },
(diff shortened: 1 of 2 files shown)


eweitz (changes requested):
Great fixes!

Blocker: like `localeCompare` needed refinement in `ScatterPlot.js`, so it does [in StudyViolinPlot.js]( https://github.com/broadinstitute/single_cell_portal_core/blob/d3a6fdec9b37f2d5c8542b7bcb6bdacd505838dd/app/javascript/components/visualization/StudyViolinPlot.js#L190).  Without that coordinated fix, labels in violin plots and box plots -- available in the "Distribution" tab in the single-gene view -- would have the old naive sorting.  And they'd be newly inconsistent with these nicely refined scatter plot legends.

ehanna4:
Per Erics feedback -fixed Violin Plot sorting as well
Screen Shot 2021-08-09 at 1 33 48 PM

eweitz (approved):

jlchang:
Emily, in your violin plot screenshot, it looks like the legend's icons (squares) reflect the annotation label's color in the border of the square with a lighter color for the fill. This seems a little different from what I see in production where the fill reflects the color and the border is black... Screen Shot 2021-08-09 at 2 43 58 PM

ehanna4:
> Emily, in your violin plot screenshot, it looks like the legend's icons (squares) reflect the annotation label's color in the border of the square with a lighter color for the fill. This seems a little different from what I see in production where the fill reflects the color and the border is black... 

that's strange i will take a look

jlchang (changes requested):
Could you double check the violin plot legend icon (squares) rendering to see if it can match what's currently used in production?
(screenshot for comparison in my comment at https://github.com/broadinstitute/single_cell_portal_core/pull/1118#issuecomment-895455624)

bistline:
@jlchang that legend behavior you're seeing only happens when you have a gene with no expression in a particular annotation value.  In the case of an empty distribution, we don't actually use a violin plot, but rather an empty box plot.  This was due to some weirdness with violin rendering that occasionally showed what looked like a positive/negative distribution, rather than the single line.

bistline (approved):

ehanna4:
> Could you double check the violin plot legend icon (squares) rendering to see if it can match what's currently used in production?
> (screenshot for comparison in my comment at [#1118 (comment)](https://github.com/broadinstitute/single_cell_portal_core/pull/1118#issuecomment-895455624))

It appears like the colors are not outlined in black when the cell is 0? My first screenshot all the cells were at 0 so I guess that would follow that none of the squares were black outlined. Does that seem reasonable?

Screen Shot 2021-08-09 at 3 00 32 PM

Screen Shot 2021-08-09 at 3 01 54 PM

jlchang (approved):
I guess I hadn't noticed the change in legend icon behavior for zero-valued results! Thanks for looking into it.

PR #1120, opened 2021-08-12
Creating anonymous publication reviewer access system (SCP-3574)

bistline:
Currently, granting access to reviewers from scientific journals to private SCP studies is mostly broken.  It relies on creating shared Google logins that multiple reviewers attempt to authenticate into SCP with.  This fails to work most of the time as Google's security rightly flags the multiple logins as suspicious, and employs 2-factor authentication to validate that the reviewer does in fact have access.  Reviewers are not allowed to have direct contact with corresponding authors, therefore they will not have access to any of the second factors.  This means login is blocked, and review is impossible.  

This update adds the `ReviewerAccess` and `ReviewerAccessSession` models to facilitate anonymous reviewer access to private studies.  The flow works as follows: a study owner visits the "Study Settings" tab for their private study, and enables reviewer access.  Once enabled, the study owner is provided with an access URL and a PIN code, which they give to their corresponding journal editor.  When reviewers are assigned, the editor provides them with the credentials, and they navigate to the access URL and enter the PIN code.  Successful authentication via the PIN creates a signed session cookie which allows limited access to the private study (reviewers can interact with visualizations, but not download data or visit the Terra workspace).  These sessions last for 4 hours, at which point they expire and cannot be used again.  The session cookies themselves only apply to a single study, meaning a reviewer can have access to multiple private studies (they are name-spaced with the study accession).  Furthermore, these are "signed" cookies, meaning that they are signed by the application and cannot be altered in any way.  Attempts to do so will cause the cookie to delete itself.

A reviewer can request a new session by re-entering the PIN at the access URL.  Study owners can also reset credentials to generate a new access URL & PIN, which will also invalidate all active reviewer sessions.  Furthermore, the reviewer access functionality itself has an expiration date, at which point all reviewer access is blocked.  Study owners can also manually turn off access at any time.

MANUAL TESTING
1. On a private study with visualizations enabled, navigate to the "Study Settings" tab, select "Yes" for "Access Enabled?", then click "Update Study Settings"
Screen Shot 2021-08-18 at 1 16 30 PM
2. Once saved, click the "Copy to clipboard" button under "Current Access Credentials" and paste into a text document to see the credentials.
Screen Shot 2021-08-18 at 1 12 14 PM
3. Paste the access URL into a **new** browser window (preferably incognito), then on the reviewer access page, enter the 10 digit PIN code and click "Submit"
Screen Shot 2021-08-18 at 1 17 46 PM
4. After submission, you should be redirected to the study overview page.  Click around the study, trying different cluster/annotation and gene search combinations
5. Open the Chrome Dev Tools window, and click into the "Application" tab, then select "Cookies" from the left.  You should see an entry for `reviewer_session_[:accession]` in the list, like so:
Screen Shot 2021-08-18 at 1 06 14 PM
6. (Optional) In the original window, select either "Yes" for "Reset Access Credentials" or "No" for "Access Enabled" and save (both will invalidate the current session)
7. (Optional) Back in the incognito window, refresh the page and note that you are redirected to the homepage with a notice that either the session has expired (if you rotated the credentials) or that you do not have permission to perform the action (if you revoked access altogether)

This PR satisfies SCP-3574.


app/views/site/_reviewer_access_fields.html.erb
```diff
@@ -0,0 +1,60 @@
...
+    <div class="col-md-6 left-border-0-5-gray">
+      <h4>Current Access Credentials <button class="btn btn-default btn-sm btn-copy" data-toggle="tooltip" title=""
+                                             data-original-title="Copy to clipboard"><i class="far fa-copy"></i></button>
+      </h4>
+      <% if access_enabled %>
+        <p id='access-credentials'>
+          <strong>Accession:</strong> <%= @study.accession %><br />
+          <strong>URL:</strong>&nbsp;<%= link_to reviewer_access_url(access_code: f.object.access_code),
+                                                 reviewer_access_url(access_code: f.object.access_code) %><br />
+          <strong>Pin:</strong>&nbsp;<%= f.object.pin %>
```
dinvlad, line 39: We should mask the PIN display and only show it for a short period of time when clicking on a "show" button. Doing so minimizes the risk of a rogue script or browser extension "grabbing" the PIN from the raw HTML..

app/models/reviewer_access.rb
```diff
@@ -0,0 +1,69 @@
+# manage anonymous access to private studies via access code & pin
+# will generate a short-lived session for reviewers with valid credentials
+class ReviewerAccess
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :access_code, type: String # UUID
+  field :pin, type: String # 6 character string
```
dinvlad, line 8: We should increase the number of characters (and possibly their type). 6-pin is considered quite "trivial" from brute-forcing standpoint. You may want to introduce other alphanumeric characters to increase the entropy too.

bistline: Sure, I can have it generate from any alphanumeric.

app/views/site/_reviewer_access_fields.html.erb
bistline: I'll look into doing that - great suggestion.

dinvlad:
Could you please point to the code generating `reviewerSession` once `access_code` and `pin` are validated?

app/models/reviewer_access.rb
dinvlad: Also, are the PIN characters submitted by the user validated somehow? I.e. is it just an "equality" comparison or do you restrict the character set on the backend also?

bistline: It's just an equality comparison, but I like where you're going with that.  I should strip non-alphanumeric characters.  The form on the front end limits the characters, so a user can't enter more that the allowed limit of 6.  I could also do the same on the backend as well, just to be sure.

bistline:
> Could you please point to the code generating `reviewerSession` once `access_code` and `pin` are validated?

@dinvlad The `reviewerSession` is generated here: https://github.com/broadinstitute/single_cell_portal_core/pull/1120/files?file-filters%5B%5D=.erb&file-filters%5B%5D=.js&file-filters%5B%5D=.rb#diff-a30d775449f9a973af282ef5fb2ede2abd479c96d780712714362f6ab55f4244R36

which is called from here: https://github.com/broadinstitute/single_cell_portal_core/pull/1120/files?file-filters%5B%5D=.erb&file-filters%5B%5D=.js&file-filters%5B%5D=.rb#diff-513aac5330921408a0a541be267b63f8af569dfa0164f40f5a13c23bbf899899R52

After that, the new UUID is appended to all API requests, which gets validated here: https://github.com/broadinstitute/single_cell_portal_core/pull/1120/files?file-filters%5B%5D=.erb&file-filters%5B%5D=.js&file-filters%5B%5D=.rb#diff-b68aea0b109098ecbbbaf443c1e11867ab5326be92603cd4d067cd05d75dc0deR21

app/models/reviewer_access.rb
dinvlad: Thanks! The main concern here is we should use a "secure equality" operator, since this is a common DoS vulnerability when a regular `==` is used. So perhaps take a look at https://github.com/ConradIrwin/secure_equals or something similar.

bistline: Gotcha.  I found this which looks pretty robust - it's part of `ActiveSupport` which is a core Rails library:

https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html

app/javascript/components/explore/ExploreTabRouter.js
```diff
@@ -109,7 +109,7 @@ function buildExploreParamsFromQuery(query) {
   exploreParams.heatmapFit = queryParams.heatmapFit ? queryParams.heatmapFit : ''
   exploreParams.bamFileName = queryParams.bamFileName ? queryParams.bamFileName : ''
   exploreParams.ideogramFileId = queryParams.ideogramFileId ? queryParams.ideogramFileId : ''
-
+  exploreParams.reviewerSession = queryParams.reviewerSession ? queryParams.reviewerSession : ''
```
eweitz, line 112: Blocker: `reviewerSession` should be passed in at a lower level.  This would avoid complicating signatures of focused data retrieval functions (e.g. `fetchExplore`, `fetchCluster`) with a parameter needed by all of them.  Reading this from the query string in the lower-level `scpApi` function would achieve that.

eweitz (changes requested):
This feature will be quite useful!  With two changes -- one trivial, one less trivial -- this code will look good to me.

app/models/reviewer_access.rb
dinvlad: Awesome, this seems great!

app/javascript/components/explore/ExploreTabRouter.js
dinvlad: Btw, consider passing `reviewerSession` in a header, since query params are not ideal for auth parameters (as they can be cached, more easily tricked with for the user etc.)

bistline: I'd thought about this, but the problem is that the action beforehand is a form POST that results in a redirect.  As far as I know there's no way to set a custom header on a redirect in HTTP.  So it's a chicken & egg problem... I can't load the study page to then set headers on API requests that would handle the auth w/o the query string parameter, because it will fail all of our normal auth/permission challenges and redirect the reviewer away from the study page.  The query string param is the only mechanism I can use on the redirect.

dinvlad: Ahh gotcha. One alternative is to only use it in query for the redirect, and then in header for any other views (if you have any).

app/views/site/_study_settings_form.html.erb
```diff
@@ -111,7 +111,16 @@
         });
     </script>
   <% end %>
-
+  <div class="form-group">
+    <h3>Reviewer Access</h3>
+    <p class="help-block">
+      Use this form to set up anonymous "reviewer" access for allowing publication reviewers to view your study while it
+      is still private.  Reviewers can interact with any enabled visualizations, but cannot download data or view the
+      associated Terra workspace.  If enabled, you can send the Access URL & Pin code to your corresponding editor to
```
eweitz, line 119: "Pin" -> "PIN", etc.

```suggestion
      associated Terra workspace.  If enabled, you can send the access URL and PIN to your corresponding editor to
```

app/javascript/components/explore/ExploreTabRouter.js
dinvlad: Another alternative is to do use an `HttpOnly Secure Samesite=Strict` cookie (unless the redirect is on a different domain?)

app/views/site/_reviewer_access_fields.html.erb
eweitz, line 39:
```suggestion
          <strong>PIN:</strong>&nbsp;<%= f.object.pin %>
```

eweitz (approved):
I found two straggling user-facing instances of "Pin" -- a case-sensitive cross-file search for "Pin" would help find any others.  Beyond that, my local functional testing showed no issues.

I approve, assuming those "Pin" become "PIN" and AppSec eventually approves.

app/javascript/components/explore/ExploreTabRouter.js
dinvlad: Or if you keep it as the query param, then please add `Cache-Control: max-age=0; no-store` header to the response as well, so it won't be cached at least (although it might still be logged and/or prone to other issues)

bistline: I think I like that approach the best.  I'm a little leery of using a cookie for this, mainly because I'm not sure how standard the behavior will be across browsers/OS, and I also don't know what implications GDPR and our privacy policy might have on that.  But setting the `Cache-Control` header should be pretty easy.

dinvlad: Yea from the usability standpoint for sure. Each has its pros and cons, although the query param one has the immediate con that those URLs will be logged (although it only goes to "us" so not a big deal overall). We'll also know if there's any potential for abuse when we pentest it.

dinvlad: If possible though, I'd still use headers and implement redirect from JS instead..

app/models/reviewer_access.rb
```diff
@@ -0,0 +1,82 @@
...
+  field :expires_at, type: Date, default: 2.months.from_now.to_date
+
+  PIN_LENGTH = 10
+  # regex to strip all non-alphanumeric characters from a user-supplied pin value
+  PIN_SANITIZER = /(\W|_)/.freeze
+
+  belongs_to :study
+  has_many :reviewer_access_sessions, dependent: :delete_all do
+    def by_session_key(session_key)
+      find_by(session_key: session_key)
```
dinvlad, line 19: I _think_ we should validate `session_key` as well, unless `find_by` does it automatically.

app/javascript/components/explore/ExploreTabRouter.js
bistline: It's still going to have the same problem of no headers on the redirect, whether initiated server-side or from the client in JS.  It would need to be present as a query string parameter at least in order to load the main page, and we would only be able to use headers for visualization API calls.

I'll mull it over some more but I can't see a way around the main stumbling block.

app/models/reviewer_access.rb
bistline: It does minimally... `find_by` and other parameterized query helpers will escape any unsafe characters to prevent injection, but it won't validate the format or length.  I can easily add a regex that will ensure that the provided key matches the pattern for a UUID.

app/javascript/components/explore/ExploreTabRouter.js
dinvlad: Ah, you right about no custom headers on redirect. The only feasible (and maybe desirable) recommended [solution](https://portswigger.net/kb/issues/00500700_session-token-in-url) seems to be the cookie I mentioned.

app/models/reviewer_access.rb
bistline: 259d81c

dinvlad: Sounds good, either that or maybe there's a ready-to-go `is_uuid` method. I was wary of introducing other vulns via regex itself 😄

app/javascript/components/explore/ExploreTabRouter.js
dinvlad: Cookies could also hurt usability in case they access multiple documents, unless you set cookie `Path=/some/document/{doc_id}` perhaps.

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -7,18 +7,23 @@ module StudyAware
...
         ##
         def check_study_view_permission
-          if !@study.public? && !api_user_signed_in?
-            head 401
-          else
-            head 403 unless @study.public? || @study.can_view?(current_api_user)
+          unless @study.public?
+            if !api_user_signed_in? && params[:reviewerSession].present?
+              reviewer = @study.reviewer_access
+              head 401 if reviewer.nil? || reviewer.expired? || !reviewer.session_valid?(params[:reviewerSession])
```
dinvlad, line 21: I have another question here: is `reviewerSession` _tied_ to a particular study? I.e. if I pass an otherwise "valid" (non-expired) session in a request for an unrelated study, will it "reject" the request? It might be helpful to also see how such a URL would look like (i.e. does it contain the study ID which is not "verified" simultaneously with the session).

bistline: In short - yes, it will reject it.  A `reviewerSession` param is tied to the `Study` through the `ReviewerAccess` object, and then each `ReviewerAccess` has many `ReviewerAccessSession` objects.  The url looks like this for the ExploreController, for instance:

```
[:hostname]/single_cell/api/v1/studies/[:id]/explore?reviewerSession=[:session_key]
```

In this case, `:id` is most usually an `:accession` ('SCP1') which is used to load the study.  This happens via `:set_study` before `:check_study_view_permissions` even fires.  Once loaded, then if the `:reviewerSession` param is present, _and_ there is no authenticated user, then it will be used to validate view permission.  The order of operations for API calls is this:

```
before_action :set_study
```

app/models/reviewer_access.rb
bistline: There is not, unfortunately.  I looked for a while.

app/javascript/components/explore/ExploreTabRouter.js
bistline: I was able to do some validation and the cookie looks like it will work.  I'm going to start porting this over to reading that, making sure to use a secure/strict cookie with the path containing the study path/accession.

dinvlad: Awesome, thanks!

app/controllers/api/v1/concerns/study_aware.rb
dinvlad: Sounds good! We'd still need to pentest this, of course, just to verify there're no access control issues (as we've done for other projects).

app/javascript/components/explore/ExploreTabRouter.js
bistline: @dinvlad there will be an issue with the cookie path.  Since this needs to be used in multiple API requests, each of which has different paths, there's no single path value I can write that will apply to all.  It will have to be blank, otherwise the cookie won't be there at all.

dinvlad: OK, then maybe cookie won't be a good solution after all, due to the issue I mentioned (accessing multiple documents). So maybe we should just keep it as a query param for now, given that you've already taken a good remediation measure of limiting its lifetime, so the risk of its leakage is minimized at least.

eweitz (approved):

dinvlad (approved):

app/models/reviewer_access.rb
```diff
@@ -0,0 +1,94 @@
+# manage anonymous access to private studies via access code & pin
+# will generate a short-lived session for reviewers with valid credentials
+class ReviewerAccess
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include ActiveSupport::SecurityUtils
+
+  field :access_code, type: String # UUID
+  field :pin, type: String # 6 character string
```
eweitz, line 9: Length was changed from 6 to 10, so let's update this comment for consistency.

```suggestion
  field :pin, type: String # 10 character string
```

app/views/site/_reviewer_access_fields.html.erb
```diff
@@ -0,0 +1,69 @@
...
+          <%= radio_button_tag 'reviewer_opts[reset]', :yes, false %> Yes
+        </label>
+        <label class="btn btn-default active">
+          <%= radio_button_tag 'reviewer_opts[reset]', :no, true %> No
+        </label>
+      </div>
+    </div>
+    <div class="col-md-6 left-border-0-5-gray">
+      <h4>Current Access Credentials <a id="copy-reviewer-creds" class="btn btn-default btn-sm btn-copy"
+                                             data-toggle="tooltip" title="" data-original-title="Show pin & copy to clipboard">
```
eweitz, line 32: Straggler:

```suggestion
                                             data-toggle="tooltip" title="" data-original-title="Show PIN & copy to clipboard">
```

bistline: Gah - thanks!

app/controllers/api/v1/concerns/api_caching.rb
```diff
@@ -4,13 +4,6 @@ module Concerns
       module ApiCaching
         extend ActiveSupport::Concern
```
devonbush, line 6: Were these just lingering no-longer used parameters?  (I'd check myself, but I don't have a coding laptop right now... grrrr...)

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -7,18 +7,25 @@ module StudyAware
...
 
         ##
         # Permission checks
         ##
         def check_study_view_permission
-          if !@study.public? && !api_user_signed_in?
-            head 401
-          else
-            head 403 unless @study.public? || @study.can_view?(current_api_user)
+          unless @study.public?
```
devonbush, line 18: nice, I like how you consolidated the .public? check

app/controllers/site_controller.rb
```diff
@@ -128,6 +130,17 @@ def update_study_settings
         set_study_permissions(@study.detached?)
         set_study_default_options
         set_study_download_options
+
+        # manage reviewer access
+        reviewer_opts = params.to_unsafe_hash['reviewer_opts']
```
devonbush, line 135: this hash seems less like an options hash, and more like actions.  so maybe rename to `reviewer_access_actions`?

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -40,7 +47,7 @@ def check_study_compute_permission
 
         def check_study_detached
           if @study.detached?
-            head 410 and return
+            head 410
```
devonbush, line 50: same about the `and return`

app/controllers/api/v1/concerns/study_aware.rb
```diff
@@ -7,18 +7,25 @@ module StudyAware
         def set_study
           @study = Study.any_of({accession: params[:study_id]},{id: params[:study_id]}).first
           if @study.nil? || @study.queued_for_deletion?
-            head 404 and return
+            head 404
```
devonbush, line 10: I'd be ok keeping the and return. It's superfluous currently, but it's a nice guard to have in case someone in the future adds any logic after the if

app/controllers/site_controller.rb
```diff
@@ -714,13 +758,22 @@ def set_study_download_options
     end
   end
 
+  def set_reviewer_access
+    @reviewer_access = ReviewerAccess.find_by(access_code: params[:access_code])
+    unless @reviewer_access.present?
+      redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]),
+                  alert: 'Invalid access code; please check the link and try again.' and return
```
devonbush, line 765: yeah, I notice we're keeping the `and return` here, which I like

app/controllers/site_controller.rb
```diff
@@ -733,17 +786,31 @@ def download_acceptance_params
...
+          alert = 'The review period for this study has expired.'
+          redirect_to merge_default_redirect_params(site_path, scpbr: params[:scpbr]), alert: alert and return
+        elsif session_key.blank? # no cookie present, so this may or may not be a reviewer
+          authenticate_user!
+        elsif !reviewer.session_valid?(session_key) # check session cookie for expiry
+          alert = 'Your review session has expired - please create a new session to continue.'
+          redirect_to merge_default_redirect_params(reviewer_access_path(access_code: reviewer.access_code),
+                                                    scpbr: params[:scpbr]), alert: alert and return
+        end
+      elsif !user_signed_in? && !@study.public?
```
devonbush, line 809: I think the study.public? check here is redundant

app/javascript/lib/scp-api.js
```diff
@@ -254,7 +254,8 @@ export async function fetchClusterOptions(studyAccession, mock=false) {
  */
 export async function fetchCluster({
   studyAccession, cluster, annotation, subsample, consensus, genes=null,
-  isAnnotatedScatter=null, isCorrelatedScatter=null, fields=[], mock=false
+  isAnnotatedScatter=null, isCorrelatedScatter=null, fields=[],  mock=false
+
```
devonbush, line 258: extra newline
```suggestion
```

app/javascript/lib/scp-api.js
```diff
@@ -340,7 +341,7 @@ export async function fetchExpressionViolin(
     annotation_scope: annotationScope,
     subsample,
     consensus,
-    genes: geneString
+    genes: geneString,
```
devonbush, line 344: trailing comma

app/models/reviewer_access.rb
```diff
@@ -0,0 +1,94 @@
...
+  field :expires_at, type: Date, default: 2.months.from_now.to_date
+
+  PIN_LENGTH = 10
+  # regex to strip all non-alphanumeric characters from a user-supplied pin value
+  PIN_SANITIZER = /(\W|_)/.freeze
+  # regex to validate a session_key is a UUID
+  UUID_REGEX = /^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/.freeze
```
devonbush, line 16: seems like there should be a ruby helper method for this

app/views/site/_reviewer_access_fields.html.erb
```diff
@@ -0,0 +1,69 @@
+<div class="well well-sm" id="reviewer-access-controls">
+  <div class="form-group row">
+    <div class="col-md-2">
+      <% access_enabled = f.object.persisted? %>
+      <%= label_tag :_destroy, 'Access Enabled?' %><br />
+      <div class="btn-group btn-toggle" data-toggle="buttons">
+        <label class="btn btn-default <%= access_enabled ? 'active' : nil %>">
+          <%= radio_button_tag 'reviewer_opts[enable]', :yes, access_enabled, disabled: !access_enabled %> Yes
```
devonbush, line 8: see previous comment about renaming this "reviewer_access_actions" or somesuch

devonbush (approved):
Looks good, and really appreciate the thoughtful discussion on this PR already.   Looking forward to trying it out on staging.  A bunch of comments, all in the realm of style preferences, but nothing blocking

app/controllers/api/v1/concerns/api_caching.rb
bistline: Yes - it all got moved to `RequestUtils` a while back.

app/controllers/site_controller.rb
bistline: Fair point.

app/controllers/api/v1/concerns/study_aware.rb
bistline: Rubocop keeps barking at me about it, so I did a little checking.  Calling `head` is analogous to calling `render`, which implicitly ends the action.  Any code after this statement will not execute.  We only need the `and return` when using `redirect_to`, as it [does not halt execution](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to).

app/controllers/site_controller.rb
bistline: See the comment about `head` - it is necessary here, but not with `head`.

app/controllers/site_controller.rb
bistline: You are correct - we've already checked for it.

app/models/reviewer_access.rb
bistline: Found the `uuid` gem to do this (it looked like it wasn't maintained, but if we only use it for `UUID.validate()` it should be fine).

app/views/site/_reviewer_access_fields.html.erb
bistline: 446268b

app/models/reviewer_access.rb
bistline: 446268b

app/javascript/lib/scp-api.js
bistline: 446268b

app/javascript/lib/scp-api.js
bistline: 446268b

app/controllers/site_controller.rb
bistline: 446268b

app/controllers/site_controller.rb
bistline: 446268b

app/views/site/_reviewer_access_fields.html.erb
bistline: 446268b

app/models/reviewer_access.rb
bistline: 446268b

PR #1121, opened 2021-08-17
Add per-label correlations for two genes (SCP-3565, SCP-3566, SCP-3591)

eweitz:
This calculates and shows correlations for each label in a selected annotation.  Correlations _within_ each annotation label -- e.g. within mast cells, within pneumocytes --  are [often more interesting](https://docs.google.com/document/d/1yPmg52pA5uRt0N6twLMas-WU5DuIkeoqvwdTyjegTjE/edit#heading=h.wmvjl9b20v8) than _across_ such labels as shown now.  This extends and refines infrastructure from #1052 towards that new end.

These new per-label correlations are shown beside each legend entry.  Performance analytics are now also logged.

Although this new feature has automated tests, given the proof of concept nature of this work, it is hidden by default behind a feature flag.  Future work includes enabling users to omit cells with zero expression, and showing adjusted p-values to e.g. warn about strong but statistically insignificant correlations.  See [internal discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1628793379039000) for more background.

Here's how it looks!

https://user-images.githubusercontent.com/1334561/129777141-61730caf-b40f-4bd5-b70c-1e7c050eb6ee.mov

https://user-images.githubusercontent.com/1334561/129777152-029b05de-752a-4c89-bda4-953d38cb7424.mov

(Per [scientific feedback](https://github.com/broadinstitute/single_cell_portal_core/pull/1121#pullrequestreview-733079518), these now omit subsampling note in plot title.)

To test:
1.  Do migration: `run ruby rails_local_setup.rb`, then `source config/secrets/.source_env.bash`, then `rails db:migrate`
2.  Sign in
3.  Open upper right nav menu -> Admin Controls -> User Roles (https://localhost:3000/single_cell/admin#users)
4.  Find yourself, set `correlation_refinements` feature flag to "True" and `raw_counts_required_backend` to "False"
5.  `rails c`
7. Populate synthetic study: SyntheticStudyPopulator.populate('monkey_correlations')
8.  Go to study
9.  Search ADCY5 and AGPAT2
10.  In "Annotation" menu, select "Category"
11. Verify that bulk correlation above plot is "Spearman rho = 0.833"
12. Verify the per-label correlations in legend are -0.18 for A and 0 for B

This satisfies SCP-3565, SCP-3566, and SCP-3591.


bistline (approved):
Looks good!

Small aside: I ran into an issue initially using `SyntheticStudyPopulator` because I had `raw_counts_required_backend` turned on for my user account.  You should probably make a note in the testing instructions to also turn that off, otherwise it will cause the expression matrix to fail saving.  This is also something we may want to address in `SyntheticStudyPopulator` more directly.

jlchang (changes requested):
Noticing the correlation plot titles indicate a subsampling level. The correlation calculation is done on all cells in the clustering so we should remove the inaccurate subsampling info.

Happily, I was able to run the testing instructions locally and validate the testing (the synthetic study doesn't allow for scientific evaluation - did what I could via the great videoclips!)

jlchang (approved):
Thanks for adjusting the correlation plot title!

Looking forward to seeing this in prod (albeit still feature flagged)

PR #1129, opened 2021-08-27
adding branding groups list page (SCP-3595, SCP-3596)

devonbush:
SCP-3595, SCP-3596 
As shown in demo, this adds a new "Collections" menu to the top nav.  
image
The menu shows all public collections as well as any private collections the user has access to.  The menu option is excluded from the top nav if no collections are visible to the user, so we can release this functionality even if there are no public collections without confusion.

The "directory" page shows more details about each branding group, and also gives edit controls if the user has access
image

This also updates the branding groups edit page to have a 'publish' checkbox.  I may have gone overboard in updating the look of the whole page to a linear form style, but I think it's appropriate for a page like this that may need a lot of explanation/attention for each field, and that we expect users to fill out more-or-less in order.
image

Note that this does NOT  replace "BrandingGroup" with Collection throughout the codebase--only in the UX.  That's a change requiring migration scripts that's not necessarily worth it, especially before we've had a month or so to confirm that 'Collection' is the best name.

TO TEST:
0. Make sure you have a couple of branding groups locally (either create through the UX or SyntheticBrandingGroupPopulator).  Edit one of them to have public: true
1.  Load the home page not signed in.  Confirm the 'Collections' menu appears, and contains the directory and the public collection
2. Go to the directory page, and confirm you see the public collection
3. Sign in as an admin
4. Confirm you now see all collections in the menu and directory
5. Use the edit page to toggle a collection published/unpublished, and confirm it saves, and the directory is updated accordingly

### DIFF
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -11,7 +11,7 @@ class BrandingGroupsController < ApplicationController
     @branding_groups = BrandingGroup.all
   end
 
-  # show a non-editable list for display and linking
+  # show a list for display and linking, editable only if the user has appropriate permissions
   def list_navigate
     @editable_branding_groups = []
     @branding_groups = BrandingGroup.visible_groups_to_user(current_user)
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -1,6 +1,6 @@
 class BrandingGroupsController < ApplicationController
   before_action :set_branding_group, only: [:show, :edit, :update, :destroy]
-  before_action do
+  before_action except: [:list_navigate] do
     authenticate_user!
     authenticate_admin
   end
@@ -9,6 +9,15 @@ class BrandingGroupsController < ApplicationController
   # GET /branding_groups.json
   def index
     @branding_groups = BrandingGroup.all
+  end
+
+  # show a non-editable list for display and linking
+  def list_navigate
+    @editable_branding_groups = []
+    @branding_groups = BrandingGroup.visible_groups_to_user(current_user)
+    if current_user.present?
+      @editable_branding_groups = current_user.available_branding_groups.to_a
+    end
   end
 
   # GET /branding_groups/1
@@ -80,7 +89,7 @@ class BrandingGroupsController < ApplicationController
 
   # Never trust parameters from the scary internet, only allow the permit list through.
   def branding_group_params
-    params.require(:branding_group).permit(:name, :tag_line, :background_color, :font_family, :font_color, :user_id,
+    params.require(:branding_group).permit(:name, :tag_line, :public, :background_color, :font_family, :font_color, :user_id,
                                            :splash_image, :banner_image, :footer_image, :external_link_url, :external_link_description)
   end
 end
--- a/app/models/branding_group.rb
+++ b/app/models/branding_group.rb
@@ -12,6 +12,7 @@ class BrandingGroup
   field :feature_flags, type: Hash, default: {}
   field :external_link_url, type: String
   field :external_link_description, type: String
+  field :public, type: Boolean, default: false
 
   # list of facets to show for this branding group (will restrict to only provided identifiers, if present)
   field :facet_list, type: Array, default: []
@@ -76,4 +77,13 @@ class BrandingGroup
   def remove_cached_images
     UserAssetService.remove_assets_from_remote("branding_groups/#{self.id}")
   end
+
+  def self.visible_groups_to_user(user)
+    if user.present?
+      user.visible_branding_groups
+    else
+      BrandingGroup.where(public: true).order_by(:name.asc)
+    end
+  end
+
 end
--- a/app/views/branding_groups/_form.html.erb
+++ b/app/views/branding_groups/_form.html.erb
@@ -11,74 +11,79 @@
     </div>
   <% end %>
 
-  <div class="form-group row">
-    <div class="col-md-6">
-      <%= f.label :name %><br>
-      <%= f.text_field :name, class: 'form-control' %>
-    </div>
-    <div class="col-md-6">
-      <%= f.label :tag_line, 'Tag Line' %><br>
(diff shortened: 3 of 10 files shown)


app/views/branding_groups/_form.html.erb
```diff
@@ -11,74 +11,79 @@
...
+  </div>
+  <div class="form-group">
+    <%= f.label :footer_image, 'Footer Image (suggested size: 185x54)' %><br/>
+    <%= f.file_field :footer_image, class: 'btn btn-default fileinput-button' %>
+    <% if @branding_group.footer_image.present? %>
+      <p class="help-block">Current file: <%= link_to "<i class='fas fa-download'></i> #{@branding_group.footer_image_file_name} (#{ number_to_human_size @branding_group.footer_image_file_size})".html_safe,
+                                                      @branding_group.footer_image.url, download: @branding_group.footer_image_file_name,
+                                                      class: 'btn btn-xs btn-primary' %></p>
+    <% end %>
+  </div>
```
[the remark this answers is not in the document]
bistline: You could do it inline without the partial but I don't think it's any better/more readable:
```
<% 
[
  {
     attribute: :banner_image,
     label: 'Banner Image (suggested size: 1273x245)'
  },
  {
     attribute: :splash_image,
     label: 'Splash Image (suggested size: 383x89)'
  },
  {
     attribute: :footer_image,
     label: 'Footer Image (suggested size: 185x54))'
  }
].each do |field| %>
  <div class="form-group">
    <%= f.label field[:attribute], field[:label] %><br/>
    <%= f.file_field field[:attribute], class: 'btn btn-default fileinput-button' %>
    <% image = @branding_group.send(field[:attribute]) %>
    <% if image.present? %>
      <p class="help-block">Current file: <%= link_to "<i class='fas fa-download'></i> #{@branding_group.send("#{field[:attribute]}_file_name")} (#{ number_to_human_size @branding_group.send("#{field[:attribute]}_file_size")})".html_safe,
                                                      image.url, download: @branding_group.send("#{field[:attribute]}_file_name"),
                                                      class: 'btn btn-xs btn-primary' %></p>
    <% end %>
  </div>
<% end %>
```

app/models/branding_group.rb
```diff
@@ -12,6 +12,7 @@ class BrandingGroup
   field :feature_flags, type: Hash, default: {}
   field :external_link_url, type: String
   field :external_link_description, type: String
+  field :public, type: Boolean, default: false
```
bistline, line 15: Not a blocker, but I think it would be good to create a migration to backfill the data for `public`.  Mongoid is weird with "default" values and pre-existing documents.  You can load the document and call the field and you will get the default value if it hasn't been set, but you cannot query on the value as it _has not been set in the document_.  It is provided by the ODM after the document is loaded and instantiated as an instance of the model.

I'm not calling it a blocker since what we really care about are branding groups where `public` is `true`, and that will only happen if the owner specifically sets that.  Also, new branding groups will get that field populated when saved for the first time.  But if we ever want to go back and find it where `public` is explicitly `false`, we'll have an issue.

app/controllers/branding_groups_controller.rb
```diff
@@ -11,6 +11,15 @@ def index
     @branding_groups = BrandingGroup.all
   end
 
+  # show a list for display and linking, editable only if the user has appropriate permissions
+  def list_navigate
+    @editable_branding_groups = []
+    @branding_groups = BrandingGroup.visible_groups_to_user(current_user)
+    if current_user.present?
+      @editable_branding_groups = current_user.available_branding_groups.to_a
```
bistline, line 19: Is there a reason we're specifically casting this to an array with `.to_a`?  The `order_by` method returns a `Mongoid::Criteria` which will respond to all the same iteration methods as an `Array` (as they are descendants of `Enumerable`).  Just curious.

app/views/layouts/_nav.html.erb
```diff
@@ -18,6 +19,20 @@
   <div class="collapse navbar-collapse" id="scp-navbar-dropdown-collapse">
     <%= render partial: '/layouts/breadcrumbs' %>
     <ul class="nav navbar-nav pull-right">
+      <li>
+        <%= scp_link_to "<span class='fas fa-plus fa-fw'></span> Create a Study".html_safe, new_study_path, class: 'check-upload', 'data-analytics-name' => 'prominent-create-study-button' %>
+      </li>
+      <% if BrandingGroup.visible_groups_to_user(current_user).any? %>
+        <li class='dropdown'><%= scp_link_to "Collections <span class='caret'></span>".html_safe, "#", class: "dropdown-toggle", "data-toggle" => "dropdown" %>
+          <ul class="dropdown-menu dropdown-menu-left">
+            <li class="left-border-0-5"><%= link_to '<span class="fas fa-globe"></span> Directory'.html_safe, collections_path %></li>
```
bistline, line 28: Just me, but I'm not crazy about the globe as the icon here.  It doesn't really convey a meaning to me that seems to indicate what this page is going to be.  Would be curious to see other options.

bistline (approved):
No blockers, just some questions/comments.

app/views/layouts/_nav.html.erb
devonbush: concur -- maybe some sort of list icon?

app/controllers/branding_groups_controller.rb
devonbush: The idea was to make sure the function always returned the same 'type' of object regardless of user being signed in or not -- either an empty array, or a full array.  You're right that since they share most of the key interaction methods, it would have been ok to return the criteria also

app/models/branding_group.rb
devonbush: ah, that's a great point -- will update

app/views/branding_groups/_form.html.erb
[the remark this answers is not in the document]
devonbush: oh, that's interesting -- I hadn't thought of that method.  You're right though that it's not particularly cleaner.  If there were 5+ of these, then that would be a strong choice

PR #1131, opened 2021-08-29
Update handling of term match text highlighting (SCP-2877)

ehanna4:
Previously a keyword search would replace the words in the text of the study with the exact search word for highlighting purposes. So if a user searched for "mOuSe" then in the study the word "mouse" would be replaced with the highlighted text "mOuSe".

This fixes the search so that the formatting/capitalization of the text in a study remains regardless of how the user typed the word.

This also removes partial word matching so if a user searches on "to" only the word "to" will result in a match not words like "on**_to_**logy" which would have previously matched. 

New keyword search respects study description/title capitalization:
Screen Shot 2021-08-30 at 2 21 07 PM

Old Style:
Screen Shot 2021-08-28 at 11 05 26 PM

New style doesn't partial match:
Screen Shot 2021-08-30 at 2 22 35 PM

Old matched partial:
Screen Shot 2021-08-28 at 11 19 33 PM

### DIFF
--- a/app/javascript/components/search/results/StudySearchResult.js
+++ b/app/javascript/components/search/results/StudySearchResult.js
@@ -15,15 +15,26 @@ export function formatDescription(rawDescription, term) {
   return shortenDescription(textDescription, term)
 }
 
+// add highlighted style around matched words keeping original capitalization of text
+function highlightWords(text, term) {
+  const words = text.split(' ')
+  words.forEach(word => {
+    if (term.toUpperCase() === word.toUpperCase()) {
+      text = text.replace(word, `<span class='highlight'>${word}</span>`)
+    }
+  }
+  )
+  return text
+}
+
 export function highlightText(text, termMatches) {
   let matchedIndices = []
   if (termMatches) {
     matchedIndices = termMatches.map(term => text.indexOf(term))
   }
   if (matchedIndices.length > 0) {
-    termMatches.forEach((term, index) => {
-      const regex = RegExp(term, 'gi')
-      text = text.replace(regex, `<span class='highlight'>${term}</span>`)
+    termMatches.forEach(term => {
+      text = highlightWords(text, term)
     })
   }
   return { styledText: text, matchedIndices }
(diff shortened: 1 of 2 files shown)


bistline (approved):

eweitz (changes requested):
This will fix a definite UX papercut!

Blocker: it looks like your screenshot for "New style doesn't partial match" doesn't account non-initial instances of " to ", e.g. "shown to improve".  I haven't tested this locally, just assessing from the screenshot.

ehanna4:
> Blocker: it looks like your screenshot for "New style doesn't partial match" doesn't account non-initial instances of " to ", e.g. "shown to improve". I haven't tested this locally, just assessing from the screenshot.

Thank you! The logic wasn't completely right actually, thanks for pointing that out, should be fixed now

eweitz (approved):

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -15,16 +15,33 @@ export function formatDescription(rawDescription, term) {
...
+    let alreadyMatched = false
+    termMatches.forEach(term => {
+      if (term.toUpperCase() === word.toUpperCase()) {
+        stylizedText = `${stylizedText} <span class='highlight'>${word}</span>`
+        alreadyMatched = true
+      }
+    })
+    if (!alreadyMatched) {
+      stylizedText = `${stylizedText} ${word}`
+    }
```
devonbush, line 32: I think you could make this a little cleaner by doing the code below.  It also has the advantage of being a bit more robust to duplicate term matches (IDK if that's actually a concern though)
```suggestion
  let stylizedWord = word
  termMatches.forEach(term => {
    if (term.toUpperCase() === word.toUpperCase()) {
      stylizedWord = `<span class='highlight'>${word}</span>`
    }
  })
  stylizedText = `${stylizedText} ${stylizedWord}`
```

devonbush (approved):
code cleanup suggestion, but non-critical

PR #1133, opened 2021-08-31
Adding history tracking to study and study file

devonbush:
This adds the[ `mongoid-history`](https://github.com/mongoid/mongoid-history) gem for tracking changes to Study and StudyFile objects.  Once this change is out and we're comfortable with it, I suspect we will want to add tracking to most of our models (e.g. User, BrandingGroup, but not DataArray).   The tricky bit is identifying which fields to track, so you don't end up with audit tables that grow too fast.

This also updates the nightly admin email with a listing of studies created, deleted, and updated.  Right now the display is rough, high level stuff.  We can see how useful it is and expand later.  Currently it's best interpreted as 'here are some things you may want to look at'.  The audit history itself is often complex and there are lots of idiosyncracies (e.g. a user update to a study description results in two update logs, since we have an after_save hook that strips tags, and reporting the real name of a deleted study qould have required more time than I had to go through the logs and get the name prior to it being overwritten with DELETE-XXX).

image

 I regret running out of time to implement a 'modification history' tab in the portal study view, but the modification history can be relatively easily explored in the rails console in the meantime to answer more detailed questions.  e.g. `HistoryTracker.where(:created_at.lt => 1.hour.ago)` will get all changes within the last hour, and `Study.find_by(accession: 'SCPXXX').history_tracks` will get the history for a given study

This PR also includes a change to the way we handle emails in development mode.  Before, development mode did not perform email delivery.  Now, development mode does send emails, but restricts them to only be sent to the developer's email.  This makes testing of email capabilities easier, and may help in making developers aware of when emails are triggered.

One other minor change -- the 'File integrity check' is moved to the top of the report if a discrepancy is detected

TO TEST:
 0. check out change, run bundle install
 1. Do some stuff to studies, update, delete, create, etc...
 2. In a rails console, run `SingleCellMailer.nightly_admin_report.deliver_now` 
 3. confirm email gets sent roughly reflecting what you did


eweitz (approved):
Cool stuff!  The code looks broadly good to me.  I'm interested to see how this new tool for time series analytics will:

* Update [growth charts for public studies with raw counts](https://docs.google.com/spreadsheets/d/1FSpP2XTrG9FqAqD9X-BHxkCZae9vxZA3cQLow8mn-bk/edit#gid=0), which would refine guidance for strategic goals like cross dataset search.

* Integrate with Mixpanel, our other big tool for time series analytics.  Both audit tables and Mixpanel seem promising and complementary for analyzing historical data for authenticated users, especially those that are also registered in Terra.

devonbush:
Test failure is an unrelated yarn test failure

app/lib/summary_stats_utils.rb
```diff
@@ -88,4 +92,104 @@ def self.ingest_run_count(start_date: Time.zone.today, end_date: Time.zone.today
     end
     ingest_jobs
   end
+
+  # returns an array of hashes, each with title, accession, study_owner,
+  # one entry in the array for each study that has been deleted during the time frame
+  def self.deleted_studies_info(start_date: Time.zone.today, end_date: Time.zone.today + 1.day)
+    deletions = HistoryTracker.trackers_by_date(Study, action: 'destroy', start_time: start_date, end_time: end_date)
+    deletion_info = deletions.map do |tracker|
+      {
```
bistline, line 101: There are several places where we're assembling this info hash - it should be extracted into a helper method that can also take extra args (like `other_studies`, `update_info`, etc).  Passing them with a double splat (`**`) would allow you to dynamically merge in as many extra hash elements like so:

```
def format_study_info(tracker, **extra_info)
  {
    title: tracker.original['name'],
    accession: tracker.original['accession'],
    study_owner: User.find_by(id: tracker.original['user_id']).try(:email)
  }.merge(extra_info)
end

SummaryStatsUtils.format_study_info(study, foo: 'bar', bing: 'baz')
=> {:name=>"My Study", :accession=>"SCP1234", :study_owner=> 'fake@email.com', :foo=>"bar", :bing=>"baz"}
```

bistline (changes requested):
Some readability/maintainability enhancements, and a few questions.

app/lib/summary_stats_utils.rb
devonbush: ah, that's a good idea

app/models/study.rb
```diff
@@ -2005,4 +2006,10 @@ def ensure_cascade_on_associations
     AnalysisMetadatum.where(study_id: self.id).delete_all
     StudyFileBundle.where(study_id: self.id).delete_all
   end
+
+  # we aim to track all fields except fields that are auto-updated.
+  # modifier is set to nil because unfortunately we can't easily track the user who made certain changes
+  # the gem (Mongoid::Userstamp) mongoid-history recommends for doing that (which auto-sets the current_user as the modifier)
+  # does not seem to work with the latest versions of mongoid
+  track_history except: [:created_at, :updated_at, :view_count, :cell_count, :gene_count, :initialized, :data_dir], modifier_field: nil
```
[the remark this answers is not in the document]
devonbush: good point on `initialized`, I'll track that.  For description, maybe let's watch it for a bit and see how noisy it is just as a learning exercise?  then I totally could see dropping it like you say

app/lib/summary_stats_utils.rb
devonbush: bah, I gave this a try, the tricky bit is each method pulls the attributes from slightly different places, since, e.g. the delete info doesn't have an actual study object to use.  So refactoring it made the code more complex.

bistline: Oh well... worth a shot

app/lib/summary_stats_utils.rb
```diff
@@ -61,6 +61,10 @@ def self.disk_usage
   # since the "filter" parameter for list_project_operations doesn't work, check dates manually.
   # defaults to current day
   def self.ingest_run_count(start_date: Time.zone.today, end_date: Time.zone.today + 1.day)
+    # avoid hanging in development mode
+    if Rails.env.development?
```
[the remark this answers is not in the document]
bistline: Well, I'm happy to report "it works on my machine", but that just makes me worry more.  Can you run `ApplicationController.papi_client.list_pipelines` without error?  If so, then there's a hole in the logic for determining when we've looked at all jobs in the specified range, and we're getting an infinite loop.  That should be addressed if it's the case.

app/lib/summary_stats_utils.rb
```diff
@@ -88,4 +88,111 @@ def self.ingest_run_count(start_date: Time.zone.today, end_date: Time.zone.today
...
+      study_file_id = update.association_chain.first['id'].to_s
+      study_file = StudyFile.find_by(id: study_file_id)
+      study_id = nil
+      if study_file.present?
+        study_id = study_file.study_id.to_s
+      elsif update['action'] == 'destroy'
+        study_id = update['original']['study_id'].to_s
+      end
+      if study_id.present? && excluded_study_ids.to_s.exclude?(study_id)
+        updates_by_id[study_id] ||= {}
```
bistline, line 192: A possible refinement
```suggestion
        updates_by_id[study_id] ||= {'file updates' => 0}
        updates_by_id[study_id]['file updates'] += 1
```

config/environments/development.rb
```diff
@@ -86,6 +86,27 @@
     authentication:       'plain',
     enable_starttls_auto: true
   }
+  # prevent emails from going to anyone except the developer
+  class DeveloperMailInterceptor
+    # send all emails to the developer's git email address
+    def self.delivering_email(message)
+      email_target = nil
+      begin
+        `git config user.email`.strip
```
bistline, line 95: We lost the assignment here:
```suggestion
        email_target = `git config user.email`.strip
```

bistline (approved):
One tiny regression that isn't worth re-approval, and a small suggestion.  Otherwise looks good!  Excited to see it live.

devonbush:
Error is unrelated to this branch -- will open a ticket since it is reproducible on development

PR #1135, opened 2021-09-01
Fix StudySearchResults test regression (SCP3627)

ehanna4:
Had added an extra space at the beginning of study descriptions which the test picked up, so fixed that logic in the code.

Also also it seemed there was a bug in the logic for shortening descriptions for display in the search results.

Previously checking `textDescription.length > descriptionCharacterLimit` for deciding if the description needed truncating was done using the unstylized `textDescription` compared to the character limit. But the display text had been taken from `styledText` which would've included the span syntax which would actually get converted out at display time. So the case could come up where the `textDescription.length` was short enough to skip the truncating and suffixing but the `styledText` was too long so you'd end up with a description that had been shortened without the suffix of "...continued"

As seen here in prod when the study has been searched it gets shortened (with no suffix) but doesn't when just shown in non-searched results:
Screen Shot 2021-09-01 at 3 06 07 PM

Screen Shot 2021-09-01 at 3 06 22 PM

The change now compares the length of `styledText` to `descriptionCharacterLimit` so that the suffix and truncation are applied to the appropriate text:

Fixed local:
Screen Shot 2021-09-01 at 3 08 03 PM

dev on local:
Screen Shot 2021-09-01 at 3 08 47 PM

### DIFF
--- a/app/javascript/components/search/results/StudySearchResult.js
+++ b/app/javascript/components/search/results/StudySearchResult.js
@@ -20,13 +20,13 @@ function highlightWords(text, termMatches) {
   let stylizedText = ''
   const words = text.split(' ')
   words.forEach(word => {
-  let stylizedWord = word
-  termMatches.forEach(term => {
-    if (term.toUpperCase() === word.toUpperCase()) {
-      stylizedWord = `<span class='highlight'>${word}</span>`
-    }
-  })
-  stylizedText = `${stylizedText} ${stylizedWord}`
+    let stylizedWord = word
+    termMatches.forEach(term => {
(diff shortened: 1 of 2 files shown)


devonbush (approved):
whoops, so this was my 'suggested code' that had the bug.  PROTIP: always test Devon's suggestions before accepting them :)

bistline:
> whoops, so this was my 'suggested code' that had the bug. PROTIP: always test Devon's suggestions before accepting them :)

nopoints

bistline (approved):
Thanks for turning this around so quick!

PR #1136, opened 2021-09-01
Clarifying study embargo language & functionality (SCP-3630)

bistline:
A study can have an "embargo" date, which will prevent any authenticated users from downloading data from that study unless they have direct access to said study (either as the owner, or a share) until a specified date.  This is often used for a study that has a corresponding publication, and the journal does not want the data to become accessible until the article has been published.  A recent support ticket highlighted a discontinuity between the displayed language and the actual functionality - the messaging for this feature states that files are "Not available until [date]".  However, the logic for checking the whether the embargo was active is date-inclusive, meaning that the displayed date was the _last day of the emabargo_, rather than the first day of availability.

This update changes that logic to match the expected functionality.  Embargoes set by study owners will expire on the specified date.  Furthermore, messaging for this feature has been updated to use "on" (which is specific), rather than "until" (which is ambiguous).  Also, the field is now displayed in forms as "Data Release Date", rather than "No Data Download Until?", and the help tooltip specifies that the data will be made available starting at 12AM on the provided date.

MANUAL TESTING
1. Sign in as an admin and go to the study settings tab on any public study
2. Set the embargo date to today and save
3. Sign out, and sign in as a non-admin user (that doesn't have a share with the above study)
4. Load the study download tab, and validate that the embargo is not in effect, and files can be downloaded

This PR satisfies SCP-3630.

### DIFF
--- a/app/views/layouts/_download_link.html.erb
+++ b/app/views/layouts/_download_link.html.erb
@@ -1,5 +1,5 @@
 <% if @user_embargoed %>
-    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Data embargo expires on <%= @study.embargo.to_s(:long) %></span>
+    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Data Release Date: <%= @study.embargo.to_s(:long) %></span>
 <% else %>
   <% if study_file.upload_file_name.nil? && study_file.human_data %>
       <%= link_to "<span class='fas fa-cloud-download'></span> External".html_safe, study_file.download_path, class: 'btn btn-primary', target: :_blank, data: {"analytics-name": 'file-download:study-single:external'} %>
--- a/app/views/layouts/_download_link.html.erb
+++ b/app/views/layouts/_download_link.html.erb
@@ -1,5 +1,5 @@
 <% if @user_embargoed %>
-    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Not available until <%= @study.embargo.to_s(:long) %></span>
+    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Data embargo expires on <%= @study.embargo.to_s(:long) %></span>
 <% else %>
   <% if study_file.upload_file_name.nil? && study_file.human_data %>
       <%= link_to "<span class='fas fa-cloud-download'></span> External".html_safe, study_file.download_path, class: 'btn btn-primary', target: :_blank, data: {"analytics-name": 'file-download:study-single:external'} %>
--- a/app/views/site/_study_settings_form.html.erb
+++ b/app/views/site/_study_settings_form.html.erb
@@ -19,7 +19,7 @@
   </div>
   <div class="form-group row">
     <div class="col-md-4">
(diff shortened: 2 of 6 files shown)


app/views/layouts/_download_link.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if @user_embargoed %>
-    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Not available until <%= @study.embargo.to_s(:long) %></span>
+    <span class="label label-warning embargoed-file"><i class="fas fa-ban"></i> Data embargo expires on <%= @study.embargo.to_s(:long) %></span>
```
jlchang, line 2: "expires on" still seems ambiguous to me. Would "Last date of data embargo" be any better?

(My pedantic spouse says "Expires <date> <time>" I suggested 12:01 am for <time>, he says 00:01 but lack of units is also odd). In any event, I'll go with whatever seems clear to most. My brain is now too alert to the issue to be useful. Sorry!

app/views/site/_study_settings_form.html.erb
```diff
@@ -19,7 +19,7 @@
   </div>
   <div class="form-group row">
     <div class="col-md-4">
-      <%= f.label :embargo, 'No Data Download Until?' %>&nbsp;<span class='fas fa-question-circle' data-toggle='tooltip' title='Prevent data file downloads until date specified below (leave blank to allow downloads)' data-placement='right'></span><br />
+      <%= f.label :embargo, 'Data Embargo Date' %>&nbsp;<span class='fas fa-question-circle' data-toggle='tooltip' title='Downloads will be prevented for everyone except study owners & shared users.  The embargo will expire on the date specified (leave blank to allow downloads)' data-placement='right'></span><br />
```
jlchang, line 22: Embargo is the act of banning (downloads in this case) so I believe we're asking for a "Data Embargo End Date" or "Data Release Date"?

app/views/studies/_form.html.erb
```diff
@@ -30,7 +30,7 @@
   </div>
   <div class="form-group row">
 		<div class="col-md-3">
-			<%= f.label :embargo, 'No Data Download Until?' %>&nbsp;<span class='fas fa-question-circle' data-toggle='tooltip' title='Prevent data file downloads until date specified below (leave blank to allow downloads)' data-placement='right'></span><br />
+			<%= f.label :embargo, 'Data Embargo Date' %>&nbsp;<span class='fas fa-question-circle' data-toggle='tooltip' title='Downloads will be prevented for everyone except study owners & shared users.  The embargo will expire on the date specified (leave blank to allow downloads)' data-placement='right'></span><br />
```
jlchang, line 33: See previous comment about "Data Embargo End Date"

jlchang (changes requested):
Non-blocking comment on "expires on".

For "Data Embargo Date", it may be clearer to ask for a "Data Embargo End Date", I can be convinced that its unnecessary but putting in as a requested change and happy to discuss :)

app/views/site/_study_settings_form.html.erb
bistline: I like `Data Release Date` - that seems much more explicit.

devonbush (changes requested):
Code changes look good, and I heartily agree with "Data Release Date" or "Release Data on" as the clearest language.  So I approve, pending the language update

jlchang (approved):
Manually tested locally and everything looks great!

PR #1137, opened 2021-09-02
Addressing UX issues with updates, required asterisk (SCP-3492)

bistline:
This update fixes small regressions with handling raw count associations when the corresponding feature flags are not enabled.

MANUAL TESTING
1. Turn off the `raw_counts_required_backend` feature flag for your account
2. Load the upload wizard for a study file that has both raw & processed matrix files
3. Confirm the asterisk for "associated raw count file" is gone
4. Confirm that updating the file (like adding a description) does not throw an error in the logs

This PR further satisfies SCP-3492.

### DIFF
--- a/app/views/studies/_expression_file_fields.html.erb
+++ b/app/views/studies/_expression_file_fields.html.erb
@@ -19,13 +19,13 @@
   <% current_values = f.object.associated_matrix_files(:raw)&.map { |sf| { label: sf.name, value: sf.id.to_s } } %>
 
   // render the raw counts association select; if replace is true, allows swapping out units dropdown in raw counts form
-  updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>)
+  updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>, <%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>)
 
   $('#study-file-<%= f.object.id.to_s %>').on('updateRawCountsSelect', function() {
-    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>)
+    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>, <%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>)
   })
 
   $('#study-file-<%= f.object.id.to_s %>').on('change', '.is_raw_counts_true, .is_raw_counts_false', function() {
-    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>)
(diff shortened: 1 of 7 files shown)


app/views/studies/_expression_file_fields.html.erb
```diff
@@ -19,13 +19,13 @@
...
 
   $('#study-file-<%= f.object.id.to_s %>').on('updateRawCountsSelect', function() {
-    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>)
+    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>, <%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>)
   })
 
   $('#study-file-<%= f.object.id.to_s %>').on('change', '.is_raw_counts_true, .is_raw_counts_false', function() {
-    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>)
+    updateRawCountsAssnSelect('#study-file-<%= f.object.id.to_s %>', <%= current_values.to_json.html_safe %>, <%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>)
   })
```
eweitz, line 30: For easier readability and maintainability, interpolated values for:

* `'#study-file-<%= f.object.id.to_s %>'`
* `<%= current_values.to_json.html_safe %>`
* `<%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>`

could each be assigned to respective JS variables, and reused.

app/views/studies/_initialize_expression_form.html.erb
```diff
@@ -102,7 +102,8 @@
       function handleRawCountsUpdate() {
         <% current_values = study_file.associated_matrix_files(:raw)&.map { |sf| { label: sf.name, value: sf.id.to_s } } %>
         const currentValues = <%= current_values.to_json.html_safe %>
-        updateRawCountsAssnSelect(expFormId_<%= allow_only %>, currentValues)
+        const isRequired = <%= User.feature_flag_for_instance(current_user, 'raw_counts_required_backend') %>
+        updateRawCountsAssnSelect(expFormId_<%= allow_only %>, currentValues, isRequired)
```
eweitz, line 106: The issue precedes this PR, but interpolating variable _names_ seems even less easily maintainable / debuggable than interpolating variable values.

eweitz (approved):
Code looks good!  I noted some nits, nothing blocking.

devonbush (approved):
agree with Eric's suggestions, and also that they are not essential for this PR

app/views/studies/_expression_file_fields.html.erb
bistline: I totally agree - that had been the original implementation, but for some reason it doesn't work.  The select inputs do not update without a page refresh, or re-rendering them by changing the `file_type` select.  I spent a while trying to debug it and decided that it was out-of-scope for this particular PR.  Given that we are actively talking about redoing many of these forms as a part of a larger refactoring, it didn't seem worth holding things up over.

app/views/studies/_initialize_expression_form.html.erb
bistline: Again - totally agreed, but with the current infrastructure on this page, it's the only thing that works.  Otherwise everything errors out, and you can't even upload files due to numerous JS errors.  Really the only solution is the large-scale refactoring we've been discussing.

PR #1139, opened 2021-09-02
Browse collections functionality (SCP-3595, SCP-3596)

devonbush:
SCP-3595, SCP-3596. This includes some work from a [now-closed PR](https://github.com/broadinstitute/single_cell_portal_core/pull/1129), but the functionality has changed enough that I thought it warranted a new PR.  This replaces the COVID-19 button with a 'Browse Collections' button that goes to the Collections list page I demoed previously.  The COVID-19 space is now listed there as a 'collection'.  

image

image

This removes the Covid-19 feature flag as it is no longer necessary

Note this work does not actually create a Covid 'collection.', that work is ticketed as SCP-3633.   See also SCP-3634 and SCP-3635 for related work. 

TO TEST:
 1. Go to home page, confirm "Browse collections" button appears
 2. Click it, and confirm it takes you to collections list page
 3. Confirm collections list page shows the covid-19 page, and any other collections that the current user has access to
 4. Confirm links on the collections page take you to the desired places
 5. Sign in as an admin
 6. Confirm you now see all collections in the directory
 7. Use the edit page to toggle a collection published/unpublished, and confirm it saves, and the directory is updated accordingly


app/models/user.rb
```diff
@@ -318,9 +318,21 @@ def email_as_id
...
+    else
+      BrandingGroup.where(user_id: self.id).order_by(:name.asc)
+    end
+  end
+
+  def visible_branding_groups
+    if self.admin?
+      BrandingGroup.all.order_by(:name.asc)
+    else
+      BrandingGroup.where(user_id: self.id).or(public: true).order_by(:name.asc)
```
bistline, line 334: Blocker: chaining these queries together will not have the intended outcome.  This is going to evaluate to "all branding groups owned by the user that ARE ALSO public".  If what you're looking to do is "give me all branding groups that are public OR owned by the user", then it will need to be written as:
```
BrandingGroup.or(public: true, user_id: self.id).order_by(:name.asc)
```
You can also use `any_of` as it is an alias for `or`, which I personally find a little more readable.  To make this a little DRYer, I'd write it as:

```
selector = self.admin? ? BrandingGroup.all : BrandingGroup.any_of(public: true, user_id: self.id)
selector.order_by(name: :asc)
```

bistline (changes requested):
Looks good, modulo the edit to `visible_branding_groups` for a user.

app/models/user.rb
devonbush: I'm pretty sure this is how `or` works. See `https://www.bigbinary.com/blog/rails-5-adds-or-support-in-active-record`   Locally, you can try something like this to confirm the behavior:
```2.6.6 :022 > Study.where(accession:'SCP41').or(accession: 'SCP29').count
 => 2 
```

app/models/branding_group.rb
```diff
@@ -76,4 +77,13 @@ def remove_branding_association
   def remove_cached_images
     UserAssetService.remove_assets_from_remote("branding_groups/#{self.id}")
   end
+
+  def self.visible_groups_to_user(user)
+    if user.present?
```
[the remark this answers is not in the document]
devonbush: I appreciate that the way you've constructed it is more Ruby-y.  I've started relying on 'truthiness' of objects less lately because I found I was writing a lot of bugs where I'd forget that e.g. an empty array is false-y but an empty hash is truth-y, or that 0 is truthy in Ruby but false-y in JS.  But that's just my senility approaching...

As for the ternary, I've started avoid ternaries except for trivial assignment is done, b/c it makes it much easier to put debug statements in and check which way the branch is going.  e.g. 
```
 foo = user.admin? ? bar : baz. 
```
works well for me, but for 
```
foo = user.admin? ? bar : Study.where(....)
```
I'd prefer the query on a separate line so it's easier to isolate.  But like you said, these are just preferences

app/views/site/index.html.erb
```diff
@@ -3,23 +3,12 @@
...
-                    padding: 10px 15px;
-                    font-size: 16.8px;
-                    display: inline-block">
-            COVID-19 Studies
+      <% if @selected_branding_group.nil? %>
+        <div class="home-nav" id="collections-tab"
+            style="float:right;">
+          <a href="/single_cell/collections" class="btn btn-primary"
+             style="display: inline-block; font-size:16px;">
+            Browse Collections <span class="fas fa-bacteria"></span><span class="fas-fa-bacteria"></span>
```
eweitz, line 11: I like the biology theme!  The detached-and-ciliated cells do seem a bit off, as the main focus of single-cell research is multicellular organisms.

Could you ask a designer for a more fitting icon?

eweitz (approved):
Code and screenshot look good!  I suggested a few non-blocking improvements.

app/views/branding_groups/_form.html.erb
```diff
@@ -11,74 +11,79 @@
...
-  <div class="row">
-    <div class="col-md-12">
-      <%= f.label :external_link_url, 'External Link URL' %>
-      <%= f.text_field :external_link_url, class: 'form-control' %>
-      <br/>
-      <%= f.label :external_link_description, 'External Link Description' %>
-      <%= f.text_field :external_link_description, class: 'form-control' %>
+
+  <div class="form-group">
+    <%= f.label :external_link_url, 'External Link URL <span class="fas fa-question-circle" data-toggle="tooltip" data-placement="top" title="This link will be available in the \'Help & Resources\' dropdown inside the collection, as well as in the Collections Directory."></span>'.html_safe %>
```
eweitz, line 33: Double quotes are mark conventional in user-facing copy for this scenario.

```suggestion
    <%= f.label :external_link_url, 'External Link URL <span class="fas fa-question-circle" data-toggle="tooltip" data-placement="top" title="This link will be available in the \"Help & Resources\" dropdown inside the collection, as well as in the Collections Directory."></span>'.html_safe %>
```

app/views/site/index.html.erb
devonbush: good idea, I'll run it by Jerome once he's back

app/models/user.rb
bistline: Huh.  Learn something every day.  I'll approve based off of that since that was my only request.

bistline (approved):

PR #1140, opened 2021-09-02
Updating to rails 6.1.4.1 re: CVE-2021-22942 (SCP-3641)

bistline:
Updates `rails` to `6.1.4.1` to address [CVE-2021-22942](https://github.com/broadinstitute/single_cell_portal_core/security/dependabot/Gemfile.lock/actionpack/open), which was identified in `actionpack`.  As this is a core dependency of `rails`, it was necessary to update `rails` rather than `actionpack`.  This is to alleviate a discovered vulnerability in using the `config.hosts` setting for preventing DNS rebinding/host header injection attacks.  While SCP is not directly vulnerable, as we use fully-qualified hostnames, this security alert is high enough severity that it should be done regardless.


eweitz (approved):
This is the first time I've realized that Ruby on Rails has a four-part version format.  Per its [maintenance policy](https://guides.rubyonrails.org/maintenance_policy.html), the uncommon fourth part is for security issues.

devonbush (approved):

PR #1145, opened 2021-09-07
Retiring obsolete 'react_explore' feature flag (SCP-3623)

bistline:
This update removes the obsolete `react_explore` feature flag, and all references to it.  The only places remaining were controlling the state of the spatial association select input, and also handling re-rendering the explore tab when updating study settings.

This update also adds a new class method to `FeatureFlag` to retire old flags from the database.  The `retire_feature_flag` method will delete the flag itself, as well as all model-based instances where this flag has been set.  This will cover _all_ SCP models that implement the `FeatureFlaggable` concern, both now and in the future.

MANUAL TESTING
1. Run `ruby_local_setup.rb && source config/secrets/.source_env.bash && rails db:migrate`
2. Boot portal as normal, including delayed_job, and sign in with an admin account
3. Load any study with visualizations enabled
4. Load the study settings tab, and update any default configuration (like point size, opacity, etc.)
5. Once the update completes, validate the explore tab has re-rendered with the update, and that there are no errors in the JS console
6. Load the Admin Config page, and edit the feature flags for your account
7. Note that the form renders w/o error, and that the `react_explore` entry is gone
8. Optional: in the Rails console, query for any user accounts that still have a value set for `feature_flags.react_explore` and confirm the return is `0`:
```
User.where(:'feature_flags.react_explore'.ne => nil).count
=> 0
```

This PR satisfies SCP-3623.

### DIFF
--- a/app/views/site/update_study_settings.js.erb
+++ b/app/views/site/update_study_settings.js.erb
@@ -8,11 +8,7 @@ closeModalSpinner('#update-study-settings-spinner', '#update-study-settings-moda
         $('#study-summary').html("<%= escape_javascript(render partial: 'study_description_view') %>");
         $('#study-download').html("<%= escape_javascript(render partial: 'study_download_data') %>");
         <% if @study.initialized? %>
-          <% if User.feature_flag_for_instance(current_user, 'react_explore')  %>
-            window.SCP.renderExploreView(document.getElementById('study-visualize'), window.SCP.studyAccession)
-          <% else %>
-            $('#study-visualize').html("<%= escape_javascript(render partial: 'study_visualize') %>");
-          <% end %>
+          window.SCP.renderExploreView(document.getElementById('study-visualize'), window.SCP.studyAccession)
         <% end %>
     <% else %>
         showMessageModal("", "An error occurred while saving: <br /><%= @study.errors.full_messages.join(', ').html_safe %>");
--- a/app/models/feature_flag.rb
+++ b/app/models/feature_flag.rb
@@ -21,4 +21,23 @@ class FeatureFlag
       hash.with_indifferent_access
     end
   end
+
+  # 'retire' a feature flag and remove all per-model flag values to prevent FeatureFlaggable#validate_feature_flags
+  # from throwing an error when trying to update existing feature flags
+  def self.retire_feature_flag(name)
+    feature_flag = FeatureFlag.find_by(name: name)
+    raise NameError, "FeatureFlag: '#{name}' does not exist" if feature_flag.blank?
+
+    # call Rails.application.eager_load! to ensure all classes are loaded in development
+    Rails.application.eager_load! if Rails.env.development?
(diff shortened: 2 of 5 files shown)


app/views/site/update_study_settings.js.erb
```diff
@@ -8,11 +8,7 @@ closeModalSpinner('#update-study-settings-spinner', '#update-study-settings-moda
         $('#study-summary').html("<%= escape_javascript(render partial: 'study_description_view') %>");
         $('#study-download').html("<%= escape_javascript(render partial: 'study_download_data') %>");
         <% if @study.initialized? %>
-          <% if User.feature_flag_for_instance(current_user, 'react_explore')  %>
-            window.SCP.renderExploreView(document.getElementById('study-visualize'), window.SCP.studyAccession)
-          <% else %>
-            $('#study-visualize').html("<%= escape_javascript(render partial: 'study_visualize') %>");
-          <% end %>
+          window.SCP.renderExploreView(document.getElementById('study-visualize'), window.SCP.studyAccession)
```
devonbush, line 11: can we now also delete that `study_visualize` partial?

devonbush (changes requested):
Looks good, love the retire helper method.  I think there's just a lingering partial that can also be deleted now

app/views/site/update_study_settings.js.erb
bistline: I don't think so... looking at the file, it has some functionality in it for IGV/Ideogram.  Likely it can get taken out, but would require refactoring that is out of scope for this particular PR, and already tracked in https://broadworkbench.atlassian.net/browse/SCP-3258.

devonbush: ok, then I'm confused on this change.  If that partial still has needed functionality beyond just calling `renderExploreView`, then shouldn't we be using it here instead of calling `renderExploreView` directly?  It seems like this file should match whatever we do in `study.html.erb`

bistline: I was correct, originally - we do still need that partial.  It has logic for setting state in regards to various things for the explore tab to render correctly on page load, in addition to the IGV- and Ideogram-specific code.  In [`study.html.erb`](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/views/site/study.html.erb#L12), we render the `_study_visualize` partial. 

Once we pick up SCP-3258, then I think we could move to remove this partial and make the explore tab 100% React.  But for now, we still need it (I even tried taking it out and calling `renderExploreView` directly but that doesn't work for a couple reasons).

devonbush (approved):
Thanks for the explanation of the code -- the edit-study-settings is only invoked after the study.html.erb is already rendered, which is why it doesn't need to use the full template

eweitz (approved):
Looks good!  `retire_feature_flag` seems convenient for upcoming clean-up.

PR #1147, opened 2021-09-08
Update defaults for selects /dropdowns on Expression Matrix files (SCP-3537)

ehanna4:
Adds an option of "Select One..." to the select for when an option has not been chosen for the fields `library_preparation_protocol`, `biosample_input_type` and `modality` .

This can be seen for `library_preparation_protocol` for new file uploads.

Also add the defaults for new files for these two fields: 
- `biosample_input_type`: "whole cell"
- `modality`: "transcriptomics: unbiased"

(The ticket requests modality: "transcriptomics: non-targeted", however this is equivalent to "transcriptomics: unbiased" in our current data model)

There are legacy studies with expression files from before SCP required: ` biosample_input_type`, `modality` and `library_preparation_protocol` that have nothing selected for these fields. Currently if a user goes to the upload/update files page the current UI displays a default in the select for each of these fields. This can be confusing for a user if they had not chosen a value in one of these already uploaded legacy files. The new UI will have "select one..." for any of these three fields for an already uploaded file where the field hadn't been chosen. 

For new file uploads SCP will have a default for `modality` and `biosample_input_type` of the most commonly used options to save the user a couple of extra clicks. There is not a most commonly chosen `library_preparation_protocol` for all users so this will stay as "Select One...".

Since all three of these fields are required by SCP now the user will be unable to save a file without selecting an option from the dropdown in either case (updating an existing file or uploading a new one)

To view the change go to the "Chicken raw counts" study and go to Upload/Edit study data. The Raw Count Matrix file "raw2_chicken_40_cells_4_genes.raw_dense.txt" is an already uploaded file that had values chosen for the three fields in question.
Screen Shot 2021-09-08 at 9 53 09 AM

The Processed Matrix file, "raw2_chicken_40_cells_4_genes.processed_dense.txt" is an already uploaded file where the user had not chosen a value for any of those three fields. In the new UI the selects will be on "Select One.." whereas in the old UI they would be populated with default values the user had not chosen.
NEW:
Screen Shot 2021-09-08 at 10 20 52 AM

OLD:
Screen Shot 2021-09-08 at 10 19 36 AM

Lastly if you click to add an expression matrix file (either raw counts or processed) the user will be presented with a form where the Biosample Input Type and Modality have defaults chosen in thier selects.
Screen Shot 2021-09-08 at 9 57 48 AM

Attempting to save when the select is set to "Select One..." will result in the error message, prompting the user to chose a valid option:
Screen Shot 2021-09-08 at 9 58 13 AM


devonbush (approved):

eweitz (changes requested):
This will be a nice UX improvement!

I think we should use sentence case instead of title case in drop-down menus, in cases where SCP and not a user provides the text.  Sentence case is easier to read, more common in UIs, and Terra does similarly in this use case.  "[Making a case for letter case](https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98)" has a good summary of the trade-offs of the two styles.  Takeaways, and more on real-world company design guidelines, are below.

Title case:
* More symmetry
* More visual prominence
* More "gravitas"
* Used by:
  * [Facebook](https://www.facebook.com) and its subsidiary [Instagram](https://www.instagram.com)
  * [Apple](http://developer.apple.com/design)

Sentence case:
* Easier to read
* Easier to define
* Friendlier
* Easier to spot proper nouns
* Used by:
  * [Google](https://material.io/design)
  * [Microsoft](https://www.microsoft.com/en-us/) and its subsidiary GitHub
  * [Twitter](https://twitter.com/observablehq)
  * [Wikipedia](https://en.wikipedia.org/wiki/Main_Page)
  * [Pinterest](https://www.pinterest.com/)
  * [LinkedIn](https://www.linkedin.com/)
  * [Stack Overflow](https://stackoverflow.com/)
  * [Yahoo](https://www.yahoo.com/)

[Amazon](https://www.amazon.com/) and [eBay](https://www.ebay.com/) are nuanced, but tend to use sentence case for verb phrases like this.  [Netflix](https://www.netflix.com/) also varies, and tends to use title case.  Terra varies even more, but for default options in drop-down select menus it uses sentence case -- e.g. click the "+" button in https://app.terra.bio/#workspaces.

app/views/studies/_expression_file_info_fields.html.erb
```diff
@@ -27,20 +27,20 @@
     </div>
     <div class="col-lg-2">
       <%= f.label :biosample_input_type, label_with_asterisk('Biosample Input Type') %><br />
-      <%= f.select :biosample_input_type, options_for_select(ExpressionFileInfo::BIOSAMPLE_INPUT_TYPE_VALUES, f.object.biosample_input_type),
-                   {}, class: 'form-control' %>
+      <%= f.select :biosample_input_type, options_for_select(ExpressionFileInfo::BIOSAMPLE_INPUT_TYPE_VALUES, f.object.study_file.name.present? ? f.object.biosample_input_type : 'Whole cell'),
```
bistline, line 30: While this will have the intended outcome of specifying a default, it also means that the "default" logic lives in the form, and not in the data model itself.  Since we are actively considering refactoring the data upload/sync UX, this kind of compartmentalized logic will lead to regressions moving forward.

I would instead posit that it would be better to specify the defaults at the model level, rather than the UI level:

```
class ExpressionFileInfo

  ...
  field :library_preparation_protocol, type: String, default: 'Transcriptomic: unbiased'
  field :biosample_input_type, type: String, default: 'Whole cell'
  field :modality, type: String
  ...
```

This means that if we execute `study_file.build_expression_file_info`, the resulting document already has the default values set, and the form logic for all of these fields doesn't need to change (except for adding the "Select one" prompts as you have done).

bistline (changes requested):
I think we should specify the defaults at the model level, rather than the UI level.  This will be more portable, and be consistent with our other models in terms of default values.

app/views/studies/_expression_file_info_fields.html.erb
devonbush: I can see this going both ways.  Agree with your point, Jon about the model being a more central place to find defaults.  But having a default at the model level is dangerous because that can lead it to being accidentally set when we actually had no information provided, and would rather the API return an error.   Having a default value means we will have to be more careful in our controller code at making sure we're preserving nils/undefined that got passed in

bistline: > But having a default at the model level is dangerous because that can lead it to being accidentally set when we actually had no information provided

Fair point, but we won't have this scenario as all of these values have been required all along, so we don't have any documents where data hasn't been set.  This will only impact new files, and we want to prompt with these new defaults.  I think my previous points still hold.

eweitz (approved):

bistline (approved):

bistline:
It looks like there's a test regression from the default updates:
```
  1) Failure:
StudyFileTest#test_expression_file_data_validates [/home/app/webapp/test/models/study_file_test.rb:107]:
--- expected
+++ actual
@@ -1 +1 @@
-{:base=>["Units is not included in the list, Biosample input type is not included in the list, Modality is not included in the list, Library preparation protocol is not included in the list"]}
+{:base=>["Units is not included in the list, Library preparation protocol is not included in the list"]}
```

This isn't a regression, just the test looking for strict equality on the error message.

PR #1149, opened 2021-09-08
SCP-3595 Collections demo UX fixups

devonbush:
SCP-3595 - this addresses the comments from demo, notably around the collections list UX, moving text to top, aligning buttons, and getting rid of broken image tags.  This also does a more thorough job of replacing branding group with collection throught the UI

image

TO TEST:
0. from the home page, click 'browse collections'
1. confirm the collections browser appears as desired
2. As admin, click around the collections UX, and confirm no stray 'branding group' references remain

### DIFF
--- a/app/views/layouts/_nav.html.erb
+++ b/app/views/layouts/_nav.html.erb
@@ -40,7 +40,7 @@
               <li><%= link_to "<span class='fas fa-flask fa-fw'></span> Analyses".html_safe, analysis_configurations_path, class: 'check-upload', id: 'analysis-nav' %></li>
               <li><%= link_to "<span class='fas fa-search fa-fw'></span> Preset Searches".html_safe, preset_searches_path, class: 'check-upload', id: 'preset-nav' %></li>
               <li><%= link_to "<span class='fas fa-dna fa-fw'></span> Species".html_safe, taxons_path, class: 'check-upload', id: 'species-nav' %></li>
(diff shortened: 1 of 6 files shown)


app/views/layouts/_nav.html.erb
```diff
@@ -53,7 +53,7 @@
             <li><%= scp_link_to "<span class='fas fa-plus fa-fw'></span> Create a Study".html_safe, new_study_path, class: 'check-upload' %></li>
             <% if current_user.available_branding_groups.any? %>
               <li role="separator" class="divider"></li>
-              <li class="dropdown-header">My Brands</li>
+              <li class="dropdown-header">My Collections</li>
```
ehanna4, line 56: Is this PR to include updating all user facing language to use 'Collections' over 'Brands'?

bistline (approved):

app/views/layouts/_nav.html.erb
devonbush: yes, that's the goal

ehanna4: There are still a few cases of "Branding Group" in `branding_groups_controller` that I think might need to be updated then

devonbush: ah, good find, I missed all the popup notifications.  fixing now

devonbush:
I think the effort to do a codebase ctrl-f replace for branding group -> collection is much higher than the typical work to retire a feature flag, along with more chances for bugs.  I also don't mind calling them branding groups just because 'collection' is such an overloaded code term.  So let's hold off ticketing it for now, and wait until the first instance of developer pain/confusion ("I spent 30 minutes trying to figure out where the code for this page was")  to think more about it

eweitz:
Sounds good!

PR #1150, opened 2021-09-09
Retiring obsolete 'faceted_search' feature flag (SCP-3621)

bistline:
Similar to #1145, this update removes the obsolete `faceted_search` feature flag.  In addition to retiring & unsetting feature flag values, this also removes the "advanced search" opt-in/out behavior.

This also includes a small enhancement to update the advanced search help modal to dynamically update the statistics regarding the number of public, conventional SCP studies.  This functionality has been added to `SummaryStatsUtils` in case we wish to include this information elsewhere.

MANUAL TESTING
* Run `ruby_local_setup.rb && source config/secrets/.source_env.bash && rails db:migrate`
* Boot portal as normal
* On the home page, note that the `?` link next to `More Facets` no longer contains opt in/out buttons when opened
* Load the Admin Config page, and edit the feature flags for your account
* Note that the form renders w/o error, and that the `faceted_search` entry is gone
* Optional: in the Rails console, query for any user accounts that still have a value set for `feature_flags.faceted_search` and confirm the return is 0:
```
User.where(:'feature_flags.faceted_search'.ne => nil).count
=> 0
```

This PR satisfies SCP-3621.


app/javascript/components/HomePageContentLegacy.js
```diff
@@ -18,7 +18,7 @@ import * as queryString from 'query-string'
 export function StudySearchView({ advancedSearchDefault }) {
```
devonbush, line 18: you can remove the `advancedSearchDefault` parameter now too

app/models/study.rb
```diff
@@ -1494,6 +1494,11 @@ def self.delete_queued_studies
   #
   ###
 
+  # get a quick count of all public studies that have compliant metadata
+  def self.public_compliant
+    self.where(public: true).select { |s| s.metadata_file&.use_metadata_convention? }.count
+  end
```
devonbush, line 1500: This can go either way, but I'd lean against putting this method here.  I think it's better in summary_stats_utils, or a similar analytics module/class.  Study is long enough even with just the basic model functionality, it would be good to have analaytics elsewhere

test/controllers/admin_configurations_controller_test.rb
```diff
@@ -14,16 +14,18 @@ def teardown
...
-
-    AdminConfigurationsController.process_feature_flag_form_data(@test_user, {feature_flag_faceted_search: '1'})
-    assert_equal({'faceted_search' => true}, @test_user.reload.feature_flags)
-
-    AdminConfigurationsController.process_feature_flag_form_data(@test_user, {feature_flag_faceted_search: '0'})
-    assert_equal({'faceted_search' => false}, @test_user.reload.feature_flags)
-
-    AdminConfigurationsController.process_feature_flag_form_data(@test_user, {faceted_search: '-'})
-    assert_equal({}, @test_user.reload.feature_flags)
+    # @test_user.update!(feature_flags: {})
```
devonbush, line 20: Instead of commenting out, I might use the `skip` command at the beginning of this test instead.  See https://www.paperlesspost.com/blog/teams/skipping-tests-minitest/

Either way, though, I like the idea of keeping it around for the next time we have toggle-able flags

devonbush (changes requested):
Good stuff, that's a lot of code I'm happy to be rid of.  Only blocking change is the extra parameter cleanup, everything else is just, like, my opinion, man.

app/models/study.rb
bistline: Oh it actually is, but I forgot to take this out after I refactored.  Will remove it now.

eweitz (approved):

test/controllers/admin_configurations_controller_test.rb
bistline: Oh actually I was thinking of `CurrentUserController`, and not `AdminConfigurationsController`.  This test is still valid, just not using `faceted_search` as the feature flag.

app/javascript/components/search/controls/SearchPanel.js
```diff
@@ -7,102 +7,59 @@ import KeywordSearch from './KeywordSearch'
...
-  function handleClick(ordering) {
-    searchState.updateSearch({ order: ordering })
-  }
-  return (
-    <>
-      <span className="facet">
-        <a onClick={() => handleClick('popular')}>Most Popular</a>
-      </span>
-      <span className="facet">
-        <a onClick={() => handleClick('recent')}>Most Recent</a>
```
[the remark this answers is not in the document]
bistline: I _completely_ agree that the defacto ordering is sub-optimal.  Also, integrating HCA results into regular results just appends to the end, which is also sub-optimal.  I would very much be in favor of prioritizing work to address this more holistically (in another PR).

devonbush (approved):

PR #1151, opened 2021-09-10
Removing cloud-billing.readonly OAuth scope (SCP-3646)

bistline:
In an effort to remove all sensitive/restricted OAuth scopes, this update removes the `cloud-billing.readonly` scope and all associated functionality.  This is due to Google's impending change to OAuth verification and the requirement that all sensitive/restricted be justified.  Any applications that do not obtain verification ahead of the 2021.09.30 deadline could potentially result in either Google displaying a warning every time a user attempts to authenticate, or block access entirely.

In addition to removing the `cloud-billing.readonly` scope, this update removes the ability to create a new Terra billing project through SCP.  This feature is hardly ever used, and is the only reason we required this particular scope.  Instead, the button to create new projects is disabled, and has a popover containing a link to the Terra billing console and a notice informing users to go there to create new billing projects.  Users can still manage billing projects through SCP (add/remove users, get storage costs, see workspaces & manage compute access).

Finally, this consolidates our omniauth/devise integration back down into one single provider for Google OAuth (back to the default `google_oauth2`), and removes obsolete validations/tests that were dependent on the dual providers.

MANUAL TESTING
* Boot portal as normal and sign in
* Select "My Billing Projects" from the profile menu (top right)
* Confirm that the billing project page loads, and that you were not asked to re-authenticate and request the billing scope
* For your `scp-development-[username]` billing project, click "Add User" and add another user (must be registered for Terra)
* Back on the billing projects page, click the red X next to the user account you just added and confirm it is removed

This PR satisfies SCP-3646.


eweitz (approved):
Code looks good!  Removing these lines seems like it will disproportionally simplify things.

I suggested a UI refinement, non-blocking as this feature has very low usage.

devonbush (approved):

devonbush:
great stuff -- really glad you proposed this path forward

PR #1152, opened 2021-09-10
Adding data-analytics-name prop to track raw counts exemption requests (SCP-3632)

bistline:
This update adds the `data-analytics-name` property to the link for requesting a raw counts exemption request.  This is required for automatically tracking exemption requests in Mixpanel.  While not an exact proxy, it will at least allow some automated tracking.  It is possible this will over-count, whereas tracking Zendesk tickets will be an exact count.

Also, as per suggestions in this PR, this update adds `rel: 'noopener noreferrer'` to all `target: :_blank` links to mitigate the risk of "tabnabbing" attacks.

MANUAL TESTING
* Boot portal as normal and sign in
* Ensure the `raw_counts_required_frontend` feature flag is turned on for your account
* Create a new study, or load the upload wizard for one w/o any raw counts expression matrices
* Open the Chrome dev tools > Network panel
* Click the "Request Exemption" link in the "Processed Matrix" tab
* Note the POST to Bard, with the event of `click:link`, and the text property of `raw-counts-exemption`.

This PR supports SCP-3632.


eweitz (approved):

devonbush (changes requested):
In the unlikely event of zendesk getting hacked, we should add `rel="noopener noreferrer"` to the link

See https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

bistline:
> In the unlikely event of zendesk getting hacked, we should add `rel="noopener noreferrer"` to the link. (and I recognize that it wasn't there before)
> 
> See https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

Sure thing - we should probably do that everywhere we say `target: :_blank` in our markup.

devonbush (approved):
Nice, good call on handling that link update in a bunch of places

PR #1153, opened 2021-09-13
(SCP-3602) Upload wizard react refactor

devonbush:
SCP-3602.  This puts the initial code for a file upload wizard using React/API instead of Rails forms.  It's behind a feature flag for ease of future release, but this merge is to an integration branch since I don't think this is ready for even staging-level testing yet.  Also, this is altering the manage_study API in a non-feature-flagged way, so we will want to test mange_study prior to releasing this, even behind a feature flag.

Currently, the cluster and image file upload forms are working, everything else is not yet implemented. See SCP-3655 for remaining work

image

TO TEST:
1. Run migration script, then set `react_upload_wizard` feature flag to true in your environment/user
2. Go to the upload wizard, and select Cluster or Image file
3. Confirm you can edit, create, save, and delete files, and they get uploaded to firecloud.


app/controllers/api/v1/study_files_controller.rb
```diff
@@ -271,52 +264,72 @@ def create
...
-              @study.default_options[:cluster] = @study_file.name
-              @study.save
+          if safe_file_params[:file_type] == 'Gene List' && name_changed
+            @precomputed_entry = PrecomputedScore.find_by(study_file_id: safe_file_params[:_id])
+            if @precomputed_entry.present?
+              logger.info "Updating gene list #{@precomputed_entry.name} to match #{safe_file_params[:name]}"
+              @precomputed_entry.update(name: study_file.name)
+            end
+          elsif safe_file_params[:file_type] == 'Cluster' && name_changed
+            @cluster = ClusterGroup.find_by(study_file_id: safe_file_params[:_id])
```
bistline, line 316: Same - this can likely just be a local variable.

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -271,52 +264,72 @@ def create
...
             end
-            @cluster.update(name: @study_file.name)
-          elsif ['Expression Matrix', 'MM Coordinate Matrix'].include?(study_file_params[:file_type]) && !study_file_params[:y_axis_label].blank?
+          elsif ['Expression Matrix', 'MM Coordinate Matrix'].include?(safe_file_params[:file_type]) && !safe_file_params[:y_axis_label].blank?
             # if user is supplying an expression axis label, update default options hash
-            @study.update(default_options: @study.default_options.merge(expression_label: study_file_params[:y_axis_label]))
+            @study.update(default_options: @study.default_options.merge(expression_label: safe_file_params[:y_axis_label]))
           end
-          render :show
+          return true
```
bistline, line 330: I don't think we need the `return` here, even though this is a helper method.

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -271,52 +264,72 @@ def create
...
         else
-          render json: {errors: @study_file.errors}, status: :unprocessable_entity
+          return false
```
bistline, line 332: Same - superfluous return.

app/javascript/components/upload/ClusteringForm.js
```diff
@@ -0,0 +1,97 @@
...
+    updateFile(file._id, { spatial_cluster_associations: newVal })
+  }
+
+  useEffect(() => {
+    if (clusterFiles.length === 0) {
+      addNewFile(DEFAULT_NEW_CLUSTER_FILE)
+    }
+  }, [clusterFiles.length])
+
+  return <div>
```
bistline, line 40: I like how this is encapsulated now - it seems like the form logic is getting a lot simpler with the componentization.

app/javascript/components/upload/UploadSteps.js
```diff
@@ -0,0 +1,55 @@
...
+  return <span>Meta-----data</span>
+}
+
+export const STEP_ORDER = ['rawCounts', 'processedExpression', 'metadata', 'clustering', 'images', 'coordinateLabels']
+
+export default {
+  rawCounts: {
+    stepTitle: 'Raw Counts',
+    formComponent: RawCountsUploadForm,
+    fileFilter: file => file.file_type === 'Expression Matrix' && file.expression_file_info?.is_raw_counts
```
bistline, line 28: We need to account for `MM Coordinate Matrix` files, but that means bundling, which I know isn't supported yet.

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/** renders save and delete buttons for a given file */
+export function SaveDeleteButtons({ file, updateFile, saveFile, deleteFile }) {
+  return <div>
+    <button type="button" className="btn btn-primary" disabled={!file.isDirty} onClick={() => saveFile(file)}>
+      Save
+      { file.submitData && <span> &amp; Upload</span> }
+    </button> &nbsp;
+    <button type="button" className="btn btn-secondary float-right" onClick={() => deleteFile(file)}>
```
bistline, line 98: We will want to disable this button if a file is parsing.

app/javascript/lib/scp-api.js
```diff
@@ -190,6 +190,70 @@ export function stringifyQuery(paramObj, sort) {
...
+ * @param {String} studyId Study id
+ * @param {FormData} studyFileData html FormData object with the file data
+*/
+export async function createStudyFile(studyId, studyFileData, mock=false) {
+  const apiUrl = `/studies/${studyId}/study_files`
+  const init = Object.assign({}, defaultInit(), {
+    method: 'POST',
+    body: studyFileData
+  })
+  delete init.headers['Content-Type']
```
bistline, line 216: Are we sure we want to delete the `Content-Type`?  Should we be setting it to `multipart/form-data` and validating that on the API side?  This seems like a possibility for something untoward to happen.

app/javascript/lib/scp-api.js
```diff
@@ -190,6 +190,70 @@ export function stringifyQuery(paramObj, sort) {
...
+ * @param {String} studyFileId Study file id
+ * @param {FormData} studyFileData html FormData object with the file data
+*/
+export async function updateStudyFile(studyId, studyFileId, studyFileData, mock=false) {
+  const apiUrl = `/studies/${studyId}/study_files/${studyFileId}`
+  const init = Object.assign({}, defaultInit(), {
+    method: 'PATCH',
+    body: studyFileData
+  })
+  delete init.headers['Content-Type']
```
bistline, line 235: Same as above.

app/javascript/lib/scp-api.js
```diff
@@ -627,7 +691,7 @@ export default async function scpApi(
   }
   if (toJson) {
     const json = await response.json()
-    throw new Error(json.error)
+    throw new Error(json.error || json.errors)
```
bistline, line 694: I've had issues with this in JS tests, so you'll want to be sure to mock this correctly when you get to writing tests for this.

app/models/study_file.rb
```diff
@@ -24,7 +24,7 @@ class StudyFile
   # constants, used for statuses and file types
   STUDY_FILE_TYPES = ['Cluster', 'Coordinate Labels' ,'Expression Matrix', 'MM Coordinate Matrix', '10X Genes File',
                       '10X Barcodes File', 'Gene List', 'Metadata', 'Fastq', 'BAM', 'BAM Index', 'Documentation',
-                      'Other', 'Analysis Output', 'Ideogram Annotations']
+                      'Other', 'Analysis Output', 'Ideogram Annotations', 'Image']
```
bistline, line 27: Thinking out loud - should this be a more specific label than "Image"?  If we get different image types later, we may want that (but maybe we also want the generality here - not really sure).

bistline (changes requested):
Really great progress!  Mostly nitpicks, nothing terribly substantive to change.

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -271,52 +264,72 @@ def create
...
-        assembly_name = update_params[:assembly]
-        update_params.delete(:assembly)
+        species_name = safe_file_params[:species]
+        safe_file_params.delete(:species)
+        assembly_name = safe_file_params[:assembly]
+        safe_file_params.delete(:assembly)
         set_taxon_and_assembly_by_name({species: species_name, assembly: assembly_name})
-
+        # clear the id so that it doesn't get overwritten (which would be a problem for new files)
+        safe_file_params.delete(:_id)
```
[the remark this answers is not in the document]
devonbush: yeah, the id gets read in the `set_study_file` hook, and so we don't need it anymore

app/javascript/lib/scp-api.js
devonbush: ah, this definitely deserves a comment.  I read that this is actually the preferred thing to do so that the browser sets the type automatically with the id of the right form boundaries.  See https://stackoverflow.com/questions/36067767/how-do-i-upload-a-file-with-the-js-fetch-api (about halfway down)

app/models/study_file.rb
devonbush: Yeah, it's a good question to keep in the back of our minds as we're doing this.  I suspect that we'll have a clear answer to this the first time we want to create a study_file_bundle with an image

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -271,52 +264,72 @@ def create
...
+        rescue => e
+          byebug
+          render json: {errors: e.message}, status: 500
+        end
+      end
+
+      # update the given study file with the request params and save it
+      # returns true/false depending on the success of the save
+      def perform_update(study_file)
+        safe_file_params = study_file_params
```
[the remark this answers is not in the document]
devonbush: `study_file_params` is a function that as far as I can tell returns a fresh object each time it is called.  e.g. after I delete params from `safe_file_params`, if I call `study_file_params` again, those params are still there.

app/javascript/components/upload/ImageStep.js
```diff
@@ -0,0 +1,72 @@
...
+const imageFileFilter = file => file.file_type === 'Image'
+
+export default {
+  title: 'Reference Images',
+  name: 'images',
+  component: ImageForm,
+  fileFilter: imageFileFilter
+}
+
+/** Renders a form for uploading one or more cluster/spatial files */
```
eweitz, line 19:
```suggestion
/** Renders a form for uploading one or more image files */
```

app/javascript/components/upload/ProcessedExpressionStep.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+export default {
+  title: 'Processed Expression',
+  name: 'processedExpression',
```
eweitz, line 5: Could we wordsmith more parallel phrases for "processed expression" and "raw counts"?  They're tightly coupled, and phrases that cohere better would make it easy for users to connect them in their minds.

Straw proposals:
* Processed counts (and raw counts)
* Raw expression (and processed expression)

app/javascript/components/upload/StepTabHeader.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+/** renders the wizard step header for a given step */
+export default function StepTitle({ step, index, currentStep, setCurrentStep, serverState, formState }) {
+  let stepFiles = []
+  if (formState && formState.files) {
+    stepFiles = formState.files.filter(step.fileFilter)
+  }
+  return <li className={step.name === currentStep.name ? 'active' : ''} onClick={() => setCurrentStep(step)}>
```
eweitz, line 9: Less dense for easier reading:

```suggestion
  const className = step.name === currentStep.name ? 'active' : ''
  return <li className={className} onClick={() => setCurrentStep(step)}>
```

app/javascript/components/upload/UploadWizard.js
```diff
@@ -0,0 +1,198 @@
...
+      return newFormState
+    })
+  }
+
+  /** Updates file fields by merging in the 'updates', does not perform any validation, and
+   *  does not save to the server */
+  function updateFile(fileId, updates) {
+    setFormState(prevFormState => {
+      const newFormState = _cloneDeep(prevFormState)
+      const fileChanged = newFormState.files.find(file => file._id === fileId)
```
eweitz, line 90: For consistency with two very similar code blocks above:

```suggestion
      const fileChanged = newFormState.files.find(f => f._id === fileId)
```

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react'
```
eweitz, line 1: This file name should use either UpperCamelCase or hyphen-case, like all other SCP JS files.

The `Utils` suffix indicates it should be in `javascript/lib` and not `javascript/components`, but I'd change the name instead of the location given this module is all components.  

Something like `UploadWidgets.js` or `UploadComponents.js` would clarify at a glance of the file name that it's React components related to upload, and not a bunch of framework-agnostic JavaScript functions.

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
...
+]
+
+/** reworks the file object to make it easier to work with
+  * maps the id property to _id, and sets undefined properties
+  * this modifies the object in-place, but also returns it for easy chaining
+  */
+export function formatFileFromServer(file) {
+  file._id = file._id.$oid
+  file.description = file.description ? file.description : ''
+  file.study_id = file.study_id.$oid
```
eweitz, line 34: This would make it more JS-idiomatic, and also warn developers that they're dealing with convenient shorthand that can't be used for e.g. posting upstream.

```suggestion
  file.studyId = file.study_id.$oid
```

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
...
+    })
+  }
+  return data
+}
+
+/** renders a basic label->value text field in a bootstrap form control */
+export function TextFormField({ label, fieldName, file, updateFile }) {
+  return <div className="form-group">
+    <label>{label}</label><br/>
+    <input className="form-control"
```
eweitz, line 60: Blocker: for accessibility (and delightful usability), please make clicking the label focus the input, whether via `for` / `id`, wrapping the `label` in the `input`, or something else.

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
...
+  'status',
+  'upload',
+  'parse_status'
+]
+
+/** reworks the file object to make it easier to work with
+  * maps the id property to _id, and sets undefined properties
+  * this modifies the object in-place, but also returns it for easy chaining
+  */
+export function formatFileFromServer(file) {
```
eweitz, line 31: I'm a big fan of this function!

app/javascript/components/upload/uploadUtils.js
```diff
@@ -0,0 +1,103 @@
...
+      <div className="file-upload-overlay">
+        Saving <FontAwesomeIcon icon={faDna} className="gene-load-spinner"/>
+      </div>
+    }
+    { file.isError &&
+      <div className="file-upload-overlay ">
+        <div className="error-message">
+          An error occurred while saving the file<br/>
+          { file.errorMessage }<br/><br/>
+          <button className="btn btn-secondary" onClick={() => updateFile(file._id, { isError: false })}>Ok</button>
```
eweitz, line 84: Like [Google](https://material.io/components/dialogs#confirmation-dialog) and [Microsoft](https://www.google.com/search?q=microsoft+%22ok+button%22):

```suggestion
          <button className="btn btn-secondary" onClick={() => updateFile(file._id, { isError: false })}>OK</button>
```

eweitz (changes requested):
Great speed on a big strategic improvement!  This will really help for more robust testability and easier UI updates.

I noted two blockers, one easy GUI issue and one substantial API issue.  Beyond that, I suggested a handful of nit fixes.

app/javascript/components/upload/uploadUtils.js
devonbush: actually, this is used for posting upstream to `study_files_controller.rb`, which is why it is camel-cased

app/javascript/lib/scp-api.js
```diff
@@ -190,6 +190,70 @@ export function stringifyQuery(paramObj, sort) {
...
+
+/**
+ * Creates a new study file
+ *
+ * @param {String} studyId Study id
+ * @param {FormData} studyFileData html FormData object with the file data
+*/
+export async function createStudyFile(studyId, studyFileData, mock=false) {
+  const apiUrl = `/studies/${studyId}/study_files`
+  const init = Object.assign({}, defaultInit(), {
```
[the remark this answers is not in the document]
eweitz: Thanks for resolving this in e6dd4f7!

eweitz (approved):

PR #1122, opened 2021-08-18
Report client-side upload validation errors in SCP UI (SCP-3546)

eweitz:
This implements a basic component to show validation errors after files are selected, but before upload.  It matures the bare native alert in #1116 into a Bootstrap alert rendered in React.

https://user-images.githubusercontent.com/1334561/129970256-081cfc70-b107-4a3b-b21a-75c18caf0a19.mov

Future work includes validating more types of errors, and logging them to Mixpanel.

To test:
* Do test steps from #1116 
* Verify that styled notice appears instead ofbare native alert

This satisfies SCP-3546.

PR #1154, opened 2021-09-14
In-app validation: cluster files, more format rules (SCP-3675, SCP-3676)

eweitz:
This adds support for cluster files and more format rules to in-app file upload validation.  It matures the prototype to a general availability candidate.  To recap, the proximate value of this feature is to:
* Reduce time to fail for uploads from 2-3 minutes to < 1 second
* Show file validation errors in SCP UI ("in-app") rather in email

Previously (#1116, #1122), we supported only 1 validation rule in 1 study file type.  Now, we support 7 rules in 2 file types.  In-app validation now handles a large portion of known failure cases.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/133274942-7bcda502-05bd-4b79-b6ff-9e2c848edf3e.mov

Analytics and automated tests are now also more robust.  Given this is a port of pre-existing functionality in SCP Ingest Pipeline, I've noted where new JavaScript code mirrors pre-existing Python code, with hyperlinks for maintainers.

To manually test:
* Do steps from #1122
* Repeat, but for cluster files

This satisfies SCP-3675 and SCP-3676.


devonbush (approved):
Looks great.  A couple thoughts, both of which are outside the scope of this PR:
1. As we add more supported file types, I suspect we will want to refactor this into a separate validation method per-file-type (and then those methods may share common sub-validation methods).  That will make it more human-readable to see and update the exact business logic for each file type
2. As discussed, I'm not sure we should be in the business of 'warning'.  I would favor just changing the spec to allow case-insensitive header matching.  But that's best done separately from this PR, if at all

test/js/lib/validate-file.test.js
```diff
@@ -45,20 +48,99 @@ describe('Client-side file validation', () => {
...
+    // Confirms this validation does not report false negatives
+    //
+    // eslint-disable-next-line max-len
+    // Mirrors https://github.com/broadinstitute/scp-ingest-pipeline/blob/af1c124993f4a3e953debd5a594124f1ac52eee7/tests/test_cell_metadata.py#L17
+    mockReadLinesAndType('metadata_bad_has_coordinates.txt')
+    const { errors } = await validateFile({}, 'metadata')
+    expect(errors).toHaveLength(1)
+  })
+
+  it('passes when no coordinates in metadata file', async () => {
```
bistline, line 134: I really appreciate the two positive tests in this suite!

bistline (approved):
Extremely excited to see this in live soon.

test/js/lib/validate-file.test.js
eweitz: Thanks is really due to @knapii-developments and @jlchang, whose bullet-proof tests I translate and refine here.

eweitz:
> As we add more supported file types, I suspect we will want to refactor this into a separate validation method per-file-type (and then those methods may share common sub-validation methods). That will make it more human-readable to see and update the exact business logic for each file type

That seems promising.  It's also closer to how Ingest Pipeline is organized, with shared code for clusters and annotations in an `annotations.py` module, and file-specific validations (and other processing code) in respectively named modules.

> As discussed, I'm not sure we should be in the business of 'warning'. I would favor just changing the spec to allow case-insensitive header matching. But that's best done separately from this PR, if at all

Yeah, we should definitely do something with warnings -- either use them (e.g. show them in the UI), or remove them.  I'll get folks' thoughts on this tomorrow at demo.

PR #1155, opened 2021-09-14
Fixing raw counts query in study report, enabling table view (SCP-3648)

bistline:
This update fixes a bug in a query for determining which studies have raw counts expression matrices.  Previously, it was only looking at `Expression Matrix` files, but not `MM Coordinate Matrix` files.  In addition, this cleans up the code for `ReportsService.study_data` to be more concise.

In addition, this PR has several (admittedly out-of-scope) updates:
1. Reduces duplicated `include` calls for various API concerns, as well as refactoring `api_base_controller.rb` to be the root-level include in many cases
2. Removes broken/obsolete code in `reports_controller.rb` and the `index.html.erb` view for the same controller
3. Adds an interactive DataTable to the two remaining reports for in-app viewing of both study data & Terra workflow submissions

MANUAL TESTING
1. Log in with an admin account, and click "Reports" from the nav menu
2. Click "View" on the line for "Study Data" to view the inline report
3. Click "Export" to download the tsv of the same data

This PR satisfies SCP-3648.


devonbush (approved):
Awesome

app/lib/reports_service.rb
```diff
@@ -8,86 +8,49 @@ class ReportsService
...
+    all_studies.each do |study|
+      study_id = study.id.to_s
+      # make a hash of study attributes using the keys from above
+      study_hash[study_id] = Hash[keys.zip(keys.map { |k| study.send(k) })]
+      study_hash[study_id][:share_count] = study.study_shares.count
+      user = study.user
+      user_domain = user.email.split('@').last
+      study_hash[study_id][:owner_domain] = user_domain
+      study_hash[study_id][:owner_email] = user.email
+      study_hash[study_id][:admin_owned] = !!user.admin # account for nil by casting to boolean
```
eweitz, line 39: Thanks for this note!

eweitz (approved):

PR #1161, opened 2021-09-16
Extend TDR indexing query (SCP-3593)

jlchang:
Revised query extends extracted data in two areas:
1. adds file_name and file_size for sequence_file and analysis_file rows
2. refactors ontology-based metadata extraction for consistency
    -  extracts .ontology_label instead of .text from HCA metadata
    - extracts .ontology (ie. ontology IDs) 
       - data model lacks ontology ID handling, using SCP conventional names
       - noted in comments that organism_age does not conform to TIM model

This update relates to SCP-3593.


bistline (approved):
Thanks for fixing this!

lib/assets/sql/tdr_indexes/query_w_analysis_file.sql
```diff
@@ -29,6 +29,19 @@ https://docs.google.com/document/d/1KJcF6aWLub--yvTIeZkd_VgTBChinuFqK2aCzSo6hgQ/
 Swagger docs for TDR-dev API
 https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/
 
+Changes:
+previous SQL extracted ".text" values, now using ".ontology_label" values
```
eweitz, line 33: These comments seem helpful overall, but "Changes" strike me as better for a commit message, or at least worth noting "from" (before) and "as of" dates.  I don't feel strongly about it, though -- if you think it's an overall benefit for maintainability let's keep it.

eweitz (approved):

lib/assets/sql/tdr_indexes/query_w_analysis_file.sql
jlchang: Good point, removed trivial change info, kept design decisions.

PR #1163, opened 2021-09-16
Refactoring TDR integration for v2 snapshot query (SCP-3677)

bistline:
This update handles necessary refactoring to deal with the new [result structure](https://github.com/broadinstitute/single_cell_portal_core/blob/development/lib/assets/sql/tdr_indexes/query_w_analysis_file.sql) for TDR queries.  This includes the following:

* Performing "double queries" to first find all projects that match the initial query, and then a second query to find all files that are included in the project.  This is to address sparsity in the ElasticSearch index for certain kinds of metadata due to the complexity of the HCA metadata graph model (like species, disease, organ, etc)
* Moving retrieving `drs_info` objects to `BulkDownloadService#generate_curl_config` as we will already have file type/size information
* Removing obsolete code

This update also addresses a previously undiscovered issue with `BulkDownloadControllerTest#test_bulk_download_should_support_a_download_id_and_TDR_files` where multiple calls were being executed to external services which subsequently failed, and those errors were suppressed.  This test has been refactored to correctly mock the external service calls, as well as handle the newly introduced functionality from this update.

MANUAL TESTING
1. Boot portal as normal and log in w/ admin account
2. Go to admin console, and create/edit configuration for `TDR Snapshot IDs` to use the new snapshot: `25740be5-b493-4185-84b0-bebf425c5072` (if creating, set the `value_type` to `string`)
3. Ensure that the `cross_dataset_search_backend` feature flag is turned on for your account
4. Back on the home page, run a search for `species: Homo sapiens` using facets
5. Scroll to the end of the results, and ensure that the following HCA studies are present and have a facet match: 
  * **A single-cell reference map of transcriptional states for human blood and tissue T cell activation**
  * **Single-cell RNA-sequencing reveals profibrotic roles of distinct epithelial and mesenchymal lineages in pulmonary fibrosis**
6. Click the "Download" button, and confirm that the file counts for both studies are as follows:
hca_results
7. (Optional): Click through to generate the `curl` command and paste into the terminal to ensure that downloads begin (there will be a longer delay than usual as we now generate access URLs for TDR files when creating the curl command).  You can hit `^C` once the download begins as there is > 2TB of data if you select everything.

This PR satisfies SCP-3677.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -695,6 +695,9 @@ module Api
       end
 
       # execute a search in TDR and get back normalized results
+      # will actually issue 2 search requests, first to extract unique project_ids that match the query,
+      # and a second to retrieve all result rows for those projects
+      # this is to address issues in sparsity of ES index with regards to some data (like species, disease, etc)
       def self.get_tdr_results(selected_facets:, terms:)
         results = {}
         begin
@@ -709,8 +712,10 @@ module Api
           logger.info "Executing TDR query with: #{query_json}"
           snapshot_ids = AdminConfiguration.get_tdr_snapshot_ids
           logger.info "Scoping TDR query to snapshots: #{snapshot_ids.join(', ')}" if snapshot_ids.present?
-          raw_tdr_results = ApplicationController.data_repo_client.query_snapshot_indexes(query_json,
-                                                                                          snapshot_ids: snapshot_ids)
+          # first request is to only retrieve project IDs, then the second is for actual row-level results
+          projects = ApplicationController.data_repo_client.query_snapshot_indexes(query_json,
+                                                                                   snapshot_ids: snapshot_ids)
+          project_ids = projects.map { |row| row['project_id'] }.uniq.compact
           added_file_ids = {}
           raw_tdr_results['result'].each do |result_row|
             results = process_tdr_result_row(result_row, results,
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -697,7 +697,7 @@ module Api
       # execute a search in TDR and get back normalized results
       # will actually issue 2 search requests, first to extract unique project_ids that match the query,
       # and a second to retrieve all result rows for those projects
-      # this is to address issues in sparsity of ES index with regards to some data (like species, disease, etc)
+      # this is to address issues in sparsity of Elasticsearch index with regards to some data (like species, disease, etc)
       def self.get_tdr_results(selected_facets:, terms:)
         client = ApplicationController.data_repo_client
         results = {}
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -307,7 +307,6 @@ module Api
           end
           # just log for now
           logger.info "Found #{@tdr_results.keys.size} results in Terra Data Repo"
-          logger.info "#{@tdr_results.map {|name, info| { name => info['facet_matches']}}}"
           @tdr_results.each do |_, tdr_result|
             @studies << tdr_result
           end
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -296,9 +296,8 @@ module Api
           @tdr_results = self.class.get_tdr_results(selected_facets: @facets, terms: @term_list)
 
           if @facets.present?
-            simple_tdr_results = self.class.simplify_tdr_facet_search_results(@tdr_results, @facets)
+            simple_tdr_results = self.class.simplify_tdr_facet_search_results(@tdr_results)
             matched_tdr_studies = self.class.match_studies_by_facet(simple_tdr_results, @facets)
-
             if @studies_by_facet.present?
               @studies_by_facet.merge!(matched_tdr_studies)
             else
@@ -630,23 +629,22 @@ module Api
       end
 
       # Simplify TDR results to be mappable for the UI badges for faceted search
-      def self.simplify_tdr_facet_search_results(query_results, search_facets)
-        simple_TDR_result = {}
-        simple_TDR_results = []
+      def self.simplify_tdr_facet_search_results(query_results)
+        tdr_results = []
         query_results.each_pair do |accession, result|
           facet_matches = result[:facet_matches]
-          simple_TDR_result[:study_accession] = accession
           if facet_matches.present?
+            simple_result = { study_accession: accession }
             facet_matches.each do |mapping|
               mapping.each do |key, val|
                 facet_name = FacetNameConverter.convert_schema_column(:tim, :alexandria, key)
-                simple_TDR_result[facet_name] = val
+                simple_result[facet_name] = val
               end
             end
           end
-          simple_TDR_results << simple_TDR_result
+          tdr_results << simple_result
         end
-        simple_TDR_results
+        tdr_results
       end
 
 
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -5,6 +5,9 @@ module Api
       include Swagger::Blocks
       include StudySearchResultsObjects
 
+      # regex to match on 'sequence_file' and 'analysis_file' entries from TDR
+      TDR_FILE_OUTPUT_TYPE_MATCH = /_file/.freeze
+
       before_action :set_current_api_user!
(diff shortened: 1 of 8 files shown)


bistline:
Test failures are unrelated to new features, and are apparently a known issue right now in Rawls.  Associated PR to address the issue: https://github.com/broadinstitute/rawls/pull/1520

jlchang (commented):
Sadly, I got "Error querying TDR: RestClient::Unauthorized -- 401 Unauthorized" so my service account doesn't have access to the snapshot :(
The local server did try to send the query:
Suppressing error reporting to Sentry: RestClient::Unauthorized:401 Unauthorized, context: {:method=>:post, :url=>"https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/search/query?limit=10000&offset=0", :payload=>"{\"query\":\"{\\\"query_string\\\":{\\\"query\\\":\\\"([TerraCore:hasSex]:\\\\\\\"unknown\\\\\\\")\\\"}}\",\"snapshotIds\":[\"25740be5-b493-4185-84b0-bebf425c5072\"]}", :retry_count=>0}

We'll need to add TDR filters to SCP search:
My local instance had never ingested a study with "homo sapiens" so that facet isn't available for selection - we had not considered populating our facets to reflect TDR content

Fortunately "Monkey Correlations" test dataset has organ = blood. The pulmonary fibrosis study doesn't have blood (only lung) but HumanTissueTcellActivation does.

Unfortunately, when I ran a query on blood locally, I didn't get HumanTissueTcellActivation back, only Monkey Correlations - due to permission issues

For addition testing, know that both PulmonaryFibrosisGSE135893 and "Monkey Correlations" have unknown for sex (sex is male for HumanTissueTcellActivation)

This following argues for using ontology IDs instead of labels:
"Monkey Correlations" and HumanTissueTcellActivation are also both 10X v2 sequencing BUT the HCA data still has the "10X v2 sequencing" label (remember EFO refactored the names?) but the portal records "10x 3' v2" as the label (for reference PulmonaryFibrosisGSE135893's library_preparation_protocol is "10X 5' v2 sequencing"

Hopefully the above info will help with testing in prod to know which HCA study to expect for what facets.

devonbush (approved):
Nice.  I had to reach deep to find ticky-tack things to comment on for this PR  :)

app/controllers/api/v1/search_controller.rb
```diff
@@ -818,6 +820,35 @@ def self.get_term_match_for_tdr_result(term, result_row)
         end
         matches
       end
+
+      # extract file information from row-level TDR results for sequence_file and analysis_file entries
+      def self.extract_file_information(result_row)
+        safe_entry = result_row.with_indifferent_access
+        output_type = safe_entry['output_type']
```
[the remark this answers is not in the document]
bistline: Ah, good catch.  I'll do that.

app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -175,7 +175,7 @@ const COLUMNS = {
   sequence: {
     title: 'Sequence',
     types: ['sequence_file'],
-    info: 'Sequence files, such as BAM or BAI files',
+    info: 'Sequence files, such as Fastq, BAM or BAI files',
```
eweitz, line 178: Blocker: per [convention](https://en.wikipedia.org/wiki/FASTQ_format), "Fastq" should be all-caps:

```suggestion
    info: 'Sequence files, such as FASTQ, BAM or BAI files',
```

app/lib/bulk_download_service.rb
```diff
@@ -57,22 +57,30 @@ def self.generate_curl_configuration(study_files:,
...
-            file_config = "url=\"#{file_info['url']}\"\n"
+      tdr_file_configs = []
+      tdr_files.map do |shortname, file_infos|
+        # pull out project manifest and process separately
+        manifests, files = file_infos.partition { |f| f['file_type'] == 'Project Manifest' }
+        manifest_info = manifests.first
+        manifest = hca_client.get_project_manifest_link(default_catalog, manifest_info['url'])
+        manifest_config = "--location\n" # add location directive to allow following 302 redirect to manifest location
+        manifest_config += "url=\"#{manifest['Location']}\"\n"
+        manifest_config += "output=\"#{shortname}/#{manifest_info['name']}\""
```
eweitz, line 68: At a glance, I suspect a multi-line string would be more readable.

app/lib/bulk_download_service.rb
```diff
@@ -57,22 +57,30 @@ def self.generate_curl_configuration(study_files:,
...
+        tdr_file_configs << manifest_config
+        # now process remainder of analysis/sequence files for download in parallel to speed up process
+        Parallel.map(files, in_threads: 100) do |file_info|
+          begin
+            drs_info = ApplicationController.data_repo_client.get_drs_file_info(file_info[:drs_id])
```
eweitz, line 73: Please open a ticket upstream so we don't need to send a bunch of parallel requests.  Client multithreading is an API code smell -- if clients (like SCP) need to do it, it's a strong hint that the API needs improvement.

Naively, I suspect SCP could avoid parallelizing requests if Data Repo API had a version of the endpoint underlying `get_drs_file_info` that accepted a comma-delimited list of DRS IDs.  So the upstream ticket would be something like "Enable passing comma-delimited list of DRS IDs to \<API endpoint name\>".

That would allow client code like SCP's to be faster, simpler, and more robust.

eweitz (changes requested):
Good stuff!  I noted a trivial blocker regarding user-facing text (search "Blocker:"), and suggested a few non-blocking refinements (and tickets for upstream refinements).

> This is to address sparsity in the ElasticSearch index for certain kinds of metadata due to the complexity of the HCA metadata graph model

Could a ticket be opened to address that upstream?  (I don't consider it a blocker.)  To my understanding, the value of TDR for Human Cell Atlas data -- and any other graph-based dataset -- is greatly diminished without this fix.

An upstream solution would improve on this local fix, as the former would be:
1. More performant
2. Less code for us to maintain
3. Applicable to all TDR clients (beyond just SCP)
4. Applicable to other graph-based datasets

(3) and (4) would especially benefit Data Sciences Platform -- and represent a nice way to give back to the broader department.

app/lib/bulk_download_service.rb
bistline: I actually did that months ago: https://broadworkbench.atlassian.net/browse/DR-1947.  Given the lack of movement on it, I doubt this is something they feel inclined to do unless we make a lot of noise, and I'm not sure this is the place we want to spend our capital getting things implemented.

Also, we had this exact same behavior for TDR files beforehand as well, just in a different place.  And we are doing the same thing for generating signed URLs for files in SCP study buckets on line 30 of this same service (though in that case, we're only generating signed URLs, and I don't think that emits an API request to GCS).  Is your objection to doing so here because it requires API calls?  And not just parallelizing method calls to an underlying SDK?

app/controllers/api/v1/search_controller.rb
```diff
@@ -771,14 +778,12 @@ def self.process_tdr_result_row(row, results, selected_facets:, terms:, added_fi
...
-        # if the output_id is a drs_id, add it and the output_type to the file_information array,
-        if row['output_id']&.starts_with?('drs')
-          drs_id = row['output_id']
-          if !added_file_ids[drs_id]
-            result[:file_information] << {
-              drs_id: drs_id,
-              file_type: row['output_type']
-            }
+        # gather file information for sequence_file and analysis_file entries
+        if row['output_type'] =~ TDR_FILE_OUTPUT_TYPE_MATCH
```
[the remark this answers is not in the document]
bistline: That was the original implementation, and Devon requested that it be constantized.  I don't have a preference one way or the other, so I'll let you two haggle over the signature here.

app/lib/bulk_download_service.rb
eweitz: Ah, thanks for noting the ticket.  I that could be noted in a code comment here, I promise not to suggest this next time!

> Is your objection to doing so here because it requires API calls? And not just parallelizing method calls to an underlying SDK?

My objection is parallelizing many network requests that could -- with a likely small amount of upstream effort -- be more quickly, reliably, and simply batched in a single request by all clients.  So if the API or SDK calls needs network I/O, and gets multi-threaded, then I ask myself "Why can't this be a single request?"  

Often, it's just a use case that hasn't been noted upstream.  I appreciate that that part's done, and agree that there are other improvements that would be even more helpful!

eweitz (approved):

PR #1165, opened 2021-09-17
Adding chunked file uploading and image support in explore tab (SCP-3602)

devonbush:
SCP-3602 The UX for the image files will definitely change (and may be feature flagged depending on timing), but the technical pieces are straightforward.

TO TEST:

1. check out branch, and turn on the 'react_upload_wizard' feature flag
2. go to upload wizard, confirm you can upload an image, associate it with a cluster, and then see it in the explore tab (note that the image preview currently requires to to refresh the page after upload -- that will get fixed in later polish)
3. Test uploading a file larger than the 10MB CHUNK SIZE (or go to UploadWizard.js and reduce CHUNK_SIZE to something smaller) and confirm it is correctly uploaded (and using the debugger network tab, confirm the chunk sizes are correct)


bistline (changes requested):
Nice to see this moving along - some suggestions for readability, and also setting proper HTTP statuses.

app/javascript/lib/scp-api.js
```diff
@@ -257,6 +304,27 @@ export async function deleteStudyFile(studyAccession, fileId, mock=false) {
...
+ * appropriate SCP readonly bearer token, and using the Google API URL that allows CORS
+ *
+ * @param {String} bucketName bucket name
+ * @param {String} fileName file name
+*/
+export async function fetchBucketFile(bucketName, fileName, mock=false) {
+  const init = {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${window.SCP.readOnlyToken}`
```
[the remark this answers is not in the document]
bistline: Yeah, we will have to put in extra error handling anyway for the ToS refactoring & PPW, so it might be expedient to start building a framework for doing that now.  The likelyhood of getting `401` or `403` when trying to generate tokens will definitely go up for private studies (public will be unaffected).

app/lib/cluster_viz_service.rb
```diff
@@ -59,10 +59,31 @@ def self.load_spatial_options(study)
...
   end
 
+  def self.load_image_options(study)
+    # grab all the image files for this study
+    image_file_info = StudyFile.where(study: study, file_type: 'Image')
+                                 .pluck(:name, :spatial_cluster_associations, :upload_file_name, :description)
+
+    # now grab any non spatial cluster files for the study that had ids specified in the image files above
+    # and put them in an id=>name hash
+    associated_clusters = StudyFile.where(study: study, :id.in => image_file_info.map{ |si| si[1] }.flatten.uniq)
```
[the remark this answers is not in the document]
bistline: True - it does.  If you want to find the middle ground, you could create hashes rather than arrays, so that we have key names to reference, rather than array indexes.  For instance:

```
attributes = [:name, :spatial_cluster_associations, :upload_file_name, :description]
image_file_info = StudyFile.where(study: study, file_type: 'Image')
                                .pluck(*attributes).map { |file_attr| Hash[attributes.zip(file_attr) } 
                                
```

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -291,60 +301,96 @@ def perform_update(study_file)
...
+            render json: {errors: "Incorrect chunk sent: expected bytes starting with #{@study_file.upload_file_size}, received #{content_range[:first_byte]}" }, status: :unprocessable_entity
+          end
         else
-          return false
+          render json: {errors: 'Missing Content-Range header'}, status: :unprocessable_entity
+        end
+
+        File.open(@study_file.upload.path, "ab") do |f|
+          f.write upload.read
+        end
```
eweitz, line 372: Out of scope, but while we're here: any idea how much it'd speed upload user journey to POST chunks from this server to upstream (whether directly to the bucket or a FireCloud proxy)?

As is, we essentially upload each file twice, serially -- once from web browser to SCP server, then again from SCP server to upstream server.  The latter only _begins_ once the former ends.  "Streaming" upload from SCP to upstream (still uploading twice, but in parallel) should effectively eliminate the time for that 2nd upload phase.

(Streaming from the web browser to upstream server would be even more optimal, but maybe the effort/value ratio wouldn't be quite as worthwhile as streaming SCP server to upstream server.)

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -291,60 +301,96 @@ def perform_update(study_file)
...
+        @study_file.save!
+
+        if @study_file.upload_file_size >= content_range[:total_size]
+          # this was the last chunk
+          complete_upload_process(@study_file, safe_file_params[:parse_on_upload])
         end
       end
 
+      def complete_upload_process(study_file, parse_on_upload)
+        @study.delay.send_to_firecloud(study_file) # send data to FireCloud if upload was performed
```
eweitz, line 385: To inform potential optimizations, could you wrap this call with timers and log that to Mixpanel (along with file type, size)?

app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -83,8 +84,10 @@ def show
             annotationList: AnnotationVizService.get_study_annotation_options(@study, current_api_user),
             clusterGroupNames: ClusterVizService.load_cluster_group_options(@study),
             spatialGroups: spatial_group_options,
+            imageFiles: image_options,
             clusterPointAlpha: @study.default_cluster_point_alpha,
-            colorProfile: @study.default_color_profile
+            colorProfile: @study.default_color_profile,
+            bucket_id: @study.bucket_id
```
eweitz, line 90: Blocker: let's use consistent case in API response objects, or add a code comment on why this is an exception.

```suggestion
            bucketId: @study.bucket_id
```

app/javascript/components/explore/ExploreDisplayTabs.js
```diff
@@ -384,10 +383,24 @@ export default function ExploreDisplayTabs({
...
             }
+            { enabledTabs.includes('images') &&
+              <div className={shownTab === 'images' ? '' : 'hidden'}>
+                <ImageTab
+                  studyAccession={studyAccession}
+                  imageFiles={exploreInfo.imageFiles}
+                  bucketName={exploreInfo.bucket_id}
+                  isCellSelecting={isCellSelecting}
+                  isVisible={shownTab === 'images'}
+                  getPlotDimensions={getPlotDimensions}
```
eweitz, line 394: Could you remind me why we pass this function as a parameter, rather than simply importing it in the component?

app/javascript/components/explore/ImageTab.js
```diff
@@ -0,0 +1,96 @@
+import React, { useState, useEffect } from 'react'
+import BucketImage from 'components/visualization/BucketImage'
+import ScatterPlot from 'components/visualization/ScatterPlot'
+
+/** Component for displaying IGV for any BAM/BAI files provided with the study */
```
eweitz, line 5:
```suggestion
/** Display reference images (e.g. tissue stains) beside spatial scatter plots */
```

app/javascript/components/upload/UploadWizard.js
```diff
@@ -22,10 +23,11 @@ import CoordinateLabelStep from './CoordinateLabelStep'
 import RawCountsStep from './RawCountsStep'
 import ProcessedExpressionStep from './ProcessedExpressionStep'
 
+const CHUNK_SIZE = 10000000
 const STEPS = [RawCountsStep, ProcessedExpressionStep, ClusteringStep, CoordinateLabelStep, ImageStep]
 
 /** shows the upload wizard */
-export default function UploadWizard({ accession, name }) {
+export default function UploadWizard({ studyAccession, name }) {
```
eweitz, line 30: FWIW, `accession` strikes me as snappier than `studyAccession`, and I somewhat wish we'd use the former at the start of our big JS refactor over a year ago.

app/javascript/components/visualization/BucketImage.js
```diff
@@ -0,0 +1,28 @@
...
+
+import { fetchBucketFile } from 'lib/scp-api'
+
+/** display an image file directly from a GCP bucket */
+export default function BucketImage({ bucketName, fileName }) {
+  const [remoteImageUrl, setRemoteImageUrl] = useState(null)
+
+  /** If there is a remote file, try to load it and render it on the page */
+  useEffect(() => {
+    console.log('useEffect!!!')
```
eweitz, line 11:
```suggestion
```

app/javascript/components/visualization/BucketImage.js
```diff
@@ -0,0 +1,28 @@
...
+    setRemoteImageUrl(null)
+    getBucketImageLocalUrl(bucketName, fileName).then(setRemoteImageUrl)
+  }, [bucketName, fileName])
+
+  if (!remoteImageUrl) {
+    return <></>
+  }
+  console.log(`remote url: ${remoteImageUrl}`)
```
eweitz, line 19:
```suggestion
```

app/javascript/components/upload/UploadWizard.js
```diff
@@ -79,25 +85,51 @@ export default function UploadWizard({ accession, name }) {
...
   }
 
   /** save the given file and perform an upload if a selected file is present */
   async function saveFile(file) {
-    updateFile(file._id, { isSaving: true })
-    const fileApiData = formatFileForApi(file)
+    let studyFileId = file._id
+    updateFile(studyFileId, { isSaving: true })
+    const fileSize = file.uploadSelection?.size
+    const isChunked = fileSize > CHUNK_SIZE
```
eweitz, line 101: It might not be strictly necessary to chunk upload when the file is small, but unless there's a significant downside to doing so, simply always chunking upload would reduce complexity a fair amount.

eweitz (changes requested):
Great engineering here!  I noted one trivial blocker, thought aloud about optimizations, and suggested a handful of maintainability refinements.

eweitz:
Could you follow up on these?

* https://github.com/broadinstitute/single_cell_portal_core/pull/1165#discussion_r712923647
* https://github.com/broadinstitute/single_cell_portal_core/pull/1165#discussion_r712944356
* https://github.com/broadinstitute/single_cell_portal_core/pull/1165#discussion_r712959830

devonbush:
@eweitz  sure, if you want more detail, let's chat tomorrow

eweitz (approved):
Text replies would be most helpful, but if needed I can chat when I return Monday.

devonbush:
ah, sorry, I had a whole slew of text replies from yesterday that I never hit 'submit' on.  I was wondering why you were ignoring them

devonbush:
merging now as Emily needs the image support in her branch

PR #1167, opened 2021-09-17
Eh strip quotes from gene search

ehanna4:
- Strip off the beginning and ending quotes from a gene search. 
- Allow searching on no genes as a means to clear out the form.

To Test:
1. Go to SCP home page and click to the Gene Search tab. Enter a gene with beginning and/or quotations like "PTEN" and search. Observe that this will return the same results as searching on just PTEN (no quotes).
2. After searching a gene in the Gene Search tab clear out the gene params by clicking the 'X' in the search. Then search on the empty gene params. Observe that the page loads the default set of studies not filtered by any genes. 

Left side shows current situation where a gene with ""s won't match even though that gene is in a study. Right side shows the new way that will show the search result for such genes.
Screen Shot 2021-09-16 at 1 15 57 PM

Previously there was no way for a user to get the gene search form back to the default state shown on the left. Right side shows the page as it's reloading after clicking search with no genes entered
Screen Shot 2021-09-17 at 4 38 42 PM

### DIFF
--- a/app/javascript/components/search/genes/GeneKeyword.js
+++ b/app/javascript/components/search/genes/GeneKeyword.js
@@ -72,6 +72,26 @@ export default function GeneKeyword({ placeholder, helpTextContent }) {
     }
   }
 
+  /** handle a user pressing the 'x' to clear the field */
+  function handleClear() {
+    setGeneArray([])
+    setInputText('')
+    geneSearchState.updateSearch(
+      { genes: '' },
+      studySearchState
+    )
+  }
+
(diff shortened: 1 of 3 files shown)


bistline (approved):
Nicely done - I fully approve of any time we can remove a modal...

devonbush (approved):
Great tests.  You should definitely present the lack-of-modal at demo. I'm fine with taking it out, but IIRC correctly it was put in there to address a specific source of user confusion, so we should make sue that's not still applicable

eweitz (approved):

PR #1169, opened 2021-09-20
Collections UX - the return (SCP-3595)

devonbush:
SCP-3595  The collections UX has now been greenlit for merge (although it should not be released without SCP-3685).  This reverts the revert commits, and adds some last-minute UX polish from Jerome.  Changes are removing the icon on the Browse Collections button, left-aligning the 'Collections' header, and adding a 'Home' link to that header.  

image

image

TO TEST:
1. Go to home page, make sure 'Browse Collections' button appears
2. Click 'Browse Collections', view the list page, and confirm the styling appears correct


app/views/branding_groups/list_navigate.html.erb
```diff
@@ -0,0 +1,44 @@
+<h2><%= link_to 'Home', site_path %> / Collections</h2>
+<span>Collections are curated groups of studies which share a common research theme. If your organization is interested in creating a collection within SCP, contact us at <%= mail_to 'scp-support@broadinstitute.zendesk.com', 'scp-support@broadinstitute.zendesk.com' %></span>
+<div class="row">
+  <div class="col-md-12">
+    <ul class="collections-list">
+      <% @branding_groups.each do |branding_group| %>
+        <li id="<%= branding_group.name_as_id %>">
+          <div class="collections-logo">
+            <% if branding_group.splash_image.filename.present? %>
+              <%= link_to image_tag(branding_group.splash_image.url), site_path(scpbr: branding_group.name_as_id) %>
```
bistline, line 10: Non-blocking: it would be nice if we had better behavior for non-existent images as opposed to the default "broken image tag" for most browsers.

bistline (approved):

app/views/branding_groups/list_navigate.html.erb
devonbush: yup, that was addressed in an earlier PR, hence the `if` above.  If there's no image, it will just be empty

bistline: Ah, that means then that I have a bunch of broken images in my local instance.  Good to know!

app/views/branding_groups/_form.html.erb
```diff
@@ -11,73 +11,83 @@
...
-                                                        class: 'btn btn-xs btn-primary' %></p>
-      <% end %>
+  <div class="form-group form-check form-check-inline">
+    <%= f.label :public, 'Publish collection <span class="fas fa-question-circle" data-toggle="tooltip" data-placement="top" title="Select to have your collection listed in the Collections Directory. Visibility of individual studies within the collection is still controlled by study permissions set by the study owner."></span>'.html_safe, class: "form-check-label" %> &nbsp;
+    <%= f.check_box :public, class: "form-check-input" %>
+  </div>
+  <div class="form-group">
+    <%= f.label :user_id, 'Associated User' %><br />
+    <%= f.select :user_id, options_from_collection_for_select(User.all, :id, :email, @branding_group.new_record? ? nil : @branding_group.user_id),
+                 {include_blank: 'Please select a user account'}, class: 'form-control' %>
```
eweitz, line 29: We don't preface other defaults with pleasantries, so this seems worth omitting for brevity and consistency.

```suggestion
                 {include_blank: 'Select a user account'}, class: 'form-control' %>
```

app/views/branding_groups/list_navigate.html.erb
```diff
@@ -0,0 +1,44 @@
+<h2><%= link_to 'Home', site_path %> / Collections</h2>
```
eweitz, line 1: As an engineer, I appreciate that "/" can separate levels in a hierarchy, but in this less technical use case I interpreted it as "and / or".  We use a chevron elsewhere for this sort of thing, and it'd be less ambiguous here, too.

app/views/branding_groups/list_navigate.html.erb
```diff
@@ -0,0 +1,44 @@
...
+<span>Collections are curated groups of studies which share a common research theme. If your organization is interested in creating a collection within SCP, contact us at <%= mail_to 'scp-support@broadinstitute.zendesk.com', 'scp-support@broadinstitute.zendesk.com' %></span>
```
eweitz, line 2: Wordsmithing:

```suggestion
<span>Collections are curated groups of studies that share a common research theme. If your organization is interested in creating a collection in SCP, contact us at <%= mail_to 'scp-support@broadinstitute.zendesk.com', 'scp-support@broadinstitute.zendesk.com' %></span>
```

eweitz (approved):

PR #1171, opened 2021-09-21
Enabling metadata upload in react refactor (SCP-3667)

devonbush:
SCP-3667   More-or-less a direct UX port of the existing metadata upload page

image

TO TEST:
1. make sure react-upload-wizard feature flag is on
2. go to upload wizard, and select Metadata step
3. confirm page more-or-less matches current production UX
4. toggle the 'convention_required' feature flag
5. confirm UX works to upload and parse a file

### DIFF
--- a/app/javascript/components/upload/MetadataStep.js
+++ b/app/javascript/components/upload/MetadataStep.js
@@ -1,0 +1,49 @@
+import React, { useEffect } from 'react'
+import metadataExplainerImage from 'images/metadata-convention-explainer.jpg'
+
+const DEFAULT_NEW_METADATA_FILE = {
+  file_type: 'Metadata'
+}
+
+const metadataFileFilter = file => file.file_type === 'Metadata'
+
+export default {
+  title: 'Metadata',
+  name: 'metadata',
+  component: MetadataForm,
+  fileFilter: metadataFileFilter
+}
+
+/** Renders a form for uploading one or more cluster/spatial files */
+function MetadataForm({
+  formState,
+  addNewFile,
+  updateFile,
+  saveFile,
+  deleteFile,
+  handleSaveResponse
+}) {
+  const metadataFile = formState.files.find(metadataFileFilter)
+
+  useEffect(() => {
+    if (!metadataFile) {
+      addNewFile(DEFAULT_NEW_METADATA_FILE)
+    }
+  }, [metadataFile?._id])
+
+  return <div>
+    <div className="row">
+      <h4 className="col-sm-12">3. Metadata</h4>
+    </div>
+    <div className="row">
+      <div className="col-md-12" id="metadata-convention-explainer">
+        <img src={metadataExplainerImage}/>
+        <a id="metadata-convention-example-link"
+          href="https://singlecell.zendesk.com/hc/en-us/articles/360060609852-Required-Metadata"
+          target="_blank" rel="noopener noreferrer">
+          View required conventional metadata
+        </a>
+      </div>
+    </div>
+  </div>
+}
--- a/app/javascript/components/upload/MetadataStep.js
+++ b/app/javascript/components/upload/MetadataStep.js
@@ -24,6 +24,7 @@ export default {
 
 /** Renders a form for uploading one or more cluster/spatial files */
 function MetadataForm({
+  serverState,
   formState,
   addNewFile,
   updateFile,
@@ -99,7 +100,10 @@ function MetadataForm({
                   checked={!file.use_metadata_convention}
                   onChange={e => updateFile(file._id, { use_metadata_convention: false })} /> No
               </label> &nbsp; &nbsp;
-              <OverlayTrigger trigger="click" rootClose placement="top" overlay={conventionIssuePopover}>
+              <OverlayTrigger trigger="click"
+                rootClose
+                placement="top"
+                overlay={<ConventionIssuePopover studyAccession={serverState.study.accession} email={userState.email}/>}>
                 <span className="action log-click"> Using conventional names is an issue for my study</span>
               </OverlayTrigger><br/>
               Learn <a href="https://singlecell.zendesk.com/hc/en-us/articles/360061006411-Metadata-Convention"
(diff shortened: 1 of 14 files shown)


app/javascript/components/upload/MetadataStep.js
```diff
@@ -0,0 +1,141 @@
...
+    Why use the metadata convention?<br/>
+    Using the metadata convention will enable your data to be discovered through our faceted search interface.
+    Using the convention means extra validation will be applied to ensure your metadata conforms to standard vocabularies.<br/>
+    <a href='https://singlecell.zendesk.com/hc/en-us/articles/360061006411-Metadata-Convention' target='_blank' rel="noopener noreferrer">Learn more</a>.
+  </Popover>
+)
+
+const conventionIssuePopover = (
+  <Popover id="convention-issues">
+    Using the convention is now required.  If this presents a particular issue for your study,
```
bistline, line 136: We should update this with the new UX from #1178.

app/javascript/components/upload/UploadWizard.js
```diff
@@ -179,43 +181,45 @@ export default function UploadWizard({ studyAccession, name }) {
...
-      <div className="col-md-10">
-        <h4>{studyAccession}: {name}</h4>
+  return <UserProvider>
+    <div className="">
+      <div className="row padded">
+        <div className="col-md-10">
+          <h4>{studyAccession}: {name}</h4>
+        </div>
+        <div className="col-md-2">
+          <a href={`/single_cell/study/${studyAccession}`}>View Study</a>
```
bistline, line 191: This link could use styling as a Bootstrap button - it is very easy to miss without it.

app/javascript/components/upload/MetadataStep.js
```diff
@@ -0,0 +1,141 @@
...
+  component: MetadataForm,
+  fileFilter: metadataFileFilter
+}
+
+/** Renders a form for uploading one or more cluster/spatial files */
+function MetadataForm({
+  formState,
+  addNewFile,
+  updateFile,
+  saveFile,
```
bistline, line 30: This works as advertised, but I feel like there is a UX regression here in terms of a successful upload/save.  There is no visual indication to the user that the action succeeded.  I'm not necessarily advocating for a modal (we have far too many, and they are very obtrusive), but the current UX (non-React) is _very_ explicit when things work, and when they don't.  I'd like to find a middle ground that gives necessary feedback to the user without taking over the page.

bistline (changes requested):
Metadata form works as advertised, but I think the lack of feedback to users on successful upload should be addressed sooner rather than later.  I thought that there was an error at first because it looked like nothing happened after the upload finished, and I only realized it worked by looking at the logs.

app/javascript/components/upload/UploadWizard.js
devonbush: This is addressed in the later (but now-merged) UX refinement PR

app/javascript/components/upload/MetadataStep.js
devonbush: yup, reasonable success/fail/parsing alerts haven't been done yet -- Jerome is still brainstorming concepts.  I'll create a separate ticket to make sure it gets addressed

app/javascript/components/upload/MetadataStep.js
```diff
@@ -0,0 +1,141 @@
...
+import { UserContext } from 'providers/UserProvider'
+import FileUploadControl, { FileTypeExtensions } from './FileUploadControl'
+import { TextFormField, SavingOverlay, SaveDeleteButtons } from './uploadUtils'
+
+const DEFAULT_NEW_METADATA_FILE = {
+  file_type: 'Metadata',
+  use_metadata_convention: true
+}
+
+const metadataFileFilter = file => file.file_type === 'Metadata'
```
ehanna4, line 16: Does this need a check if file exists? 
ie `file?.file_type`

app/javascript/components/upload/MetadataStep.js
```diff
@@ -0,0 +1,141 @@
...
+      </div>
+
+    </div>
+    { file &&
+      <div className="row top-margin" key={file._id}>
+        <div className="col-md-12">
+          <form id={`metadataForm-${file._id}`}
+            className="form-terra"
+            acceptCharset="UTF-8"
+            onSubmit={() => { console.log('submitting'); return false}}>
```
ehanna4, line 70: Should this` console.log` be removed?

app/javascript/components/upload/MetadataStep.js
```diff
@@ -0,0 +1,141 @@
...
+    Using the convention means extra validation will be applied to ensure your metadata conforms to standard vocabularies.<br/>
+    <a href='https://singlecell.zendesk.com/hc/en-us/articles/360061006411-Metadata-Convention' target='_blank' rel="noopener noreferrer">Learn more</a>.
+  </Popover>
+)
+
+const conventionIssuePopover = (
+  <Popover id="convention-issues">
+    Using the convention is now required.  If this presents a particular issue for your study,
+    please contact us at <a href='mailto:scp-support@broadinstitute.zendesk.com'>scp-support@broadinstitute.zendesk.com</a>
+    so we can help or make any updates to the convention.
```
ehanna4, line 138: need a space between the link and 'so'

ehanna4:
Unsure if this is intended (maybe since the metadata file is still uploading?) but the cluster files are normal while the metadata file name is italicized 
Screen Shot 2021-09-26 at 9 29 28 PM

ehanna4 (approved):
Overall looks good! Tested clicking around the UX and uploading a file. Was a "bad file" from the test data in SCP-core but I'm figuring the file validation is not included in this PR? (Looks like more convo in other comments about this success/fail upload being addressed in a different ticket)

Left a couple of comments but otherwise looks good 👍

devonbush:
Now updated with the zendesk form UX 
image

devonbush:
> Unsure if this is intended (maybe since the metadata file is still uploading?) but the cluster files are normal while the metadata file name is italicized
> Screen Shot 2021-09-26 at 9 29 28 PM

I think this is probably because you were using a 'bad' file, and so it would get deleted from the server.  In any case, we have a separate ticket to clean up that UX for showing why things are or are not italicized

PR #1174, opened 2021-09-23
updating branding group image capability (SCP-3685)

devonbush:
SCP-3685 This adds a checkbox so that users can delete collection images and restore defaults.  I've also double checked in both development and production that images can be changed.  
image

This also adds a minor UX enhancement so that if a branding group does not have images specified, the form just shows 'default' on the display
image

TO TEST:
1. Go to the 'edit collection' page for a collection that has one or more images
2. If the collection has an existing image, check the 'Clear/Reset to default' checkbox next to the image
3. save the form
4. confirm the image is no longer associated with a branding group
5. add an image to the collection and save
6. change the image and save
7. confirm the change appears

### DIFF
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -57,7 +57,15 @@ class BrandingGroupsController < ApplicationController
   # PATCH/PUT /branding_groups/1.json
   def update
     respond_to do |format|
-      if @branding_group.update(branding_group_params)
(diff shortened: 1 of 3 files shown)


ehanna4 (approved):
Looks good to me - tested uploading new pictures and clearing to default and saw the appropriate images.

devonbush:
Leyla approved (although I can't add her as reviewer due to org constraints)

PR #1176, opened 2021-09-24
Instrumenting react-select (SCP-3678)

devonbush:
SCP-3678

TO TEST:
  1. Go to explore tab, and open developer tools to network tab
  2. change cluster
  3. observe log event menu:change with text and label properties
  4. change annotation, subsample, and other dropdowns and confirm event log

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -81,7 +81,14 @@ export function logClick(event) {
 
   if (typeof event.isTrigger !== 'undefined') {return}
 
+
   const target = $(event.target)
+  // don't double-log clicks we have otherwise instrumented
+  const EXCLUDED_CLICK_SELECTOR = '.labeled-select'
+  if (target.closest(EXCLUDED_CLICK_SELECTOR).length) {
+    return
+  }
+
   // we use closest() so we don't lose clicks on, e.g. icons within a link/button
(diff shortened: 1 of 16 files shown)


app/javascript/lib/metrics-api.js
```diff
@@ -174,31 +181,51 @@ function logClickButton(target) {
...
-    labels = Array.from(document.querySelector(`label[for='${id}']`))
+  if (!label) {
+    const id = element.id
+    if (id) {
+      label = document.querySelector(`label[for='${id}']`)
+    }
+  }
+  if (!label) { // traverse parents looking for label
+    let parent = element
+    while (parent = parent.parentNode) {
```
bistline, line 205: Are we sure this won't get stuck in an infinite loop?  I'm assuming yes, because at a certain point you'll reach the top-level node in the DOM and have nowhere to go...

bistline (approved):

app/javascript/lib/metrics-api.js
devonbush: yup, that's the idea (and this structure was in our code before)

ehanna4 (approved):
The menu:change event is coming through as expected but I notice for Spatial Groups and Annotation the value chosen isn't being passed on to mixpanel like it is for Clustering. Is this intended?

devonbush:
build failure is related to Azul

PR #1182, opened 2021-09-28
Add File Download Component for Upload Wizard refactor (SCP-3666)

ehanna4:
Add a React component for downloading files from the Wizard and add it to the currently available file types: Metadata, Clustering, Spatial transcriptomics and Reference Images. 

To test go to the new upload wizard for a study and choose any of those file types above and if there is a file you should be able to download it and everything should appear very similar to the current wizard.

There is one piece of logic not ported into this component yet from the existing one which is for Sequence Data when a user has an external link to human data. Since the Sequence Data hasn't been done in the refactor yet I didn't implement this yet since I wouldn't be able to test it nicely. Will follow up and update that when it's available. Unless there are objections.

File that's downloadable and one that is not processed to the bucket yet:
Screen Shot 2021-09-28 at 7 27 03 AM

Tool Tip:
Screen Shot 2021-09-27 at 6 22 11 PM


bistline (approved):

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,44 @@
...
+
+  /** If there is a remote file, try to load it */
+  useEffect(() => {
+    setRemoteImageUrl(null)
+    getBucketLocalUrl(bucketName, fileName).then(setRemoteImageUrl)
+  }, [bucketName, fileName])
+
+  /** get the remote file path */
+  async function getBucketLocalUrl(bucketName, fileName) {
+    const response = await fetchBucketFile(bucketName, fileName)
```
devonbush, line 18: `fetchBucketFile` downloads the file, and we're calling this method every time this component is loaded.  which means we're downloading every file, regardless of whether the user clicks on it.  I don't think that's correct

app/javascript/components/upload/ClusteringFileForm.js
```diff
@@ -27,6 +28,11 @@ export default function ClusteringFileForm({
           file={file}
           updateFile={updateFile}
           allowedFileTypes={FileTypeExtensions.plainText}/>
+        <FileDownloadControl
+          file={file}
+          handleSaveResponse={handleSaveResponse}
```
devonbush, line 33: you don't need to pass in 'handleSaveResponse' here

app/javascript/components/upload/ClusteringStep.js
```diff
@@ -98,7 +100,9 @@ export function ClusteringUploadForm({
         updateFile={updateFile}
         saveFile={saveFile}
         deleteFile={deleteFile}
-        handleSaveResponse={handleSaveResponse}/>
+        handleSaveResponse={handleSaveResponse}
```
devonbush, line 103: you don't need to pass in 'handleSaveResponse' here

app/javascript/components/upload/ClusteringStep.js
```diff
@@ -98,7 +100,9 @@ export function ClusteringUploadForm({
...
+        bucketName={formState.study.bucket_id}
```
devonbush, line 104: since you've passed in 'bucketName' as a parameter to the ClusteringUploadForm, you can just use that here

app/javascript/components/upload/ImageFileForm.js
```diff
@@ -36,6 +36,11 @@ export default function ImageFileForm({
               file={file}
               updateFile={updateFile}
               allowedFileTypes={FileTypeExtensions.image}/>
+            <FileDownloadControl
+              file={file}
+              handleSaveResponse={handleSaveResponse}
```
devonbush, line 41: same, don't need handleSaveResponse

app/javascript/components/upload/MetadataStep.js
```diff
@@ -86,6 +88,11 @@ function MetadataForm({
                   file={file}
                   updateFile={updateFile}
                   allowedFileTypes={FileTypeExtensions.plainText}/>
+                <FileDownloadControl
+                  file={file}
+                  handleSaveResponse={handleSaveResponse}
```
devonbush, line 93: same :)

devonbush (changes requested):
General structure looks great.  I think the fetching sequence is off though, because this fetches the file as soon as the button is visible, instead of when the button is clicked on.

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,44 @@
...
+    const respBlob = await filefrombucket.blob()
+    return URL.createObjectURL(respBlob)
+  }
+
+  if (!file.upload_file_name) {
+    return null
+  } else {
+    if (!file.upload_file_name && file.human_data) {
+      return null
+      // TODO once the Sequence Data tab is added update this section for handling external human data files
```
eweitz, line 29: Could you note which ticket this TODO corresponds ticket (and ticket the task if isn't already)?  We typically note the ticket beside the TODO, e.g.:

`TODO (SCP-1234): Once the Sequence Data tab...`

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      return <p> <label> Link to file </label>
+        <br></br>
+        {!file.generation ? <span className="label label-warning no-download-available" data-toggle="tooltip"
+          title='You will be able to download this file once it has been uploaded to our remote data store. Check back soon.'>
```
eweitz, line 34: Trimmed, and with less technical detail:

```suggestion
          title='You can download this file once it has been fully uploaded. Check back soon.'>
```

eweitz (approved):
Code looks good, modulo Devon's blocker!  I noted some nits.

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { fetchBucketFile } from 'lib/scp-api'
+import { bytesToSize } from 'lib/stats'
+
+/** renders a file download control for the given file object */
+export default function FileDownloadControl({ file, bucketName }) {
+  const fileName = file.upload_file_name
+
+  /** Load file when button is clicked */
+  const handleOnClick = () => {
```
devonbush, line 10: I think it's more conventional to name methods like this either `handleClick` or `handleDownloadClick` 
```suggestion
  const handleDownloadClick = () => {
```

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,41 @@
...
+    getBucketLocalUrl(bucketName, fileName).then(value => window.open(value))
```
devonbush, line 11: for clarity, and since you declared the `getBucketLocalUrl` function async
```suggestion
    const localUrl = await getBucketLocalUrl(bucketName, fileName)
    window.open(url)
```

app/javascript/components/download/FileDownloadControl.js
```diff
@@ -0,0 +1,41 @@
...
+      return null
+    // TODO (SCP-3719): Once the Sequence Data tab is added update this section for handling external human data files
+    } else {
+      return <p>
+        <br></br>
+        {!file.generation ? <span className="label label-warning no-download-available" data-toggle="tooltip"
+          title='You can download this file once it has been fully uploaded. Check back soon.'>
+          {<span className="fas fa-ban"></span> } Awaiting remote file
+        </span> :
+          <a onClick={() => handleOnClick()} >
```
devonbush, line 34: add classes so the link gets a cursor pointer even though it doesn't have an href element
```suggestion
          <a onClick={() => handleDownloadClick()} className="btn action" >
```

app/javascript/styles/_global.scss
```diff
@@ -791,3 +791,8 @@ figcaption.ck-editor__editable {
 .min-width-100 {
   min-width: 100%;
 }
+
+.file-upload-download-container {
+  display: flex;
+  align-items: center;
+}
```
devonbush, line 798: this seems like a set of styles that will be useful for a lot of form layouts.  We already have a class 'flexbox' for `display: flex`.  Maybe change this class name to `.flexbox-align-center` or something similar so we can reuse it elsewhere

app/javascript/components/download/FileDownloadControl.js
devonbush: Also, since this is opening an arbitrary file, we should be careful about security.  Reading https://stackoverflow.com/questions/40593632/use-window-open-but-block-use-of-window-opener, it seems like the arguments below are the equivalent of `rel="noopener noreferrer"` 
`window.open(url, '_blank', 'noopener,noreferrer)`

devonbush (changes requested):
Looks great -- really like the use of window.open to solve the loading issue.  A few style nits and one security concern

app/javascript/components/download/FileDownloadControl.js
ehanna4: I actually find the `then` structure more clear in this case. Since `getBucketLocalUrl` returns a promise the `then` just consume the value returned when the promise fulfills and opens the window using that value which feels simpler than having to await and make `handleDownloadClick` `async` and then needing to `await` and `async` the `onclick` in the <a> as well. Please let me know though if you feel strongly about this and I can be persuaded haha.

devonbush (approved):

PR #1187, opened 2021-09-29
Fixing bulk download of "root" DirectoryListing objects (SCP-3682)

bistline:
This update fixes a bug where directories at the root level of a GCP bucket could not download via bulk download.  Users could select "all data", which works, but any requests to download individual directories fail and do not return any files.  The issue was that we were not generating the output map of accessions/files to folders on the host machine.

This update also addresses a newly discovered issue where multiple directories at the same path, but with different file types, cannot be downloaded discretely.  This update now appends the `file_type` value to directory-based bulk download requests to disambiguate which directory is being requested.

MANUAL TESTING 
1. Before starting, download either of the folders from [here](https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/directory_listing_data?project=broad-singlecellportal-staging) to your local machine
2. Boot and log in as normal
3. Create a new study, but don't upload any files
4. Go back to "My Studies", and click on "Show Details" for the new study
5. On the study detail page, click "View Google Bucket"
6. Upload either of the folders downloaded earlier to the bucket at the root level
7. Back in the portal, return to "My Studies", and click "Sync Workspace"
8. Sync the directory you uploaded before
9. Sign out, and sign in with a different user account, and load the study
10. Click "Bulk Download", and then use the download link for the directory at the bottom

This PR satisfies SCP-3682.

### DIFF
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -311,7 +311,8 @@ module Api
         # if this is a single-study download, allow for DirectoryListing downloads
         if valid_accessions.size == 1 && params[:directory].present?
           study_accession = valid_accessions.first
-          directories = self.class.find_matching_directories(params[:directory], study_accession)
+          directory_name, file_type = params[:directory].split('--')
+          directories = self.class.find_matching_directories(directory_name, file_type, study_accession)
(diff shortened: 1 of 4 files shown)


eweitz (approved):

devonbush (approved):
nice

PR #1189, opened 2021-09-30
Upload sequence files in react upload wizard (SCP-3721)

devonbush:
SCP-3721.  Adding sequence file upload capability.  No validation yet -- that will come later.  I think we'll need a document summarizing all the validation rules we want to apply within and across forms

image

 Also updates the way file ids are tracked after to address the bug Jon found where the bundle children would be reset after saving the parent.  The issue was the the id of the parent file changes when it's saved from `newFile-<x>` to a mongo GUID.  So I needed to make sure we were tracking the old id as well so the kids' bundle parent ids could be updated to the GUID after save

 Also addresses a comment from Eric in a previous PR about how `uploadUtils.js` was a bad filename in just about every way possible, so it's now split into two more sensibly organized files -- `upload-utils.js` for utility functions and `form-components.js` for small shared components.

TO TEST:
1 . Go to upload wizard for a study (make sure react upload wizard feature flag is on)
2. Go to sequence file step
3. add and save fastq file
4. add a file and set its type to BAM
5. set the species and assembly, and confirm the assembly menu updates with the species
6.  choose a BAI file
7.  confirm BAI file cannot be saved without first saving BAM file
8. save BAM file
9. observe BAI file and description are preserved
10. save BAI file


bistline (approved):

app/javascript/components/upload/SequenceFileForm.js
```diff
@@ -0,0 +1,185 @@
...
+  saveFile,
+  deleteFile,
+  addNewFile,
+  handleSaveResponse,
+  sequenceFileTypes,
+  fileMenuOptions,
+  associatedBaiFile
+}) {
+  const speciesOptions = fileMenuOptions.species.map(spec => ({ label: spec.common_name, value: spec.id }))
+  const selectedSpecies = speciesOptions.find(opt => opt.value === file.taxon_id)
```
eweitz, line 20: [The more I think about `taxon_id`](https://github.com/broadinstitute/single_cell_portal_core/pull/1181#discussion_r717723757), the more I think it's like `study_id` -- i.e., worth avoiding in client code.  It's used a lot for sequence files.

Like any study and `accession`, any file with an associated species (taxon) has a unique `ncbi_taxid` and `scientific_name`, e.g. "9606" and "Homo sapiens".  Unlike `taxon_id`, which is an opaque database ID tied to a particular MongoDB instance, both `ncbi_taxid` and (moreso) `scientific_name` are interoperable, memorable, and often known off-hand to users and developers.  

For better interoperability and often easier debugging, I suggest we use `ncbi_taxid` instead of database IDs as the canonical identifier for species throughout our API and clients.  I'd actually moreso prefer something like adding a `name` field with kebab-case formatting of `scientific_name` values.  Other bioinformatics APIs like Ensembl, NCBI, and MyGene.info all use scientific name or taxid as the canonical identifier in APIs.

This is relevant around now while we're doing a big refactor, have improved interoperability with moves from study accessions, and are considering more work in the CLI client.  I don't expect these changes in this PR, but figured here makes sense to outline a possible related improvement.

eweitz (approved):
Code looks good!  Well, except for something out of scope -- see my comment below.

app/javascript/components/upload/SequenceFileForm.js
devonbush: makes sense to me.  I think the relative impact for this change with respect to our application codebase would be small (but it's also a relatively minor change--doesn't need a migration or anything like that).  And you're right on that it would make the CLI more comprehensible

PR #1190, opened 2021-09-30
Upload wizard gene list form (SCP-3723)

devonbush:
SCP-3723.  Adding a gene list form to the new react wizard.  I'm curious to get some feedback here and at demo of renaming 'Gene Lists' externally to 'Mean Expression Lists' or something similarly more descriptive and less easily confused with 10x gene lists.  Maybe something even more generic like "Analysis tables"?

image

 Nothing much to see here technically--very similar to the other forms.  At this stage, I'm very much erring on the side of making the forms easy to independently edit and change rather than DRYing them out.  As the UX crystallizes we can start factoring out more common UX (e.g. the form heading styles).  

This PR also cleans out some redundant divs.  TextFormField is already wrapped in a 'form-group' div

### DIFF
--- a/app/javascript/components/upload/ClusteringFileForm.js
+++ b/app/javascript/components/upload/ClusteringFileForm.js
@@ -41,9 +41,8 @@ export default function ClusteringFileForm({
             </label>
           </div>
         }
-        <div className="form-group">
-          <TextFormField label="Description / Figure Legend (this will be displayed below cluster)" fieldName="description" file={file} updateFile={updateFile}/>
-        </div>
+        <TextFormField label="Description / Figure Legend (this will be displayed below cluster)" fieldName="description" file={file} updateFile={updateFile}/>
+
         <div className="row">
           <div className="col-md-4">
             <TextFormField label="X Axis Label" fieldName="x_axis_label" file={file} updateFile={updateFile}/>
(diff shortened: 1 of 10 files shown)


bistline:
> I'm curious to get some feedback here and at demo of renaming 'Gene Lists' externally to 'Mean Expression Lists' or something similarly more descriptive and less easily confused with 10x gene lists. Maybe something even more generic like "Analysis tables"?

In the database, these are called `PrecomputedScore`, as originally the thought was that we would have many different kinds, but it has only ever been precomputed expression values by some kind of annotation/heuristic.  I'm not crazy about having the term "Mean" in the title, because the type of collapse/computation done is completely independent of all the functionality surrounding these files.  These are essentially expression matrices that are rolled up to "samples" that are lower resolution than individual cells - usually, they represent some kind of annotation which is reflected in a cluster or metadata value, where the columns represent populations of cells.  That being said, study owners can put whatever they want - the "precomputed" denotes that these are all self contained, independent of all other files in the study, and cannot be viewed in any other context.

Waaay back when this model was first added, they were actually called "Marker Gene Lists", but then we dropped the "Marker" later on.  I'd be very curious what @jlchang thinks these should be called.

bistline (approved):

eweitz (approved):

devonbush:
merging now to get the functionality in.  Will definitely continue discussion around the best names

PR #1196, opened 2021-10-05
Complete adding download links (SCP-3719)

ehanna4:
Add download links to the remaining new file types and update the style to be ready for file validation.

To test: flip on the feature flag for upload wizard and all the file types should have a download (other than human data sequence files which just have their url as is seen today in the wizard)

Human Data:
Screen Shot 2021-10-05 at 3 38 24 PM

Not Human Data:
Screen Shot 2021-10-05 at 3 37 00 PM

Coordinate Labels that is still uploading:
Screen Shot 2021-10-05 at 3 35 52 PM

Reference image:
Screen Shot 2021-10-05 at 3 20 04 PM

### DIFF
--- a/app/javascript/components/download/FileDownloadControl.js
+++ b/app/javascript/components/download/FileDownloadControl.js
@@ -27,11 +27,11 @@ export default function FileDownloadControl({ file, bucketName }) {
     } else {
       return <p>
         <br></br>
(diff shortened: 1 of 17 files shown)


ehanna4:
@devonbush it looks like newly added files that aren't yet saved now don't have a `file.upload_file_name` and instead have a `file.name` is that correct?

PR #1199, opened 2021-10-07
Enable ingest of larger matrix files through larger boot disk (SCP-3772)

jlchang:
PR to increase boot disk to 300 GB so large matrix uploads do not run out of disk.

Sentry reference: https://sentry.io/organizations/broad-institute/issues/2693851858/?project=1424198

This supports SCP-3772


jlchang:
Build errors were in tests for Azul integration (3rd party dependency) and unlikely to be caused by this PR.

PR #1230, opened 2021-11-01
Plotting sorted by expression (SCP-3779)

devonbush:
This sorts the trace we input to plotly in ascending order of gene expression, so that expressed values will be plotted on top of non-expressed values.  This adds a new synthetic study 'chicken_overplot' to test the behavior.   This adds about 1 second to plot time per 1MM cells plotted in localhost testing

BEFORE (note the 'A' is grey)
image

AFTER (note the 'A' is now expressed)
image

This also cleans out an unused file from the chicken_raw_counts study.

TO TEST:
1.  in console, run `SyntheticStudyPopulator.populate('chicken_overplot')`
2. view the study,  and confirm the biosample overplotting -- hiding the red trace 'A' should reveal the purple 'A' behind it
3. search for gene Foo, and confirm that the 'A' shows an expression gradient.
4. comment out line 286 in `ScatterPlot.js` (the `expressionsWithIndices.sort...` line), and confirm the 'A' would be greyed out without the sorting


app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -274,11 +278,36 @@ function getPlotlyTraces({
       styles: legendEntries
     }]
   } else {
+    // for non-clustereed plots, we pass in a single trace with all the points
```
devonbush, line 281:
```suggestion
    // for non-clustered plots, we pass in a single trace with all the points
```

db/seed/synthetic_studies/chicken_overplot/study_info.json
```diff
@@ -0,0 +1,24 @@
+{
+  "study": {
+    "name": "Chicken overplotted",
+    "description": "<p>Synthetic chicken raw counts with overplotting.  The 'A' has overplotting (with the cells that are the 'D' in chicken_raw_counts).  Jean Chang, Eric Weitz, Jon Bistline, Eno-Abasi Augustine-Akpan, Devon Bush, Vicky Horst.  Journal of SCP Synthetic Studies.  Volume 1, p10, October 2020.  Assesing the benefits of raw count data for single cell transcriptomics analysis.</p><p>Expected study gene expression patterns:</p><p>Foo - gradient for 'spatial', top (hi) to bottom (low); gradient for 'cluster', distal (hi) to center (lo)</p><p>Bar - A: 2, B: 3, C: 4, D: 5</p><p>Baz - gradient for 'spatial', roughly bottom (hi) to top (low); gradient for 'cluster', center (hi) to distal (lo)</p><p>Qux: same value (5) for all cells</p><p>Reference screenshots of expected visualizations can be found at: https://github.com/broadinstitute/single_cell_portal_core/pull/772</p>",
```
devonbush, line 4:
```suggestion
    "description": "<p>Synthetic chicken raw counts with overplotting.  The 'A' has overplotting (with the cells that are the 'D' in chicken_raw_counts).  Jean Chang, Eric Weitz, Jon Bistline, Eno-Abasi Augustine-Akpan, Devon Bush, Vicky Horst.  Journal of SCP Synthetic Studies.  Volume 1, p10, October 2020.  Assesing the benefits of raw count data for single cell transcriptomics analysis.</p><p>Expected study gene expression patterns:</p><p>Foo - gradient for 'spatial', top (hi) to bottom (low); gradient for 'cluster', distal (hi) to center (lo)</p><p>Bar - A: 2, B: 3, C: 4</p><p>Baz - gradient for 'spatial', roughly bottom (hi) to top (low); gradient for 'cluster', center (hi) to distal (lo)</p><p>Qux: same value (5) for all cells</p><p>Reference screenshots of expected visualizations can be found at: https://github.com/broadinstitute/single_cell_portal_core/pull/772</p>",
```

eweitz (commented):
Code looks generally good, but I'm concerned about performance.  Could you post some local timing data for how long `getPlotlyTraces` takes on a large plot with and without this enhancement?

If this adds >= 0.5 s for a 1M cell plot, I think it'd be worth briefly trying some optimizations.  First view could perhaps be faster via `for` loops instead of `map` and fewer iterations. (I see 6 iterations over the sorted array, and wonder if we could achieve the same faster in 2.) Later views could perhaps be faster via caching.

If this adds < 0.5 s for a 1M cell plot, then I'd be OK merging without trying to optimize.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -274,11 +278,36 @@ function getPlotlyTraces({
...
     }]
   } else {
+    // for non-clustered plots, we pass in a single trace with all the points
+    let colors
+    if (isGeneExpressionForColor) {
+      // sort the points by order of expression
+      const expressionsWithIndices = data.expression.map((val, i) => [val, i])
+      expressionsWithIndices.sort((a, b) => a[0] - b[0])
+      // now that we know the indices, reorder the other data arrays
+      trace.x = expressionsWithIndices.map(ei => data.x[ei[1]])
```
bistline, line 288: I'm with @eweitz on this - I think calling `map` repeatedly will not scale well for large plots.  Can we reduce this to a single `map` call, and update all traces at the same time?  For example:
```suggestion
      expressionsWithIndices.map(ei => {
        trace.x.concat(data.x[ei[1]])
        trace.y.concat(data.y[ei[1]])
        ...
      })
```

bistline (changes requested):
I think the approach is sound, but will need optimization for it to scale for production.

app/javascript/components/visualization/ScatterPlot.js
eweitz: Implementing this on the server would be fastest by a dramatic margin.  That way, we could sort _once per file info update_, rather than (at best) once per page load.  Cache invalidation could presumably just reuse what we already have in place for this endpoint.

bistline: I'm not sure we can do that, since all of the visualization data will depend on the cluster/annotation/gene requested.  We'd have to pre-cache everything, which will never scale.  If we wanted to do this server side at runtime, I'd had the idea previously that we could first plot all of the 0 values, and then plot the significant expression.  Since we only store the significant expression values in Mongo, we can infer all the 0 values from a quick set difference of `(cells in current plot - expressed cells for genes)`, put those in the traces _first_, and then the actual expression values.  It's not quite sorted by expression, but it does put all of the non-zero values last.

eweitz: Good point.  Sorting at ingest seems fastest, in that case, but your proposal is likely better effort/value.

devonbush: Interesting, my thought was that separate maps are better since then the array allocations (which I suspect are by far the most time-intensive part of this operation) can happen all at once, rather than via a series of pushes. But you're right that investigation is warranted

devonbush:
Let me know your thoughts on including the perf testing code -- I didn't think it was worth a feature flag, but I've used this code 3 or 4 different times now, so figured it might be worth putting somewhere

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -271,6 +274,27 @@ def self.get_selected_subsample_threshold(param, cluster)
           end
           subsample
         end
+
+        # returns a random sample of data, useful for frontend performance testing
+        # the only parameters are the number of cells in the response, and whether to include expression data
+        # this method either randomly generates the data
+        def self.generate_fixed_size_plot_data(num_cells, inlcude_expression)
+          num_annots = 10
+          genes = num_genes.times.map{ |n| "gene#{n}" }
```
eweitz, line 283: Where is `num_genes` defined?  20000 strikes me as a decent default.

eweitz (approved):
Code looks more performant!  For our future convenient reference, could you post some performance numbers as a comment in this PR?  Per demo chat, the performance impact seems worthwhile.

I suggested a few more optimizations, and noted several rough areas of the (rather neat!) new performance testing method.  None are blocking.

devonbush:
removed the perf code, I think Eric made good points that it's very quick-and-dirty, and so just as likely to mislead as to illuminate unless you know exactly what kinds of cases its appropriate for (e.g. javascript array sorting tests :-)

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -274,11 +278,47 @@ function getPlotlyTraces({
...
+      colors = new Array(data.expression.length)
+      for (let i = 0; i < expressionsWithIndices.length; i++) {
+        // now that we know the indices, reorder the other data arrays
+        trace.x[i] = data.x[expressionsWithIndices[i][1]]
+        trace.y[i] = data.y[expressionsWithIndices[i][1]]
+        if (is3D) {
+          trace.z[i] = data.z[expressionsWithIndices[i][1]]
+        }
+        trace.cells[i] = data.cells[expressionsWithIndices[i][1]]
+        trace.annotations[i] = data.annotations[expressionsWithIndices[i][1]]
```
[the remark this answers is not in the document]
devonbush: It's not faster (interpreter probably made this optimization for me), but it is more readable

bistline (approved):

bistline:
Build failure is actually a green build, but timed out on the cleanup.

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP277/chicken-overplotted?genes=Foo#study-visualize).

PR #1231, opened 2021-11-02
Storing all facets/filters from Azul service (SCP-3808)

bistline:
In order to allow users to discover external data effectively via Azul, we need to pre-populate our `SearchFacet` entries will all possible facet/filter values from Azul.  This will enable users to select values from the filter lists that will correspond with an entry in Azul.  In addition, this update introduces three notable changes to executing searches:

1. A new attribute called `merged_filters` will store filter values from Azul alongside all entries from SCP studies (sourced from BigQuery)
2. An instance method on `SearchFacet` objects called `filters_for_user(user)` that will take into account the `cross_dataset_search_backend` feature flag
3. Changes to query string delimiters for search facets:
  a. Facet entries are now delimited with `|` instead of `+`
  b. Filter values are now delimited with `;` instead of `,`

Changes 1 & 2 are needed to prevent sign-in users from seeing a large list of filter values that will return empty searches, as they likely will not have the `cross_dataset_search_backend` feature flag enabled.  Change 3 is needed as filter values from Azul contain our current delimiters (`+` and `,`) within actual values (e.g. `CD4+ T cell` or `CD8-positive, alpha-beta T cell` for `cell_type`).  Changing our delimiter characters proved simpler than adding in complicated logic to double URL-encode/decode values.  As of right now, there are no entries in Azul that contain either `|` or `;` for a filter value.  If at some point such entries are added, this logic will need to be revisited.

MANUAL TESTING
1. Pull this branch, run migrations via `rails db:migrate` and boot as normal
2. On the home page, open the `disease` facet, and confirm it looks as normal for your instance (something like below)
Screen Shot 2021-11-02 at 11 47 27 AM
3. Log in with an admin account, and ensure the `cross_dataset_search_backend` feature flag is enabled for this account
4. Go back to the home page and open the `disease` facet, and confirm that it has many more entries now:
Screen Shot 2021-11-02 at 11 47 52 AM
 5. Select `Alzheimer disease` from the list, execute the search, and confirm one of the returned studies is `Single-cell RNA Sequencing of human microglia from post mortem Alzheimers Disease CNS tissue.`
 6. (Optional): log out, and log in with an account that does not have `cross_dataset_search_backend` enabled, and confirm the `disease` filter list has reverted back to normal

This PR satisfies SCP-3808.


app/lib/azul_search_service.rb
```diff
@@ -117,4 +117,27 @@ def self.extract_azul_file_info(file)
     end
     file_info.with_indifferent_access
   end
+
+
+  # retrieve all possible facet/filter values present in Azul
+  # this is done by executing an empty search and requestin only 1 project, then retrieving the
```
eweitz, line 123:
```suggestion
  # this is done by executing an empty search and requesting only 1 project, then retrieving the
```

eweitz (approved):
The new delimiters seem more robust!

app/controllers/api/v1/search_controller.rb
```diff
@@ -32,7 +32,9 @@ class SearchController < ApiBaseController
           parameter do
             key :name, :facets
             key :in, :query
-            key :description, 'User-supplied list facets and filters, formatted as: "facet_id:filter_value+facet_id_2:filter_value_2,filter_value_3"'
+            key :description, 'User-supplied list facets and filters.  Delimit facets from filters with ":", filter ' \
+                              'values with ";" and facet entries with "|"'
+            key :example, 'facet_id:filter_value|facet_id_2:filter_value_2;filter_value_3'
```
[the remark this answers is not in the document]
bistline: I agree - and I did spend some time thinking about this.  The main reason why I didn't change it as it would require us to version the _entire_ API up to `v2`.  For something this small, it seemed like a waste of effort.  Instead, I'm going to open a ticket for a feature request to allow versioning of individual API endpoints.  I'm not sure exactly how we'd do that given the current structure, but it seems like the "whole API" version isn't going to scale well for us.

bistline:
Build error was due to partial outage on Terra earlier today (MySQL lock timeout issue)

app/controllers/api/v1/search_controller.rb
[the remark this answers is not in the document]
devonbush: big fan of how you updated the syntax rather than adding complex character escaping logic.  Good call on opening a ticket -- I think it's worth a discussion that includes Leyla for how much we're willing to spend on backwards API compatibility (e.g. do we need to maintain full test suites on all API versions, etc...)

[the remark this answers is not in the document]
devonbush: FWIW I might have switched the semi-colon and pipe operators, so that it would be `facet_id:filter_value;facet_id_2:filter_value_2|filter_value_3`
I think that's because the pipe feels more like an 'or' character to me, and semi-colon is more like an 'and'.  But that's just me, and probably not worth changing at this point

app/models/search_facet.rb
```diff
@@ -267,9 +268,11 @@ def self.get_table_schema(table_name: CellMetadatum::BIGQUERY_TABLE, column_name
 
   # update all search facet filters after BQ update
   def self.update_all_facet_filters
```
devonbush, line 270: What's the performance/reliability of fetching the facets from Azul?  This means we have to make a query to azul every time a metadata file is updated, which seems like more than we need, and will make some tests more brittle.  Wondering if it's worth adding a `field :cached_azul_facets, type: {}` and then adding a parameter to this method (defaulting to false) for whether to refresh the cached azul facets too

app/models/search_facet.rb
```diff
@@ -12,6 +12,7 @@ class SearchFacet
   field :identifier, type: String
   field :filters, type: Array, default: []
   field :public_filters, type: Array, default: [] # filters for public studies only
+  field :merged_filters, type: Array, default: [] # filters for XDSS, includes all SCP & external facet filters
```
devonbush, line 15:
```suggestion
  field :filters_with_external, type: Array, default: [] # filters for XDSS, includes all SCP & external facet filters
```
might be more explicit about what's included.  or maybe rename :filters to :scp_filters (trickier since that'll be a migration)

app/views/api/v1/search/_search_facet_config.json.jbuilder
```diff
@@ -9,7 +9,7 @@ else
 end
 json.set! :id, search_facet_config.identifier
 json.set! :links, search_facet_config.ontology_urls
-json.set! :filters, user_signed_in? ? search_facet_config.filters : search_facet_config.public_filters
+json.set! :filters, search_facet_config.filters_for_user(current_user)
```
devonbush, line 12: nice

devonbush (approved):
Good stuff!  A few suggestions, none blocking.

app/models/search_facet.rb
bistline: The performance for that particular query is very good - around 1s.  We're only asking for a single project, so the response time is quick.  As to the reliability, this might be an instance where the health check is a good thing, and if there's any kind of issue/error, we skip updating the filters from Azul.

Those values don't change super often, but the Azul team said it is possible for those values to update at least on a weekly basic, though bi-weekly is more normal.  That being said, we don't get a huge number of metadata files daily.  According to [Mixpanel](https://mixpanel.com/project/2120588/view/19059/app/insights#report/9671053/file-ingests-scp/~(columnWidths~(bar~())~description~'Number*20of*20ingests*20of*20study*20files*20--*20gene*20expression*20matrices*2c*20cluster*20files*2c*20and*20metadata*20files*20--*20in*20Single*20Cell*20Portal*2c*20done*20via*20Ingest*20Pipeline.~displayOptions~(chartType~'line~plotStyle~'stacked~analysis~'linear~value~'absolute)~sorting~(bar~(sortBy~'column~colSortAttrs~(~(sortBy~'value~sortOrder~'desc)~(sortBy~'value~sortOrder~'desc)))~line~(sortBy~'value~sortOrder~'desc~valueField~'averageValue~colSortAttrs~(~(sortBy~'value~sortOrder~'desc~valueField~'averageValue)))~table~(sortBy~'column~colSortAttrs~(~(sortBy~'label~sortOrder~'asc)~(sortBy~'label~sortOrder~'asc)))~insights-metric~(sortBy~'value~sortOrder~'desc~valueField~'totalValue~colSortAttrs~(~))~pie~(sortBy~'value~sortOrder~'desc~valueField~'totalValue~colSortAttrs~(~)))~timeComparison~null~liftComparison~null~querySamplingEnabled~false~dashboard~(id~647255~title~'Single*20Cell*20Portal*20over*20time~creator_name~'Eric*20Weitz~can_update_basic~true~can_share~true~allow_staff_override~false~is_superadmin~false)~sections~(show~(~(dataset~'!mixpanel~value~(name~'ingest~resourceType~'events)~resourceType~'events~profileType~null~search~'~dataGroupId~null~math~'total~perUserAggregation~null~property~null))~group~(~(dataset~'!mixpanel~value~'fileType~resourceType~'events~profileType~null~search~'~dataGroupId~null~propertyType~'string~typeCast~null~unit~null))~time~(~(dateRangeType~'in*20the*20last~unit~'day~window~(unit~'month~value~3))))~legend~(individualSelectedSegments~(selected~(~(key~'Metadata)))))), the most we've seen in the past 3 months is 23, and that was a pretty big outlier.  5-10 is a much more normal range.

I think adding pre-flight and error handling to updating these facets will suffice in terms of reliability & brittleness.

PR #1234, opened 2021-11-03
Add tab analytics (SCP-3700)

ehanna4:
Add analytics for tabs
- Add "referringTab" property for events that occur from within a tab i.e. clicking a button while in a tab
- Add "tabDisplayText" property to record the displayed text (different than "text" property as it won't be overwritten by the "data-analytics-name" field)
- Add "tabListName" to the new upload wizard

To test:
Pull the branch and click about the upload wizard and observe the Network tab to see the events and see the new properties being passed in the payload

Screen Shot 2021-11-02 at 5 14 56 PM

Screen Shot 2021-11-02 at 5 10 44 PM
)

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -335,10 +335,10 @@ function getAnalyticsPageName() {
  * gets the tab name for analytics
  */
 function getTabProperty() {
-  if (window.location.href.match(/\?tab=/)) {
+  if (window.location.href?.match(/\?tab=/)) {
     return window.location.href.split('?tab=')[1]
   } else {
-    return window.location.hash.replace(/#/, '')
+    return window.location.hash?.replace(/#/, '')
   }
 }
 
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -163,6 +163,7 @@ export function logClickLink(target) {
   if (parentTabList.length > 0) {
     // Grab name of tab list and add to props
     props.tabListName = parentTabList[0].attributes['data-analytics-name'].value
+    props.tabDisplayText = target.innerText.trim()
     log('click:tab', props)
   } else {
     log('click:link', props)
@@ -331,6 +332,22 @@ function getAnalyticsPageName() {
 }
 
 /**
+ * gets the tab name for analytics
+ */
+function getTabProperty() {
+  const url = window.location.toString()
+  if (url.includes('step=')) {
+    // if in an upload wizard tab
+    return url.replace(/^(.*?)?step=/, '').trim().replace(/\s/g, '-')
+  } else if (url.includes('#study')) {
+    // if in a study tab
+    return url.replace(/^(.*?)#/, '').trim().replace(/\s/g, '-')
+  } else {
+    return ''
+  }
+}
+
+/**
  * Log metrics to Mixpanel via Bard web service
  *
  * Bard docs:
@@ -339,13 +356,18 @@ function getAnalyticsPageName() {
  * @param {String} name
  * @param {Object} props
  */
-export function log(name, props={}) {
+export function log(name, props = {}) {
   props = Object.assign(props, {
     appId: 'single-cell-portal',
     appPath: getAnalyticsPageName(),
     appFullPath: getAppFullPath(),
     env
   }, getDefaultProperties())
+
+  const tab = getTabProperty()
+  if (tab) {
+    props['tab'] = tab
+  }
 
   props['timeSincePageLoad'] = Math.round(performance.now())
 
@@ -418,7 +440,7 @@ export function log(name, props={}) {
   firing the log event
 */
 export function startPendingEvent(
-  name, props={}, completionTriggerPrefix, fromPageLoad
+  name, props = {}, completionTriggerPrefix, fromPageLoad
 ) {
   const startTime = fromPageLoad ? 0 : performance.now()
   const pendingEvent = {
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -366,7 +366,7 @@ export function log(name, props = {}) {
 
   const tab = getTabProperty()
   if (tab) {
-    props['tab'] = tab
+    props['referringTab'] = tab
   }
 
   props['timeSincePageLoad'] = Math.round(performance.now())
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -335,8 +335,8 @@ function getAnalyticsPageName() {
  * gets the tab name for analytics
  */
 function getTabProperty() {
-  if (window.location.href.match(/\?step=/)) {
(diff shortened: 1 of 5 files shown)


app/javascript/lib/metrics-api.js
```diff
@@ -330,6 +331,22 @@ function getAnalyticsPageName() {
...
+  const url = window.location.toString()
+  if (url.includes('step=')) {
+    // if in an upload wizard tab
+    return url.replace(/^(.*?)?step=/, '').trim().replace(/\s/g, '-')
+  } else if (url.includes('#study')) {
+    // if in a study tab
+    return url.replace(/^(.*?)#/, '').trim().replace(/\s/g, '-')
+  } else {
+    return ''
+  }
```
eweitz, line 347: Blocker: I believe this logs an empty string as the value for the tab, which seems useless.  The `if (tab)` on line 386 would also be superfluous in this case.

Instead, let's either omit the tab property if it's empty, or remove that `if (tab)` condition.  The former seems preferable.

```suggestion
  } else {
    return null
  }
```

app/javascript/lib/metrics-api.js
```diff
@@ -339,14 +356,19 @@ function getAnalyticsPageName() {
  * @param {String} name
  * @param {Object} props
  */
-export function log(name, props={}) {
+export function log(name, props = {}) {
```
eweitz, line 359: I hadn't realized [JavaScript conventionally puts a space around "=" in default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)!  That differs from Python and Ruby convention, unfortunately, but I agree we should use this style in JS.

app/javascript/lib/metrics-api.js
```diff
@@ -330,6 +331,22 @@ function getAnalyticsPageName() {
   return getSCPContext().analyticsPageName
 }
 
+/**
+ * gets the tab name for analytics
+ */
+function getTabProperty() {
```
bistline, line 337: Non-blocking: While this totally works, it seems overly complicated.  Using `window.location.hash` will always return the open tab as we have JS that will append this to the URL on every click.  In the case of the upload wizard, I think there's a simpler way to get the info.  For the final case, `window.location.hash` returns a string, so the `else` clause will cover that as well and return `''`.
```suggestion
function getTabProperty() {
  if (window.location.href.match(/\?step=/)) {
    return window.location.href.split('?step=')[1]
  } else {
    return window.location.hash.replace(/#/, '')
  }
}
```

bistline (approved):
Non-blocking suggestion for a simpler version of `getTabProperty()`

app/javascript/lib/metrics-api.js
```diff
@@ -339,14 +356,19 @@ function getAnalyticsPageName() {
...
   props = Object.assign(props, {
     appId: 'single-cell-portal',
     appPath: getAnalyticsPageName(),
     appFullPath: getAppFullPath(),
     env
   }, getDefaultProperties())
 
+  const tab = getTabProperty()
+  if (tab) {
+    props['referringTab'] = tab
```
eweitz, line 369: Blocker: We use `referringDomain` to mean "the _previous_ domain".  I could see analysts interpreting "referring tab" as "the previous tab"; that's how I first interpreted it.

So for consistency and simplicity, I think we should go with:

```suggestion
    props['tab'] = tab
```

app/javascript/lib/metrics-api.js
```diff
@@ -330,6 +331,22 @@ function getAnalyticsPageName() {
...
+ * gets the tab name for analytics
+ */
+function getTabProperty() {
+  const url = window.location.toString()
+  if (url.includes('step=')) {
+    // if in an upload wizard tab
+    return url.replace(/^(.*?)?step=/, '').trim().replace(/\s/g, '-')
+  } else if (url.includes('#study')) {
+    // if in a study tab
+    return url.replace(/^(.*?)#/, '').trim().replace(/\s/g, '-')
```
eweitz, line 344: I hadn't considered this implementation route, but it strikes me as reasonable!  

It's a bit less directly connected to the UX (and thus a bit less intuitive for analysts like, say, a PM), but also likely more stable than using some normalized string version of the tab's displayed text.

It'd be nice if the tab's displayed text were also logged, but if that's substantial effort then maybe it's worth deferring.

eweitz (changes requested):
This will really enrich our capability to assess upload wizard UX!

I noted two trivial blockers, and suggested a non-blocking nice-to-have.

app/javascript/lib/metrics-api.js
ehanna4: An empty string should be false-y same as null ([link](https://frontend.turing.edu/lessons/module-1/js-truthy-falsy-expressions.html?ads_cmpid=6451354298&ads_adid=76255849919&ads_matchtype=b&ads_network=g&ads_creative=517671727591&utm_term=&ads_targetid=dsa-19959388920&utm_campaign=&utm_source=adwords&utm_medium=ppc&ttv=2&gclid=Cj0KCQjw5oiMBhDtARIsAJi0qk15NUZ5cRTIFFBAqeQxnfGzlZ6nxKfsANGro2qJ6XNqVxGs5kubivsaAiNEEALw_wcB)) but I agree that `null` is a much better/clearer return value for this

app/javascript/lib/metrics-api.js
ehanna4: I went back and forth on this, because if you click from one tab to another in the wizard the tab you clicked _from_ gets logged as `referringTab` so I thought it would make it less confusing for that, but I see what you mean about that being less clear for button clicks within a tab. I'll look into making this most clear

app/javascript/lib/metrics-api.js
ehanna4: We can alter the display name in Mixpanel so someone like a PM can have the nice human styled names for these properties even if the raw data version is less nice

app/javascript/lib/metrics-api.js
ehanna4: Oh nice! much cleaner!

eweitz (approved):

app/javascript/lib/metrics-api.js
eweitz: By "the tab's displayed text", I mean the property value, not the property name.

Needing to mentally map the tab name to something that appears in the URL (rather than the main GUI) isn't especially intuitive, and I could see folks not making the connection and getting stuck.

devonbush (approved):
Looks good as is.  But I wonder if we should just rename 'step' to 'tab' in the upload wizard so that it's clearer?  And then when we make other tabbed pages with a similar pattern, we wouldn't be stuck calling them 'steps'

app/javascript/lib/metrics-api.js
eweitz: Ah, I didn't know empty string was falsy -- thanks!

devonbush (approved):

eweitz:
I've confirmed this works in staging.

PR #1235, opened 2021-11-03
Fixing broken COVID-19 view (SCP-3854)

bistline:
This update removes obsoleted feature flag code on the special COVID-19 view.  It caused unauthenticated users to not be able to view the COVID-19 collection, and instead saw the generic 500 error page.  This code is no longer used anywhere, and is superseded by [this data](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/views/layouts/application.html.erb#L236) in the main application layout.

MANUAL TESTING
1. Boot as normal, and click the "Browse Collections" button w/o signing in
2. Select the COVID-19 collection
3. Confirm the page renders and executes the preset search for COVID-19 studies (you may not have any in your local instance, and could see an empty search)

This PR satisfies SCP-3854.

### DIFF
--- a/app/views/site/covid19.html.erb
+++ b/app/views/site/covid19.html.erb
@@ -61,8 +61,6 @@
   </div>
 </div>
 
-<input type="hidden" id="feature-flags" value='<%= current_user.feature_flags_with_defaults.to_json %>'/>
-
 <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
     $(document).ready(function() {
       var main = $('.home-page-fix');


ehanna4 (approved):

eweitz (approved):

bistline:
All of our builds are now erroring with the same issue of timeouts - this appears to be affecting all our PRs right now.  Hopefully this will take care of itself, otherwise we'll need to open a ticket w/ Travis.

app/views/site/covid19.html.erb
```diff
@@ -61,8 +61,6 @@
   </div>
 </div>
```
devonbush, line 62: Ah, yup, this was my fault. But we should update this to match `application.html.erb` rather than deleting it.  that will keep things in sync between the searches on the covid space and the regular home page
```suggestion
</div>

<input type="hidden" id="user-provider-info" value='<%=
  {
    email: current_user&.email,
    featureFlagsWithDefaults: FeatureFlaggable.feature_flags_for_instances(@selected_branding_group, current_user)
  }.to_json %>'/>
```

devonbush (changes requested):

devonbush (approved):

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/single_cell/covid19).

PR #1236, opened 2021-11-03
Frontend cache no longer caches subsampled data (SCP-3803)

devonbush:
(SCP-3803) Because we subsample each annotation separately to ensure feature preservation, this means we cannot use caching on subsampled data -- the cells and cell order may be totally different. 

This means we can only trust the cache where the subsample is explicitly 'all' -- since the default subsample may or may not be 'all'.  Note that that raises the priority of preloading the 'explore_info' data onto the page--that change would enable us to never have to just request the 'default'.  By always knowing exactly what we're requesting, we can then cache more effectively (and eliminate some brutally complex caching scenarios)

TO TEST:
1. load human raw counts synthetic study in explore tab
2. set subsampling to 1000
3. change annotation to 'subcluster'
4. observe the 8 colors are separated 2 per quadrant (without this bugfix, they would be spread across all quadrants)

### DIFF
--- a/app/javascript/components/explore/plot-data-cache.js
+++ b/app/javascript/components/explore/plot-data-cache.js
@@ -289,7 +289,7 @@ export function createCache() {
     // we don't cache anything for annotated/correlated scatter since the coordinates are different per annotation/gene
 
     if (!isAnnotatedScatter && !isCorrelatedScatter) {
-      if (subsample != 'all') {
+      if (subsample !== 'all') {
         // we also don't cache for subsampled views, since the cells and ordering may be different across annotations
         fields.push('coordinates', 'cells', 'annotation')
         if (genes.length) {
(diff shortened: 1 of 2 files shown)


app/javascript/components/explore/plot-data-cache.js
```diff
@@ -288,12 +287,21 @@ export function createCache() {
...
     const promises = [] // API call promises
     // we don't cache anything for annotated/correlated scatter since the coordinates are different per annotation/gene
+
     if (!isAnnotatedScatter && !isCorrelatedScatter) {
-      const cacheEntry = cache._findOrCreateEntry(studyAccession, cluster, subsample)
-      Fields.cellsAndCoords.addFieldsOrPromise(cacheEntry, fields, promises)
-      Fields.annotation.addFieldsOrPromise(cacheEntry, fields, promises, annotation.name, annotation.scope)
-      if (genes.length) {
-        Fields.expression.addFieldsOrPromise(cacheEntry, fields, promises, genes, consensus)
+      if (subsample != 'all') {
```
eweitz, line 292:
```suggestion
      if (subsample !== 'all') {
```

eweitz (approved):
Please add manual testing instructions, though.

bistline (approved):

devonbush:
Jon got a clean build locally, build hanging in Travis confirmed as unrelated to this change

PR #1239, opened 2021-11-05
Use Sentence Case (SCP-3785)

ehanna4:
Following the UX guidelines outlined for Terra here: https://www.sketch.com/s/cca3a834-985b-4165-bbae-21f615aebd2d make things like table columns, buttons, tabs, form titles, etc. use sentence casing

To test: Load the UI and click around observing buttons, form titles, tabs. etc. should all use sentence casing ("Looks like this" rather than "Looking Like This")

Examples:
Screen Shot 2021-11-05 at 12 28 07 PM
Screen Shot 2021-11-05 at 12 24 29 PM


bistline (changes requested):
I think the prevalence of title case headers in this codebase shows which form I've preferred historically, but I get this is an effort for standardization, so I'll get on board.  To that end - if you select items from the profile (top-right) menu when you're signed in (especially as an admin), most of the page headers have not been converted.  For example:

My Annotations
My Terra Billing Projects
Admin Control Panel
Preset Searches
...

If you look at all of the `index.html.erb` views, you'll find the majority of these `h1` tags.  Some have `h2` or `h3` sub-headers as well that should be converted.

devonbush (approved):
Looks good, thanks for kicking off this effort.  It's not critical to me that this gets all changes everywhere--I'm fine if a few slip through the cracks now to be caught later.  Agree this is a great move towards standardization, and the important thing is to make sure we keep this in mind going forward

eweitz (approved):
I approve contingent on SCP-3785 being done this sprint or next, as this will almost break some (perhaps many) queries in Mixpanel.  

Beyond that and Jon's notes, this looks good!  I'm glad we're aligning with [our partners](https://github.com/broadinstitute/single_cell_portal_core/pull/1147#pullrequestreview-749553117) on this aspect of style.

bistline (approved):

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/single_cell).

PR #1240, opened 2021-11-08
Resize legend upon resizing scatter plot (SCP-3795, SCP-3807)

eweitz:
This resizes the legend upon resizing the scatter plot.   Previously (#1211), the legend width and height did not update upon searching a gene, nor upon manually resizing the window.  Now it resizes to better fit updated dimensions.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/140761314-1f3bd940-ebcc-4304-a39b-a4ec4476cae9.mov

An automated test was also added.  Tests involving window size seem very complicated, so I skip that here.  Instead, this new test adds automatic verification of the legend itself, as well as a basic legend interaction.  

The "mouse salmonella" synthetic study has also been enriched to give an easily reproducible test case for a cluster group with many labels, include long labels and labels with unintuitive formatting.  (This will be more relevant for my next PR to handle long labels, but seemed unproblematic to include here.)

To test:
* Go to a study with > 2 labels in a group-based annotation. Labels should ideally be short.
* Resize window height
* Verify legend resizes appropriately
* Resize window width
* Verify legend resizes appropriately
* Search a gene
* Verify legend resizes appropriately

This satisfies SCP-3795 and SCP-3807.


app/javascript/styles/_scatterPlot.scss
```diff
@@ -0,0 +1,80 @@
...
+}
+
+.scatter-graph {
+  margin-left: 40px;
+  display: flex;
+}
+
+.scatter-legend {
+  float: right;
+  overflow: scroll;
```
bistline, line 13: While having the visual indication that the legend can scroll is nice, when it doesn't need to scroll, setting `overflow: scroll` can have weird behavior depending on browser/OS.  I noticed in your video that the scrollbars don't show, but on my machine they are visible the entire time, even when not needed.  This persists both on my external monitor and laptop screen (even when the external is disconnected).

Screen Shot 2021-11-08 at 10 09 22 AM

bistline (approved):
Code looks good, but we might want to have one more round of revision/discussion re: legend scrollbar behavior.

test/js/explore/scatter-tab.test.js
```diff
@@ -182,4 +185,81 @@ describe('getNewContextMap correctly assigns contexts', () => {
...
+
+    // findByTestId would be more kosher, but the numerical leaf ("1") is
+    // assigned randomly in `ScatterPlot`.  This ID has been observed with the
+    // value "study-scatter-25", but relying on that seems brittle.
+    //
+    // Consider changing `ScatterPlot` to use a deterministic `data-testid`.
+    // await screen.findByTestId('study-scatter-1-legend')
+
+    await waitFor(() => {
+      container.querySelectorAll('#study-scatter-1-legend').length > 0
```
devonbush, line 249: Can this be `container.getElementById('#study-scatter-1-legend') != null`? If so, I think it's cleaner

test/test_data/mock_responses/scatter-plot.test-data.js
```diff
@@ -0,0 +1,819 @@
+/** @fileoverview Mock responses for scatter plot tests
```
devonbush, line 1: To me it seems a little more expensive to maintain than we ned to have a full-fledged mock response just to test different label configurations--now any time any part of the explore API changes, we have to update this file too.  I would move the existing MOCK_CLUSTER_RESPONSE to this file, and then make these responses just diffs of it.  I don't feel too strongly about this though, since it is nice having the MOCK_CLUSTER_RESPONSE in the same file as the tests...

```
MANY_LABELS_CLUSTER_RESPONSE = Object.assign(_cloneDeep(MOCK_CLUSTER_RESPONSE), {
   <<extra labels>>
})

**Review by devonbush — APPROVED** [review]
Good stuff, as usual, I defer to you whether you want to tackle the prod blockers in this PR or the merge down to development

File: app/javascript/components/visualization/ScatterPlot.js [hunk 4]
```diff
@@ -196,7 +208,7 @@ function RawScatterPlot({
     if (scatterData && !isLoading) {
       processScatterPlot()
     }
-  }, [shownTraces])
+  }, [shownTraces, dimensionProps])
```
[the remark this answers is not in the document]
eweitz: Good point regarding performance.  Decoupling as you (rightly) propose entails a non-trivial refactor of `processScatterPlot`.  I've opened SCP-3865 to track that.

test/js/explore/scatter-tab.test.js
eweitz: Surprisingly to me, it cannot: `TypeError: container.getElementById is not a function`.

app/javascript/styles/_scatterPlot.scss
eweitz: Could you try again?

Thanks for demo-ing the issue for me.  Devon suggested `overflow: auto`, and I've updated accordingly.

I suspect you have “Show scroll bars” set to “Always” in System Preferences -> General.  I just updated my preferences to that, and could reproduce the issue, and confirmed `overflow: auto` fixes it for me.  [Here's a detailed exploration](https://kilianvalkhof.com/2021/css-html/you-want-overflow-auto-not-overflow-scroll/) that helped me reliably reproduce what you saw.

bistline: Auto works!  Totally solved the issue; scrollbars only show when needed.

bistline: Also, my scrollbar preference is set to `Automatically based on mouse or trackpad`

test/test_data/mock_responses/scatter-plot.test-data.js
eweitz: Testing maintainability is complex.  Diff'd mocks strike me as harder reason about in terms of fidelity to real data.  `MOCK_CLUSTER_RESPONSE` and `MANY_LABELS_MOCKS.CLUSTER_RESPONSE` are from two different studies, and using diff'd mocks instead of full mocks would require the maintainer to build more context and reason more about whether the diffs are really representative.

I'll keep an eye on these to assess where the two approaches help and hurt maintainability.

PR #1241, opened 2021-11-08
Handle many long, odd labels in custom scatter plot legend (SCP-3818)

eweitz:
This refines the new custom group scatter plot legend to handle legends that have labels that are long or many in number.  It takes this by using hiding text that would overflow behind an ellipsis (...), and showing the full text upon hovering.  It also visually compresses related content like "(52 points)" to simply "52", and likewise showing more context on hover.

Here's a video of how it looks:

https://user-images.githubusercontent.com/1334561/140769939-c40a795e-4efb-4634-b1a8-39adb00c79ab.mov

To accommodate label sets of cells that have potentially many cells, while ensuring the number of points is always displayed, labels for sets with _few_ cells are often truncated before they would overlap.  The fixed width handles sets with up to 7 digits (i.e. single-digit-millions of cells).

Automatically testing such layout handling is complicated, so I've added a synthetic dataset to ease manually testing.

To test:
* Populate "mouse_salmonella" study, e.g. via `echo "SyntheticStudyPopulator.populate('mouse_salmonella')" | bundle exec rails console -e development`.
* Navigate to it in web browser
* In "Clustering" dropdown menu, select "cluster_many_long_odd_labels.tsv"
* Verify long labels do not overflow the legend, and each are contained in one line
* Hover over a truncated label
* Verify full label appears nearby in small box
* Hover over number at right in a label
* Verify it shows the number of points with more context

This satisfies SCP-3818.


app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -196,7 +208,7 @@ function RawScatterPlot({
     if (scatterData && !isLoading) {
       processScatterPlot()
     }
```
devonbush: Consolidating the useEffects as you've done here makes me worry about performance.  Now we have to rebuild all the traces (which now takes a little longer for some plots thanks to the expression sorting) every time the graph gets resized.  We could include caching and if/else optimizations inside processScatterData, but I'd just as soon keep them separate, and have one effect for 'cosmetic change' and one for 'data change'

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -94,7 +94,10 @@ function LegendEntry({
       onClick={() => toggleSelection()}
     >
       <div className="scatter-legend-icon" style={iconStyle}></div>
-      <div className="scatter-legend-entry">{entry}</div>
+      <div className="scatter-legend-entry">
+        <span className="legend-label" title={entry}>{entry}</span>
+        <span className="num-points" title={`${numPoints} points have this label`}>{numPoints}</span>
```
devonbush, line 99: I wonder if 'X points are in the {entry} group'  or just 'X points are in this group' might be a clearer title.  Probably a good demo mini-discussion (the former might run into length issues)

devonbush (approved):

app/javascript/styles/_scatterPlot.scss
```diff
@@ -63,6 +66,21 @@
 .scatter-legend-entry {
   position: relative;
   top: 5px;
+  overflow: scroll;
```
bistline, line 69: This has the same problem with `overflow` as the other PR - this should be set to `auto`.

bistline (approved):
Approving with a caveat - all of the `overflow` values need to be set to `auto` to avoid the same issues as seen in #1240

app/javascript/styles/_scatterPlot.scss
eweitz: [Done over in the base branch](https://github.com/broadinstitute/single_cell_portal_core/pull/1240#discussion_r744824672).

eweitz: [Done over in the base branch](https://github.com/broadinstitute/single_cell_portal_core/pull/1240#discussion_r744824672).

app/javascript/components/visualization/ScatterPlot.js
eweitz: Good point regarding performance. Decoupling as you (rightly) propose entails a non-trivial refactor of processScatterPlot. I've opened SCP-3865 to track that.

app/javascript/components/visualization/controls/ScatterPlotLegend.js
eweitz: Good suggestions.  Per demo feedback, I went with "`${numPoints} points in this group`" (8855a28).

PR #1242, opened 2021-11-08
cancel upload button (SCP-3760)

devonbush:
As shown in demo, this puts a 'cancel' button where the 'delete' button was when you are uploading a file. (it does not appear if you are just saving a file).  

image

On clicking the button, a confirm dialog appears

image

This also fixes some bugs, most notably a missing negative in the feature flag branch, but also a bug where non-text, non-gzipped files could not be uploaded as they would fail validation

TO TEST:
1. Open upload wizard with the `react_upload_wizard` flag set to true
2. Go to the Miscellaneous/Other tab and select a very large file (you can create one with `mkfile -n 10g temp_10GB_file` in the terminal)
3. press "Save & Upload"
4. Observe cancel button appears in place of 'Delete'
5. Press 'Cancel' 
5. Observe confirmation dialog appears
6. Press 'Yes'
7. Observe upload stops, and file form is cleared

### DIFF
--- a/app/javascript/components/upload/MiscellaneousStep.js
+++ b/app/javascript/components/upload/MiscellaneousStep.js
@@ -13,7 +13,7 @@ const miscFileFilter = file => miscFileTypes.includes(file.file_type)
 
 export default {
   title: 'Miscellaneous / Other',
-  header: 'Documentation &amp; Other Files',
+  header: 'Documentation & Other Files',
   name: 'misc',
   component: MiscellaneousForm,
   fileFilter: miscFileFilter
--- a/app/javascript/components/upload/MiscellaneousStep.js
+++ b/app/javascript/components/upload/MiscellaneousStep.js
@@ -13,7 +13,7 @@ const miscFileFilter = file => miscFileTypes.includes(file.file_type)
 
 export default {
   title: 'Miscellaneous / Other',
-  header: 'Documentation & Other Files',
+  header: 'Documentation & other files',
   name: 'misc',
   component: MiscellaneousForm,
   fileFilter: miscFileFilter
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -16,7 +16,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCheckCircle, faExclamationTriangle, faChevronLeft, faChevronRight } from '@fortawesome/free-solid-svg-icons'
 
 import { formatFileFromServer, formatFileForApi, newStudyFileObj } from './upload-utils'
-import { createStudyFile, updateStudyFile, deleteStudyFile, fetchStudyFileInfo, sendStudyFileChunk } from 'lib/scp-api'
+import { createStudyFile, updateStudyFile, deleteStudyFile, fetchStudyFileInfo, sendStudyFileChunk, RequestCanceller } from 'lib/scp-api'
 import MessageModal from 'lib/MessageModal'
 import UserProvider from 'providers/UserProvider'
 import ErrorBoundary from 'lib/ErrorBoundary'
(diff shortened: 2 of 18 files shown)


app/javascript/components/upload/UploadWizard.js
```diff
@@ -171,10 +178,24 @@ export function RawUploadWizard({ studyAccession, name }) {
...
+    requestCanceller.cancel()
+    const fileId = requestCanceller.fileId
+
+    deleteFileFromForm(fileId)
+    // wait a brief amount of time to minimize the timing edge case where the user hits 'cancel' while the request
+    // completed.  In that case the file will have been saved, but we won't have the new id yet.  We need to wait until
+    // the id comes back so we can address the delete request properly
+    // We also don't await the delete since if it errors, it is likely because the file never got created or
+    // got deleted for other reasons (e.g. invalid form data) in the meantime
+    setTimeout(() => deleteFileFromServer(requestCanceller.fileId), 500)
```
bistline, line 192: I think this is causing the behavior I'm seeing:

1. Click `Save & Upload` on a large file
2. Wait a few seconds
3. Click `Cancel`
4. Dialog shows briefly, then automatically closes itself
5. Click `Cancel` again, and dialog stays open

I get that we have an edge case here regarding whether or not the file persisted, but the timeout is resulting in a weird UX.

bistline (approved):
Works as advertised, but I'm getting weird UX when trying to cancel an upload (relatively) soon after uploading:

https://user-images.githubusercontent.com/729968/140970834-5f76222d-0472-492a-af00-9532acf77213.mov

I think it has something to do with the `setTimeout` call, but I'm not 100% sure on that.  Either way, I don't think it's blocking because the feature works, but if there's a way to deal with this, it'd be great.

app/javascript/components/upload/UploadWizard.js
devonbush: yeah, I wonder if the dialog close happens if you happen to have the dialog open during the exact moment when an individual chunk completes.  I'll take a look

app/javascript/components/upload/MiscellaneousStep.js
```diff
@@ -13,7 +13,7 @@ const miscFileFilter = file => miscFileTypes.includes(file.file_type)
 
 export default {
   title: 'Miscellaneous / Other',
-  header: 'Documentation &amp; Other Files',
+  header: 'Documentation & Other Files',
```
eweitz, line 16:
```suggestion
  header: 'Documentation & other files',
```

eweitz: I could be missing something, and please correct me if I'm wrong Devon or @ehanna4, but it looks like this escaped the letter case sweep in #1239.  The purpose / title of the ticket motivation that (SCP-3785) was "Prefer sentence case in upload wizard".   

Whether it's here or in a separate PR before making this GA, please ensure the redesigned upload wizard uses sentence case instead of title case.

app/javascript/components/upload/UploadWizard.js
```diff
@@ -216,39 +247,45 @@ export function RawUploadWizard({ studyAccession, name }) {
...
+      return newFormState
+    })
+  }
+
+  /** deletes the file from the server */
+  async function deleteFileFromServer(fileId) {
+    await deleteStudyFile(studyAccession, fileId)
+    setServerState(prevServerState => {
+      const newServerState = _cloneDeep(prevServerState)
+      newServerState.files = newServerState.files.filter(f => f._id != fileId)
```
eweitz, line 284:
```suggestion
      newServerState.files = newServerState.files.filter(f => f._id !== fileId)
```

eweitz (approved):
Code looks good overall!  I suggested a few nit fixes.

app/javascript/components/upload/UploadWizard.js
devonbush: yup, confirmed that that's the problem -- if you have the cancel dialog open during the time when the first chunk resolves, it clears itself.  This is because the file gets updated with the new id from the server, and we use the id as the `key` property for the file list components.  So when the key changes, it gets treated as a new component, and the cancel modal open/close state gets reset.

I'm going to open a separate ticket for this, because it's a non-trivial fix to change how we're id'ing all the components.  But it's an important change since this id change issue will likely come up in other areas too.

eweitz:
I've confirmed this works in staging.  (I had to add a file extension, e.g. `mkfile -n 2g temp_2GB_file.txt`, which seems expected.)

PR #1243, opened 2021-11-09
Remove jquery-ui from Gems (SCP-3837)

ehanna4:
This removes the redundant `jquery-ui` that we had by removing it from the gems. Also removing the jquery datapicker in favor of using the form date field provided by ruby.

Tested by running the tests locally and they passed, able to pick a date, gene auto complete works, no errors in the console.

Screen Shot 2021-11-09 at 2 40 36 PM
Screen Shot 2021-11-09 at 2 37 59 PM


eweitz (approved):
Code looks good!  

In addition to ironing out possible `datepicker` errors as SCP-3837 notes, this will also slightly improve performance of all page loads -- because it means we load and parse less JavaScript.

devonbush (approved):
awesome, great simplification

bistline (approved):

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP277/chicken-overplotted#study-settings).

PR #1244, opened 2021-11-09
Upload wizard bugfixes (SCP-3758)

devonbush:
SCP-3758 Fixing bugs found during[ upload wizard testing ](https://docs.google.com/document/d/1o3YEQtYI0J9ifMekUf_ryVeHmoB02imz7A9D2xfMSnQ/edit#)

 A. Custom names are now allowed for Gene List files
 B. Bam files no longer fail validation if they don't have a gzip suffix
 C. Custom cluster names now no longer overwrite the upload_file_name

TO TEST:
1. with react_upload_wizard feature flag on, go to the upload wizard
2. Add a new cluster file, and set the cluster name to something other than the file name
3. hit Save & Upload 
4. Observe the file is saved and the filename is preserved
5. Go to the Sequence Files tab, and attempt to select the `sample1.bam` file from the mouse_brain synthetic study
6. confirm that it allows you to use that file
7. Go to the Precomputed Expression Stats tab
8. Confirm that 'name' is now independently editable outside of the file name

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -300,6 +300,12 @@ module Api
           end
         end
 
+        # Somewhere in Rails automagic, the upload filename is set as the study file name
+        # We want to have it be the upload_file_name sent by the frontend
+        if safe_file_params[:upload] && safe_file_params[:upload_file_name]
+          safe_file_params[:upload].original_filename = safe_file_params[:upload_file_name]
+        end
+
         # manually check first if species/assembly was supplied by name
         species_name = safe_file_params[:species]
         safe_file_params.delete(:species)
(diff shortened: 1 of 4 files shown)


eweitz (approved):

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -300,6 +300,12 @@ def perform_update(study_file)
           end
         end
 
+        # Somewhere in Rails automagic, the upload filename is set as the study file name
+        # We want to have it be the upload_file_name sent by the frontend
+        if safe_file_params[:upload] && safe_file_params[:upload_file_name]
+          safe_file_params[:upload].original_filename = safe_file_params[:upload_file_name]
```
bistline, line 306: I wish we didn't have to do this, but I agree this is a classic case of Rails completely obfuscating functionality and making it impossible to know what's actually going on.

bistline (approved):
All the fixes appear to work as advertised.

eweitz:
This seems to have a (perhaps orthogonal) bug in step 5.  I followed up with Devon.

Otherwise, I've verified this works as expected [on staging](https://singlecell-staging.broadinstitute.org/single_cell/studies/5f32debdb067340cbb7cf742/upload)

PR #1245, opened 2021-11-10
Scatter plot legend analytics

eweitz:
This adds analytics for the new custom legend for reference group scatter plots.  Clicks on legend entries indicate the user is gaining scientific insight, especially so in spatial transcriptomics.  [Tracking these clicks](https://mixpanel.com/s/4E3GG3) is essential to know if our redesigned UX achieves our expected outcomes.

Here's an example showing these analytics:

https://user-images.githubusercontent.com/1334561/141139596-aa946425-7e37-4ad7-bbe1-b63cf88ffe92.mov

To test:
* Open study that has a cluster group plot
* Go to Network panel in DevTools
* In "Filter" box, enter "bard method:POST"
* Click a label in the legend
* Verify that an event named `click:scatterlegend:single` is posted
* Verify that event includes properties `label`, `numPoints`, `numLabels`, `wasShown`, `iconColor`, and `hasCorrelations`

This helps satisfy SCP-3712.

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -236,12 +236,8 @@ function RawScatterPlot({
   useEffect(() => {
     const jqScatterGraph = $(`#${graphElementId}`)
     jqScatterGraph.on('plotly_selected', plotPointsSelected)
-    jqScatterGraph.on('plotly_legendclick', logLegendClick)
-    jqScatterGraph.on('plotly_legenddoubleclick', logLegendDoubleClick)
     return () => {
       jqScatterGraph.off('plotly_selected')
-      jqScatterGraph.off('plotly_legendclick')
-      jqScatterGraph.off('plotly_legenddoubleclick')
       Plotly.purge(graphElementId)
     }
   }, [])
@@ -597,22 +593,4 @@ export function get3DScatterProps({
 /** get the appropriate plotly dragmode option string */
 function getDragMode(isCellSelecting) {
   return isCellSelecting ? 'lasso' : 'lasso, select'
-}
-
-let currentClickCall = null
-
-/** we don't want to fire two single click events for a double click, so
- * we wait until we've confirmed a click isn't a double click before logging it.
- * Unfortunately (despite the docs indicating otherwise), there doesn't seem to be
- * a way of getting the text of the clicked annotation
- */
-function logLegendClick(event) {
-  clearTimeout(currentClickCall)
-  currentClickCall = setTimeout(() => log('click:scatterlegend:single'), 300)
-}
-
-/** log a double-click on a plotly graph legend */
-function logLegendDoubleClick(event) {
-  clearTimeout(currentClickCall)
-  log('click:scatterlegend:double')
 }
--- a/app/javascript/components/visualization/controls/ScatterPlotLegend.js
+++ b/app/javascript/components/visualization/controls/ScatterPlotLegend.js
@@ -91,7 +91,7 @@ function LegendEntry({
     updateShownTraces(label, wasShown, numLabels)
 
     const legendEntry = {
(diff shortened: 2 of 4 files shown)


app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -102,6 +111,20 @@ function LegendEntry({
   )
 }
 
+/** Handle click on "Show all" or "Hide all" button */
+function showHideAll(showOrHide, labels, updateShownTraces) {
```
ehanna4, line 115: I'm seeing double events for the show/hide all. One is these events that are added here and the other is a `click:link` event. Not sure it's a problem but just wanted to call it out that it might be adding noise to have both?

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -598,21 +594,3 @@ export function get3DScatterProps({
 function getDragMode(isCellSelecting) {
   return isCellSelecting ? 'lasso' : 'lasso, select'
 }
-
```
devonbush, line 601: hooray for getting rid of this crap!

app/javascript/components/visualization/controls/ScatterPlotLegend.js
devonbush: concur.  the links already have a data-analytics-name on them, so I don't think the extra property logging is worth the extra complexity

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -84,8 +87,14 @@ function LegendEntry({
...
-    const state = !isShown
-    updateShownTraces(label, state, numLabels)
+    const wasShown = !isShown
+    updateShownTraces(label, wasShown, numLabels)
+
+    const legendEntry = {
+      label, numPoints, numLabels, wasShown, iconColor,
+      hasCorrelations
+    }
+    log('click:scatterlegend:single', legendEntry)
```
devonbush, line 97: I'm not sure we need a custom event logging code here--this is what logClickOther() was built for.  You can add the `log-click` class to the legend entries and a data-analytics-name and achieve what you want
  Incidentally, the fact that we have to do either of these approaches seems like a side effect of not using links/buttons for the entries.  Is that an accessibility concern?  Should we change those legend entries to be links or at least have role="button"?

devonbush (changes requested):
I always feel weird arguing for less analytical effort -- "Information is bad -- Knowledge is an illusion, we should have less of it".  But I think there's a lot of long term benefit to keeping our analytics lightweight and standardized.  So unless there's a clear need for the extra logging, I would just use the existing click event logging framework

app/javascript/components/visualization/controls/ScatterPlotLegend.js
eweitz: Good catch!  I've eliminated the double logging (94522ee).

app/javascript/components/visualization/controls/ScatterPlotLegend.js
eweitz: I hadn't considered tab-navigability!  I've added `role="button"` as you suggest (94522ee).

Regarding custom event logging: this does two things.  

First, trivially, it eases product analysis by giving [seamless compatibility in Mixpanel for the old event that tracks legend clicks](https://mixpanel.com/s/4E3GG3).  We could adjust the query, of course; or we could just as simply do what's already done here.

Second, more substantively, logging this as a custom event gives us a way to track related properties.  The threshold for what counts as "related" is subjective, but I'd suggest that data that's already at hand (e.g. `numPoints`, `hasCorrelations`) qualifies and has analytics value.

ehanna4 (approved):
Looks good, tested locally

devonbush (approved):

PR #1250, opened 2021-11-11
Adding mtx file testing

devonbush:
Testing of happy path of mtx file upload.  Creates a new upload-wizard-test-utils.js file to consolidate common test patterns.  

Also fixes a UX papercut where the message "A file named already exists in your study" would appear if you hovered over the barcodes file save button before selecting a file.  This was a result of the features and barcodes file objects being created, both with a name of '', and therefore duplicating each other.  

TO TEST:
1. Run 'yarn test' -- confirm everything passes
2. in the new upload wizard, select 'Sparse Matrix (.mtx)'
3. Hover over the 'Save' buttons for the features and barcodes files.  Confirm two messages appear: 'You must save the parent file' and 'You must select a file'. but NOT 'a file named already exists'


test/js/upload-wizard/upload-new-study.test.js
```diff
@@ -1,60 +1,48 @@
...
-    const studyInfoSpy = jest.spyOn(ScpApi, 'fetchStudyFileInfo')
-    // pass in a clone of the response since it may get modified by the cache operations
-    studyInfoSpy.mockImplementation(params => {
-      const response = _cloneDeep(EMPTY_STUDY)
-      return Promise.resolve(response)
-    })
+  beforeAll(() => {
+    jest.restoreAllMocks()
+    // This test is long, so allow extra time
+    jest.setTimeout(30000)
```
eweitz, line 19: Blocker: 30 seconds is > 10x too long.  Running all JS tests (`time yarn test`) without this just took me 28 seconds, not including this.  This would let that double, which is a developer-experience bug.

If fixing that DX bug is infeasible (see other more specific Blocker comment), then please at least add more detail on why these tests are so long in a code comment here.  Maybe also move them to another directory, or otherwise give an easy hook to exclude them so running big batches of local JS tests doesn't take twice as long by default.  (We'd still always run them in CI, of course.)

test/js/upload-wizard/upload-new-study.test.js
```diff
@@ -1,60 +1,48 @@
...
-    await testProcessedUpload({ createFileSpy, saveButton })
-    await testMetadataUpload({ createFileSpy, saveButton })
-    await testClusterUpload({ createFileSpy, saveButton })
+    await testRawCountsUpload({ createFileSpy })
+    await testProcessedUpload({ createFileSpy })
+    await testMetadataUpload({ createFileSpy })
+    await testClusterUpload({ createFileSpy })
+    await testSpatialUpload({ createFileSpy })
+    await testCoordinateLabelUpload({ createFileSpy })
+    await testSequenceFileUpload({ createFileSpy })
```
eweitz, line 40: Blocker: these shouldn't need to run synchronously, which I suspect is the cause of their excessive slowness.  Could these be wrapped in [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) or [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) instead?

Or could these [Jest tests run in parallel or conconcurrently](https://freecontent.manning.com/the-value-of-concurrency-in-tests/)?

test/js/upload-wizard/upload-wizard-test-utils.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
```
eweitz, line 1: I'm a fan of test helpers!

eweitz (changes requested):
I'm glad test coverage is improving -- it's a main benefit of the upload refactor!

I have blocking concerns about performance, though.  Details below.  (I had also begun marking up letter case bugs, but I see there are conflicts with base branch that I suspect fix these.)

test/js/upload-wizard/upload-new-study.test.js
devonbush: The whole idea of this test is to test that you can create a study as a user would.  And a user would run them in sequence--and the state of the wizard at each step depends on previous steps.  I agree that we'll want to start parallelizing our jest suites eventually though

test/js/upload-wizard/upload-new-study.test.js
devonbush: Good point, I was just noticing occasional timeouts at 5 seconds, and I went very conservative so as to avoid sporadic fails.  I'll drop it to 10 seconds.

devonbush: I also think that we should be prepared for our front-end testing to eventually take almost as long as our backend testing -- Right now our frontend coverage is somewhat minimal, and as you note, it still takes 30 seconds.  I like your later comment about looking to parallelize these

devonbush: Thinking more about this, I realize that I wasn't really doing any explicit testing of the wizard state over time to benefit from the serialized form.  So I've added more explicit checks after everything is uploaded

test/js/upload-wizard/upload-new-study.test.js
```diff
@@ -1,60 +1,48 @@
...
-  it('renders the raw counts form and allows updates and save', async () => {
-    const studyInfoSpy = jest.spyOn(ScpApi, 'fetchStudyFileInfo')
-    // pass in a clone of the response since it may get modified by the cache operations
-    studyInfoSpy.mockImplementation(params => {
-      const response = _cloneDeep(EMPTY_STUDY)
-      return Promise.resolve(response)
-    })
+  beforeAll(() => {
+    jest.restoreAllMocks()
+    // This test is long, so allow extra time
```
eweitz:
```suggestion
    // This test is long, as it uploads in sequence like users would.  So allow extra time.  
```

Thanks for clarifying the intent, and thus why the tests run long.  Briefly documenting that here will (one hopes) avoid setting off my alarm bells next time I see this.

test/js/upload-wizard/upload-new-study.test.js
eweitz: Great comments -- that works.  I've suggested a quick comment that would help.

test/js/upload-wizard/upload-new-study.test.js
eweitz: Makes sense.  Per your note, I've suggested expanding a code comment slightly to ensure it's clear to maintainers.

eweitz (approved):
Looks good!  Thanks for explaining, and being open to (later) parallelization.  I suggested a trivial clarification in a comment; non-blocking.

bistline (approved):

test/js/upload-wizard/upload-new-study.test.js
```diff
@@ -1,34 +1,24 @@
...
+
+const processedFileName = 'example_processed_dense.txt'
+const rawCountsFileName = 'example_raw_counts.txt'
 
 describe('creation of study files', () => {
   beforeAll(() => {
     jest.restoreAllMocks()
-    // This test is long, so allow extra time
-    jest.setTimeout(30000)
+    // This test is long--running all steps in series, so allow extra time
```
devonbush, line 20:
```suggestion
    // This test is long--running all steps in series as if it was a user uploading a new study from scratch--so allow extra time
```

PR #1252, opened 2021-11-12
Default to 10 results instead of 5 (SCP-3769)

ehanna4:
previously we showed 5 results per page, this PR changes that to 10.

To test pull this branch and view the home page and see that there are 10 results on the page rather than 5.

Screen Shot 2021-11-11 at 2 07 18 PM

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -38,7 +38,7 @@ class Study
 
   # pagination
   def self.per_page
-    5
+    10
   end
 
   # associations and scopes


eweitz (approved):
Nice!

Tangentially, I wonder if doubling the number of items on each page makes search and pagination much slower.  I recall Jon noting performance issues with Azul search integration, and if this will make things half as fast as that already-concerningly-slow change, then maybe related optimizations are worth prioritizing as a separate task.

devonbush:
Yup, we should definitely measure this.

devonbush (approved):

bistline (approved):

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/).

PR #1253, opened 2021-11-12
Temporary fix for the workspace deletion integration test (SCP-3886)

ehanna4:
As of 11/10 the integration tests have failed on deleting the workspace due to the response message from Rawls/Orch being strange. After conversing with App Services and checking the DB etc. it is concluded it is likely the workspaces are indeed being deleted but that the messaging is just getting messed up. Therefore while [App Services sorts out the workspace deletion infrastructure](https://broadworkbench.atlassian.net/browse/AS-1022?atlOrigin=eyJpIjoiODNjYThhOTEwOTYyNDNkZmJiNWIxZDIxOTFlNDE4MjMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) this keep the test from always failing while still checking for a good response code.

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -385,7 +385,7 @@ function addHoverLabel(trace, annotName, annotType, genes, isAnnotatedScatter, i
   trace.hovertemplate = groupHoverTemplate
 }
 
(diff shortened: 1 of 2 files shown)


devonbush (approved):

devonbush:
Thanks for investigating this -- will be great to have it fixed!

bistline (approved):

bistline:
Love the green build!

PR #1256, opened 2021-11-15
fixing UX papercuts from file id changes (SCP-3760)

devonbush:
This fixes a bug Jon found in the cancel button, where the cancel dialog would close of its own accord if it was open during the time the first chunk upload completed.  This also fixes a related bug where if a user was uploading a new file other than the first file in the list, after the upload completed, that file would be shown as collapsed.

TO TEST:
 1. Open a study with the new upload wizard
 2. go to the Miscellaneous / other tab
 3. upload a file
 4. upload another file
 5. Confirm that after the second file upload completes, the panel of the uploaded file does not spontaneously collapse.

### DIFF
--- a/app/javascript/components/upload/ClusteringStep.js
+++ b/app/javascript/components/upload/ClusteringStep.js
@@ -91,7 +91,7 @@ export function ClusteringUploadForm({
     { clusterFiles.length > 1 && <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_CLUSTER_FILE}/> }
     { clusterFiles.map(file => {
       return <ClusteringFileForm
-        key={file._id}
+        key={file.oldId ? file.oldId : file._id}
         file={file}
         allFiles={formState.files}
         updateFile={updateFile}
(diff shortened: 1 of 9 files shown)


eweitz (approved):
Nice refinement; code looks good.

Making the same change across 9 files is a bit code smelly.  How feasible is abstraction here and in the other overlapping parameters in `*FileForm`?

devonbush:
Agree--that would be nice.  It's tricky--because it's a key property, I can't just have the ExpandedFileForm (the shared component that handles top-level rendering of the file forms) handle it.  So I'd need to create a new component like "FileListIterator" that takes a file form component and an array of file objects.  Which then gets tricky because then a general component will have to handle some form-specific properties like `rawCountsOptions` and finding the children for each sequence file.

anyway, it's probably doable with suitable passthroughs, but definitely non-trivial and I suspect the result would be code that is harder to read

bistline (approved):
Agree with Eric re: duplication, but the fix works as expected.

devonbush:
test failures are Azul

PR #1257, opened 2021-11-15
Flip the color scale (SCP-2945)

ehanna4:
This flips the colorscale so that the high expression values are the more high contrast colors and the lower expression values are lower contrast colors. This improves visual signal for the plots.

To test pull this branch and view a study with a Scatter plots for “numeric” annotations and Scatter plots for gene expression.

Be sure to look at "Reds" and also any other colorscale.

Before:
Screen Shot 2021-11-15 at 3 02 48 PM
After:
Screen Shot 2021-11-15 at 3 03 46 PM

Before:
Screen Shot 2021-11-15 at 3 37 21 PM
After:
Screen Shot 2021-11-15 at 3 38 57 PM

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -328,6 +328,7 @@ export function getPlotlyTraces({
       Object.assign(trace.marker, {
         showscale: true,
         colorscale: appliedScatterColor,
+        reversescale: true,
         color: colors,
         colorbar: { title, titleside: 'right' }
       })
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -234,10 +234,10 @@ function getLegendEntries(data, pointSize, labelCorrelations) {
 }
 
 
-  // Reverse the continous colorscale as appropriate for high contrast color to correspond to high expression
-  function shouldReverseScale(scatterColor) {
-    return scatterColor === 'Reds' ? false : true
-  }
+/** Reverse the continuous colorscale so high contrast color corresponds to high expression */
+function shouldReverseScale(scatterColor) {
+  return scatterColor !== 'Reds'
+}
 
 /** get the array of plotly traces for plotting */
 export function getPlotlyTraces({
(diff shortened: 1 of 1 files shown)


bistline (dismissed):
This seems to work for every single-color scale _except_ `Reds`.  I have no idea why, but for some reason this is inverted on my machine.  I don't have caching turned on, so it isn't that.

Examples from Monkey correlations synthetic study (I confirmed all numeric annotations are behaving this way for me)

Greys:
Screen Shot 2021-11-15 at 4 37 31 PM

Reds:
Screen Shot 2021-11-15 at 4 37 25 PM

ehanna4:
@bistline 

> This seems to work for every single-color scale except Reds. I have no idea why, but for some reason this is inverted on my machine. I don't have caching turned on, so it isn't that.

huh I'm also seeing that now, I will look into that, thanks for catching that

jlchang:
> @bistline
> 
> > This seems to work for every single-color scale except Reds. I have no idea why, but for some reason this is inverted on my machine. I don't have caching turned on, so it isn't that.
> 
> huh I'm also seeing that now, I will look into that, thanks for catching that

Historically, `reds` was the **only** color profile that worked well for gene expression plots (for some reason it had greys at zero and reds for higher values) so it's not surprising, given prior knowledge, that its color scale would be in the "wrong orientation" for a general flip of all color profiles.

Because `reds` is the default, we'd want to make sure it stays useful :)

devonbush (approved):

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -233,6 +233,12 @@ function getLegendEntries(data, pointSize, labelCorrelations) {
   return legendEntries
 }
 
+
+  // Reverse the continous colorscale as appropriate for high contrast color to correspond to high expression
+  function shouldReverseScale(scatterColor) {
+    return scatterColor === 'Reds' ? false : true
+  }
```
eweitz, line 240: More succinctly, and with formatting that doesn't throw lint warnings:

```suggestion
/** Reverse the continuous colorscale so high contrast color corresponds to high expression */
function shouldReverseScale(scatterColor) {
  return scatterColor !== 'Reds'
}
```

eweitz (approved):
Looks good!  I tested locally.  I suggest some non-functional refinements.

jlchang (approved):
Looks good on functional review. The multi-color profiles are all now oriented so lighter colors are near zero which is exactly the behavior we're hoping for. (And `reds` are now looking good too!)

PR #1262, opened 2021-11-16
fixing expression info saving bugs (SCP-3758)

devonbush:
SCP-3758 Fixes the display of legacy expression files that do not have expression_file_info on them.

This also fixes two bugs that had for the most part canceled each other out in the past.  First is that we weren't sending expression_file_info back from the server after saving a file.  And we didn't notice it because react-select apparently preserves the previous value if you pass 'undefined' as a new value, so the forms looked fine after the save, even though the underlying object was lacking the info.  

TO TEST:
 1.  Delete the expression_file_info object for an existing expression matrix. e.g. `Study.find_by(accession: 'SCP__').study_files.find_by(file_type: 'Expression Matrix').update(expression_file_info: nil)`
 2. View the study in the new upload wizard
 3. click on the Processed matrices tab
 4. observe that the file appears, but the dropdowns are not populated (except species)
 5. fill in the dropdowns, and observe the save button become enabled.
 6. Click save
 7. confirm the file remains intact with the values populated after the save is complete.
 8. update the file description, and confirm you can save again
 9. reload the page, and confirm the file appears with all saved info.

### DIFF
--- a/app/javascript/components/upload/ExpressionFileForm.js
+++ b/app/javascript/components/upload/ExpressionFileForm.js
@@ -136,7 +136,12 @@ export default function ExpressionFileForm({
 /** render a dropdown for an expression file info property */
 function ExpressionFileInfoSelect({ label, propertyName, rawOptions, file, updateFile }) {
   const selectOptions = rawOptions.map(opt => ({ label: opt, value: opt }))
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good!

bistline (approved):

devonbush:
build failures related to google service outages

PR #1265, opened 2021-11-16
Allowing for multiple collection curators, studies in multiple collections (SCP-3849, SCP-3867, SCP-3872)

bistline:
Currently, each collection may only have one "curator" (user who controls assigning studies to collections), and a study may only belong to one collection.  Also, curators are not allowed to edit the collection themselves, and must go through an SCP admin to make any updates to the collection itself (like the tag line, or visual characteristics).

This update allows for collections to have multiple curators (who can now directly edit the collection), and for studies to belong to multiple collections.  This is done by changing the associations from `has_many -> belongs_to` to `has_and_belongs_to_many` for the `Study`, `BrandingGroup` (collections), and `User` models.  This way, we can access all the studies for a collection directly from the collection itself, and vice versa.  This update also includes a new integration test suite to cover all new functionality.  In addition, a reversible data migration has been added to migrate from the existing association structure to the new one.

There is also now a visual indication on the Collections page as to whether a collection has been published or not.  This only shows if a user is signed in, and is a curator for at least one collection (otherwise, the only collections visible will all be published, and the indicator is not necessary).  Example:

Screen Shot 2021-11-18 at 10 30 39 AM

MANUAL TESTING
1. Pull this branch, and run the migration via `rails db:migrate`
2. Enter the Rails console and run `SyntheticBrandingGroupPopulator.populate_all`
3. Sign in as an admin, and select "Branding Groups" from the profile menu
4. For each new synthetic collection, in the "curators" text field, add another non-admin user account:
Screen Shot 2021-11-16 at 6 11 25 PM 
5. Sign out, and sign in as the non-admin
6. Go to the "My Studies" page, and add any study to both new collections at the same time
7. From the "Browse Collections" page, select both new collections and ensure that they contain the study you just added
8. From the "Browse Collections" page, confirm that the non-admin account can see & click the "Edit" button for the new collections:
Screen Shot 2021-11-18 at 10 28 48 AM
 
This PR satisfies SCP-3849, SCP-3867, and SCP-3872.


app/models/branding_group.rb
```diff
@@ -69,7 +86,8 @@ def set_name_as_id
   # remove branding association on delete
```
devonbush, line 86: This function is not needed -- Rails/Mongoid will automatically remove the study-side associations on destroy

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -55,12 +59,14 @@ def self.create_branding_group(branding_group_config, user)
     end
 
     branding_group = BrandingGroup.new(branding_group_config.dig('branding_group'))
-    branding_group.user ||= user
+    branding_group.user_ids = [user.id]
```
devonbush, line 62: I think it's generally considered better form to operate on the collections where possible, as that makes the corresponding triggers more efficent.
```suggestion
    branding_group.users = [user]
```

app/controllers/branding_groups_controller.rb
```diff
@@ -90,6 +101,16 @@ def destroy
...
   end
 
+  # helper to merge in the list of curator IDs into the :user_ids parameter
+  # will prevent curator from removing themselves from the collection
+  def self.merge_curator_params(curator_list, collection, user)
+    curators = curator_list.split(',').map(&:strip)
+    user_ids = curators.map { |email| User.find_by(email: email)&.id }.compact
+    # ensure current user cannot remove accidentally remove themselves from the list
+    user_ids << user.id if collection.present? && collection.users.include?(user) && !user_ids.include?(user.id)
+    user_ids
```
devonbush, line 111: Let's refactor this method to return and operate on `users` rather than user_ids.  Unless you're making high-performance plucks (and you know I loooove my plucks), it's generally better and faster to just operate on the objects themselves

app/controllers/branding_groups_controller.rb
```diff
@@ -65,6 +65,17 @@ def update
...
         # delete the param since it is not a real model param
         clean_params.delete("reset_#{image_name}")
+
+        # handle updates to the list of curators
+        curators = params[:curator_emails].split(',').map(&:strip)
+        user_ids = curators.map { |email| User.find_by(email: email)&.id }.compact
+        # ensure current user cannot remove accidentially remove themselves from the list
+        if @branding_group.users.include?(current_user)
+          user_ids << current_user.id unless user_ids.include?(current_user.id)
+        end
```
devonbush, line 75: consolidate the logic
```suggestion
        if @branding_group.users.include?(current_user) &&  !user_ids.include?(current_user.id)
          user_ids << current_user.id
        end
```

app/views/branding_groups/list_navigate.html.erb
```diff
@@ -21,10 +21,13 @@
             <% end %>
           </div>
           <div class="collections-actions">
-            <% if @editable_branding_groups.include?(branding_group) %>
+            <% if branding_group.can_edit?(current_user) %>
               <%= scp_link_to "<i class='fas fa-search'></i> Info".html_safe, branding_group_path(branding_group), class: "btn btn-xs btn-info #{branding_group.name_as_id}-show" %>
               <%= scp_link_to "<i class='fas fa-edit'></i> Edit".html_safe, edit_branding_group_path(branding_group), class: "btn btn-xs btn-primary #{branding_group.name_as_id}-edit" %>
-              <%= scp_link_to "<i class='fas fa-trash'></i> Destroy".html_safe, branding_group_path(branding_group), method: :delete, class: "btn btn-xs btn-danger delete-btn #{branding_group.name_as_id}-delete" %>
+              <% if branding_group.can_destroy?(current_user) %>
```
devonbush, line 27: I like that you've added this here.  Unguarded deletes make me nervous though, is there an easy way to put this behind a confirm dialog?

db/migrate/20211110220355_set_new_collection_associations.rb
```diff
@@ -0,0 +1,44 @@
+# since we have changed the basic association types, we can no longer use the dynamic methods of :branding_group_id or
+# :user_id to reference back to the parent models.  Since the data exists in the documents, we can call :attributes to
+# load existing data and create maps with which to set new associations
+class SetNewCollectionAssociations < Mongoid::Migration
+  def self.up
+    # create maps showing existing associations
+    study_map = Study.all.map do |s|
```
devonbush, line 7: This is going to load all studies into memory in prod, which should be ok, but just to be safe, I might rewrite (also more concisely) as
```suggestion
    study_map = ActiveRecordUtils.pluck_to_hash(Study, [:id, :branding_group_id])
    study_map.each do |study_hash|
       if study_hash[:branding_group_id].present? && study_hash[:branding_group_id].is_a?(BSON::ObjectId)
         Study.find(study_hash[:id]).branding_groups.push(BrandingGroup.find(study_hash[:branding_group_id])
     end
```

and then something similar for the users.  This will have a few more queries to the DB (two per study-branding group pair), but much easier to read, and given that we only have a couple hundred or so studies in branding groups, that's much better than loading all studies into memory

test/api/search_controller_test.rb
```diff
@@ -1,11 +1,14 @@
...
 require 'user_tokens_helper'
 require 'bulk_download_helper'
+require 'test_helper'
 
 class SearchControllerTest < ActionDispatch::IntegrationTest
   include Devise::Test::IntegrationHelpers
   include Requests::JsonHelpers
   include Requests::HttpHelpers
+  include Minitest::Hooks
+  include ::TestInstrumentor
```
devonbush, line 11: nice, thanks for updating this test

devonbush (changes requested):
This is good stuff, taking our Collections to a new level of sophistication.  I think you can simplify a lot fo this code by operating on the associations (e.g. `BrandingGroup.users`) rather than the id properties (e.g. `BrandingGroup.user_ids`).  By using the associations, Rails+Mongoid will handle the reciprocal relationship bookkeeping for you.

app/controllers/branding_groups_controller.rb
```diff
@@ -65,6 +65,17 @@ def update
         end
         # delete the param since it is not a real model param
         clean_params.delete("reset_#{image_name}")
+
+        # handle updates to the list of curators
+        curators = params[:curator_emails].split(',').map(&:strip)
```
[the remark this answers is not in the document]
bistline: That's because it is!  I refactored into a class method (`merge_curator_params`) and then forgot to remove this.

app/lib/synthetic_branding_group_populator.rb
```diff
@@ -42,7 +42,11 @@ def self.populate(branding_group_config, user: User.first)
     synthetic_studies_ids = SyntheticStudyPopulator.collect_synthetic_studies.pluck(:id)
     study_regex = /#{branding_group_config.dig('study_title_regex')}/i
     study_list = Study.where(name: study_regex, :id.in => synthetic_studies_ids)
-    study_list.update_all(branding_group_id: branding_group.id) if study_list.any?
+    study_list.each do |study|
+      study.branding_group_ids << branding_group.id
+      study.save!
+    end
+    branding_group.update(study_ids: study_list.pluck(:id)) if study_list.any?
```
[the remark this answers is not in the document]
bistline: I'll try again, but what I found was that it didn't... It certainly does when we're using the controllers to take form data - it updates both sides - but for some reason the synthetic populator didn't.  I'll take another look at it and see if I can figure out what's up with that.

app/models/branding_group.rb
```diff
@@ -60,6 +60,23 @@ def facets
...
+  # determine if user can edit branding group (all portal admins & collection curators)
+  def can_edit?(user)
+    return false if user.nil?
+
+    !!(user.admin? || users.include?(user))
+  end
+
+  # determine if a user can destroy a branding group (only portal admins)
+  def can_destroy?(user)
+    user.nil? ? false : user.admin
```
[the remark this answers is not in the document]
bistline: ooooh fancy

app/models/branding_group.rb
```diff
@@ -60,6 +60,23 @@ def facets
...
+  # list of curator emails
+  def curator_list
+    users.map(&:email)
+  end
+
+  # determine if user can edit branding group (all portal admins & collection curators)
+  def can_edit?(user)
+    return false if user.nil?
+
+    !!(user.admin? || users.include?(user))
```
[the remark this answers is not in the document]
bistline: love it

app/controllers/branding_groups_controller.rb
bistline: OK I'll try to get that working

app/views/branding_groups/list_navigate.html.erb
bistline: Yeah sure

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: FWIW I tested out all this code (minus the assignments) in production to see how well it performed.  I'm confident we won't have issues with hanging.  But I take the point.

app/models/branding_group.rb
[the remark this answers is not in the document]
bistline: This actually should be `!!(user && (user.admin? || users.include?(user)))`, because otherwise if the user is `nil` it returns `nil`.  While this is still "falsey", I prefer to constrain the response to `true` or `false`.

app/controllers/branding_groups_controller.rb
```diff
@@ -90,6 +97,16 @@ def destroy
     end
   end
 
+  # helper to merge in the list of curators into the :users parameter
+  # will prevent curator from removing themselves from the collection
+  def self.merge_curator_params(curator_list, collection, user)
+    curators = curator_list.split(',').map(&:strip)
+    users = curators.map { |email| User.find_by(email: email) }.compact
+    # ensure current user cannot remove accidentally remove themselves from the list
```
eweitz, line 105:
```suggestion
    # ensure current user cannot accidentally remove themselves from the list
```

app/controllers/branding_groups_controller.rb
```diff
@@ -90,6 +97,16 @@ def destroy
...
+    users << user if collection.present? && collection.users.include?(user) && !users.include?(user)
```
eweitz, line 106: Can this be more succinctly expressed via [safe navigation operator](https://stackoverflow.com/questions/36812647/what-does-ampersand-dot-mean-in-ruby)?

```suggestion
    users << user if collection&.users.include?(user) && !users.include?(user)
```

app/views/studies/_form.html.erb
```diff
@@ -60,9 +60,14 @@
...
-          <%= f.select :branding_group_id, options_from_collection_for_select(current_user.available_branding_groups, 'id', 'name', @study.branding_group_id),
-                       {include_blank: 'No Selected Brand'}, class: 'form-control' %>
+          <%= f.label :branding_group_ids, 'Add to a collection' %>
+          <span class='fas fa-question-circle' data-toggle='tooltip'
+                title='Add this study to a collection you are a curator of (select all that apply)'
+                data-placement='right'>
+          </span><br />
+          <%= f.select :branding_group_ids, options_from_collection_for_select(
+            current_user.available_branding_groups, 'id', 'name', @study.branding_group_ids
+          ), { include_blank: '-- No Selected Collection --' }, class: 'form-control', multiple: true %>
```
eweitz, line 70: Sentence case for consistency:

```suggestion
          ), { include_blank: '-- No selected collection --' }, class: 'form-control', multiple: true %>
```

app/views/studies/_form.html.erb
```diff
@@ -86,7 +91,7 @@
                       id: 'add-external-resource'  %>
   </div>
   <div class="form-group">
-    <%= f.submit 'Create study', class: 'btn btn-lg btn-success', id: 'save-study' %>
+    <%= f.submit 'Save Study', class: 'btn btn-lg btn-success', id: 'save-study' %>
```
eweitz, line 94: Sentence case for consistency:

```suggestion
    <%= f.submit 'Save study', class: 'btn btn-lg btn-success', id: 'save-study' %>
```

db/migrate/20211110220355_set_new_collection_associations.rb
```diff
@@ -0,0 +1,33 @@
+# since we have changed the basic association types, we can no longer use the dynamic methods of :branding_group_id or
+# :user_id to reference back to the parent models.  Since the data exists in the documents, we can call :attributes to
+# load existing data and create maps with which to set new associations
+class SetNewCollectionAssociations < Mongoid::Migration
+  def self.up
+    # create maps showing existing associations
+    study_list = ActiveRecordUtils.pluck_to_hash(Study, [:id, :branding_group_id])
+    study_map = study_list.map do |entry|
+      if entry[:branding_group_id].present? && entry[:branding_group_id].is_a?(BSON::ObjectId)
```
eweitz, line 9: Would safe navigation operator suffice here?  (I suppose the answer will match my other question.)

```suggestion
      if entry[:branding_group_id]&.is_a?(BSON::ObjectId)
```

db/migrate/20211110220355_set_new_collection_associations.rb
```diff
@@ -0,0 +1,33 @@
...
+        { entry[:id].to_s => entry[:branding_group_id] }
+      end
+    end.compact.reduce({}, :merge)
+    collection_map = BrandingGroup.all.map do |col|
+      {
+        col.id.to_s => {
+          users: [User.find(col.attributes[:user_id])],
```
eweitz, line 16: I believe `filter` would make the array literal unnecessary, and perhaps be more idiomatic.

```suggestion
          users: User.filter(col.attributes[:user_id]),
```

eweitz (approved):
Looks good!  I noted a user-facing letter case inconsistency, a comment typo, and suggested some (perhaps naive) refinements.

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: e6441de

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: `filter` is not a class method for `User`, so that won't work.  We could do `User.where(id: id).to_a` though, which I kinda like.

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: I think so?  Lemme try that.

app/views/branding_groups/list_navigate.html.erb
bistline: e6441de

app/controllers/branding_groups_controller.rb
bistline: e6441de

app/models/branding_group.rb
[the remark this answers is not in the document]
bistline: e6441de

app/models/branding_group.rb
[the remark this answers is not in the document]
bistline: e6441de

app/lib/synthetic_branding_group_populator.rb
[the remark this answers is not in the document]
bistline: e6441de

app/controllers/branding_groups_controller.rb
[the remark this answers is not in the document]
bistline: e6441de

app/controllers/branding_groups_controller.rb
bistline: It would have to be `collection&.users&.include?(user)` to prevent a `NoMethodError` if the collection isn't present, and I'm not a fan of multiple nil-safe operators in a row.  I find them to be somewhat obtuse.  I could do this, however:

```
curators = curator_list.split(',').map(&:strip)
users = curators.map { |email| User.find_by(email: email) }.compact
return users if collection.nil?

# ensure current user cannot accidentally remove themselves from the list if this is an update
users << user if collection.users.include?(user) && !users.include?(user)
users
```

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: cfb3f20

db/migrate/20211110220355_set_new_collection_associations.rb
bistline: cfb3f20

app/views/studies/_form.html.erb
bistline: 70eb474

app/views/studies/_form.html.erb
bistline: 70eb474

app/controllers/branding_groups_controller.rb
bistline: 70eb474

app/controllers/branding_groups_controller.rb
bistline: 70eb474

db/migrate/20211110220355_set_new_collection_associations.rb
```diff
@@ -0,0 +1,34 @@
...
+          studies: Study.where(:id.in => study_map.select {|_, c_id| c_id == col.id }.keys)
+        }
+      }
+    end.compact.reduce({}, :merge)
+    # perform new assignments for collections
+    # this will also update all associated studies/users via :has_and_belongs_to_many association on each
+    BrandingGroup.all.each do |collection|
+      associations = collection_map[collection.id.to_s]
+      collection.update(studies: associations[:studies], users: associations[:users])
+    end
```
devonbush, line 26: Not worth changing now, but reading your comment, I realize this migration probably would have been cleaner with a migration class like we use for featureFlags
```
class StudyMigrator
     include Mongoid::Document
     belongs_to :branding_group, optional: true,
  end 
```
and 
```
class BrandingGroupMigrator
     include Mongoid::Document
     belongs_to :user
  end 
```

Then you could just do:
```
BrandingGroupMigrator.all.each |bgm|
  BrandingGroup.find(bgm._id).update!(
    users: [bgm.user],
    studies: StudyMigrator.where(branding_group_id: bgm._id).to_a
  )
end
```

devonbush (approved):

eweitz:
I've verified this on staging.  See [here](https://singlecell-staging.broadinstitute.org/single_cell?scpbr=human-research-group), [here](https://singlecell-staging.broadinstitute.org/single_cell?scpbr=mouse-research-group) and [here](https://singlecell-staging.broadinstitute.org/single_cell/studies/619d29ecb067340ea23015f0/edit?scpbr=mouse-research-group).  

Anyone on the SCP team can independently validate this by signing into the non-admin account registeredinscpandterra@gmail.com using instructions [here](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365349124700?thread_ts=1600365317.124600&cid=CBEHTH601).

PR #1267, opened 2021-11-18
fixing graph resizing (SCP-3897)

devonbush:
This fixes a bug introduced by the fix of the create annotations capability.  We weren't passing the actual width and height to the resize useEffect -- the reason resizing was working before was that we were redrawing the graph on every rerender.

This also fixes a bug with user annotation caching on expression plots

TO TEST:
1. Open a study in the explore tab
2. view a cluster
3. resize your browser
4. confirm the plot redraws
5. sign in, and click 'create annotation'
6. confirm that you can select points for a labelled annotation
__________
1. Open a study with a user annotation
2. go to explore tab
3. search for a gene
4. confirm the expression plot renders


eweitz (approved):
Nice!  I've locally verified that expression plots render, resizing works, and "Create annotation" works and user annotations render.

ehanna4 (approved):
Followed steps provided and everything worked as expected

PR #1268, opened 2021-11-18
fixing colorscale behavior when Reds are default (SCP-2945)

devonbush:
Prior to this fix, The 'Reds' colorscale would be inverted when it was rendered as the default colorscale, but not inverted if it was explicitly selected.  See e.g. https://singlecell-staging.broadinstitute.org/single_cell/study/SCP1/breast-milk-alexandria-staging#study-visualize.  

TO TEST:
1. Load a study and go to the explore tab
2. do a single gene search
3. observe the Reds scale is not inverted


ehanna4 (approved):
Nice catch!

eweitz (approved):
Looks good, verified locally.  Great catch!

PR #1269, opened 2021-11-18
Integrating HCA results from Azul into bulk download UX (SCP-3812, SCP-3815)

bistline:
This update integrates the preceding work leveraging searches via Azul into the bulk download UX.  This also includes updates to our analytics reporting to Mixpanel to aid in distinguishing study- from search-based bulk download requests.  This update also includes refactoring that has improved latency on XDSS requests, as we no longer ask for file-based information as a part of a search request.  This logic has not been moved into the `BulkDownloadService`, as normal search responses contain summary-level information about files (including types, formats, counts, and total bytes) that are sufficient to use in the bulk download modal.

MANUAL TESTING
1. Pull this branch, run any remaining migrations via `bin/rails db:migrate`
2. Enter the Rails console and run `SearchFacet.update_all_facet_filters` (this is to ensure you have all filter values from Azul)
3. Boot as normal, sign in with an admin account, and make sure the `cross_dataset_search_backend` feature flag is turned on
4. On the home page, execute a search for `species:Homo sapiens` and `disease:HIV infectious disease`
5. Ensure the following HCA studies appear, and match correctly on both facets:
Screen Shot 2021-11-18 at 3 38 40 PM
6. Click "Download", and select all files entries for the HCA project `An Atlas of Immune Cell Exhaustion in HIV-Infected Individuals Revealed by Single-Cell Transcriptomics`
7. Click through to "Authorize", and paste the curl command into a terminal window
8. Ensure that the request succeeds, and files begin downloading into a directory called `ImmuneCellExhaustianHIV` (you can hit `CTRL^C` once the download starts)
9. (Optional) View [this](https://mixpanel.com/project/2085496/view/19055/app/insights#report/26971107/cross-dataset-search-requests) report in Mixpanel (dev analytics) and took for your download request to see that the line for `ImmuneCellExhaustianHIV` with `135` for `numAzulFiles` has an entry (there may be more depending on other usage)

### DIFF
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -246,10 +246,10 @@ module Api
             key :required, false
           end
           parameter do
-            key :name, :source
+            key :name, :context
             key :type, :string
             key :in, :query
-            key :description, 'Source of download request, either "study" or "search"'
+            key :description, 'Context of the download, its scope: either "study" or "global"'
             key :required, false
           end
           response 200 do
@@ -284,7 +284,7 @@ module Api
         azul_files = {}
 
         # determine if this a single-study bulk download (from download tab) or from home page search
-        search_source = params[:source] || 'search'
+        search_context = params[:context] || 'global'
 
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
@@ -355,7 +355,7 @@ module Api
                                                                            study_bucket_map: bucket_map,
                                                                            output_pathname_map: pathname_map,
                                                                            azul_files: azul_files,
-                                                                           source: search_source)
+                                                                           context: search_context)
         end_time = Time.zone.now
         runtime = TimeDifference.between(start_time, end_time).humanize
         logger.info "Curl configs generated for studies #{valid_accessions}, #{files_requested.size + directory_files.size} total files"
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -245,6 +245,13 @@ module Api
             key :description, 'Name of directory folder to download (for single-study bulk download only), can be "all"'
             key :required, false
           end
+          parameter do
+            key :name, :source
+            key :type, :string
+            key :in, :query
+            key :description, 'Source of download request, either "study" or "search"'
+            key :required, false
+          end
           response 200 do
             key :description, 'Curl configuration file with signed URLs for requested data'
             key :type, :string
@@ -275,6 +282,9 @@ module Api
         valid_accessions = []
         file_ids = []
         azul_files = {}
+
+        # determine if this a single-study bulk download (from download tab) or from home page search
+        search_source = params[:source] || 'search'
 
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
@@ -344,7 +354,8 @@ module Api
                                                                            user: current_api_user,
                                                                            study_bucket_map: bucket_map,
                                                                            output_pathname_map: pathname_map,
-                                                                           azul_files: azul_files)
+                                                                           azul_files: azul_files,
+                                                                           source: search_source)
         end_time = Time.zone.now
         runtime = TimeDifference.between(start_time, end_time).humanize
         logger.info "Curl configs generated for studies #{valid_accessions}, #{files_requested.size + directory_files.size} total files"
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -284,7 +284,7 @@ module Api
         azul_files = {}
 
         # determine if this a single-study bulk download (from download tab) or from home page search
-        search_context = params[:context] || 'global'
+        search_context = params[:context] || 'unknown'
 
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -57,14 +57,14 @@ module Api
         half_hour = 1800 # seconds
 
         totat = current_api_user.create_totat(half_hour, api_v1_bulk_download_generate_curl_config_path)
-        valid_params = params.permit({ file_ids: [], tdr_files: {} }).to_h
+        valid_params = params.permit({ file_ids: [], azul_files: {} }).to_h
 
         # for now, we don't do any permissions validation on the param values -- we'll do that during the actual download, since
         # quota/files/permissions may change between the creation of the download and the actual download.
         auth_download = DownloadRequest.create!(
           auth_code: totat[:totat],
           file_ids: valid_params[:file_ids],
-          tdr_files: valid_params[:tdr_files],
+          azul_files: valid_params[:azul_files],
           user_id: current_api_user.id
         )
         auth_code_response = {
@@ -274,7 +274,7 @@ module Api
       def generate_curl_config
         valid_accessions = []
         file_ids = []
-        tdr_files = {}
+        azul_files = {}
 
         # branch based on whether they provided a download_id, file_ids, or accessions
         if params[:download_id]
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -245,6 +245,13 @@ def drs_info
             key :description, 'Name of directory folder to download (for single-study bulk download only), can be "all"'
             key :required, false
           end
+          parameter do
+            key :name, :source
+            key :type, :string
+            key :in, :query
+            key :description, 'Source of download request, either "study" or "search"'
+            key :required, false
+          end
```
eweitz, line 254:
```suggestion
          parameter do
            key :name, :context
            key :type, :string
            key :in, :query
            key :description, 'Context of the download, its scope: either "study" or "global"
            key :required, false
          end
```

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -274,15 +281,18 @@ def drs_info
       def generate_curl_config
         valid_accessions = []
         file_ids = []
-        tdr_files = {}
+        azul_files = {}
+
+        # determine if this a single-study bulk download (from download tab) or from home page search
+        search_source = params[:source] || 'search'
```
eweitz, line 287:
```suggestion
        # determine if this a single-study bulk download (from download tab) or from global search in home page
        search_context = params[:context] || 'global'
```

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -344,7 +354,8 @@ def generate_curl_config
                                                                            user: current_api_user,
                                                                            study_bucket_map: bucket_map,
                                                                            output_pathname_map: pathname_map,
-                                                                           tdr_files: tdr_files)
+                                                                           azul_files: azul_files,
+                                                                           source: search_source)
```
eweitz, line 358:
```suggestion
                                                                           context: search_context)
```

app/lib/bulk_download_service.rb
```diff
@@ -22,7 +23,8 @@ def self.generate_curl_configuration(study_files:,
                                        user:,
                                        study_bucket_map:,
                                        output_pathname_map:,
-                                       tdr_files: nil)
+                                       azul_files: nil,
+                                       source: 'study')
```
eweitz, line 27:
```suggestion
                                       context: 'study')
```

app/lib/bulk_download_service.rb
```diff
@@ -13,7 +13,8 @@ class BulkDownloadService
   #   - +user+ (User) => User requesting download
   #   - +study_bucket_map+ => Map of study IDs to bucket names
   #   - +output_pathname_map+ => Map of study file IDs to output pathnames
-  #   - +tdr_files+ => Hash of tdr accessions to arrays of file information including name & url for each file
+  #   - +azul_files+ => Hash of Azul file summary objects
+  #   - +source+ => Origin of bulk download request (either search or study-based download)
```
eweitz, line 17: Which context a user searched, the scope of the search: "study" or "global"
```suggestion
  #   - +context+ => Context of bulk download request ("study" for single-study, or "global" for search across all studies)
```

app/lib/bulk_download_service.rb
```diff
@@ -88,9 +101,12 @@ def self.generate_curl_configuration(study_files:,
...
       studyAccessions: studies.map(&:accession),
-      numTdrStudies: tdr_studies.count,
-      tdrAccessions: tdr_studies,
-      numTDRFiles: tdr_file_configs.count
+      numAzulStudies: azul_studies.count,
+      azulAccessions: azul_studies,
+      numAzulFiles: azul_file_configs.count,
+      numAzulAnalysisFiles: azul_analysis_files,
+      numAzulSequenceFiles: azul_sequence_files,
+      source: source
```
eweitz, line 109:
```suggestion
      context: context
```

app/views/site/_study_download_data.html.erb
```diff
@@ -158,7 +158,8 @@
                 '/single_cell/api/v1/bulk_download/generate_curl_config?' +
                 `accessions=${accession}` +
                 `&auth_code=${authCode}` +
-                `&directory=${downloadObjectName}`
+                `&directory=${downloadObjectName}` +
+                `&source=study`
```
eweitz, line 162:
```suggestion
                `&context=study`
```

eweitz (changes requested):
Code looks good overall, but I think we should revise an analytics property for greater consistency across events.

What this logs as `source`, the `search` Mixpanel event already logs with the property name [`context`](https://mixpanel.com/project/2120588/view/19059/app/lexicon#transformations/event-properties/context), with values "study" or (instead of "search") "global".  Search is of course semantically different from download at a superficial layer, but this is a good opportunity to reuse vocabulary for consistency in our analytics UX.  Search and download have underlying similarities we can leverage to ease usage analysis.

Below, I suggest specifically what would need to change to do that.

app/controllers/api/v1/bulk_download_controller.rb
bistline: 735d0f0

app/controllers/api/v1/bulk_download_controller.rb
bistline: 735d0f0

app/controllers/api/v1/bulk_download_controller.rb
bistline: 735d0f0

app/lib/bulk_download_service.rb
bistline: 735d0f0

app/lib/bulk_download_service.rb
bistline: 735d0f0

app/lib/bulk_download_service.rb
bistline: 735d0f0

app/views/site/_study_download_data.html.erb
bistline: 735d0f0

eweitz (approved):

jlchang (approved):
Functional review per testing instructions worked fine. Some additional queries for Pulmonary fibrosis and Tcell activation projects also worked (queries we could not do via our TDR integration but expected to work with HCA Azul).

Noticed the analysis files include both loom and contributor-generated files - we might want to separate those in future since _we_ will only be able to work programmatically with the loom files (for things like subsetting).

bistline:
> Functional review per testing instructions worked fine. Some additional queries for Pulmonary fibrosis and Tcell activation projects also worked (queries we could not do via our TDR integration but expected to work with HCA Azul).
> 
> Noticed the analysis files include both loom and contributor-generated files - we might want to separate those in future since _we_ will only be able to work programmatically with the loom files (for things like subsetting).

We don't get a ton of information from the `projects` or `files` responses in terms of content of files.  The file "format" is usually something like `txt.gz` or `loom`, and the `contentDescription` will say something like `['Matrix']`.  However, when we run the files query, we can restrict on `fileSource: DCP/2 Analysis`, but then we likely will not get any primary sequence data.  Many entries have `fileSource: null` making it even more problematic.  So while I agree that this would be good, I'm not sure how much control we will have.

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -274,15 +281,18 @@ def drs_info
       def generate_curl_config
         valid_accessions = []
         file_ids = []
-        tdr_files = {}
+        azul_files = {}
+
+        # determine if this a single-study bulk download (from download tab) or from home page search
+        search_context = params[:context] || 'global'
```
devonbush, line 287: I'm leery about a default here inflating stats one way or the other.  I'd lean towards a default of 'unknown' or maybe some logic like `(azul_files.any? || study_accessions.count > 1) ? 'global' : 'study'`

devonbush (approved):
one non-blocking concern about analytics -- otherwise looks good

app/controllers/api/v1/bulk_download_controller.rb
bistline: Fair point, let me think about that a bit.

bistline: 4d0862f

eweitz: In what cases would `context` be undefined?  It strikes me as a minor analytics bug.

bistline:
Build failure is unrelated - all other ingest tests passed, test failure is likely a race condition.

eweitz:
I've verified this [on staging](https://singlecell-staging.broadinstitute.org/single_cell?type=study&page=1&facets=species%3ANCBITaxon_9606%3Bdisease%3AMONDO_0005109).

(I had to update the Mixpanel report to query on `context = global` instead of `source = study`, which we updated in instrumentation code per review.)

PR #1270, opened 2021-11-19
consolidating scatter plot tests, adding cache test (SCP-3897)

devonbush:
This adds tests for two of the bugs we found prior to yesterdays release-- the scatter color reversal, and the caching of user annotations. 

It also consolidates the scatter plot tests and test data into the test/js/visualization directory

TO TEST:
 1. run `yarn test scatter-plot`
 2. observe passing
 3. run `yarn test plot-data-cache`
 4. observe passing


test/js/visualization/scatter-plot.test.js
```diff
@@ -1,43 +1,112 @@
...
+        genes: [],
+        dimensionProps: BASIC_DIMENSION_PROPS
+      }}/>
+  ))
+
+  // findByTestId would be more kosher, but the numerical leaf ("1") is
+  // assigned randomly in `ScatterPlot`.  This ID has been observed with the
+  // value "study-scatter-25", but relying on that seems brittle.
+  //
+  // Consider changing `ScatterPlot` to use a deterministic `data-testid`.
```
bistline, line 77: I agree with this comment - having a function that would deterministically assign DOM ids would be more robust/testable.

bistline (approved):

test/js/visualization/scatter-plot.test.js
```diff
@@ -67,6 +136,24 @@ describe('getPlotlyTraces handles expression graphs', () => {
...
+    // check that it doesn't reverse the default (Reds)
+    expect(traces[0].marker.reversescale).toEqual(false)
+
+    plotData.scatterColor = 'Reds'
+    traces = getPlotlyTraces(plotData)
+    expect(traces[0].marker.reversescale).toEqual(false)
+
+    plotData.scatterColor = 'Greens'
+    traces = getPlotlyTraces(plotData)
+    expect(traces[0].marker.reversescale).toEqual(true)
```
eweitz, line 155: I understood this because I thought hard and had context, but otherwise I'd think this first block is redundant.

Please add a couple more comments for context -- perhaps:

```suggestion
    // Verify it doesn't reverse the default (Reds) when it's explicitly specified
    plotData.scatterColor = 'Reds'
    traces = getPlotlyTraces(plotData)
    expect(traces[0].marker.reversescale).toEqual(false)

    // Verify it reverses non-default color scales
    plotData.scatterColor = 'Greens'
    traces = getPlotlyTraces(plotData)
    expect(traces[0].marker.reversescale).toEqual(true)
```

eweitz (approved):
Looks good!  I suggested a trivial maintainability improvement.

test/js/visualization/scatter-plot.test.js
devonbush: Agree that it could use more comments.  I think the important thing is to note that we're not reversing Reds.  And it's not because it is/isn't default, it's just that it's the only color scale that happens to be good for our use un-reversed.  I'll add some clarification comments

PR #1271, opened 2021-11-22
Validate cell names are unique for Metadata and Cluster files (SCP-3767)

ehanna4:
This adds client-side validation to check that the cell names in a Metadata or Cluster file are all unique. If there is a duplicated cell name it is displayed to the user with a request to fix the duplicates.

To test:
- pull this branch
- go to the upload wizard in a study you can add files to
- try to upload the file that is included in this PR: `test/js/lib/validate-file-content.test.js`
- you should see an error message indicating the duplicate cell names

Error message is added to the list of errors if a file has multiple things wrong:
Screen Shot 2021-11-22 at 12 00 43 PM

Error message is shown for a file with duplicated cell names
Screen Shot 2021-11-22 at 11 58 27 AM


eweitz (changes requested):
Kudos on getting this working in an unfamiliar area of the codebase!  Jumping into new patterns is _hard_.

I might be missing something, but it seems this implementation unnecessarily skips reusing existing constructs like `readLinesAndType` and `sniffDelimiter`, and then re-implements them (with some unaccounted edge cases) as part of a function in `io.js`.  

That causes this validation to have rather different inner workings from existing validations, which in turn requires unduly repetitive and hard-to-maintain additions to many validation tests.

The top of `parseFile` has a line:

```
const { lines, mimeType } = await readLinesAndType(file, 2)
```

I can't suggest this inline due to GitHub PR UI limitations, but the changes I suggest below would need `parseFile` updated to something roughly like:

```
const { lines, mimeType } = await readLinesAndType(file)
const headerLines = lines.slice(0, 2)
```

Then in `parseLines` replace all uses of `lines` with `headerLines`, _except_ in the call to `validateUniqueCellNamesWithinFile`.  

Benefits:
* Eliminate the many calls to `mockCatchDuplicateCellNames` in unrelated tests
* Simplify the logic of this new rule by eliminating unnecessary use of `await`, promises, `Error` and related `try` / `catch`.
* Simplify parameters by eliminating unnecessary new `file` parameter in several functions
* Simplify architecture for client-side file validation, by reusing existing constructs

I think the essence of your validation is sound.  I hope these details help integrate it smoothly into the existing patterns!

bistline:
Functionality works as advertised, but I agree with Eric about reusing the existing infrastructure.  I'm holding off on approving/requesting changes until that is dealt with.  But so far it works!

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -322,10 +349,11 @@ async function parseFile(file, fileType) {
         return { table, delimiter, issues: [['error', 'cap:format:no-newlines', 'File does not contain newlines to separate rows']] }
       }
       // if there are no encoding issues, and this isn't a gzipped file, validate content
-      delimiter = sniffDelimiter(lines, mimeType)
+      delimiter = sniffDelimiter(headerLines, mimeType)
       table = lines.map(line => line.split(delimiter))
-
-      issues = await validateCapFormat(table, fileType)
+      const headerTable = table.slice(0, 2)
+      issues = await validateCapFormat(headerTable, fileType, file)
```
eweitz, line 355: This function doesn't have a third parameter.

```suggestion
      issues = await validateCapFormat(headerTable, fileType)
```

eweitz (approved):
Much cleaner!  I suggest an optimization and maintainability refinement. 

Now that `tables` is potentially many, many lines instead of just 2, performance is more important than elegance.  Traditional `for` loops are [much faster](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arraymap-vs-for-vs-forof) than modern `map` iteration, thus my suggestion below.

bistline (approved):

test/js/lib/validate-file-content.test.js
```diff
@@ -64,6 +63,13 @@ describe('Client-side file validation', () => {
     expect(summary).toBe('Your file had 1 error')
   })
 
+  it('catches duplicate cell names in cluster file', async () => {
+    mockReadLinesAndType({ fileName: 'cluster_duplicate_cell_names.txt' })
+    const { errors, summary } = await validateFileContent({ name: 'cluster_duplicate_cell_names.txt' }, 'Cluster')
```
devonbush, line 68: I'd like to minimize the amount of test data files we have to maintain, as it's tricky to keep track of which files are used where, and makes reviewing test cases difficult.  So I would shorten the file down to 4-5 rows, and then inline the content similar to how we inlined the content in the `catches text file with .gz suffix` 
(And I know this is different than the other cluster files in this test, but I think Eric merged those while I was on vacation :) )

devonbush (approved):
nice!  This will be a big win for our users -- just one suggestion on inlining the test content for easier reviewability

ehanna4:
Build Failure due to ongoing known issue with secrets and Travis

PR #1272, opened 2021-11-22
Adding support for numeric facets in Azul (SCP-3823)

bistline:
This update adds support for using numeric facets like `organism_age` with XDSS requests against Azul.  This facet can be used by itself, or in conjunction with other facets.  Search requests are converting the age in years into seconds before executing, similar to how BigQuery searches are run.

In addition, this update adds significant mocking to the `AzulSearchServiceTest` and `HcaAzulClientTest` as part of an effort to reduce errors/false negatives during CI runs, usually caused by service interruptions in Azul.  Also, HCA search results now link out directly to the corresponding project in the HCA data browser.

MANUAL TESTING
Note: you may need to run `SearchFacet.update_all_facet_filters` in the Rails console if you have been working on a branch that was based off of versions of `development` before today (Nov. 22)
1. Pull this branch and sign in with an account that has `cross_dataset_search_backend` set to true
2. Run a query for `organism_age:1-5 years` and ensure the following HCA projects are returned
Screen Shot 2021-11-22 at 3 54 19 PM
3. Refine the query to add `organ:Pancreas` and ensure only one match is returned
Screen Shot 2021-11-22 at 3 55 14 PM
4. (Optional) Click on the HCA Pancreas project title and ensure it opens on the correct page in the HCA data browser

This PR satisfies SCP-3832.

### DIFF
--- a/app/javascript/components/search/results/StudySearchResult.js
+++ b/app/javascript/components/search/results/StudySearchResult.js
@@ -165,7 +165,7 @@ function inferredBadge(study, termMatches) {
 function studyTypeBadge(study) {
   if (study.study_source === 'HCA') {
     // Display a badge indicating this result came from Azul
-    return <span className="badge badge-secondary study-type" data-toggle="tooltip" data-placement="right"
+    return <span className="badge badge-secondary study-type" data-toggle="tooltip"
       title={'Study from Human Cell Atlas'}> Human Cell Atlas </span>
   }
 }
@@ -182,7 +182,8 @@ export default function StudySearchResult({ study }) {
       <div key={study.accession}>
         <label htmlFor={study.name} id="result-title" className="study-label">
           {study.study_source === 'SCP' ? <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} ></a> :
-            <span dangerouslySetInnerHTML={displayStudyTitle} />
+            <a href={`https://data.humancellatlas.org/explore/projects/${study.hca_project_id}`} target="_blank"
+               dangerouslySetInnerHTML={displayStudyTitle} title="View in HCA Data Browser" data-toggle="tooltip"></a>
           }
           {inferredBadge(study, termMatches)}
         </label>
--- a/app/javascript/components/search/results/StudySearchResult.js
+++ b/app/javascript/components/search/results/StudySearchResult.js
@@ -176,15 +176,17 @@ export default function StudySearchResult({ study }) {
   const studyTitle = highlightText(study.name, termMatches).styledText
   const studyDescription = formatDescription(study.description, termMatches)
   const displayStudyTitle = { __html: studyTitle }
-
+  let studyLink = <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} ></a>
+  if (study.study_source !== 'SCP') {
+    studyLink = <a href={`https://data.humancellatlas.org/explore/projects/${study.hca_project_id}`} target="_blank"
+                   dangerouslySetInnerHTML={displayStudyTitle} title="View in HCA Data Browser" data-toggle="tooltip"
+                   rel="noreferrer"></a>
+  }
   return (
     <>
       <div key={study.accession}>
         <label htmlFor={study.name} id="result-title" className="study-label">
-          {study.study_source === 'SCP' ? <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} ></a> :
-            <a href={`https://data.humancellatlas.org/explore/projects/${study.hca_project_id}`} target="_blank"
-               dangerouslySetInnerHTML={displayStudyTitle} title="View in HCA Data Browser" data-toggle="tooltip"></a>
-          }
+          {studyLink}
           {inferredBadge(study, termMatches)}
         </label>
         <div>
--- a/app/lib/azul_search_service.rb
+++ b/app/lib/azul_search_service.rb
@@ -86,7 +86,7 @@ class AzulSearchService
         # gotcha where sampleDisease is called disease in Azul response objects
         azul_name = 'disease' if azul_name == 'sampleDisease'
         field_entries = result[result_field].map { |entry| entry[azul_name] }.flatten.uniq
-        if facet[:filters].is_a? Hash
+        if facet[:filters].is_a? Hash # this is a numeric facet, and we only have one right now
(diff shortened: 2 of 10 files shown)


test/integration/lib/azul_search_service_test.rb
```diff
@@ -28,57 +24,129 @@ class AzulSearchServiceTest < ActiveSupport::TestCase
...
-    assert_includes results.keys, @hca_project_shortname
-    project = results[@hca_project_shortname]
-    # will always be project manifest file
-    manifest = project[:file_information].detect { |f| f[:file_type] == 'Project Manifest' }
-    assert manifest.present?
+  test 'should search Azul using numeric facets' do
+    facets = [
+      {
+        id: 'organism_age', filters: { min: 1, max: 5, unit: 'years' }
+      }.with_indifferent_access
```
eweitz, line 75: Is there any more elegant way to enable JSON-like format?  Needing `.with_indifferent_access` on every declaration is noisy.

test/integration/hca_azul_client_test.rb
```diff
@@ -6,25 +6,47 @@ class HcaAzulClientTest < ActiveSupport::TestCase
...
+    @azul_is_ok = false
+    begin
+      api_up = @hca_azul_client.api_available?
+      project = @hca_azul_client.projects(query: @query_json, size: 1)
+      if api_up && get_entries_from_response(project, :projects).present?
+        @azul_is_ok = true
+      end
+    rescue RestClient::Exception => e
+      puts "Error in determining if Azul is healthy: #{e.message}"
+    end
```
eweitz, line 43: I suspect we don't currently do it, but how feasible would it be to log such exceptions from the `test` environment to Mixpanel and/or Sentry?

My vague understanding is that Azul becomes not-fully-available for a while a day or two per week.  That's concerning for a service we'll (entirely?) depend on for global study search.

Metrics on the frequency of such degradations would inform decisions around reliability engineering allocation, etc.

eweitz (approved):
Code looks good!

I suggest metrics on upstream reliability, and quixotically inquire / gripe about a Ruby-ism.

test/integration/lib/azul_search_service_test.rb
bistline: Totally agreed - this is my major gripe about Ruby, Hash objects, and Symbols.  Because `{ foo: 'bar' } != { 'foo' => 'bar' }`, we need  `with_indifferent_access`.  Enabling this as the default project-wide would require monkey patching the `Hash` class, and replacing it with `ActiveSupport::HashWithIndifferentAccess`.  I have no idea what the downstream consequences of this would be, especially with 3rd party libraries, so I've always resigned myself to needing to use `with_indifferent_access` all over the place.

But I agree - it's a pain, especially when I forget, and tests start failing all over the place.

test/integration/hca_azul_client_test.rb
bistline: Feasible: very.  But I don't think Mixpanel would be the appropriate place for this.  Since this is specifically regarding the `test` environment, and is an actual error, Sentry would be the obvious place to log this.  We would also have the added benefit of having parameters/stack traces as well, in addition to more easily accessing times/counts of any outages.  Also, I don't think merging non-production analytics into our existing Mixpanel dashboards is a habit we want to get into.  Would Sentry tracking suffice for this?

Also, regarding Azul stability: it is unclear to me how much of a problem this actually is.  They only release catalogs on a monthly basis, and do normal deployments on a similar schedule to us.  That doesn't account for a lot of the problems we've had with builds lately.  In addition, when looking at recent build failures, only a portion of them were related to Azul.  Of those that were, it seems that we also had a possible issue with idempotency in some of the tests (i.e. they sometimes fail depending on the order in which they're run), which this PR tries to address.  My hope is that this PR will address those lurking issues, and will then allow us a good, clean signal on how much of an issue service interruptions from Azul are.

bistline: Ah, I see where you're going with this now.  I agree - that makes sense, and seems like a good discussion topic.

bistline:
Test failure is tracked in SCP-3909

app/javascript/components/search/results/StudySearchResult.js
```diff
@@ -182,7 +182,8 @@ export default function StudySearchResult({ study }) {
       <div key={study.accession}>
         <label htmlFor={study.name} id="result-title" className="study-label">
           {study.study_source === 'SCP' ? <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} ></a> :
-            <span dangerouslySetInnerHTML={displayStudyTitle} />
```
devonbush, line 185: This is now a little long for a ternary construct for my taste.   So maybe something up above like
```
let studyLink = <a href={study.study_url} dangerouslySetInnerHTML={displayStudyTitle} ></a>
if (study.study_source !== 'SCP') {
  studyLink = <a href={`https://data.humancellatlas.org/ex...
}
```

test/integration/lib/azul_search_service_test.rb
devonbush: double concur.  IIRC the author of Ruby has said the handling of symbols and hashes is one of his biggest regrets in the design of the language.   And yeah, I don't know of any workaround that isn't even worse

devonbush (changes requested):
minor touchups, then this is good to go.  Nice work -- numeric facets are always tricky.  I remember our initial implementation of numeric facets in advanced search took about as long as all the other features put together

app/lib/azul_search_service.rb
```diff
@@ -86,11 +86,15 @@ def self.get_facet_matches(result, facets)
         # gotcha where sampleDisease is called disease in Azul response objects
         azul_name = 'disease' if azul_name == 'sampleDisease'
         field_entries = result[result_field].map { |entry| entry[azul_name] }.flatten.uniq
-        facet[:filters].each do |filter|
-          match = field_entries.select { |entry| filter[:name] == entry || filter[:id] == entry }
-          results_info[facet_name] ||= []
-          if match.any? && !results_info[facet_name].include?(filter)
-            results_info[facet_name] << filter
+        if facet[:filters].is_a? Hash
```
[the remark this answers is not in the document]
bistline: Yes, that's exactly the reason.

devonbush (approved):

PR #1273, opened 2021-11-22
Update `react-scripts` to fix alert about `immer` (SCP-3893)

eweitz:
This fixes a [false-positive critical security vulnerability in `immer`](https://github.com/broadinstitute/single_cell_portal_core/security/dependabot/yarn.lock/immer/open).  That was a dependency of an old version of `react-scripts`, which we use in development -- and not production.  Nevertheless, the dependency can't easily be moved from `dependencies` to `devDependencies`.  Details: facebook/create-react-app#10411.

I drafted a more comprehensive set of dependency updates, but those caused deeply cryptic problems.  This is a minimal change that fixes the most-reportedly-severe error.

To test:
* Pull branch
* `yarn install --force`
* Restart `rails s`, verify no errors upon startup
* In another terminal, restart `bin/webpack-dev-server`, verify no errors upon startup
* In local SCP, navigate to a study that has visualizations
* Save a change to a JavaScript file, e.g. open `ScatterPlot.js`, add `console.log('ok')` atop `RawScatterPlot`
* In `webpack-dev-server` terminal, verify your change triggered a recompilation
* In local SCP, open DevTools console, verify "ok" was printed to console
* Revert change to `ScatterPlot.js`

This relates to SCP-3893.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,12 +37,12 @@
     "react-bootstrap": "^0.33.1",
     "react-compound-slider": "^3.0.0-beta.1",
     "react-dom": "^16.12.0",
+    "react-notifications-component": "3.1.0",
     "react-plotly.js": "2.5.1",
-    "react-scripts": "^3.4.0",
+    "react-scripts": "^4.0.3",
     "react-select": "^3.1.0",
     "react-switch": "6.0.0",
     "react-table": "^7.1.0",
-    "react-notifications-component": "3.1.0",
     "sass": "^1.37.5",
     "spearman-rho": "^1.0.6",
     "spin.js": "^4.0.0",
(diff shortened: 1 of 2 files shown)


bistline (approved):

ehanna4 (approved):
Tested following the outlined steps and got the results expected.

eweitz:
The SCP team investigated the build failure, and determined it was a false positive with some newly-introduced log noise.

eweitz:
After merge, Dependabot decreased reported vulnerabilities from 11-12 to 9, but retained its basic alert about `immer`.

The previous alert had a clause that suggested updating to `immer` 8.0.1 (via updating `react-scripts`) would suffice, but the new message clarifies that’s not so.

As noted above, though, this vulnerability is a false positive.  It's used only in development, and not production, and so there are no known vulnerabilities in our use case.

PR #1276, opened 2021-11-23
Prevent saving `ReviewerAccess` if `expires_at` is blank (SCP-3908)

bistline:
This update addresses a corner case where a user can unset the `expires_at` date for a `ReviewerAccess` object from the "Settings" tab, causing that access object to error out any time a reviewer attempts to view the study after authenticating with the PIN.  Now, if the `expires_at` field is blank, it will prevent saving, and display an appropriate error message.

This update also addresses an issue where any updates to the "Settings" tab would cause a `ReviewerAccess` object to be created, as a necessary parameter was being omitted due to the control being disabled.  Now, objects are only persisted when `enabled` is set to `yes`.  In addition, this update fixes a bug where the `expires_at` timestamp could not be changed, due to the order in which associations were declared.  The `StudyDetail` model has a callback that will go update the parent study on `after_save`, and this callback clobbered any incoming changes to the `ReviewerAccess` object.

MANUAL TESTING
1. Boot as normal and sign in
2. Load any study, and go to the Settings tab
3. Click `Update study settings` w/o changing any values, and ensure that reviewer access has not been enabled
4. Click `Yes` under `Access Enabled?` and ensure reviewer access has been turned on (you will see credentials under `Current reviewer access credentials`)
5. In the `Expiration date` datepicker, clear out the values (the easiest way is to select each date part individually and press `Delete`)
6. Click `Update study settings` and confirm that the following error message is shown, and that the `Expiration date` datepicker is now showing an error state:
Screen Shot 2021-11-23 at 3 33 53 PM
Screen Shot 2021-11-23 at 3 33 58 PM

This PR satisfies SCP-3908.

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -679,6 +679,12 @@ class Study
         errors.add(:base, "Share Error - #{msg}")
       end
     end
+    # get errors for reviewer_access, if any
+    if study.reviewer_access.present? && !study.reviewer_access.valid?
+      study.reviewer_access.errors.full_messages.each do |msg|
+        errors.add(:base, msg)
+      end
+    end
   end
 
   # XSS protection
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -213,24 +213,19 @@ class Study
   has_many :external_resources, as: :resource_links, dependent: :destroy
   accepts_nested_attributes_for :external_resources, allow_destroy: true
 
-  # Anonymous Reviewer Access
-  has_one :reviewer_access, dependent: :delete_all
-  accepts_nested_attributes_for :reviewer_access, allow_destroy: true
+
 
   # DownloadAgreement (extra user terms for downloading data)
   has_one :download_agreement, dependent: :delete_all
   accepts_nested_attributes_for :download_agreement, allow_destroy: true
 
   # Study Detail (full html description)
-  #
-  # NOTE: THIS ASSOCIATION MUST ALWAYS BE DECLARED LAST
-  #
-  # This is due to the after_save :set_study_description_text callback which will cause the parent study
-  # to update itself with the plain-text version of the description
-  # this will clobber any incoming changes to other associations, since the study is always saved first
-  # therefore, put any new associations above :study_detail
   has_one :study_detail, dependent: :delete_all
   accepts_nested_attributes_for :study_detail, allow_destroy: true
+
+  # Anonymous Reviewer Access
+  has_one :reviewer_access, dependent: :delete_all
+  accepts_nested_attributes_for :reviewer_access, allow_destroy: true
 
   # field definitions
   field :name, type: String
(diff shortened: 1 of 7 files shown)


eweitz (approved):

bistline:
Test failures:

1. `FAIL test/js/download/download-selection-modal.test.js` is fixed in #1272
2. `SyntheticStudyPopulatorTest#test_should_be_able_to_populate_a_study_with_convention_metadata ` is tracked in SCP-3909

app/models/study.rb
```diff
@@ -81,6 +81,7 @@ def persisted
       available.reject(&:new_record?)
     end
   end
+  accepts_nested_attributes_for :study_files, allow_destroy: true
```
devonbush, line 84: interesting, I haven't used allow_destroy before.  Reading up on it, I like the rails approach of making you have to specify the ids of  models to destroy explicitly, rather than the "oh, you accidentally passed the study back with an empty array for study_files?  I'll delete them all" approach of some other frameworks

app/models/study.rb
```diff
@@ -209,15 +211,26 @@ def non_primary_data
...
-  # Anonymous Reviewer Access
-  has_one :reviewer_access, dependent: :delete_all
+  # Study Detail (full html description)
+  #
+  # NOTE: THIS ASSOCIATION MUST ALWAYS BE DECLARED LAST
+  #
+  # This is due to the after_save :set_study_description_text callback which will cause the parent study
+  # to update itself with the plain-text version of the description
+  # this will clobber any incoming changes to other associations, since the study is always saved first
+  # therefore, put any new associations above :study_detail
```
devonbush, line 231: Good comment.  I just spent a few minutes trying to think if there was a cleaner way to structure the updates so we wouldn't have this order-specific dependency.  

Would it work if we updated `set_study_description_text` in StudyDetail to something like:
```
  study_object = Study.find(self.study_id)
  text_description = self.plain_text_description
  study_object.update!(description: text_description)
  self.study.description = text_description # make sure the update above isn't clobbered by a study save if this is happening in a hook
```
By avoiding the reload and save, and just updating the object in the DB directly instead of off the association, we might be able to avoid gotchas like this in the future.  But I'm not sure this will work

devonbush (approved):
Looks good -- one suggestion for making this code a bit less brittle going forward, but I'm not sure if the suggestion will work

ehanna4 (approved):
Tested manually following the steps provided and the error messages appear as described improving the usability of this feature 👍 👍 👍

app/models/study.rb
bistline: This isn't a change, just a reordering of declarations because of the issue I found with `StudyDetail` objects.

app/models/study.rb
bistline: Well whaddya know... that works like a charm.  Nice catch.

PR #1278, opened 2021-11-29
breaking out separate vendor.js bundle (SCP-3927)

devonbush:
Splitting rarely-updated node module dependencies from application code is a well-known technique for improving site performance for both developers and end users.  See https://rewind.com/blog/optimizing-your-rails-webpacker-compilation/ for the example configuration this is based on.  In local testing on my and Eric's machine, it sped up javascript recompilation time from 2-3sec to 0.4-0.7secs

TO TEST:
1. Stop any existing rails server or webpack dev server
2. Check out branch
2. Restart rails server and webpack dev server
3. load site with chrome debugger active
4. make a change to a JS file
5. observe recompile time in webpack-dev-server is (hopefully) < 1second
6.  in 'view source' note that there is now an `application-[guid].chunk.js` file and a `vendors~application-[guid].chunk.js` file

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -375,4 +375,13 @@ module ApplicationHelper
   def label_with_asterisk(label_name)
     "#{label_name} <i class='text-danger'>*</i>".html_safe
   end
+
+  # helper for rendering split chunks tags with a nonce
+  # based off of nonced_javascript_pack_tag in secure_headers
+  # see https://github.com/github/secure_headers/blob/main/lib/secure_headers/view_helper.rb
+  def nonced_javascript_pack_with_chunks_tag(*args, &block)
+    opts = extract_options(args).merge(nonce: _content_security_policy_nonce(:script))
+
+    javascript_packs_with_chunks_tag(*args, **opts, &block)
+  end
 end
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -375,4 +375,13 @@ def get_page_name
   def label_with_asterisk(label_name)
     "#{label_name} <i class='text-danger'>*</i>".html_safe
   end
+
+  # helper for rendering split chunks tags with a nonce
+  # based off of nonced_javascript_pack_tag in secure_headers
+  # see https://github.com/github/secure_headers/blob/main/lib/secure_headers/view_helper.rb
+  def nonced_javascript_pack_with_chunks_tag(*args, &block)
```
bistline, line 382: Took me a minute to realize what this was doing - nice.

bistline (approved):

app/helpers/application_helper.rb
devonbush: yeah, this definitely more technical Ruby code than I usually write, but I wanted it to match the built-in helpers as close as possible

devonbush:
build failure is travis config issue

PR #1279, opened 2021-11-30
Enhance caching of report runs (SCP-3790)

devonbush:
This updates ReportTimePoint with a cache, so that summary stats can be cached and retrieved without database calls when they are relatively fresh. 

This also refines the query in SummaryStatsUtils.study_counts.  Previously, we were loading all public studies into memory, and then running a separate DB query for each study to pull its metadata file, which explains the production slowdown we observed.

This also adds mongo logs to standard out in development.  this can definitely be chatty, but is great for seeing cases like the above where we are running more queries than necessary, and so will be helpful as we tackle more performance issues.

TO TEST:
1. prior to checking out this branch, load the homepage in your local instance
2. click the (?) next to the More Facets button, and note the `(X of Y) public studies` 
3. check out this branch
4. restart your rails server, and load the homepage
5. confirm the `(X of Y)` public studies numbers are the same
6. in your server console, confirm you can see mongo query start/stop log messages like ```MONGODB | 104.155.132.18:27017 req:130 conn:1:1 sconn:4243 | single_cell_portal_development.find | STARTED | {"find"=>"studies", "filter"=>{"queued_for_deletion"=>false, "public"=>true}, "$db"=>"single_cell_portal_development", "lsid"=>{"id"=><BSON::Binary:0x70169748647360 type=uuid data=0x569ba5b4777c4865...>}}```
and
``` MONGODB | 104.155.132.18:27017 req:129 | single_cell_portal_development.find | SUCCEEDED | 0.096s```

### DIFF
--- a/app/models/report_time_point.rb
+++ b/app/models/report_time_point.rb
@@ -9,19 +9,63 @@ class ReportTimePoint
   include Mongoid::Document
   include Mongoid::Timestamps
 
+  ## Names of defined reports
+  STUDY_COUNTS = {
+    name: 'study_counts',
+    proc: Proc.new { SummaryStatsUtils.study_counts }
+  }
+  WEEKLY_RETURNING_USERS = {
+    name: 'Weekly Returning Users',
+    proc: Proc.new { SummaryStatsUtils.weekly_returning_users }
+  }
+
   field :name, type: String
-  field :date, type: Date
+  field :date, type: DateTime
   field :value, type: Hash
 
   validates_presence_of :name, :date, :value
 
-  ## Custom Reporting Methods
+  after_save :do_record_change_update
+  after_destroy :do_record_change_update
 
-  # get a weekly count of users that have logged into the portal
-  def self.weekly_returning_users
-    today = Date.today
-    one_week_ago = today - 1.weeks
-    user_count = User.all.select {|user| (user.last_sign_in_at >= one_week_ago || user.current_sign_in_at >= one_week_ago) && (user.last_sign_in_at < today || user.current_sign_in_at < today) }.size
-    self.create!(name: 'Weekly Returning Users', date: today, value: {count: user_count, description: "Count of returning users from #{one_week_ago} to #{today}"})
+  index({name: 1})
+
+  @@latest_reports = {}
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: app/models/report_time_point.rb [hunk 1]
```diff
@@ -9,19 +9,63 @@ class ReportTimePoint
   include Mongoid::Document
   include Mongoid::Timestamps
 
+  ## Names of defined reports
+  STUDY_COUNTS = {
+    name: 'study_counts',
+    proc: Proc.new { SummaryStatsUtils.study_counts }
+  }
+  WEEKLY_RETURNING_USERS = {
+    name: 'Weekly Returning Users',
```
eweitz, line 18: It's internal, but why not match wider convention?

```suggestion
    name: 'Weekly returning users',
```

eweitz (approved):
Looks good!  I verified locally.  Below for easy reference are my measurements per our chat.

These changes yield no obvious local speedup.  Below are times for the home page in my local SCP.  Reported times are the value of "Finished" in seconds per Chrome DevTools network tab.  It seems to slow home page load, but I suspect the averages are not significantly different given the variance.

`development` as of 11/29 (bf5cc10b)
* Average: 15.3
* Samples: 12.8, 14.9, 18.8, 13.9, 16.3

`db-cache-stats` as of 11/30 (f7a82f6, 14:53:39 -0500):
* Average: 17.5
* Samples: 16.8, 17.8, 23.4, 15.3, 14.4

You mentioned you expect this change to speed up production.  I imagine we'd see the speedup in [home page load time](https://mixpanel.com/s/3YOpIe).  

Do you foresee anywhere else we could observe the speedup in systematic measurements?

app/models/report_time_point.rb
devonbush: it would require a migration script, since these names are in the database and that report has been running for a while

devonbush:
As far as I know, this should only really impact production.  It might be observable in staging, but I expect that staging has few enough public studies that the penalty of querying each one individually might get lost in the noise as it does for you locally.

ehanna4 (approved):
Looks good to me, ran the manual testing steps as outlined and saw the results described👍

devonbush:
travis environment still corrupted, so build failure is false positive

PR #1280, opened 2021-12-02
Refine related genes ideogram (SCP-3928)

eweitz:
This improves related genes ideogram accuracy, relevance, layout, and performance.  It does this largely via a gene cache.

## Accuracy
Previously, genes occasionally incorrectly displayed as interacting.  WikiPathways often returns nodes for non-genes, and Ideogram would query MyGene.info using the non-gene labels as gene symbols, _which would sometimes unexpectedly return gene symbols from MyGene.info_, and thus spuriously display as interacting genes in the related genes ideogram.  

Ideogram now has a complete list of known genes, and robustly detects if a WikiPathways node is a gene.

## Relevance
Results are now more complete, consistent, and relevant.  There had been a threshold of 20 genes.  This was applied to unordered upstream data, which sometimes caused inconsistent results across queries.  Now Ideogram orders all returned genes by popularity, using [Gene Hints](https://github.com/broadinstitute/gene-hints) data, and shows up to 40.  Gene Hints measures popularity using Wikipedia page views and PubMed citation counts for genes over time, giving a decent assessment of global interest.

## Layout
Simply showing 40 triangle annotations is fine, but showing so many _labeled triangles_ is overcrowded.  To reduce cluttering, Ideogram adds more margin to its label-collision algorithm.  Popular genes get priority.

## Performance
Ideogram now fetches a < 2 MB file with compact genome coordinates, IDs, and symbols sorted by global popularity.  This is done once per user, and cached in the client via the web [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) API.

In addition to enabling much of the benefits outlined in sections above, this also permits certain requests to convert from blocking to non-blocking.  This often makes related gene ideogram searches roughly 0.25 - 0.5 seconds faster.

You can see the raw gene cache by going to Chrome DevTools, then "Application" panel, then "Cache Storage".

## Screenshots
Below is an example of how these changes look.  It's usually subtle and sometimes striking.  This case highlights how the updated result set is often more complete.

Old:
Related_genes_ideogram_1-30-0_GNAQ__SCP_2021-12-01

New:
Related_genes_ideogram_1-32-0_GNAQ__SCP_2021-12-01

A new synthetic study provides an easy way to manually test related genes ideogram for human.  These updates also incidentally fix a layout bug with the inferCNV ideogram legend labels.

To test:
* Install JS package updates: `yarn install --force`
* Populate new "human_all_genes" study, e.g. via `echo "SyntheticStudyPopulator.populate('human_all_genes')" | bundle exec rails console -e development`
* Navigate to it in web browser
* Search a gene
* Hover over triangles, i.e. annotations, to show tooltips / cards:
  * Interacting genes (purple) should all include a section that starts with "Interacts with"
  * Paralogous genes (pink) should all include a line that starts with "Paralog of"

This satisfies SCP-3928.


devonbush (approved):
Nice, big fan of having a new synthetic study with a realistic gene list!

jlchang (approved):
Functional review had no surprises - to be honest, I'm not sure I could detect the changes, as they are somewhat subtle. 
- search for `ACAP3` did turn up >20 paralogs so I can confirm that this PR supports >20 triangle indicators
- search for `BCL2` also turns up >20 triangle indicators. Apparently interacting genes are generally more popular than paralogs as none of the pink triangles have labels. This makes sense because genes known to be interacting have probably been better studied (paralogs may be identified computationally and not have as much research as their more famous related gene - ie. they are likely to be less popular). 

Glad to have a synthetic study to support better ideogram testing!

ehanna4 (approved):
👍  I'm still facing issues locally running the synthetic studies but looked at an alternate study and looked good and the code change makes sense to me.

PR #1227, opened 2021-10-28
Enable server-side profiling for any page (SCP-3797)

eweitz:
This adds a configuration setting for profiling server-side code with `rack-mini-profiler`, to help optimization engineering.  Setting `config.profile_performance = true` will add a small "speed badge" at top left on any page.  Clicking that opens more data to explore.  

This complements existing profiling via SCP's `GenericProfiler` ([PR](https://github.com/broadinstitute/single_cell_portal_core/pull/969), [docs](https://docs.google.com/document/d/1oSJklX9bsO4AGEwHwvZurZqnVTd6kWNz4OiZ4Zu2GtE/edit#)).  Whereas `GenericProfiler` is great for investigating timing data for a specific method, [`rack-mini-profiler`](https://github.com/MiniProfiler/rack-mini-profiler#rack-mini-profiler) is great for investigating granular timing data for a general page -- often comprised of functions that are not otherwise feasible to profile.

Here's how this profiling looks when enabled:

https://user-images.githubusercontent.com/1334561/139253987-f723f5ac-5e86-4ea0-98b0-d39f555855a5.mov

Adding `?pp=flamegraph` to the URL will show a flamegraph, and `?pp=help` will show how to do even more.  More details and particular findings are [documented here](https://docs.google.com/presentation/d/1rvenYVO6mODe2thUDmWvFCveA5xh7lpE218i_XAixm0/edit).

To test:
* Pull this branch, `bundle install`, `rails s`
* Go to a page, see no profiling button at top left of page
* In `config/environments/development.rb`, set `config.profile_performance = true`
* Restart `rails`
* Go to a page, see profiling button at top left of page.  Click on it, verify more detail is shown.
* In `config/environments/development.rb`, set `config.profile_performance = false`

This relates to SCP-3797.

PR #1281, opened 2021-12-02
adding caching to explore endpoint (SCP-3828)

devonbush:
This adds caching to the explore controller endpoint, with the hope of speeding up initial explore rendering.

This also shifts our study api cache invalidation strategy from 'attempt to be smart' to 'attempt to be careful'.  Our API endpoints are growing more complex and interdependent as we add new features to the explore tab, and tracking cache invalidation rules at a level more granular than the entire study requires quite deep knowledge.  So since study files themselves update relatively infrequently, and caching isn't essential for reasonable performance (especially since we have ClusterCacheService to regenerate automatically), I think opting for accuracy over speed is the right tradeoff.

TO TEST:
1. Open a study in the explore tab
2. reload the page
3. Observe in your console that the 'explore' and 'clusters' controller responses are cached
4. Update a cluster file (such as by changing the name, or changing the file)
5. after save and/or parse, reload the study
6. observe that the explore and clusters controller both are not cached, and instead hit the database
7. View a study with user annotations, not signed in
8. observe the user annotations are not shown in the annotation menu
9. sign in to the user who created the annotations
10. observe the user annotations are visible

(on a personal note, I think when I wrote `exclude?` on this test, it was the first time I ever got it right the first time on whether the method name is exclude, excludes, exclude?, or excludes?)

### DIFF
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -5,9 +5,15 @@ module Api
       # the goal of this controller is to be business-logic free, and only amalgamate calls
       # to other controller/service methods when needed to save server round-trips
       class ExploreController < ApiBaseController
+        include Concerns::ApiCaching
         before_action :set_current_api_user!
+
         before_action :set_study
         before_action :check_study_view_permission
+
+        before_action :check_api_cache!
+        after_action :write_api_cache!
+
 
         swagger_path '/studies/{study_id}/explore' do
           operation :get do
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -11,8 +11,9 @@ module Api
         before_action :set_study
         before_action :check_study_view_permission
 
-        before_action :check_api_cache!
-        after_action :write_api_cache!
+        # don't cache cluster_options, as those are user-dependent.
+        before_action :check_api_cache!, except: :cluster_options
+        after_action :write_api_cache!, except: :cluster_options
 
 
         swagger_path '/studies/{study_id}/explore' do
@@ -65,37 +66,9 @@ module Api
 =end
 
         def show
-          default_cluster = @study.default_cluster
-          ideogram_files = ExpressionVizService.get_infercnv_ideogram_files(@study)
-
-          if default_cluster.present?
-            cluster = {
-              numPoints: default_cluster.points,
-              isSubsampled: default_cluster.subsampled?
-            }
-          else
-            cluster = nil
-          end
-          spatial_group_options = ClusterVizService.load_spatial_options(@study)
-          image_options = ClusterVizService.load_image_options(@study)
-          bam_bundle_list = @study.study_file_bundles.where(bundle_type: 'BAM').pluck(:original_file_list)
-
-          explore_props = {
-            cluster: cluster,
-            taxonNames: @study.expressed_taxon_names,
-            inferCNVIdeogramFiles: ideogram_files,
-            bamBundleList: bam_bundle_list,
-            uniqueGenes: @study.unique_genes,
-            geneLists: @study.precomputed_scores.pluck(:name),
-            annotationList: AnnotationVizService.get_study_annotation_options(@study, current_api_user),
-            clusterGroupNames: ClusterVizService.load_cluster_group_options(@study),
-            spatialGroups: spatial_group_options,
-            imageFiles: image_options,
-            clusterPointAlpha: @study.default_cluster_point_alpha,
-            colorProfile: @study.default_color_profile,
-            bucketId: @study.bucket_id
-          }
-
+          # because this endpoint is cached for performance, we explicitly exclude user-specific information
+          # (e.g. user annotations) from it
+          explore_props = self.class.get_study_explore_info(@study, nil)
           render json: explore_props
         end
 
@@ -140,6 +113,39 @@ module Api
           render json: {
             bamAndBaiFiles: @study.get_bam_files,
             gtfFiles: @study.get_genome_annotations_by_assembly
+          }
+        end
+
+        def self.get_study_explore_info(study, user)
+          default_cluster = study.default_cluster
+          ideogram_files = ExpressionVizService.get_infercnv_ideogram_files(study)
+
+          if default_cluster.present?
+            cluster = {
+              numPoints: default_cluster.points,
+              isSubsampled: default_cluster.subsampled?
+            }
+          else
+            cluster = nil
+          end
+          spatial_group_options = ClusterVizService.load_spatial_options(study)
+          image_options = ClusterVizService.load_image_options(study)
+          bam_bundle_list = study.study_file_bundles.where(bundle_type: 'BAM').pluck(:original_file_list)
+
+          {
+            cluster: cluster,
+            taxonNames: study.expressed_taxon_names,
+            inferCNVIdeogramFiles: ideogram_files,
+            bamBundleList: bam_bundle_list,
+            uniqueGenes: study.unique_genes,
+            geneLists: study.precomputed_scores.pluck(:name),
+            annotationList: AnnotationVizService.get_study_annotation_options(study, user),
+            clusterGroupNames: ClusterVizService.load_cluster_group_options(study),
+            spatialGroups: spatial_group_options,
+            imageFiles: image_options,
+            clusterPointAlpha: study.default_cluster_point_alpha,
+            colorProfile: study.default_color_profile,
+            bucketId: study.bucket_id
           }
         end
       end
--- a/app/javascript/components/visualization/controls/ClusterSelector.js
+++ b/app/javascript/components/visualization/controls/ClusterSelector.js
@@ -50,7 +50,7 @@ export default function ClusterSelector({
 
   return (
     <div className="form-group">
-      <label className="labeled-select">Clustering2
+      <label className="labeled-select">Clustering
         <Select options={clusterOptions}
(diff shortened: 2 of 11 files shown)


eweitz (approved):
This speeds up `/explore` by 2-3 seconds in my local SCP.  Thank you!  

Below are more detailed metrics, measured in seconds via Network panel in Chrome DevTools.  Times for `/explore` are the all-inclusive values reported in the "Time" column.  Times for "Study Overview" are the "Finished" values at bottom of Network panel after reloading page.

**Before**
`development` as of 12/2 (5797978)

`/explore`:
* Average: 5.2
* Samples: 4.5, 5.5, 4.9, 5.9, 5.4

Study Overview: 
* Average: 14.7
* Samples: 12.8, 15.1, 14.9, 15.6, 14.9

**After**
`db-api-cache-updates` as of 12/2 (b881abd)

`/explore`:
* Average: 2.5
* Samples: 2.3, 2.2, 3.0, 2.8, 2.4

Study Overview:
* Average: 12.9
* Samples: 11.9, 12.7, 14.8, 13.5, 11.6

I often see uncached times much slower than reported above -- e.g. 20 seconds or more to "Finished".  In those cases, this performance optimization would help even more!  

It's also nice, of course, that our users will give a slight performance boost (presumably ~0.2 s, as things are snappier on production).

It surprises me that cached `/explore` takes 2.5 seconds.  I wonder what's taking that much time!  I tried to profile with [`rack-mini-profiler` flame graphs](#1227), but it seems they don't show flamegraphs for API calls.  Maybe I'll try `GenericProfiler` ([PR](https://github.com/broadinstitute/single_cell_portal_core/pull/969), [docs](https://docs.google.com/document/d/1oSJklX9bsO4AGEwHwvZurZqnVTd6kWNz4OiZ4Zu2GtE/edit#)).

app/javascript/components/visualization/controls/ClusterSelector.js
```diff
@@ -50,7 +50,7 @@ export default function ClusterSelector({
 
   return (
     <div className="form-group">
-      <label className="labeled-select">Clustering
+      <label className="labeled-select">Clustering2
```
bistline, line 53: Is this supposed to say `Clustering2`, or was this just a debug thing?

app/models/study_file.rb
```diff
@@ -983,28 +983,13 @@ def invalidate_cache_by_file_type
...
-      when 'Ideogram Annotations'
-        @cache_key = "#{accession}/#{study_name}.*render_cluster"
-      else
-        @cache_key = nil
+    # because of the complex interactions of the expression, explore, cluster, and annotation controllers,
+    # we play it safe and invalidate the study's api cache on any file upload that isn't 'other'
+    file_types_not_impacting_cache = ['Documentation', 'Other']
+    cache_key = nil
+    if file_types_not_impacting_cache.exclude?(self.file_type)
+      cache_key = "#{accession}/#{study_name}"
```
bistline, line 991: Since all of our viz data is now in the API, I think all we want here is `#{accession}`.  The `url_safe_name` attribute is no longer used for any visualization requests, so I think excluding it will be more precise.  Taking this into account, I think that we can merge this with `api_cache_removal_key` and just clear caches based on the accession to DRY up this code.

bistline (changes requested):
Glad this is being done - I know caching is an issue for us at times.  I think we can simplify this a bit to make it more maintainable.

app/javascript/components/visualization/controls/ClusterSelector.js
devonbush: leftover debug, good catch

eweitz (approved):
This looks good to me, modulo Jon's requested changes.

With API call split up as a side-effect of this optimization, the "Finished" time is now less representative of user-perceived load time for plots.  So we'd need SCP-3836 to reliably measure relevant speed-ups here, or to have measured using `timeSincePageLoad` in my prior measurements.  That said, `/explore` times seem similar to my last set of measurements, and I believe this retains the previous speed-up.

Thanks for digging in and optimizing here!

devonbush:
This now also includes logic to invalidate caches after any study settings change (e.g. default annotation)

app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -5,10 +5,17 @@ module Visualization
...
       # to other controller/service methods when needed to save server round-trips
       class ExploreController < ApiBaseController
+        include Concerns::ApiCaching
         before_action :set_current_api_user!
+
         before_action :set_study
         before_action :check_study_view_permission
 
+        # don't cache cluster_options, as those are user-dependent.
+        before_action :check_api_cache!, except: :cluster_options
```
bistline, line 15: I just found a corner case that we haven't considered - updating the study default options.  I changed both the default cluster/annotation, and I saw a message that said it was clearing the caches, and then re-caching the new "default" settings.  However, on page reload, the old defaults were still there, despite the values being changed in the study settings tab.  Looking at the logs, despite it saying that caches are cleared, the downstream viz requests still hit the cache, even if I comment out the line to pre-cache the defaults, so there is definitely a race condition somewhere.  I've tried ~10 times, and only once did it do the right thing.
```
Started PATCH "/single_cell/study/SCP13/snuc-seq/update_settings" for 127.0.0.1 at 2021-12-06 16:18:20 -0500
Processing by SiteController#update_study_settings as JS
  Parameters: {"authenticity_token"=>"[FILTERED]", "study"=>{"name"=>"sNuc-Seq", "embargo"=>"", "public"=>"1", "cell_count"=>"1402", "default_options"=>{"cluster"=>"DG", "annotation"=>"SUB-CLUSTER--group--study", "cluster_point_size"=>"6", "cluster_point_border"=>"false", "cluster_point_alpha"=>"1.0", "expression_label"=>"log(TPM)"}, "reviewer_access_attributes"=>{"expires_at"=>"2022-02-06"}}, "reviewer_access_actions"=>{"enable"=>"no", "reset"=>"no"}, "accession"=>"SCP13", "study_name"=>"snuc-seq"}
FireCloud API request (GET) https://api.firecloud.org/api/workspaces/single-cell-portal-development/snuc-seq-mongo-3-6/acl with tracking identifier: 6033f4a8e2413917a1d0ddcd
Disabling reviewer access in SCP13
  Rendering site/update_study_settings.js.erb
  Rendered site/_reviewer_access_fields.html.erb (Duration: 1.2ms | Allocations: 374)
  Rendered site/_study_settings_form.html.erb (Duration: 375.3ms | Allocations: 13942)
  Rendered site/_study_title_bar.html.erb (Duration: 0.4ms | Allocations: 49)
  Rendered layouts/_external_resources_view.html.erb (Duration: 0.8ms | Allocations: 144)
2021-12-06 16:18:28 -0500: Deleting caches for SCP13
2021-12-06 16:18:28 -0500: caches successfully cleared
  Rendered site/_study_description_view.html.erb (Duration: 83.1ms | Allocations: 2452)
  Rendered layouts/_download_link.html.erb (Duration: 1.4ms | Allocations: 347)
  Rendered layouts/_download_link.html.erb (Duration: 1.1ms | Allocations: 338)
  Rendered layouts/_download_link.html.erb (Duration: 1.4ms | Allocations: 336)
  Rendered layouts/_download_link.html.erb (Duration: 1.6ms | Allocations: 338)
  Rendered layouts/_download_link.html.erb (Duration: 1.7ms | Allocations: 338)
  Rendered layouts/_download_link.html.erb (Duration: 1.2ms | Allocations: 339)
  Rendered layouts/_download_link.html.erb (Duration: 1.0ms | Allocations: 338)
  Rendered layouts/_download_link.html.erb (Duration: 0.9ms | Allocations: 339)
  Rendered layouts/_download_link.html.erb (Duration: 1.0ms | Allocations: 338)
  Rendered layouts/_download_link.html.erb (Duration: 1.0ms | Allocations: 339)
  Rendered layouts/_download_link.html.erb (Duration: 0.9ms | Allocations: 338)
  Rendered site/_study_download_data.html.erb (Duration: 58.9ms | Allocations: 6534)
  Rendered site/update_study_settings.js.erb (Duration: 1167.9ms | Allocations: 49514)
Completed 200 OK in 7427ms (Views: 1169.0ms | MongoDB: 4.5ms | Allocations: 360440)


Started GET "/single_cell/api/v1/studies/SCP13/explore" for 127.0.0.1 at 2021-12-06 16:18:29 -0500
Processing by Api::V1::Visualization::ExploreController#show as JSON
  Parameters: {"study_id"=>"SCP13", "explore"=>{}}
Reading from API cache: _single_cell_api_v1_studies_SCP13_explore_
Completed 200 OK in 683ms (Views: 0.2ms | MongoDB: 0.3ms | Allocations: 39979)


Started GET "/single_cell/api/v1/studies/SCP13/clusters/_default?fields=coordinates%2Ccells%2Cannotation" for 127.0.0.1 at 2021-12-06 16:18:31 -0500
Started GET "/single_cell/api/v1/studies/SCP13/annotations" for 127.0.0.1 at 2021-12-06 16:18:31 -0500
  Parameters: {"fields"=>"coordinates,cells,annotation", "study_id"=>"SCP13", "cluster_name"=>"_default", "cluster"=>{}}
Reading from API cache: _single_cell_api_v1_studies_SCP13_clusters__default_cluster_name__default_fields_coordinates_cells_annotation
Completed 200 OK in 456ms (Views: 0.2ms | MongoDB: 0.3ms | Allocations: 39968)
```

bistline (changes requested):
See comment about updating study settings - a race condition is coming into play with clearing caches and re-rendering results.

devonbush:
travis build still broken

PR #1282, opened 2021-12-02
Make the new wizard a bit more accessible (SCP-3789)

ehanna4:
This PR starts the work to make the upload wizard more accessible making the steps in the list keyboard accessible and adds alt-tags to images. 
Improving the accessibility score given by [lighthouse](https://medium.com/accessibility-a11y/do-an-accessibility-audit-using-google-chrome-lighthouse-30ff7887901c) from a 71% 
Screen Shot 2021-12-02 at 2 43 53 PM
to a 90%
Screen Shot 2021-12-02 at 2 51 21 PM

To test this change start with a different branch and trying "tabbing" around the upload wizard. You will see you cannot tab down the list of steps. Then pull this branch and try tabbing again and you will see you can use your keyboard to navigate the steps.

You can also run the lighthouse report yourself by right clicking to inspect the page then choose Lighthouse and generate an accessibility report.


app/javascript/components/upload/StepTabHeader.js
```diff
@@ -38,14 +38,14 @@ export default function StepTabHeader({
...
   }
 
-  return <li className={className}>
-    <div onClick={() => setCurrentStep(step)} className="step-number">
-      {badgeContent}
-    </div>
-    <div>
-      <a className="action link" onClick={() => setCurrentStep(step)}>
+  return <li className={className} role="tab">
+    <button className="unset-background-border-style" onClick={() => setCurrentStep(step)} >
```
devonbush, line 42: This is a good upgrade to the markup, we just need to update our tab-detection in metrics-api.js to also catch buttons that are under tablists.  So I would extract the logic in lines 161-167 in metrics-api.js into a helper function, and then make sure that helper function is called from both `logClickLink` and `logClickButton`

app/javascript/styles/_uploadWizard.scss
```diff
@@ -122,3 +122,8 @@
 .prev-next-buttons {
   margin-left: auto;
 }
+
+.unset-background-border-style {
```
devonbush, line 126: this seems generally useful, let's move it to _global.scss

devonbush (changes requested):
Good stuff, just one minor update to make sure we're catching all the juicy analytics

app/javascript/components/upload/StepTabHeader.js
ehanna4: I think I've addressed this as you described in the last commit, please let me know if that fits what you were thinking!

app/javascript/lib/metrics-api.js
```diff
@@ -149,6 +149,22 @@ function getClassListAsArray(target) {
   return 'classList' in target? Array.from(target.classList) : []
 }
 
+/** Log a tab click if click was on a tab otherwise return false */
```
devonbush, line 152: changing this function signature to `updatePropertiesIfTab(target, props, eventName)` and having it return eventName if not a tab click, and `click:tab` if it is a tab click will let you avoid the funky `alreadyLoggedTabClick` ternary expressions you have below.

app/javascript/lib/metrics-api.js
```diff
@@ -159,23 +175,17 @@ export function logClickLink(target) {
...
-  if (parentTabList.length > 0) {
-    // Grab name of tab list and add to props
-    props.tabListName = parentTabList[0].attributes['data-analytics-name'].value
-    props.tabDisplayText = target.innerText.trim()
-    log('click:tab', props)
-  } else {
-    log('click:link', props)
-  }
+  const alreadyLoggedTabClick = logIfTabProperty(target, props)
+  alreadyLoggedTabClick ? undefined : log('click:link', props)
```
devonbush, line 179:
```suggestion
  const eventName = updatePropertiesIfTab(target, props, 'click:link')
  log(eventName, props)
```

devonbush (changes requested):
works great, just some function signature polishing will help make the code easier to read

app/javascript/components/upload/StepTabHeader.js
eweitz, line 42: We often use classes only styling selectors, but conventionally omit `-style`.

```suggestion
    <button className="unset-background-border" onClick={() => setCurrentStep(step)} >
```

app/javascript/components/upload/WizardNavPanel.js
```diff
@@ -25,15 +25,17 @@ function RawWizardNavPanel({
...
     </ul>
     <ul className="upload-wizard-steps" role="tablist" data-analytics-name="upload-wizard-secondary-steps">
-      <li className="other-header" onClick={() => setOthersExpanded(!othersExpanded)}>
-        <div className="step-number">
-          <span className="badge highlight">+</span>
-        </div>
-        <div>
-          <a className="action link" role="link">
+      <li className="other-header" role="tab" >
+        <button className="unset-background-border-style" onClick={() => setOthersExpanded(!othersExpanded)} >
```
eweitz, line 29:
```suggestion
        <button className="unset-background-border" onClick={() => setOthersExpanded(!othersExpanded)} >
```

app/javascript/lib/metrics-api.js
```diff
@@ -149,6 +149,22 @@ function getClassListAsArray(target) {
   return 'classList' in target? Array.from(target.classList) : []
 }
 
+/** Log a tab click if click was on a tab otherwise return false */
+function logIfTabProperty(target, props) {
+  // Check if target is a tab that's not a part of a menu
+  const parentTabList = $(target).closest('[data-analytics-name][role="tablist"]')
+  console.log('parentTabList', parentTabList)
```
eweitz, line 156:
```suggestion
```

app/javascript/lib/metrics-api.js
eweitz: To clarify, I understand this as suggesting the function would return `eventName` _and_ the updated log props.

app/views/layouts/application.html.erb
```diff
@@ -180,7 +180,7 @@
   <% if @selected_branding_group.present? && @selected_branding_group.footer_image.file %>
     <%= image_tag @selected_branding_group.footer_image.url, class: 'pull-left' %>
   <% else %>
-    <%= link_to image_tag('broad-logo-white.png'), "https://www.broadinstitute.org", class: 'pull-left' %>
+    <%= link_to image_tag('broad-logo-white.png', alt:"Broad logo"), "https://www.broadinstitute.org", class: 'pull-left' %>
```
eweitz, line 183:
```suggestion
    <%= link_to image_tag('broad-logo-white.png', alt:"Broad Institute logo"), "https://www.broadinstitute.org", class: 'pull-left' %>
```

eweitz (approved):
Code looks good!  I agree with Devon's suggested function refactor.

Beyond my various nits, I wonder why there's so much diff noise with spacing.  I suspect either you or @devonbush have a VS Code extension that isn't configured quite right.

app/views/layouts/application.html.erb
devonbush: it could, but I would just as soon avoid multiple return values if we can help it.  it can update the properties in-place, which is why the method is called 'update...'    A more purely functional approach might, as Eric said, be a method called `getEventPropsWithTabsApplied()` that returns both and leaves the passed-in props unchanged.  Both are totally reasonable approaches

app/javascript/lib/metrics-api.js
devonbush: it could, but I would just as soon avoid multiple return values if we can help it. it can update the properties in-place, which is why the method is called 'update...' A more purely functional approach might, as Eric said, be a method called getEventPropsWithTabsApplied() that returns both and leaves the passed-in props unchanged. Both are totally reasonable approaches

bistline (approved):
Nice work - although I can't help noticing that we get a 25% for performance from Lighthouse...

app/views/layouts/application.html.erb
ehanna4: The words in the image are "Broad Institute" so I'd like to leave it as is. Also I imagine that a screenreader would read out Broad as the word pronunciation aka /brôd/ so I think having the word "Institute" as well would help clarify this logo is for the institute rather than for something wide 😅

devonbush (approved):
nice, your current refactor is much better than what I originally suggested

PR #1284, opened 2021-12-03
removing unused search_xds javascript directory (SCP-3927)

devonbush:
This UX overhaul got de-prioritized, and is no longer on the roadmap.  So we should save ourselves the trouble of seeing duplicate files everywhere.  

TO TEST:
1. confirm frontend still builds/compiles
2. confirm homepage is unchanged


eweitz (approved):

devonbush:
Travis build still broken due to problems on their end

PR #1288, opened 2021-12-09
Refactor CSFV to use chunked reads

devonbush:
This is mostly a pure refactor, with no behavioral changes expected, except a few noted below.  The goal was to lay the groundwork for parsing and validation of larger files with larger lines, and also make it easier to add parsing functions for specific file types.  Although many file types share common validations, making a separate function per fileType hopefully will make it easier to read and comprehend the validations that are happening, at the cost of a few extra lines of code.  

CHANGED BEHAVIOR: 
  1. Only leading and trailing quotes are now trimmed from values prior to validation, to match the behavior of ingest-pipeline.  So e.g. for a column header of `dfafe"dfaf`, the quote will be preserved.  But for a column header of `"stuff"` the quotes will be stripped.
  2. For duplicate cells, the error message will only display the first 10 duplicate cells, to avoid making the message display unreadable
  3. Minor changes to logging properties of the file.  For example, we now log 'linesRead' rather than numRows, since we may not always be reading the whole file.  We also now log 'isGzipped'

TO TEST:
  1.  run `yarn test` - confirm all tests pass
  2. in the upload wizard, experiment with adding various right/wrong files (including some from the test/test_data/validation directory).   Compare the validation messages with what is displayed on staging when you select the same files


app/javascript/lib/validation/chunked-line-reader.js
```diff
@@ -0,0 +1,101 @@
...
+    this.updateHasMoreLines()
+  }
+
+  /**
+   * iterates over the remaining lines in the file, calling func(line, lineNum, isLastLine) for each line
+   * (lineNum is 0-indexed).
+   * maxBytesPerLine can be set to avoid reading the entire file into memory in the event the file is missing
+   * proper newlines
+  */
+  async iterateLines(func, maxLines=Number.MAX_SAFE_INTEGER, maxBytesPerLine=1000*1000*1024) {
```
bistline, line 33: Rather than defining the default value as an expression, I would assign this to a constant and reference it.

```suggestion
  const MAX_BYTES_PER_LINE = 1000*1000*1024
  async iterateLines(func, maxLines=Number.MAX_SAFE_INTEGER, maxBytesPerLine=MAX_BYTES_PER_LINE) {
```

bistline (approved):
This appears to work fine in practice - I'm getting passes/failures on uploads as I would expect, but running `yarn ui-test` is consistently erroring with this:
```
Summary of all failing tests
 FAIL  test/js/upload-wizard/upload-new-study.test.js (19.422 s)
  ● creation of study files › allows upload of all common file types in sequence

    TypeError: Cannot read property 'split' of undefined

      190 |   delimiters.forEach(delimiter => {
      191 |     const numFieldsLine1 = line1.split(delimiter).length
    > 192 |     const numFieldsLine2 = line2.split(delimiter).length
          |                                  ^
      193 |
      194 |     if (numFieldsLine1 !== 1 && numFieldsLine1 === numFieldsLine2) {
      195 |       bestDelimiter = delimiter

      at forEach (app/javascript/lib/validation/validate-file-content.js:192:34)
          at Array.forEach (<anonymous>)
      at sniffDelimiter (app/javascript/lib/validation/validate-file-content.js:190:14)
      at getParsedHeaderLines (app/javascript/lib/validation/validate-file-content.js:318:21)
      at Object.parseMetadataFile [as Metadata] (app/javascript/lib/validation/validate-file-content.js:277:40)
      at parseFile (app/javascript/lib/validation/validate-file-content.js:380:47)
      at validateFileContent (app/javascript/lib/validation/validate-file-content.js:393:32)
      at validateSelectedFile (app/javascript/components/upload/FileUploadControl.js:55:28)
      at handleFileSelection (app/javascript/components/upload/FileUploadControl.js:80:23)


Test Suites: 1 failed, 38 passed, 39 total
Tests:       1 failed, 134 passed, 135 total
Snapshots:   0 total
Time:        28.289 s, estimated 36 s
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

jlchang (approved):
I ran `yarn test` and everything passed. Also, the described behavioral changes to quote-handling and duplicate-cell messaging make sense to me.We don't have great test coverage of our ingest pipeline quote-handling behavior ... there are a couple of ways that the quote handling in this PR would have different outcomes in conjunction with some other metadata convention rules. I think we haven't implemented those rules in CSFV though so it isn't testable yet? (The particular rule I'm thinking of is "only alphanumeric and underscore characters are allowed in column headers")

Let me know if that particular rule is already in play. If it is, then my manual test case isn't being detected and we should take a look together. Perhaps the right thing to do is that I make sure the test cases used in this repo are at least manually tested against ingest and vice versa? If that's worth doing, we should keep a checklist of manual test files for cross-repo validation.

devonbush:
Thanks Jean--you are correct--we have not yet implemented the 'only alphanumeric and underscore' rule for headers yet in CSFV.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
   return issues
 }
 
-/** Get properties about this validation run to log to Mixpanel */
-function getLogProps(fileObj, fileType, errorObj) {
-  const { file, table, delimiter } = fileObj
-  const { errors, summary } = errorObj
+/** parse a metadata file, and return an array of issues, along with file parsing info */
+export async function parseMetadataFile(chunker, mimeType) {
+  const { parsedHeaders, delimiter } = await getParsedHeaderLines(chunker, mimeType, 2)
```
eweitz, line 284: Let's call this `headers`.  Assuming it's parsed is reasonable, like assuming it's a header for an SCP study file.  Using this name would also let us consistently refer to "headers" as parsed headers throughout this high-level module, rather than "parsedHeaders" in some functions and "headers" in other functions.

```suggestion
  const { headers, delimiter } = await getParsedHeaderLines(chunker, mimeType, 2)
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
+/** reads in the specified number of header lines, sniffs the delimiter, and returns the
+ * lines parsed by the sniffed delimiter
+ */
+export async function getParsedHeaderLines(chunker, mimeType, numHeaderLines=2) {
+  const headerLines = []
+  await chunker.iterateLines((line, lineNum, isLastLine) => {
+    headerLines.push(line)
+  }, 2)
+  if (headerLines.length < numHeaderLines || headerLines.some(hl => !hl)) {
+    throw new ParseException('format:cap:missing-header-lines', `Your file is missing newlines or is missing some of the required ${numHeaderLines} header lines`)
```
eweitz, line 326: This line is too long, e.g. it requires horizontal scrolling in GitHub to read.

```suggestion
    throw new ParseException(
      'format:cap:missing-header-lines', 
      `Your file is missing newlines or is missing some of the required ${numHeaderLines} header lines`
    )
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
   }
+  const delimiter = sniffDelimiter(headerLines, mimeType)
+  const parsedHeaders = headerLines.map(l => parseLine(l, delimiter))
+  return { parsedHeaders, delimiter }
```
eweitz, line 330: This briefer name would provide a cleaner interface outside this function:

```suggestion
  const headers = headerLines.map(l => parseLine(l, delimiter))
  return { headers, delimiter }
```

Related to that goal, for consistency _within_ this function, I could also see renaming `headerLines` to `unparsedHeaderLines`.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
+  return parseResult
 }
 
-/** Validate a local file, return list formatted of any detected errors */
+/** Validate a local file, return { errors, summary } object, where errors is an array of errors, and summary
+ * is a message like "Your file had 2 errors"
+ */
 export async function validateFileContent(file, fileType) {
-  const { table, delimiter, issues } = await parseFile(file, fileType)
+  const { fileInfo, issues } = await parseFile(file, fileType)
```
eweitz, line 410: Blocker: Towards performance tracking:

```suggestion
  const startTime = performance.now()
  const { fileInfo, issues } = await parseFile(file, fileType)
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
 
   const errorObj = formatIssues(issues)
-  const fileObj = { file, table, delimiter }
-  const logProps = getLogProps(fileObj, fileType, errorObj)
+  const logProps = getLogProps(fileInfo, errorObj)
   log('file-validation', logProps)
```
eweitz, line 414: Blocker: this more robust CSFV now takes longer, so tracking performance is worthwhile.  

```suggestion
  logProps.perfTime = Math.round(performance.now() - startTime)
  log('file-validation', logProps)
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -147,24 +160,25 @@ function validateEqualCount(headers, annotTypes) {
...
+  } else {
+    dataObj.duplicateCellNames.add(cell)
   }
-  if (duplicates.size > 0) {
-    const nameTxt = (duplicates.size > 1) ? 'names' : 'name'
-    const dupString = [...duplicates].join(', ')
-    const msg = `Cell names must be unique within a file. Please fix the following duplicated cell ${nameTxt}: ${dupString}`
+  if (isLastLine && dataObj.duplicateCellNames.size > 0) {
+    const nameTxt = (dataObj.duplicateCellNames.size > 1) ? 'duplicates' : 'duplicate'
+    const dupString = [...dataObj.duplicateCellNames].slice(0, 10).join(', ')
```
eweitz, line 180: Smart!  This had been a worry in the back of my mind.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -212,12 +224,11 @@ function sniffDelimiter(lines, mimeType) {
...
  */
-async function validateCapFormat(table, fileType) {
+function validateCapFormat([headers, annotTypes]) {
   let issues = []
-
-  // Trim spaces and quotes
-  const headers = table[0].map(header => clean(header))
-  const annotTypes = table[1].map(type => clean(type))
+  if (!headers || !annotTypes) {
+    return [['error', 'cap:format:no-header', 'File does not have at least 2 non-empty header rows']]
```
eweitz, line 230: Blocker: The use of `cap:format` instead of `format:cap` is a bug that had slipped by in earlier work.  

And is "at least" correct?  To my memory, metadata and cluster files must have two _and only two_ such rows.

```suggestion
    return [['error', 'format:cap:no-header', 'File does not have 2 non-empty header rows']]
```

eweitz, line 230: Terminology nit: the reason I used "cap" instead of "header rows" to refer to the first two lines in metadata and cluster files is that (IIRC) SCP documentation only calls the first row the header row, and never calls the second row a header.

So this message strikes me as a case of slight term-overloading, perhaps like "cluster" or "annotation".

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
+  let issues = validateCapFormat(parsedHeaders, delimiter)
+  issues = issues.concat(validateNoMetadataCoordinates(parsedHeaders))
+  // add other header validations here
 
-  const numRows = table.length
-  const numColumns = table[0].length
+  const dataObj = {} // object to track multi-line validation concerns
+  await chunker.iterateLines((line, lineNum, isLastLine) => {
+    const parsedLine = parseLine(line, delimiter)
+    issues = issues.concat(validateUniqueCellNamesWithinFile(parsedLine, isLastLine, dataObj))
```
eweitz, line 293: Similar to `headers` vs. `parsedHeaders`, let's refer to the lines we predominantly use in this high-level module as simply `lines` instead of `parsedLines`.

```suggestion
  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
    const line = parseLine(rawLine, delimiter)
    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -147,24 +160,25 @@ function validateEqualCount(headers, annotTypes) {
 
 /**
  * Verify cell names are each unique for a cluster or metadata file
+ * creates and uses 'cellNames' and 'duplicateCellNames' properties on dataObj to track
+ * cell names between calls to this function
  */
-function validateUniqueCellNamesWithinFile(table) {
+function validateUniqueCellNamesWithinFile(parsedLine, isLastLine, dataObj) {
```
eweitz, line 166:
```suggestion
function validateUniqueCellNamesWithinFile(line, isLastLine, dataObj) {
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -147,24 +160,25 @@ function validateEqualCount(headers, annotTypes) {
...
-    const cell = table[i][0]
-    if (cellNames.has(cell)) {
-      duplicates.add(cell)
-    } else {
-      cellNames.add(cell)
-    }
+  dataObj.cellNames = dataObj.cellNames ? dataObj.cellNames : new Set()
+  dataObj.duplicateCellNames = dataObj.duplicateCellNames ? dataObj.duplicateCellNames : new Set()
+
+  const cell = parsedLine[0]
```
eweitz, line 172:
```suggestion
  const cell = line[0]
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
 
-  const numRows = table.length
-  const numColumns = table[0].length
+  const dataObj = {} // object to track multi-line validation concerns
+  await chunker.iterateLines((line, lineNum, isLastLine) => {
+    const parsedLine = parseLine(line, delimiter)
+    issues = issues.concat(validateUniqueCellNamesWithinFile(parsedLine, isLastLine, dataObj))
+    // add other line-by-line validations here
+  })
+  return { issues, delimiter, numColumns: parsedHeaders[0].length }
```
eweitz, line 296:
```suggestion
  return { issues, delimiter, numColumns: headers[0].length }
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
-    numColumns,
-    numTableCells: numRows * numColumns,
-    delimiter: friendlyDelimiter,
-    fileName: file.name,
-    fileSize: file.size,
-    fileMimeType: file.type
-  }
+/** parse a cluster file, and return an array of issues, along with file parsing info */
+export async function parseClusterFile(chunker, mimeType) {
+  const { parsedHeaders, delimiter } = await getParsedHeaderLines(chunker, mimeType, 2)
```
eweitz, line 301:
```suggestion
  const { headers, delimiter } = await getParsedHeaderLines(chunker, mimeType, 2)
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -276,99 +279,138 @@ function validateClusterCoordinates(headers) {
...
+  // add other header validations here
+
+  const dataObj = {} // object to track multi-line validation concerns
+  await chunker.iterateLines((line, lineNum, isLastLine) => {
+    const parsedLine = parseLine(line, delimiter)
+    issues = issues.concat(validateUniqueCellNamesWithinFile(parsedLine, isLastLine, dataObj))
+    // add other line-by-line validations here
+  })
+
+  return { issues, delimiter, numColumns: parsedHeaders[0].length }
```
eweitz, line 314:
```suggestion
  issues = issues.concat(validateClusterCoordinates(parsedHeaders))
  // add other header validations here

  const dataObj = {} // object to track multi-line validation concerns
  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
    const line = parseLine(rawLine, delimiter)
    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
    // add other line-by-line validations here
  })

  return { issues, delimiter, numColumns: headers[0].length }
```

app/javascript/lib/validation/validate-file-content.js
eweitz, line 330: Returning `headers` instead of `parsedHeaders` would provide a cleaner interface outside this function:

```suggestion
  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
    rawHeaderLines.push(line)
  }, 2)
  if (rawHeaderLines.length < numHeaderLines || rawHeaderLines.some(hl => !hl)) {
    throw new ParseException('format:cap:missing-header-lines', `Your file is missing newlines or is missing some of the required ${numHeaderLines} header lines`)
  }
  const delimiter = sniffDelimiter(rawHeaderLines, mimeType)
  const headers = rawHeaderLines.map(l => parseLine(l, delimiter))
  return { headers, delimiter }
```

eweitz (changes requested):
Great engineering!  I'm surprised something like this doesn't exist as a widespread JavaScript library, but I couldn't find anything in my research.  That suggests we're fundamentally innovating.

I note a few trivial blockers, related to performance and tracking.  Beyond that, I raise a handful of non-blocking maintainability nits.

eweitz (approved):

eweitz (commented):
I thumbed too fast -- please do add those request changes for `perfTime` ([here](https://github.com/broadinstitute/single_cell_portal_core/pull/1288#discussion_r767911594), [here](https://github.com/broadinstitute/single_cell_portal_core/pull/1288#discussion_r767908588)) when you have a moment.  I suspect they got lost among my many other comments!

eweitz:
Ah, gotcha.  I assumed it was overlooked given no "outdated" tag in my suggestion, and then myself overlooked your totally-reasonable alternate implementation.  My bad, looks good!

PR #1293, opened 2021-12-13
Client-side validation for Dense Matrix (first piece of SCP-3766)

ehanna4:
Splitting the validation of expression matrices into 2 PRs for now to try to start getting initial feedback as I work on the validation for the Sparse Matrix

This portion covers the validation for Dense Matrices:
- validate that the cell names are all unique (first row)
- validate that the next row has a gene name in the first column 
- validate that all values in the remainder of the columns are numeric
- validate the number of columns match

The following screenshot shows a file receiving all four the error messages
Screen Shot 2021-12-13 at 1 21 51 PM

To test:
- Pull this branch
- I used the Expression Matrix file from the test data folder `expression_matrix_example_bad.txt` with the following changes: I deleted the last few columns from the last row, changed a '0' value in the matrix to be 'test', and replaced 'cell_002' with 'cell_001'.


app/javascript/lib/validation/validate-file-content.js
```diff
@@ -144,17 +144,72 @@ function validateEqualCount(headers, annotTypes) {
...
 }
 
+/**
+ * Verify "GENE" is present as the first column in the first row for an Expression Matrix file
+ */
+function validateGeneInHeader(firstRowTable) {
+  const issues = []
+
+  if (firstRowTable[0][0] !== 'GENE') {
+    const msg = 'Dense matrices require the first value of the file to be "GENE". ' +
```
bistline, line 154: Technically, there are 2 cases that we allow as per `scp-ingest-pipeline`:

1. First keyword is `GENE` (See [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/expression_files/dense_ingestor.py#L215))
2. There is no `GENE` keyword, and the first line has exactly 1 entry less than each successive line (also referred to as "R formatted", since this is what the output of `write.table` will do).  See [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/expression_files/dense_ingestor.py#L129).

Looking at SCP-3766, I see this isn't called out in that ticket, so obviously there was no way for you to know that - but we should maintain feature parity with `scp-ingest-pipeline`, where possible.

bistline (changes requested):
The `GENE` header validation looks good, but we need to add in support for R-formatted expression matrices (no `GENE` header, and first line has exactly 1 entry less than each successive line).

app/javascript/lib/validation/validate-file-content.js
bistline: That being said - the validation does work as expected for the `GENE` header.

ehanna4: oo thanks for catching that! I will make that update

ehanna4: oh looking at the ingest-pipeline code it also appears "gene" would be acceptable as well - as in it is not case sensitive - is that correct?

bistline: Yes - you are correct.  Good catch!  At one point, we also accepted an empty string (`""`), but in looking at the code that doesn't appear to be the case anymore.

devonbush (changes requested):
Marking as request changes, as `validateColumnNumberAndValue` will need refactoring to account for the new chunked line reader method of reading files.  And we should wait to release this and those other CSFV changes until after the new year to ensure a relaxing holiday for everyone :)

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -144,17 +144,72 @@ function validateEqualCount(headers, annotTypes) {
...
+    const msg = `All rows must have the same number of columns - ` +
+    `please ensure the number of columns for ${rowText}: ${rowsWithIncorrectColumnNumbers.join(', ')}, ` +
+    `matches the file-header-specificed number of ${correctNumberOfColumns} columns-per-row.`
+    issues.push(['error', 'format:mismatch-column-number', msg])
+  }
+
+  if (rowsWithNonNumericValues.length > 0) {
+    const rowText = rowsWithNonNumericValues.length > 1 ? 'rows' : 'row'
+    const msg = `All values (other than the first column and header row) in a dense matrix file must be numeric. ` +
+    `Please ensure all values in ${rowText}: ${rowsWithNonNumericValues.join(', ')}, are numbers.`
```
eweitz, line 196: Blocker: this UI seems like it will poorly handle when many (e.g. nearly all) rows have this problem.

Please see my other comment about this problem for a possible solution.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -144,17 +144,72 @@ function validateEqualCount(headers, annotTypes) {
...
+ * Verify cell names are each unique for a file
+ */
+function validateUniqueCellNamesWithinFile(table, fileType) {
   const issues = []
 
   const cellNames = new Set()
   const duplicates = new Set()
   for (let i = 0; i < table.length; i++) {
-    const cell = table[i][0]
+    const cell = fileType === 'Expression Matrix' ? table[i]: table[i][0]
```
eweitz, line 212: A space before the colon is conventional in ternaries.

```suggestion
    const cell = fileType === 'Expression Matrix' ? table[i] : table[i][0]
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -348,15 +403,25 @@ async function parseFile(file, fileType) {
...
+      if (fileType === 'Expression Matrix') {
+        const firstRowTable = table.slice(0, 1)
+        issues = issues.concat(validateUniqueCellNamesWithinFile(firstRowTable[0], fileType))
+        issues = issues.concat(validateGeneInHeader(firstRowTable))
+        issues = issues.concat(validateColumnNumberAndValues(table))
+      } else if (['Cluster', 'Metadata'].includes(fileType)) {
+        const headerTable = table.slice(0, 2)
+        issues = await validateCapFormat(headerTable, fileType)
+        issues = issues.concat(validateUniqueCellNamesWithinFile(table, fileType))
+      }
```
eweitz, line 424: I'd organize this differently, for greater simplicity, coverage, and sync CSFV extensibility.

* `validateUniqueCellNamesWithinFile` applies to all file types handled here.  Per your changes in this PR, that function now has special handling for dense matrices, so this special-casing for the first argument (`firstRowTable` / `table`) seems more coherent to do within that function.

* `validateGeneInHeader` is fundamentally a cap validation, i.e. it requires only the first line(s) in a file.  Thus, it'll be a good candidate for sync CSFV.  Grouping all cap validations into `validateCapFormat` would let us more easily apply all of them to sync.  So like we can handle file type-specific validations _within_ `validateUniqueCellNamesWithinFile`, I'd do similarly in `validateCapFormat`.  (I'd also update its docstring to note that a cap is the first "few" lines, and applies to all the main SCP file types.)

* `validateColumnNumberAndValues` should apply to all three file types handled here.

Apologies that I can't be more specific; the GitHub review UI is limiting.  I'm happy to chat if it'd help!  The above notes would simplify this block to:

```suggestion
      const cap = table.slice(0, fileType === 'Expression Matrix' ? 1 : 2)
      
      issues = await validateCapFormat(cap, fileType)
      issues = issues.concat(validateUniqueCellNamesWithinFile(table, fileType))
      issues = issues.concat(validateColumnNumberAndValues(table))
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -348,15 +403,25 @@ async function parseFile(file, fileType) {
...
       }
-      const headerTable = table.slice(0, 2)
-      issues = await validateCapFormat(headerTable, fileType)
-      issues = issues.concat(validateUniqueCellNamesWithinFile(table))
+
+      if (fileType === 'Expression Matrix') {
+        const firstRowTable = table.slice(0, 1)
+        issues = issues.concat(validateUniqueCellNamesWithinFile(firstRowTable[0], fileType))
+        issues = issues.concat(validateGeneInHeader(firstRowTable))
+        issues = issues.concat(validateColumnNumberAndValues(table))
```
eweitz, line 419: Blocker: this validation seems relevant for cluster and metadata files, too.  See my comments in that function for detail.

```suggestion
        issues = issues.concat(validateColumnNumberAndValues(table))
```

eweitz (changes requested):
CSFV for dense expression matrices will really improve upload UX!  Good judgement call splitting the ticket up into multiple PRs.

Below, I note blockers related to likely UI issues, validation coverage, analytics, and code organization.  Validation is surprisingly complex to get right!

Beyond that, we'll also want to integrate Devon's chunked-read updates.  Such work strikes me as best for a subsequent PR.

app/javascript/lib/validation/validate-file-content.js
ehanna4: It looks like there is actually a bug currently in the code with the `sniffdelimiter` function such that R-formatted expression matrices are not being handled correctly. Because of that I am thinking I will keep this PR to only allow exp matrices with "gene" headers and make a followup ticket to address the bug in the delimiter and then allow R-formatting as well. Please let me know if that seems reasonable, thanks!

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -144,17 +144,72 @@ function validateEqualCount(headers, annotTypes) {
...
+
+  return issues
+}
 
 /**
- * Verify cell names are each unique for a cluster or metadata file
+ * Verify dense matrix column values and column numbers
  */
-function validateUniqueCellNamesWithinFile(table) {
+function validateColumnNumberAndValues(table) {
```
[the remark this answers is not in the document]
ehanna4: Adding validation to cluster and metadata files feels out of scope for this ticket. I will make a followup ticket if that seems reasonable.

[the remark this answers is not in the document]
eweitz: Good point!  That's reasonable.

bistline (approved):

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -399,6 +478,25 @@ function validateClusterCoordinates(headers) {
...
+
+  const dataObj = {} // object to track multi-line validation concerns
+  await chunker.iterateLines((line, lineNum, isLastLine) => {
+    const parsedLine = parseLine(line, delimiter)
+    issues = issues.concat(validateValuesAreNumeric(parsedLine, isLastLine, lineNum, dataObj))
+    issues = issues.concat(validateColumnNumber(parsedLine, headers, lineNum, isLastLine, dataObj))
+    issues = issues.concat(validateUniqueCellNamesWithinFile(parsedLine, isLastLine, dataObj))
+    issues = issues.concat(validateMetadataLabelMatches(headers, parsedLine, isLastLine, dataObj))
+    issues = issues.concat(validateGroupColumnCounts(headers, parsedLine, isLastLine, dataObj))
+    // add other line-by-line validations here
```
eweitz, line 495: Similar to how we say `headers` instead of `parsedHeaders`, I'd suggest we refer to the lines we predominantly use in this high-level module as simply `lines` instead of `parsedLines`.  That's also consistent with how this module refers to the return value of `parseLine` elsewhere.

```suggestion
  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
    const line = parseLine(rawLine, delimiter)
    issues = issues.concat(validateValuesAreNumeric(line, isLastLine, lineNum, dataObj))
    issues = issues.concat(validateColumnNumber(line, headers, lineNum, isLastLine, dataObj))
    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
    issues = issues.concat(validateMetadataLabelMatches(headers, line, isLastLine, dataObj))
    issues = issues.concat(validateGroupColumnCounts(headers, line, isLastLine, dataObj))
    // add other line-by-line validations here
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -177,9 +177,89 @@ function validateEqualCount(headers, annotTypes) {
...
+/**
+ *
+ * Validate all values are numbers outside first column cell name
+ */
+function validateValuesAreNumeric(parsedLine, isLastLine, lineNum, dataObj) {
+  const issues = []
+
+  // skip first column
+  const lineWithoutFirstColumn = parsedLine.slice(1)
+  console.log('parsedLine after', parsedLine)
```
eweitz, line 205:
```suggestion
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -177,9 +177,89 @@ function validateEqualCount(headers, annotTypes) {
...
+
+/**
+ *
+ * Validate all values are numbers outside first column cell name
+ */
+function validateValuesAreNumeric(parsedLine, isLastLine, lineNum, dataObj) {
+  const issues = []
+
+  // skip first column
+  const lineWithoutFirstColumn = parsedLine.slice(1)
```
eweitz, line 204: Per rationale in my other comment beginning "Similar to".
```suggestion
function validateValuesAreNumeric(line, isLastLine, lineNum, dataObj) {
  const issues = []

  // skip first column
  const lineWithoutFirstColumn = line.slice(1)
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -177,9 +177,89 @@ function validateEqualCount(headers, annotTypes) {
...
+
+/**
+ * Verify dense matrix column numbers match header column numbers
+ */
+function validateColumnNumber(parsedLine, isLastLine, headers, lineNum, dataObj) {
+  const issues = []
+  // use the first header row to determine the correct number of columns all rows should have
+  const correctNumberOfColumns = headers[0].length
+
+  if (correctNumberOfColumns !== parsedLine.length) {
```
eweitz, line 238: Per rationale outlined in my comment beginning "Similar to".

```suggestion
function validateColumnNumber(line, isLastLine, headers, lineNum, dataObj) {
  const issues = []
  // use the first header row to determine the correct number of columns all rows should have
  const correctNumberOfColumns = headers[0].length

  if (correctNumberOfColumns !== line.length) {
```

eweitz (approved):
This looks good!  Thanks for incorporating the feedback.

I suggest a few maintainability improvements, and still think [some organization refinements](https://github.com/broadinstitute/single_cell_portal_core/pull/1293#discussion_r768847323) would help, but I don't consider those things blocking.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -235,29 +315,35 @@ function sniffDelimiter([line1, line2], mimeType) {
...
+ * Verify cap format for a cluster, metadata, or expression matrix file
  *
- * The "cap" of an SCP study file is its first two lines, i.e.:
+ * The "cap" of an SCP study file is its first "few" lines that contain structural data., i.e.:
  *  - Header (row 1), and
  *  - Annotation types (row 2)
  *
  * Cap lines are like meta-information lines in other file formats
  * (e.g. VCF), but do not begin with pound signs (#).
  */
```
devonbush, line 326: BLOCKER: we're trying to avoid branches based on file type in individual validation methods so that they are easier to reason about.  So I would leave this method as-is, and just create a new `validateDenseHeader` method that checks that the header line exists and then calls `validateGeneInHeader`

devonbush (changes requested):
Nice!  thanks for updating this to use the new, hopefully simpler-to-implement, validation code framework.  

One blocker about not having filetype-sensitive parse methods that is hopefully easy to fix, otherwise looks great.

devonbush (approved):

ehanna4:
I see the same errors on the other PR branch also unrelated to the code that is failing, and don't see these failures on the PR that has test fixes so I think it is safe to figure these are unrelated test issues.

PR #1300, opened 2022-01-04
Tooltip for Disabled Download button (SCP-3817)

ehanna4:
Add a tooltip for the Download button on the home page that works even when the button is disabled. This adds a wrapper around the button so the tooltip will be on the wrapper allowing the button to be properly disabled as well, without losing the tooltip.

Currently when the Download button is disabled the tooltip doesn't appear and the user just sees a `Ø` type indication that the button is not clickable when they hover over it. This improvement will have the proper hints when a user hovers over the button even when it's disabled. 

To see the difference:
- Go to [SCP production](https://singlecell.broadinstitute.org/single_cell) and hover over the `Download` button when you are not signed in, or when you have not done a valid search. There won't be a properly displayed tooltip
- Then pull this branch and go to the home page and hover on the `Download` button and you should be able to see 3 distinct messages for:
- Not signed in
- signed in but invalid search
- signed in and done a valid search

Screen Shot 2022-01-04 at 3 07 45 PM
Screen Shot 2022-01-04 at 3 09 09 PM
Screen Shot 2022-01-04 at 3 09 29 PM


test/js/download.test.js
```diff
@@ -45,7 +45,7 @@ describe('Download components for faceted search', () => {
     wrapper.find('#download-button > span').simulate('mouseenter')
     const tooltipHint = wrapper.find('OverlayTrigger').prop('overlay').props['children']
 
-    expect(tooltipHint).toBe('To download, first do a search')
+    expect(tooltipHint).toBe('To download, first do a valid search')
```
eweitz, line 48:
```suggestion
    expect(tooltipHint).toBe('To download, first do a search')
```

test/js/download/download-button.test.js
```diff
@@ -30,7 +30,7 @@ describe('Download button for faceted search', () => {
     expect(screen.getByRole('button')).toHaveTextContent('Download')
     expect(screen.getByRole('button')).toBeDisabled()
     fireEvent.mouseOver(screen.getByText('Download'))
-    expect(screen.getByRole('tooltip')).toHaveTextContent('To download, first do a search')
+    expect(screen.getByRole('tooltip')).toHaveTextContent('To download, first do a valid search')
```
eweitz, line 33:
```suggestion
    expect(screen.getByRole('tooltip')).toHaveTextContent('To download, first do a search')
```

app/javascript/components/search/controls/download/DownloadButton.js
```diff
@@ -29,30 +27,31 @@ export default function DownloadButton({ searchResults={} }) {
...
+  </button>
+
   let hint = 'Download files for your search results'
-  if (!active) {
+  if (downloadDisabled) {
     if (userContext.accessToken === '') {
       hint = 'To download, please sign in'
     } else {
-      hint = 'To download, first do a search'
+      hint = 'To download, first do a valid search'
```
eweitz, line 48:
```suggestion
      hint = 'To download, first do a search'
```

eweitz (changes requested):
Helpful UI polish!  I tested this locally.

Blocker: this draft changes the user-facing term "search" to "valid search".  I suspect the latter will confuse users.  

While simply home page load does trigger a search API request, that's an internal implementation detail that we hide from users (and our analytics).  To the user, loading the home page is not a search; they haven't searched for anything yet.  

So I think we should omit the "valid" qualifier.  That's clearer and simpler for the user, sidestepping the need to answer the question "How does a valid search differ from an invalid search?".

ehanna4:
> Blocker: this draft changes the user-facing term "search" to "valid search". I suspect the latter will confuse users.

@eweitz I found the original language confusing which was why I added 'valid'. If a user does a search that has no results the button is still disabled, and it seemed unhelpful to have the tooltip say simply to run a search since, they just did but doing so didn't un-disable the button. However, I see your point about introducing this possibly problematic idea of what a valid or invalid search might be. Basically I wanted something that told the user "run a search that produces results" but thought 'valid' was more succinct. If you feel strongly I can put the language back how it was or if there is a better word than 'valid' also open to that haha

>While merely loading the home page does trigger a search API request, that's an internal implementation detail that we hide from users (and our analytics). To the user, loading the home page is not a search; they haven't searched for anything yet.

I didn't think adding 'valid' exposed this implementation detail since if the user has performed no search at all the tooltip still says they must perform a valid search to download.

eweitz:
Ah, gotcha!  In that case I'd opt for the phrase "search that returns results" over "valid search".   (I've updated my suggestions above accordingly.)  The former is a bit wordier, but not confusing.  I don't think most would consider a search that returns no results to necessarily be invalid.

Separately, the PR description notes:

> Go to [SCP production](https://singlecell.broadinstitute.org/single_cell) and hover over the Download button when you are not signed in, or when you have not done a valid search. There won't be a properly displayed tooltip

When I do that, I see what looks like a proper tooltip.  Could you help me understand how this one differs?

ehanna4:
https://user-images.githubusercontent.com/54322292/148125462-d6edfe0a-b43b-492c-bf1d-85087cdcef89.mov

@eweitz this is how it looks for me on prod

eweitz:
Whoa!  Seeing is believing, thank you.  I don't experience that bug on my machine, FWIW.  (I wonder if other reviewers do.)

bistline (approved):
I agree with Eric's comments, so I'll approve modulo those changes.  I might suggest the phrase `To download, first perform a valid search` instead.

eweitz:
To clarify, my comments are mostly concerned that the term "valid" is confusing here.

If one does a well-formed query that returns no results, the search is still valid, even though it has no hits.  Something like "To download, first do a search that returns results" avoids such confusion while still being fairly brief.

eweitz (approved):

PR #1302, opened 2022-01-05
Migrating all Rails-based tests to use FactoryBot, SelfCleaningSuite (SCP-3061)

bistline:
Currently, in the Rails-based testing section of SCP, there are a mix of test construction techniques.  Roughly half are older-style tests that rely on data populated at the beginning of each CI run from `db/seeds.rb`.  The other half are newer-style tests that leverage both `FactoryBot` and our internal `SelfCleaningSuite` that handle CRUDing all necessary data for the current suite (including cleanup).  These newer style tests are highly preferable to the older ones as they require no pre-test setup, and developers can run them at any time while working on feature development as they are completely self-contained.

This update migrates all of the older-style tests to this newer framework.  This means that any Rails-based test will now adhere to the following rules:
1. Test suites are now fully idempotent (can be run as many times as needed in any order, and return the same results, regardless of database state)
2. Test suites handle setting their own state, and do not require running `rake db:seed` to populate data
3. Test suites clean up after themselves to remove any extraneous data that might potentially skew downstream results

The main caveat is that some test suites now run slower, as data seeding can sometimes take 10-20s, depending on the kind of data being created (a study w/ Terra workspace takes the longest, and is required for some integration testing).  This can add a non-trivial amount of time to an entire CI run, though the exact amount is not currently known as we lack sufficient data to estimate the impact.  However, given that we intend to eventually parallelize tests as much as possible, this should not be viewed as a blocker.

Additionally, it is still necessary for developers to start `Delayed::Job` and reindex their MongoDB test schema before running tests locally (using the `bin/run_tests.sh` script will handle this for you).

MANUAL TESTING
1. Pull this branch, and then clean out your test schema by running the following commands from the project root:
```
./rails_local_setup.rb && source config/secrets/.source_env.bash
bundle exec bin/rails runner -e test "StudyCleanupTools.destroy_all_studies_and_workspaces"
bundle exec bin/rails runner -e test "BigQueryClient.clear_bq_table"
bundle exec rake RAILS_ENV=test db:purge
bundle exec rake RAILS_ENV=test db:mongoid:create_indexes
```
2. Restart `Delayed::Job` for the test environment:
```
bin/delayed_job restart test -n 6
```
3. Select any Rails-based test suite, and run it 2+ times in a row to ensure idempotency (`SearchControllerTest` and `SiteControllerTest` are good examples):
```
bin/rails test test/api/search_controller_test.rb
bin/rails test test/controllers/site_controller_test.rb
```

This PR satisfies SCP-3061.


lib/test_data_populator.rb
```diff
@@ -0,0 +1,99 @@
...
+        study_file = StudyFile.create!(attributes)
+        puts "done, pushing file to bucket #{study.bucket_id}"
+        study.send_to_firecloud(study_file)
+      end
+    end
+    study.reload # refresh study state on return
+  end
+
+  # clears out any existing search facets and re-creates defaults of species, disease, and organism_age
+  def self.create_search_facets
```
devonbush, line 54: maybe `create_sample_search_facets` to clarify that it's not re-creating everything from the json schema?

lib/test_data_populator.rb
```diff
@@ -0,0 +1,99 @@
...
+      name: 'cluster_example.txt',
+      path: 'test/test_data/cluster_example.txt',
+      file_type: 'Cluster'
+    }
+  }.with_indifferent_access.freeze
+
+  FILE_TYPES = EXAMPLE_FILES.keys.freeze
+
+  # add example files to an existing, non-detached study and push to workspace bucket
+  def self.add_files_to_study(study, file_types: FILE_TYPES)
```
devonbush, line 33: this is nice functionality. I think we can include this capability in our study factory.  If we make a new 'predefined_file_types' transient argument to the study factory, and if present, add the files with the code below in the existing `after(:create)` method.  That keeps more of our study-testing code in the same place.

devonbush: (and you may want to still keep this function accessible as an independent method for the cases where someone wants to create a study with no files, test something, add files, and then test something else

test/big_query_helper.rb
```diff
@@ -3,3 +3,32 @@ def get_bq_row_count(study)
   bq_dataset = ApplicationController.big_query_client.dataset(CellMetadatum::BIGQUERY_DATASET)
   bq_dataset.query("SELECT COUNT(*) count FROM #{CellMetadatum::BIGQUERY_TABLE} WHERE study_accession = '#{study.accession}'", cache: false)[0][:count]
 end
+
+# seed BQ with 5 rows of data for a given study/metadata file
+def seed_bq(study, metadata_file)
```
devonbush, line 8: I don't think you need the 'metadata_file' argument, since you could get that from just calling study.metadata_file.  
Also, consider 'seed_example_bq_data' as a name that indicates we're not parsing the metadata file

devonbush (changes requested):
VERY COOL!  This is an impressive amount of work.  Minor changes -- I'd love for your new file-addition code to be rolled into the FactoryBot interface so it's easier to find, and then just some name suggestion tweaks.

lib/test_data_populator.rb
bistline: Yeah let me kick the tires on that - it's a fair point.

bistline: Actually, I think I'm going to leave the code in place and reference it from the `Study` factory.  This way the same code is exercised/tested, and it can be used both via tests or manually.

bistline:
Looking at runtimes in the [Actions](https://github.com/broadinstitute/single_cell_portal_core/actions/workflows/run-all-tests.yml) tab, this update appears to only add roughly 2-3m of total runtime to builds (from ~31m to ~33-34m).

devonbush (approved):
nice!

test/api/directory_listings_controller_test.rb
```diff
@@ -26,33 +33,29 @@ class DirectoryListingsControllerTest < ActionDispatch::IntegrationTest
...
-    puts "#{File.basename(__FILE__)}: #{self.method_name}"
     execute_http_request(:get, api_v1_study_directory_listing_path(study_id: @study.id, id: @directory_listing.id))
     assert_response :success
     # check all attributes against database
     @directory_listing.attributes.each do |attribute, value|
       if attribute =~ /_id/
         assert json[attribute] == JSON.parse(value.to_json), "Attribute mismatch: #{attribute} is incorrect, expected #{JSON.parse(value.to_json)} but found #{json[attribute.to_s]}"
       elsif attribute =~ /_at/
-        assert DateTime.parse(json[attribute]) == value, "Attribute mismatch: #{attribute} is incorrect, expected #{value} but found #{DateTime.parse(json[attribute])}"
+        # ignore timestamps as fprmatting & drift on milliseconds can cause comparison errors
```
eweitz, line 49:
```suggestion
        # ignore timestamps as formatting & drift on milliseconds can cause comparison errors
```

test/api/external_resources_controller_test.rb
```diff
@@ -26,33 +31,29 @@ class ExternalResourcesControllerTest < ActionDispatch::IntegrationTest
...
-    puts "#{File.basename(__FILE__)}: #{self.method_name}"
     execute_http_request(:get, api_v1_study_external_resource_path(study_id: @study.id, id: @external_resource.id))
     assert_response :success
     # check all attributes against database
     @external_resource.attributes.each do |attribute, value|
       if attribute =~ /_id/
         assert json[attribute] == JSON.parse(value.to_json), "Attribute mismatch: #{attribute} is incorrect, expected #{JSON.parse(value.to_json)} but found #{json[attribute.to_s]}"
       elsif attribute =~ /_at/
-        assert DateTime.parse(json[attribute]) == value, "Attribute mismatch: #{attribute} is incorrect, expected #{value} but found #{DateTime.parse(json[attribute])}"
+        # ignore timestamps as fprmatting & drift on milliseconds can cause comparison errors
```
eweitz, line 47:
```suggestion
        # ignore timestamps as formatting & drift on milliseconds can cause comparison errors
```

test/api/reports_controller_test.rb
```diff
@@ -11,7 +11,35 @@ class ReportsControllerTest < ActionDispatch::IntegrationTest
 
   before(:all) do
     @user = FactoryBot.create(:api_user, test_array: @@users_to_clean)
```
eweitz, line 13: Out of curiosity, is there a way to share `before(:all)` between files?  At a quick glance, this `@user` line looks the same across all tests here; perhaps other lines are shared too.

test/api/search_controller_test.rb
```diff
@@ -2,45 +2,70 @@
...
-  NO_DISEASE_FILTER = { id: 'MONDO_0000001', name: 'disease or disorder' }
+  HOMO_SAPIENS_FILTER = { id: 'NCBITaxon_9606', name: 'Homo sapiens' }.freeze
+  NO_DISEASE_FILTER = { id: 'MONDO_0000001', name: 'disease or disorder' }.freeze
 
   # shorthand accessors
   FACET_DELIM = Api::V1::SearchController::FACET_DELIMITER
   FILTER_DELIM = Api::V1::SearchController::FILTER_DELIMITER
 
+  before(:all) do
+    @random_seed = SecureRandom.uuid
```
eweitz, line 23: Could you help me understand why `@random_seed` is used in this test but not others in this PR?

test/api/study_file_bundles_controller_test.rb
```diff
@@ -26,42 +37,38 @@ class StudyFileBundlesControllerTest < ActionDispatch::IntegrationTest
...
-    puts "#{File.basename(__FILE__)}: #{self.method_name}"
     execute_http_request(:get, api_v1_study_study_file_bundle_path(study_id: @study.id, id: @study_file_bundle.id))
     assert_response :success
     # check all attributes against database
     @study_file_bundle.attributes.each do |attribute, value|
       if attribute =~ /_id/
         assert json[attribute] == JSON.parse(value.to_json), "Attribute mismatch: #{attribute} is incorrect, expected #{JSON.parse(value.to_json)} but found #{json[attribute.to_s]}"
       elsif attribute =~ /_at/
-        assert DateTime.parse(json[attribute]) == value, "Attribute mismatch: #{attribute} is incorrect, expected #{value} but found #{DateTime.parse(json[attribute])}"
+        # ignore timestamps as fprmatting & drift on milliseconds can cause comparison errors
```
eweitz, line 53:
```suggestion
        # ignore timestamps as formatting & drift on milliseconds can cause comparison errors
```

test/api/study_files_controller_test.rb
```diff
@@ -26,77 +32,69 @@ class StudyFilesControllerTest < ActionDispatch::IntegrationTest
...
     execute_http_request(:get, api_v1_study_study_file_path(study_id: @study.id, id: @study_file.id))
     assert_response :success
+    @study_file.reload # gotcha in case 'should parse study file' test has already run
     # check all attributes against database
     @study_file.attributes.each do |attribute, value|
       if attribute =~ /_id/
         assert json[attribute] == JSON.parse(value.to_json), "Attribute mismatch: #{attribute} is incorrect, expected #{JSON.parse(value.to_json)} but found #{json[attribute.to_s]}"
       elsif attribute =~ /_at/
-        assert DateTime.parse(json[attribute]) == value, "Attribute mismatch: #{attribute} is incorrect, expected #{value} but found #{DateTime.parse(json[attribute])}"
+        # ignore timestamps as fprmatting & drift on milliseconds can cause comparison errors
```
eweitz, line 49:
```suggestion
        # ignore timestamps as formatting & drift on milliseconds can cause comparison errors
```

test/integration/lib/bulk_download_service_test.rb
```diff
@@ -4,12 +4,24 @@
 
 class BulkDownloadServiceTest < ActiveSupport::TestCase
   include Minitest::Hooks
+  include SelfCleaningSuite
   include TestInstrumentor
```
eweitz, line 8: These three includes seem present in most, if not all, tests in this PR.

If feasible, I think abstracting them to a share file and perhaps importing that one thing here (and the 15+ equivalent other places) would be more maintainable than importing these three things.

eweitz (approved):
Great foundational work to enable parallelization!  That will be a phase change for testing.

This code looks good.  Thanks for the many small readability improvements: shorter lines, obviating need to escape quotes, space between "#" and first character in comment, etc.

I suggest a few (admittedly conjectural, and all non-blocking) abstractions to improve maintainability.  I also suggest a fix for a repeated code-comment typo.

test/api/reports_controller_test.rb
bistline: No, not that I know of.  It has to do with order of precedence for includes/requires.  The only way I would know how do it would be to add a helper and use `require` at the top of the test, but then the `include MiniTest::Hooks` declaration would be missing and the whole thing will error out.  I've tried similar things before and it always barfs.

test/integration/lib/bulk_download_service_test.rb
bistline: See above re: require/include precedence.  I'm not aware of a way to do this that won't result in errors.  Doesn't mean it isn't possible, I'm just not sure how it could be done.

bistline: @eweitz actually I take that back - I think I know a way how to do it.  Will need to run some tests but I think I can do includes in helpers if I do it inside class definitions for the base classes of each test (`ActiveSupport::TestCase`, `ActionDispatch::IntegrationTest`, and `ActionController::TestCase`).

PR #1305, opened 2022-01-07
Upload wizard visualization status indicators (SCP-3969)

devonbush:
SCP-3969  This adds indicators for whether a study has visualizations enabled for clusters and/or expression data.  It also polls the server every 10 seconds so that the indicators can be updated without need page refreshes, and spinners can be shown for when a file is parsing.  

The approach for this polling was to keep it as simple as possible--every 10 seconds it asks the server to resend all information about the study and study files.  This avoids needing complex logic of inter-file dependencies in determining what to poll.  Once the client receives the updated study data, it stores it in the 'serverState' variable, and it also matches each study form file object to the corresponding server file state.  The server state is pointed to in a 'serverFile' property on each form file.  So in rough [MVVM ](https://www.educative.io/collection/page/5429798910296064/5725579815944192/5161641743220736 )terms, the serverState/serverFile is the model, the formState/file is the view model, and the components are the view.  

Two new spinners indicate when a file is awaiting remote, or parsing:  
image (9)

image (7)

Visual indicators show that the study is missing files needed for visualization
image (8)

TO TEST:
1. Open an existing study with visualizations (e.g. male mouse brain), in the upload wizard, confirm the indicators are green
2. Create a new study -- confirm the indicators show no visualizations yet
3. add expression matrices.  Confirm the spinners behave as expected
4. add a metadata file, confirm spinners behave as expected during loading, and once parsing completes, 'expression visuals' indicator should be green
5. add a cluster file, confirm 'cluster visuals' indicator is green once parsing completes.


devonbush:
got some seemingly unrelated test fails, rerunning now

eweitz (changes requested):
Exciting stuff!   I tested this locally and found some seeming bugs:

> Open an existing study with visualizations (e.g. male mouse brain), in the upload wizard, confirm the indicators are green

When I do that,  I see a "parsing" spinner icon (at right, beside "Save") even though the study has no files parsing.

<details>
<summary>Click for screenshot</summary>
1__Preexisting_study_with_parsing_spinner_bug__async_status_update__SCP_2021-01-10
</details>

> Create a new study -- confirm the indicators show no visualizations yet

When I do that, I see the indicators are orange as expected, but I see a "parsing" spinner here too.

<details>
<summary>Click for screenshot</summary>
2__New_study_no_files_uploaded_with_parsing_spinner_bug__async_status_update__SCP_2021-01-10
</details>

Finally, after I upload raw and processed matrices and they complete ingest, then the "parsing" spinner goes away but the "no expression visuals" indicator stays orange.  

That's reasonable, but the tooltip for that indicator only says "A processed matrix file is required for gene expression visualization" -- which seems likely to confuse users who have uploaded and completely ingested a processed matrix.  Would noting more conditions they need to satisfy make sense?  (Is a metadata file also needed?)

<details>
<summary>Click for screenshot</summary>
3__New_study_parse_completed_expression_box_still_orange_bug__async_status_update__SCP_2021-01-10
</details>

app/javascript/components/upload/WizardNavPanel.js
```diff
@@ -50,5 +57,75 @@ function RawWizardNavPanel({
...
+    </OverlayTrigger>
+  </div>
+  return <div className="viz-notifications">
+    {clusterVizIndicator}
+    {expressionVizIndicator}
+  </div>
+}
+
+/** gets the popup message based on whether there are files parsing */
+function expressionHelpContent(isExpressionParsing) {
```
bistline, line 113: Technically, to view expression-based visualizations, you need all 3 main file types: expression, metadata, and clustering.  All of our non-genomic visualizations require a cluster to view.

bistline (approved):
Looks good and functions as advertised, but the message/logic for having expression-based visualizations enabled also requires a cluster file to be parsed.  Since this change is so minor, I'm going to pre-approve modulo that change.

app/javascript/components/upload/WizardNavPanel.js
```diff
@@ -50,5 +57,75 @@ function RawWizardNavPanel({
...
+
+/** gets the popup message based on whether there are files parsing */
+function clusteringHelpContent(isClusteringParsing) {
+  return <Popover id="cluster-viz-upload-info" className="tooltip-wide">
+    <div> A metadata file and a clustering file are required for cluster visualization</div>
+    { isClusteringParsing && parsingMessage }
+  </Popover>
+}
+
+const parsingMessage = <div> Some files which will impact visualization are currently being processed </div>
```
eweitz, line 128: Let's end sentences with a period.  (The other changes here are a bit more subjective.)

```suggestion
const parsingMessage = <div>Some files that will impact visualization are being processed.</div>
```

app/javascript/lib/scp-api.js
```diff
@@ -200,8 +200,11 @@ export function stringifyQuery(paramObj, sort) {
  *
  * @param {String} studyAccession Study accession
 */
-export async function fetchStudyFileInfo(studyAccession, mock=false) {
-  const apiUrl = `/studies/${studyAccession}/file_info`
+export async function fetchStudyFileInfo(studyAccession, includeOptions=true, mock=false) {
```
eweitz, line 203: Remind me -- does `file_info` leverage our Rails file cache?  

Maybe it's just because I'm on a development environment, but about 1/3 of my poll requests have > 5 s TTFB.   Even though it's used to assess current status of an occasionally-rapidly-changing object on the server, those objects will only be changing -- i.e. ingesting / parsing -- for a very brief phase of their life.  Cache updates would be rare, and not a notable performance burden in the parsing phase.

So caching seems like it'd be a big scalability improvement, if not done already.

eweitz (approved):
Live status updates are an awesome UX improvement.  I hope we can use them everywhere, eventually!

I made a handful of suggestions below for user-facing text, scalability, etc.  None of it's blocking.  

I tested this locally and it worked well.

devonbush:
In addition to addressing Jon's comment, I also just added some logic to better handle parse failures -- previously the UX was very confusing if the parse failed.  Now if the parse fails, you see this
image

eweitz (changes requested):
Trivial blocker I just noticed: the "Save" button is blue on dev, without any unsaved changes in local file options UI.  It's white on staging in the same conditions, which to my understanding is the expected color in that state.

Unexpected blue color for "Save" on dev:
Blue_save_dev__SCP_2021-01-11

Expected white color [on staging](https://singlecell-staging.broadinstitute.org/single_cell/studies/61b23412b067340e9f9279d3/upload?tab=rawCounts):
White_save_dev__SCP_2021-01-11

devonbush:
I think there are enough issues flagged in this PR that I will mothball it until the next developer choice days unless feedback from demo suggests this is worth bringing into regular sprint status.

devonbush:
Adding better handling of the case where a user goes offline or session expires
image

eweitz (commented):
Is the ["Save" blocker](https://github.com/broadinstitute/single_cell_portal_core/pull/1305#pullrequestreview-849613135) resolved?

eweitz (approved):
Looks good!  Thanks for following up via Slack.

Also -- considering someone noted this feature'd be nice while this has been brewing here -- kudos on the good product instinct.

PR #1312, opened 2022-01-13
improving scroll display of facet accordion (SCP-3945)

devonbush:
SCP-3945.  This updates the scroll behavior on the facet accordion to mimic the scroll behavior of facets not in the accordion.  Previously, there was a mismatch between the max heights that meant a user would have to scroll down  inside the accordion to see the apply button if there were more than ~7 options. Now the apply button is not hidden behind a scrollbar.
image

In the event that the user's monitor is very short (<650px) they would have to scroll (similar to how they would have to scroll for other facets), but the scrolling is more clearly indicated, since the bottom of the accordion is no longer visible, and easier to accomplish, since it is scrolling the whole document. 
image

TO TEST:
1. View home page
2. open 'more facets'
3. expand library preparation protocol, confirm apply button is visible

### DIFF
--- a/app/javascript/styles/_search.scss
+++ b/app/javascript/styles/_search.scss
@@ -202,23 +202,6 @@ nav.search-links, #search-panel {
     }
   }
 
-  #facets-accordion {
-    background: #eff2f5;
-    overflow-y: scroll;
-    overflow-x: hidden;
-    .facet {
-      width: 98%;
-      display: block;
-      margin: 5px 2px;
-      .filters-box-searchable {
-        position: relative;
-        width: 98%;
-        border: none;
-        margin-top: -2px;
-      }
-    }
-  }
-
   .filters-box-header {
     margin: 5px 0px;
     display: flex;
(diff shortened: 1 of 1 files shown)


bistline (approved):

ehanna4 (approved):
Overall looks good, however I am seeing an issue with the `organism age` (screenshot included) that I'm not seeing on Prod or Staging but am seeing locally with both this branch and the development branch. Because I'm also seeing it with the development branch I'm not marking this as a blocker for this PR but just wanted to call it out and wondering if someone can confirm if they also see the issue locally.
Screen Shot 2022-01-15 at 9 04 16 AM

devonbush:
@ehanna4  The organism age filter looks ok for me locally both on this branch and development.  Are you just clikcing on the age filter to get that screenshot above?  Do you have anything else selected?

devonbush:
confirmed Emily's issue was confined to her environment (misconfugred facet in her local DB)

PR #1315, opened 2022-01-18
Check user Terra Terms of Service status on sign-in (SCP-3640)

bistline:
In preparation for Terra's planned changes to Terms of Service acceptance, this update adds a check upon successful sign in for all users to determine if they need to accept any updated Terms of Service from Terra (similar to how SCP checks for it's own ToS updates on sign in).  In this updated flow, any users who meet the following criteria will have a message displayed notifying them of the situation, and prompt them to visit Terra to accept the updated terms:
1. User must be registered for Terra
2. User has yet to accept the current Terra Terms of Service

Note that this modal only shows on sign in, and not on every page load.  Expected failure states for users would be showing the `Your session may have timed out. Please sign in again` modal on most pages, as any API requests to orchestration would return `401/403` when issued with user credentials.  This likely will be confined to study-related pages, but could potentially happen other places as well.

New ToS alert:
Screen Shot 2022-01-18 at 1 48 22 PM

MANUAL TESTING
1. Boot as normal
2. Sign in with any account that is registered for Terra, and confirm no alert is shown
3. In `app/models/user.rb`, line 356, add `return true` to short-circuit the ToS check
4. Sign out, then sign in again and confirm the above modal is shown

This PR satisfies SCP-3640.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -25,11 +25,19 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       # update a user's FireCloud status
       @user.delay.update_firecloud_status
       sign_in(@user)
+      @alert = nil
+      # check if user needs to accept updated Terra ToS
+      if @user.must_accept_terra_tos?
+        @alert = 'We have detected that Terra has updated their terms of service - please visit ' \
+                 "<a href='https://app.terra.bio' target='_blank'>Terra</a> to accept the updated terms.  Some " \
+                 'features may not function correctly until you do so.'
+      end
       if TosAcceptance.accepted?(@user)
         MetricsService.merge_identities_in_mixpanel(@user, cookies)
-        redirect_to request.env['omniauth.origin'] || site_path
+        requested_path = request.env['omniauth.origin'] || site_path
+        redirect_to requested_path, alert: @alert
       else
-        redirect_to accept_tos_path(@user.id)
+        redirect_to accept_tos_path(@user.id), alert: @alert
       end
     else
       redirect_to new_user_session_path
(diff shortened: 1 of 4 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -25,11 +25,19 @@ def google_oauth2
       # update a user's FireCloud status
       @user.delay.update_firecloud_status
       sign_in(@user)
+      @alert = nil
+      # check if user needs to accept updated Terra ToS
+      if @user.must_accept_terra_tos?
+        @alert = 'We have detected that Terra has updated their terms of service - please visit ' \
```
eweitz, line 31: Trimmed, and casing [Terms of Service](https://terra.bio/about/terms-of-service/) like Terra does:

```suggestion
        @alert = 'Terra has updated their Terms of Service.  Please visit ' \
```

eweitz (approved):

app/models/user.rb
```diff
@@ -351,6 +351,32 @@ def update_firecloud_status
     end
   end
 
+  # determine if user needs to accept an updated Terra Terms of Service agreement
+  def must_accept_terra_tos?
```
devonbush, line 355: Consider changing this method to just be 
 `has_accepted_latest_terra_tos?` and then having the callbacks_controller be the place where we have the logic for not calling the method if they aren't registered.  That lets this method focus purely on the terra ToS logic, and the controller owns the 'when do we prompt' logic

devonbush (approved):
Nice -- rails alerts are so great for things like this.  Just one optional suggestion.

app/models/user.rb
bistline: TL;DR - you're right, but because of weirdness in how we emit/handle errors from `FireCloudClient`, not making the registration check in the User instance method leads to a corner case where a non-registered user would have `must_accept_terra_tos?` return `true`, rather than `false`.  Because we only emit `RuntimeError` from orchestration API requests, we don't know if the real HTTP response code was `401`, `403` or `404`.  The controller check sidesteps this, but it leaves me with a bad feeling knowing that we have a method that has a known issue that external logic avoids.  Hence the pre-flight check for registration status.

Ideally, we'd do a larger refactoring to emit correct errors from `FireCloudClient`, like `HcaAzulClient` and `DataRepoClient` do, because then we could catch the `404` and handle appropriately.

bistline:
Test failure tracked in SCP-3995.

PR #1316, opened 2022-01-18
Log file validation trigger to Mixpanel (SCP-3704)

bistline:
This update adds the `trigger` property to all Mixpanel `file-validation` events.  This will enable admins to ascertain which view/page a file validation event was triggered from (currently, the upload wizard or the sync page).  This information will be useful for determining next steps for extending client-side validation to other parts of the application.

MANUAL TESTING
1. Boot as normal and sign in
2. Create a new study, or go to the upload wizard for an existing study
3. Open the network tab in Chrome Dev Tools
4. Click into the `Clustering` section of the upload wizard, and upload `cluster_bad_no_coordinates.txt` from `test/test_data/validation`
5. Find the `file-validation` event in the Network tab, and confirm that the `trigger` property is set to `upload` as seen below:
Screen Shot 2022-01-18 at 4 19 33 PM

This PR satisfies SCP-3704.

### DIFF
--- a/app/javascript/lib/validation/validate-file-content.js
+++ b/app/javascript/lib/validation/validate-file-content.js
@@ -12,6 +12,7 @@ import { log } from 'lib/metrics-api'
 import { readFileBytes } from './io'
 import ChunkedLineReader from './chunked-line-reader'
 import { PARSEABLE_TYPES } from 'components/upload/upload-utils'
+import getSCPContext from 'providers/SCPContextProvider'
 
 // from lib/assets/metadata_schemas/alexandria_convention_schema.json (which in turn is from scp-ingest-pipeline/schemas)
 export const REQUIRED_CONVENTION_COLUMNS = [
@@ -672,10 +673,20 @@ function formatIssues(issues) {
   return { errors, warnings, summary }
 }
 
+/** determine trigger for file-validation event (e.g. upload vs. sync) **/
+function getValidationTrigger() {
+  const pageName = getSCPContext().analyticsPageName
+  if (pageName === 'studies-initialize-study') {
+    return 'upload'
+  } else if (pageName === 'studies-sync-study') {
+    return 'sync'
+  }
+}
 
 /** Get properties about this validation run to log to Mixpanel */
 function getLogProps(fileInfo, errorObj, perfTime) {
   const { errors, warnings, summary } = errorObj
+  const trigger = getValidationTrigger()
 
   // Avoid needless gotchas in downstream analysis
   let friendlyDelimiter = 'tab'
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Short and sweet!  

You and I had discussed how deferring implementation of the `cli` trigger made sense, given it'd require delving into a rarely-touched codebase for something that's rarely-used.  

Considering further, `cli` seems more appropriate for another property.  Like this SCP GUI, the SCP CLI _also_ has "upload" and "sync" validation (or ought to), and so the complementary upcoming `logger` property -- not this `trigger` property -- seems best to distinguish GUI vs. CLI.  

So using `trigger` and `logger` would give us "trigger: upload, logger: gui", "sync gui", "upload cli", and "sync cli".  (Of course neither of those sync CSFV contexts are implemented yet, with the latter not even roadmapped.)

bistline:
> Considering further, `cli` seems more appropriate for another property. Like this SCP GUI, the SCP CLI _also_ has "upload" and "sync" validation (or ought to), and so the complementary upcoming `logger` property -- not this `trigger` property -- seems best to distinguish GUI vs. CLI.
> 
> Using `trigger` and `logger` would give us "trigger: upload, logger: gui", "sync gui", "upload cli", and "sync cli". (Of course neither of those sync CSFV contexts are implemented yet, with the latter not even roadmapped.)

I like where you're going with that - and I agree.  I think the two fields combined would give us the most accurate picture.  Personally, I would prefer `client` over `logger` as it seems more apposite, but I'm not sure if that will conflict with an existing property in Mixpanel (it doesn't look like it, but it is also an over-used term).

devonbush (approved):
Approving because it satisfies the ticket definition.  But I think what you've got here is much more generally useful.  Consider:
rename `getValidationTrigger` to `getUserReadablePageName()`, and have it just return the analyticsPageName if none of the ifs are true
rename `trigger` in `getLogProps` to something like `readablePageName`

devonbush:
'Trigger' has the difficulty that the scope is unclear.  In the context of an event, is the 'trigger' which button they clicked, which page they were on, or which app they were using? So I lean towards using a property name that's more descriptive about what it is, and then let mixpanel reports be the ones to assign meaning like 'trigger'

bistline:
> 'Trigger' has the difficulty that the scope is unclear. In the context of an event, is the 'trigger' which button they clicked, which page they were on, or which app they were using? So I lean towards using a property name that's more descriptive about what it is, and then let mixpanel reports be the ones to assign meaning like 'trigger'

@devonbush @eweitz I'm not sure how to proceed now given these comments.  Are further changes needed?  There doesn't seem to be consensus on what to call things.  I would opt to leave things as-is as it satisfies the specific AC for the ticket, but there seems to be differing opinions now.

eweitz:
Per chat among participants here, we’ve got rough consensus to merge, and maybe a possible retro / engineering strategy topic to discuss the reasonable non-blocking concerns that have been raised.

PR #1317, opened 2022-01-19
Ignore empty/incomplete HCA project entries in bulk download (SCP-4009)

bistline:
This update fixes two bugs found recently:
- If a user selects an HCA project in the bulk download modal, but selects no file types, and empty entry is created in the `DownloadRequest#azul_files` hash
- `BulkDownloadService#generate_curl_config` assumes that HCA project manifest are always present, even if user has not requested them

MANUAL TESTING
1. Boot as normal, and sign in with an account with XDSS enabled
2. Run a query for `species:Homo sapiens` and `disease:COVID-19`
3. Open the bulk download selection table, and select only the following entries:
Screen Shot 2022-01-19 at 12 25 36 PM
4. Proceed all the way through to generating the curl config and pasting the command into the terminal
5. Confirm that the request succeeds and begins downloading files into a directory called `LethalCovidLungAtlas` (you can cancel out once files start as this means the request worked)

This PR satisfies SCP-4009.

### DIFF
--- a/app/controllers/api/v1/bulk_download_controller.rb
+++ b/app/controllers/api/v1/bulk_download_controller.rb
@@ -59,12 +59,17 @@ module Api
         totat = current_api_user.create_totat(half_hour, api_v1_bulk_download_generate_curl_config_path)
         valid_params = params.permit({ file_ids: [], azul_files: {} }).to_h
 
+        # discard any empty Azul entries - this can happen if row is selected, but no file types are checked
+        if valid_params[:azul_files].present?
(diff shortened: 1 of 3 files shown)


eweitz (approved):

bistline:
Test failure is tracked in SCP-3995

jlchang (approved):
Confirmed bug is fixed for both the test case described as well as the originating search on `prostate gland`.

Thanks for finding and squashing this bug so quickly, Jon!

ehanna4 (approved):
followed steps and worked as expected👍

PR #1319, opened 2022-01-19
fixing error display when study is behind a download agreement (SCP-4006)

devonbush:
SCP-4006.  Yikes, our API error handling has a LOT of edge cases.  I've created SCP-4010 to handle a more substantive upgrade, and hopefully this is a step in the right direction.  there are two basic problems:
1.  on the frontend, we don't want/need custom handling for 401/403 errors on every UI component since they are almost always the user getting signed out.  (since the UX shouldn't be showing options the user isn't allowed to do)
2. The frontend can't rely on a single, canonical format of error response, since devise, rails, and our base API controller all report differently.

In any case, this updates the UX for if you try to do a bulk download including a study that needs a download agreement
image

TO TEST:
1.  Add a download agreement to a study
2. sign in as a user who is not the study owner and who has not accepted the agreement
3. do a search which includes the study
4. click 'Download'
5. observe the modal above
6. Sign out
7. In `DownloadButton.js`, line 24, set `active = true` to enable the UX download even if you are not signed in
8. Reload the page, and click 'Download'
9. confirm a popup appears suggesting your session has expired


app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -373,12 +376,16 @@ def self.check_accession_permissions(valid_accessions, user)
...
         if accessions_by_permission[:forbidden].any? || accessions_by_permission[:lacks_acceptance].any?
-          error_msg = "Forbidden: cannot access one or more requested studies for download. "
+          error_msg = "Forbidden: cannot access one or more requested studies for download.\n "
           if accessions_by_permission[:forbidden].any?
             error_msg += "You do not have permission to view #{accessions_by_permission[:forbidden].join(', ')}"
           end
           if accessions_by_permission[:lacks_acceptance].any?
-            error_msg += "#{accessions_by_permission[:lacks_acceptance].join(', ')} require accepting a download agreement that can be found by viewing that study and going to the 'Download' tab"
+            error_msg += "Download agreement required for #{accessions_by_permission[:lacks_acceptance].join(', ')}. \n\n"
+            error_msg += "Visit the 'Download' tab at these url(s) and accept the agreement to enable download\n"
```
eweitz, line 385:
```suggestion
            error_msg += 'Visit the "Download" tab at these URL(s) and accept the agreement to enable download:\n'
```

app/javascript/lib/scp-api.js
```diff
@@ -767,31 +767,47 @@ export default async function scpApi(
...
   }
   throw new Error(response)
 }
 
+/** custom class for handling json-api style API errors */
+class ApiError extends Error {
+  /** get a new instance based on an already-parsed-to-json http response */
+  constructor(jsonResponse, httpStatus, path) {
+    const rawString = jsonResponse.errors.map(err => err.detail).join('.\n')
+    const message = `API Error calling ${path} (${httpStatus}): ${rawString}`
```
eweitz, line 803:
```suggestion
    const message = `API error calling ${path} (${httpStatus}): ${rawString}`
```

bistline (approved):
Approving as this works as advertised and addresses concerns from ticket.

Non-blocking: as per our discussion, it might be preferable to allow the request to continue, and simply note either in the UI or in the curl configuration that some studies were excluded due to the non-acceptance of the agreement.  But given how rare these are, it may not be worth it, at least in the context of this PR.

app/javascript/lib/scp-api.js
```diff
@@ -767,31 +767,47 @@ export default async function scpApi(
...
+        Authentication failed<br/>
+        Your session may have timed out. Please sign in again.<br/><br/>
+      </div>,
+      'api-auth-failure',
+      {
+        source: 'api',
+        url,
+        isError: true,
+        messageType: 'error-client',
+        statusCode: response.status
```
eweitz, line 782: It'd be cool if we had an overview of the various event names and properties we use to log errors to Mixpanel.  My hunch is that we've accreted several overlapping but unnecessarily inconsistent ways to do that.

This code is pre-existing, so it's definitely not a blocker.  It just caught my attention because I've been querying `error` events recently, and sends a different set of data to our analytics.

app/javascript/lib/MessageModal.js
```diff
@@ -15,6 +15,24 @@ export function showMessage(message, key, logProps) {
   functionHolder.showMessage(message, key, logProps)
 }
 
+/** take a json-api error object and render a modal with the contents.
+ * For now, this just uses the 'detail' field of each error.
+ */
+export function showJsonApiErrorMessage(error, key, logProps) {
```
eweitz, line 21: Out of curiosity,  does this trigger [an `error-modal` event log](https://mixpanel.com/s/3EDujS) as well?  How / why would this error modal logging differ from that error modal logging?

app/controllers/api/v1/bulk_download_controller.rb
```diff
@@ -130,7 +130,10 @@ def summary
           # only validate accessions, if present.  TDR/HCA-only downloads will not have SCP accessions present
           self.class.check_accession_permissions(valid_accessions, current_api_user) if valid_accessions.any?
         rescue ArgumentError => e
-          render json: e.message, status: 403 and return
+          # there's a good argument that this should return a 403 rather than 422 when there are studies requested that the user
+          # does not have access to.  However, our current default handling of 401/403 errors in the UX makes that undesirable
+          # once SCP-4010 is addressed, this can be restored to 403.
+          render json: { errors: [{status: 422, detail: e.message, title: 'Invalid studies requested' }] }, status: 422 and return
```
eweitz, line 136:
```suggestion
          # The response uses JSON API formatting -- details: https://jsonapi.org/format/#error-objects
          render json: { errors: [{status: 422, detail: e.message, title: 'Invalid studies requested' }] }, status: 422 and return
```

app/javascript/lib/MessageModal.js
```diff
@@ -15,6 +15,24 @@ export function showMessage(message, key, logProps) {
   functionHolder.showMessage(message, key, logProps)
 }
 
+/** take a json-api error object and render a modal with the contents.
+ * For now, this just uses the 'detail' field of each error.
```
eweitz, line 19: The formal spec is nice and readable; linking it here would be convenient reference.

```suggestion
 * For now, this just uses the 'detail' field of each error.
 * Details on JSON API error object format: https://jsonapi.org/format/#error-objects
```

eweitz (approved):
Hooray JSON API specification!

I suggest refinements for user-facing text and maintainability, nothing blocking.  I also grumble about not being able to wrap my head around our various approaches of logging errors to Mixpanel.

devonbush:
Thanks for the good suggestions, I've included them.  And agree, when we do the upgrade to our error infrastructure, we'll want to systematize the logging as well

PR #1320, opened 2022-01-20
Speed up filtering and resizing scatter plot (SCP-3865)

eweitz:
This optimizes performance and improves separation of concerns for recent legend updates, per a [review comment](https://github.com/broadinstitute/single_cell_portal_core/pull/1240#discussion_r744861850).

Previously, resizing the window would re-compute traces, and updating traces would re-compute layout (among several other things).  Now, resizing and updating traces avoid such unnecessary processing.  This speeds up those interactions.  

It also extracts code from the `processScatterPlot` function, separating reusable chunks into their own functions.  And it cleans up some needless redundancy in the `traceArgs` parameter.

To test:
1. Open `ScatterPlot.js`
2. At top of `getPlotlyLayout` function, add `console.log('in getPlotlyLayout')`
3. At top of `getPlotlyTraces` function, add `console.log('in getPlotlyTraces')`
4. Open Console panel in DevTools, filter to console output that contains "getPlotly"
5. Open a study that has a scatter plot
6. Confirm that `in getPlotlyLayout` and `in getPlotlyTraces` were logged to DevTools Console
7. Resize window
8. Confirm log of `in getPlotlyLayout` and not `in getPlotlyTraces`
9. Click on a label in the legend to filter shown traces
10. Confirm log of `in getPlotlyTraces` and not `in getPlotlyLayout`
11. Change "Clustering" and "Annotation"
12. Confirm log of `in getPlotlyTraces` and `in getPlotlyLayout`
13. Checkout `development`
14. Do steps 1-10, but note how `in getPlotlyLayout` and `in getPlotlyTraces` are always both logged

This satisfies SCP-3865.

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -81,8 +81,21 @@ function RawScatterPlot({
     setHiddenTraces(newShownTraces)
   }
 
+  /** Update layout, without recomputing traces */
+  function resizePlot() {
+    let scatter = Object.assign({}, scatterData)
+    const widthAndHeight = getScatterDimensions(scatter, dimensionProps)
+    scatter = Object.assign(scatter, widthAndHeight)
+    const layout = getPlotlyLayout(scatter)
+
+    Plotly.relayout(graphElementId, layout)
+
+    setScatterData(scatter)
+  }
+
   /** Process scatter plot data fetched from server */
   function processScatterPlot(clusterResponse=null) {
+    console.log('processScatterPlot')
     let [scatter, perfTimes] =
       (clusterResponse ? clusterResponse : [scatterData, null])
 
@@ -158,7 +171,7 @@ function RawScatterPlot({
   }, [cluster, annotation.name, subsample, consensus, genes.join(','), isAnnotatedScatter])
 
   const widthAndHeight = getScatterDimensions(scatterData, dimensionProps, genes)
-  // Handles custom scatter legend updates and window resizing
+  // Handles custom scatter legend updates
   useUpdateEffect(() => {
     // Don't update if graph hasn't loaded
     if (scatterData && !isLoading) {
@@ -166,7 +179,17 @@ function RawScatterPlot({
     }
     // look for updates of individual properties, so that we don't rerender if the containing array
     // happens to be a different instance
-  }, [hiddenTraces.join(','), widthAndHeight.height, widthAndHeight.width])
+  }, [hiddenTraces.join(',')])
+
+  // Handles window resizing
+  useUpdateEffect(() => {
+    // Don't update if graph hasn't loaded
+    if (scatterData && !isLoading) {
+      resizePlot()
+    }
+    // look for updates of individual properties, so that we don't rerender if the containing array
+    // happens to be a different instance
+  }, [widthAndHeight.height, widthAndHeight.width])
 
   // Handles Plotly `data` updates, e.g. changes in color profile
   useUpdateEffect(() => {
--- a/app/javascript/components/visualization/ScatterPlot.js
+++ b/app/javascript/components/visualization/ScatterPlot.js
@@ -92,20 +92,10 @@ function RawScatterPlot({
     const layout = getPlotlyLayout(scatter)
 
     const traceArgs = {
-      axes: scatter.axes,
-      data: scatter.data,
-      annotName: scatter.annotParams.name,
-      annotType: scatter.annotParams.type,
-      genes: scatter.genes,
-      isAnnotatedScatter: scatter.isAnnotatedScatter,
+      genes,
+      isAnnotatedScatter,
       isCorrelatedScatter,
       scatterColor,
-      dataScatterColor: scatter.scatterColor,
-      pointAlpha: scatter.pointAlpha,
-      pointSize: scatter.pointSize,
-      showPointBorders: scatter.showClusterPointBorders,
-      is3D: scatter.is3D,
-      labelCorrelations,
       hiddenTraces,
       scatter
     }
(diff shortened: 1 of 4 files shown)


bistline:
Build failure is tracked in SCP-3995

bistline (approved):

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -81,37 +81,50 @@ function RawScatterPlot({
...
-    const layout = getPlotlyLayout(scatter)
+  /** Update layout, without recomputing traces */
+  function resizePlot() {
+    const scatter = updateScatterLayout()
+    Plotly.relayout(graphElementId, scatter.layout)
+    setScatterData(scatter)
+  }
 
+  /** Update legend counts and recompute traces, without recomputing layout */
+  function updateCountsAndGetTraces(scatter) {
```
devonbush, line 104: two different verbs in a method name often indicate something is amiss...     I think this is best just inlined in `processScatterPlot`--this doesn't really add any substantial logic to `getPlotlyTraces()`, and so having to follow an extra method call makes it harder to read.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -167,16 +180,26 @@ function RawScatterPlot({
     })
   }, [cluster, annotation.name, subsample, consensus, genes.join(','), isAnnotatedScatter])
 
+  // Handles custom scatter legend updates
+  useUpdateEffect(() => {
```
devonbush, line 184: nice

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -295,21 +318,18 @@ function shouldReverseScale(scatterColor) {
...
-  pointAlpha,
-  pointSize,
-  is3D,
   hiddenTraces,
   scatter
 }) {
+  const { axes, data, pointAlpha, pointSize, is3D } = scatter
+  const dataScatterColor = scatter.scatterColor
+  const annotName = scatter.annotParams.name
+  const annotType = scatter.annotParams.type
```
devonbush, line 331: It's nice that now the arguments to this function are all clearly independent state objects--that makes life easier for the caller, but at the expense of more of our code exposed to the `scatter` 'god object'.  I wonder if the best of both worlds is moving the deconstruction to the function parameters like: 
```
function getPlotlyTraces({
  genes, 
  isAnnotatedScatter, 
  scatter: {axes, data, pointAlpha, pointSize, is3D, 
    scatterColor: dataScatterColor, 
    annotParams: {name: annotName, type: annotType}
  }
})```

**Review by devonbush — APPROVED** [review]
Nice, in addition to the performance benefit, this also makes the code easier to follow in many ways.   Some code suggestions, none of which rise to blocking, but I'd appreciate if at least the comments got some extra love

**Comment by eweitz** [comment]
Emily noted that the log noise isn't actually reduced -- production logs the same amount.  Following her good catch, I've updated the test instructions, description, and title.

File: app/javascript/components/visualization/ScatterPlot.js [hunk 4]
```diff
@@ -81,37 +81,50 @@ function RawScatterPlot({
...
 
-  /** Process scatter plot data fetched from server */
-  function processScatterPlot(clusterResponse=null) {
-    let [scatter, perfTimes] =
-      (clusterResponse ? clusterResponse : [scatterData, null])
-
+  /** Get new, updated scatter object instance, and new layout */
+  function updateScatterLayout(scatter=null) {
+    if (!scatter) {
+      scatter = Object.assign({}, scatterData) // New instance forces render
```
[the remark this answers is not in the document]
eweitz: More specifically, it forces re-rendering the legend with updated dimensions (without recomputing traces).  Merely `scatter = scatterData` retains the reference to the old instance, and causes the legend to get out-of-sync with updated dimensions.

Would it make sense to simply update the comment with that?

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -81,37 +81,50 @@ function RawScatterPlot({
     setHiddenTraces(newShownTraces)
   }
 
-  /** Process scatter plot data fetched from server */
-  function processScatterPlot(clusterResponse=null) {
-    let [scatter, perfTimes] =
-      (clusterResponse ? clusterResponse : [scatterData, null])
-
+  /** Get new, updated scatter object instance, and new layout */
```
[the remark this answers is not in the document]
eweitz: Per my [other comment](https://github.com/broadinstitute/single_cell_portal_core/pull/1320#discussion_r789849490), this suggestion isn't quite correct.  If no `scatter` is passed in, it assigns the re-computed width and height to a new instance of the `scatterData` state object.  Having a new instance is important as it forces re-rendering the legend without re-computing traces or counts.

The design here lets us call this function when initializing the entire plot (e.g. loading page, or changing 
"Clustering" or "Annotation") _or_ when specifically only updating the layout (and not traces/counts) upon resizing.

ehanna4 (approved):
Following the new test steps and comparing to prod looks good and outputs as expected.

app/javascript/components/visualization/ScatterPlot.js
eweitz: Agreed, the complicated name is a code smell.  This partly aimed to fix a misleading name.  Previously, we put a setter (`setCountsByLabel`) in a getter (`getPlotlyTraces`), which seemed worse.

To your latter point: inlining it in `processScatterPlot` would require extra parameters and more complexity in the callers (`processScatterPlot` and a `useUpdateEffect`) to retain these performance optimizations.  That seems like a wash in terms of readability.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -81,37 +81,50 @@ function RawScatterPlot({
...
-      dataScatterColor: scatter.scatterColor,
-      pointAlpha: scatter.pointAlpha,
-      pointSize: scatter.pointSize,
-      showPointBorders: scatter.showClusterPointBorders,
-      is3D: scatter.is3D,
-      labelCorrelations,
       hiddenTraces,
       scatter
     }
     const [traces, labelCounts] = getPlotlyTraces(traceArgs)
```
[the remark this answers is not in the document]
eweitz: Agreed, done in de38b69.

app/javascript/components/visualization/ScatterPlot.js
eweitz: I didn't know this was possible!  The syntax is a bit complex and dense, but seems like an excellent compromise.  Done in de38b69.

PR #1321, opened 2022-01-20
Fixing `BulkDownloadControllerTest` for single-study download (SCP-3995)

bistline:
This update (attempts) to address a recurring issue where the test `BulkDownloadControllerTest#test_should_honor_directory_parameter_for_single-study_bulk_download` fails with a mock expectation error where an unexpected argument is passed to the mock.  This test did not fail reproducibly (and almost never locally), but did seem to have an increased error rate in CI when the test was run towards the end of a build.  This fix now simplifies the test to reduce the number of files per directory from 5 to 1, and reuses the same file list each time rather than regenerating it every assertion.  Repeated local runs of single test, whole suite, and complete build have shown that the test does not fail anymore with this particular error.

MANUAL TESTING
1. Pull this branch, and run `rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Run the corresponding test with `bin/rails test test/api/bulk_download_controller_test.rb -n /honor/`
3. Optional: repeatedly run the test using the while loop below (you can exit at any time with `^C`:

```
code=0
while [ $code -eq 0 ]
do
  bin/rails test test/api/bulk_download_controller_test.rb -n /honor/
  code=$?
done
```

This PR satisfies SCP-3995.

### DIFF
--- a/test/api/bulk_download_controller_test.rb
+++ b/test/api/bulk_download_controller_test.rb
@@ -142,10 +142,15 @@ class BulkDownloadControllerTest < ActionDispatch::IntegrationTest
                               public: false,
                               user: @user,
                               test_array: @@studies_to_clean)
-    file_types = %w[csv xlsx]
-    file_types.each do |file_type|
-      files = 1.upto(5).map { |i| "#{file_type}/file_#{i}.#{file_type}" }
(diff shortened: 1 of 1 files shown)


devonbush (approved):

ehanna4 (approved):
followed steps as outlined and saw the test run multiple runs and succeed 👍

PR #1322, opened 2022-01-20
fixing case of download agreement test (SCP-4006)

devonbush:
Minitest scoping is always an adventure.  This both cleans up the delete method, and nil-safes the production check just in case we end up with orphans in the future.

TO TEST:
1.  run `rails test test/integration/download_agreement_test.rb`
2. observe success
3. on line 68, add `assert false` to make the test fail after the download agreement is created
4. run  `rails test test/integration/download_agreement_test.rb`
5. observe test fails
6. remove the `assert false`
7. run  `rails test test/integration/download_agreement_test.rb`
8. observe test succeeds (previously, it would fail due to the agreement not being cleaned

### DIFF
--- a/test/integration/download_agreement_test.rb
+++ b/test/integration/download_agreement_test.rb
@@ -20,6 +20,12 @@ class DownloadAgreementTest < ActionDispatch::IntegrationTest
     sign_in_and_update @user
   end
 
+  # we use after(:each) because after(:all) will not have @download_agreement in scope
+  after(:each) do
+    @download_agreement&.destroy
+    @download_acceptance&.destroy
+  end
+
   test 'should enforce download agreement' do
     # ensure normal download works
     get download_file_path(accession: @study.accession,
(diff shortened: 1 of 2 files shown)


test/integration/download_agreement_test.rb
```diff
@@ -20,6 +20,12 @@ class DownloadAgreementTest < ActionDispatch::IntegrationTest
     sign_in_and_update @user
   end
 
+  # we use after(:each) because after(:all) will not have @download_agreement in scope
+  after(:each) do
+    @download_agreement&.destroy
```
bistline, line 25: I think it might be simpler/faster to call this:
```
DownloadAgreement.destroy_all
DownloadAcceptance.destroy_all
```

If nothing else, it skips callbacks and ensures empty collections, which is what we want.

bistline (approved):

test/integration/download_agreement_test.rb
devonbush: I thought about that -- but if we want our tests to be parallelizable, we want to avoid destroy_all, since that could destroy instances currently in use by other tests

bistline: Fair point - though I don't think these two classes will impact other tests.  But your hunch is likely correct.  We'll have to refactor all `destroy_all` calls when we parallelize.

eweitz (approved):

PR #1323, opened 2022-01-20
Add logger property for front or back app call (SCP-3991)

ehanna4:
Add the logger property to mixpanel calls to indicate whether it was the front or backend of the app making the call.
Screen Shot 2022-01-20 at 2 26 20 PM

To test:
- pull this branch
- do anything clicky in the UI (i.e. click a button or change pages)
- do something back-endy like downloading files
- 
Then check out dev Mixpanel 
(Toggling between Mixpanel envs is done with the dropdown in the upper right)
Screen Shot 2022-01-20 at 2 33 19 PM

Check out the Events tab and filter by the logger property with value of each 'app-backend' and 'app-frontend' 
Screen Shot 2022-01-20 at 2 32 57 PM
Screen Shot 2022-01-20 at 2 33 07 PM

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -365,7 +365,8 @@ export function log(name, props = {}) {
     appId: 'single-cell-portal',
     appPath: getAnalyticsPageName(),
     appFullPath: getAppFullPath(),
-    env
+    env,
+    logger: 'app-frontend'
   }, getDefaultProperties())
 
   const tab = getTabProperty()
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Good stuff!  I understand that Jean is adding `logger: "ingest-pipeline"` over there, and we'll defer `logger: "cli"` for now.

bistline (approved):

ehanna4:
@eweitz correct! I am making the cli ticket as the final AC of this ticket

devonbush (approved):

PR #1328, opened 2022-01-24
Updating bulk download labels & HCA filtering (SCP-3982)

devonbush:
SCP-3982.  The functional change is to only show finalized loom files as 'analysis' files in the UX.  Note that we still pass along every file we find to the UI components in case we want to later refine the UI with message like "[X] contributor generated files may be available here". But the non-finalized loom analysis files are now classified as "contributor_analysis_files and hidden in the UI. 

The info popups have also been improved -- the SCP types link to the appropriate zendesk documentation, and the HCA analysis popup links to the HCA info page. 

image

TO TEST:
1. sign in and have XDSS enabled
2. do a search that yields both HCA and SCP studies (`cow lung`) worked in my environment
3. Confirm the SCP file column info popups link appropriately
4. Confirm the HCA 'analysis' column info popup links appropriately
5. Confirm zero or one file shows up in the analysis column for each HCA study

### DIFF
--- a/app/javascript/components/search/controls/download/DownloadSelectionTable.js
+++ b/app/javascript/components/search/controls/download/DownloadSelectionTable.js
@@ -169,7 +169,7 @@ const COLUMNS = {
   analysis: {
     title: 'Analysis',
     types: ['analysis_file'],
-    info: 'Expression matrix files, including processed or raw count files',
+    info: 'Loom files of finalized raw counts.  This does not include contributor generated files or intermediate files. Those can be downloaded directly from HCA',
     default: true
   },
   sequence: {
--- a/app/javascript/components/search/controls/download/DownloadSelectionTable.js
+++ b/app/javascript/components/search/controls/download/DownloadSelectionTable.js
@@ -144,7 +144,7 @@ const COLUMNS = {
     types: ['Expression Matrix', 'MM Coordinate Matrix', '10X Genes File', '10X Barcodes File'],
     info: <span>
       Expression matrix files, including processed or raw count files.<br/>
-      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610292-Matrix-Files" target="_blank" rel="nofollow noreferrer">
+      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610292-Matrix-Files" target="_blank" rel="noopener noreferrer">
         Matrix file documentation
       </a>
     </span>,
@@ -155,7 +155,7 @@ const COLUMNS = {
     types: ['Cluster'],
     info: <span>
       Clustering coordinate files, including 2D and 3D clustering, as well as spatial.<br/>
-      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060609792-Cluster-Files" target="_blank" rel="nofollow noreferrer">
+      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060609792-Cluster-Files" target="_blank" rel="noopener noreferrer">
         Cluster file documentation.
       </a>
     </span>,
@@ -166,7 +166,7 @@ const COLUMNS = {
     types: ['Metadata'],
     info: <span>
       The listing of all cells in the study, along with associated metadata such as species, cell type, etc.<br/>
-      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610232-Metadata-File-Overview" target="_blank" rel="nofollow noreferrer">
+      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610232-Metadata-File-Overview" target="_blank" rel="noopener noreferrer">
         Metadata file documentation.
       </a>
     </span>,
@@ -182,7 +182,7 @@ const COLUMNS = {
     title: 'Analysis',
     types: ['analysis_file'],
     info: <span>
-      Loom files of <a href="https://broadinstitute.github.io/warp/docs/Pipelines/Optimus_Pipeline/Loom_schema/" target="_blank" rel="nofollow noreferrer">finalized raw counts.</a><br/>
+      Loom files of <a href="https://broadinstitute.github.io/warp/docs/Pipelines/Optimus_Pipeline/Loom_schema/" target="_blank" rel="noopener noreferrer">finalized raw counts.</a><br/>
        This does not include contributor generated files or intermediate files, which can be downloaded directly from HCA
     </span>,
     default: true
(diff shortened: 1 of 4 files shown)


app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -145,19 +142,34 @@ const COLUMNS = {
   matrix: {
     title: 'Matrix',
     types: ['Expression Matrix', 'MM Coordinate Matrix', '10X Genes File', '10X Barcodes File'],
-    info: 'Expression matrix files, including processed or raw count files',
+    info: <span>
+      Expression matrix files, including processed or raw count files.<br/>
+      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610292-Matrix-Files" target="_blank" rel="nofollow noreferrer">
```
eweitz, line 147: Out of curiosity, why `noreferrer`?  Insofar as referrer is ever valuable, I can see it being valuable here.

app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -169,7 +181,10 @@ const COLUMNS = {
   analysis: {
     title: 'Analysis',
     types: ['analysis_file'],
-    info: 'Expression matrix files, including processed or raw count files',
+    info: <span>
+      Loom files of <a href="https://broadinstitute.github.io/warp/docs/Pipelines/Optimus_Pipeline/Loom_schema/" target="_blank" rel="nofollow noreferrer">finalized raw counts.</a><br/>
+       This does not include contributor generated files or intermediate files, which can be downloaded directly from HCA
```
eweitz, line 186: Full stop polish:

```suggestion
      Loom files of <a href="https://broadinstitute.github.io/warp/docs/Pipelines/Optimus_Pipeline/Loom_schema/" target="_blank" rel="nofollow noreferrer">finalized raw counts</a>.<br/>
       This does not include contributor generated files or intermediate files, which can be downloaded directly from HCA.
```

app/javascript/lib/InfoPopup.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { Popover, OverlayTrigger } from 'react-bootstrap'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons'
+import _uniqueId from 'lodash/uniqueId'
+/**
```
eweitz, line 6:
```suggestion

/**
```

eweitz (changes requested):
Code looks mostly good!

Only blocker: a true-positive Azul-related [build failure](https://github.com/broadinstitute/single_cell_portal_core/runs/4916488842?check_suite_focus=true#step:4:4207).  It seems a test assertion needs to be updated.

```
  1) Failure:
AzulSearchServiceTest#test_should_get_summary_file_information_from_project_shortnames [/home/app/webapp/test/integration/lib/azul_search_service_test.rb:159]:
--- expected
+++ actual
@@ -1 +1 @@
-["source", "count", "upload_file_size", "file_format", "accession", "project_id", "file_type"]
+["source", "count", "upload_file_size", "file_format", "accession", "project_id", "is_intermediate", "file_type"]


418 runs, 1720 assertions, 1 failures, 0 errors, 1 skips
```

Beyond that, I suggest a few copyediting refinements, and ask about `noreferrer`.

app/javascript/components/search/controls/download/DownloadSelectionTable.js
bistline: This is a security enhancement to avoid [exploits](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) to `window.opener` in Javascript.

devonbush: Thanks for calling this out.  `noreferrer` became best practice since older versions of firefox didn't support noopener. https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/.  (Incidentally, I just realized I'd put 'nofollow' instead of noopener by mistake).    I see your point about us wanting to preserve the referrer so Zendesk can see where it came from.  But I think the security need outweighs that -- since Zendesk is not inside our FISMA boundary, we need to treat it as a site which may become compromised.  

Reading a bit more, it looks like as of 2021, most major browsers do noopener by default, so this `rel` attribute which was best practice, may no longer be needed given enough time for everyone to update

app/lib/azul_search_service.rb
```diff
@@ -180,26 +180,41 @@ def self.extract_file_information(result)
         upload_file_size: file_summary['totalSize'],
         file_format: file_summary['format'],
         accession: short_name,
-        project_id: project_id
+        project_id: project_id,
+        is_intermediate: file_summary['isIntermediate']
       }
       content = file_summary['contentDescription']
+      assigned_file_type = 'other'
+      puts file_summary
```
bistline, line 188: Debug call

bistline (changes requested):
Looks good, except two small issues in `azul_search_service.rb`.  One is just a debug logging call, but the other looks like it will in fact cause a regression (double assignment of variable).

app/javascript/components/search/controls/download/DownloadSelectionTable.js
eweitz: Good point on using `noreferrer` for security likely being out-dated advice.  Some detail:

[Firefox has supported `rel="noopener"`](https://caniuse.com/rel-noopener) since FF 52 was released in early 2017.  Earlier versions were end-of-lifed long ago, and thus should likely be considered insecure and unsupported in SCP.  

So it seems that article's security argument has not been valid for almost 5 years, and we can securely pass referrer to sites outside our FISMA boundary.

app/javascript/components/search/controls/download/DownloadSelectionTable.js
```diff
@@ -145,19 +142,34 @@ const COLUMNS = {
   matrix: {
     title: 'Matrix',
     types: ['Expression Matrix', 'MM Coordinate Matrix', '10X Genes File', '10X Barcodes File'],
-    info: 'Expression matrix files, including processed or raw count files',
+    info: <span>
+      Expression matrix files, including processed or raw count files.<br/>
+      <a href="https://singlecell.zendesk.com/hc/en-us/articles/360060610292-Matrix-Files" target="_blank" rel="noopener noreferrer">
+        Matrix file documentation
```
eweitz, line 148: Like the others:

```suggestion
        Matrix file documentation.
```

eweitz (approved):

bistline (approved):

PR #1330, opened 2022-01-24
Nil-safing expression- and cluster-based visualization methods (SCP-3990)

bistline:
This update adds various safeguards to expression- and cluster-based visualization methods in the various controllers/service libraries responsible for handling visualization requests, as well as general linting/refactoring for concision.  This is handled mainly through guard clauses to return/render if a top-level entity (such as a `ClusterGroup` object, or genes hash) is not found.  This is mainly an effort to allow normal error handling to catch these instances and let the UI gracefully rescue from these errors, as opposed to server-side errors that cause the Explore tab to break and stop rendering visualizations.

There is also one method (`ClusterVizService.transform_coordinates`) that has been removed entirely as it is no longer used.

MANUAL TESTING
**NOTE**: Before beginning, make sure that caching is disabled by running `bin/rails dev:cache` from the project root directory until the message `Development mode is no longer being cached.` is displayed.

The best way to test this particular ticket is through the Swagger API documentation:

1. Boot as normal, and get the accession of any study with clustering & gene expression data (e.g. the synthetic study `Human HIV and Tuberculosis, blood study`)
2. Open the Swagger API documentation from the `Help and resources` menu, and sign in via the "Authorize" button at the top
3. Go down to the "Expression" [data rendering](https://localhost:3000/single_cell/api/v1#/Visualization/study_visualization_expression_show) endpoint
4. Enter the study accession from above, and the following parameters:
```
genes: AGPAT2
annotation_name:  cell_type__ontology_label
annotation_type: group
annotation_scope: study
```
5. Confirm the request executes as expected and returns `200`
6. Enter `does-not-exist` for the `cluster` value and re-run the request
7. Confirm the response code is `404` and the error message is `"error": "Requested cluster not found"`

This PR satisfies SCP-3990.

### DIFF
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -89,65 +89,31 @@ class ClusterVizService
   # only options allowed are 1000, 10000, 20000, and 100000
   # will only provide options if subsampling has completed for a cluster
   def self.subsampling_options(cluster)
-    num_points = cluster.points
-    if cluster.is_subsampling?
-      []
-    else
-      ClusterGroup::SUBSAMPLE_THRESHOLDS.select {|sample| sample < num_points}
-    end
+    return [] if cluster.nil? || cluster.is_subsampling?
+
+    ClusterGroup::SUBSAMPLE_THRESHOLDS.select { |sample| sample < cluster.points }
   end
 
   # return an array of values to use for subsampling dropdown scaled to number of cells in study
   # only options allowed are 1000, 10000, 20000, and 100000
   # will only provide options if subsampling has completed for a cluster
   def self.default_subsampling(cluster)
-    num_points = cluster.points
-    if num_points < 100000
-      return nil
-    else
-      ClusterGroup::SUBSAMPLE_THRESHOLDS.select{|val| val <= 100000}.max
-    end
-  end
+    return nil if cluster.nil? || cluster.points < 100000
 
-  # Convert cluster group data array points into JSON plot data for Plotly
-  def self.transform_coordinates(coordinates, study, cluster_group, selected_annotation)
-    plot_data = []
-    plot_type = cluster_group.is_3d? ? 'scatter3d' : 'scattergl'
-    coordinates.sort_by {|k,v| k}.each_with_index do |(cluster, data), index|
-      cluster_props = {
-        x: data[:x],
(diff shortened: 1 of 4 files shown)


app/lib/cluster_viz_service.rb
```diff
@@ -250,21 +220,24 @@ def self.load_cluster_group_data_array_points(study, cluster, annotation, subsam
 
   # returns an array of the values for the given annotation, sorted by the cells of the given cells array
   def self.get_annotation_values_array(study, cluster, annotation, cells, subsample_annotation, subsample_threshold)
-    annotation_array = []
-    annotation_hash = {}
+    return [] if study.nil? || cluster.nil?
+
     # Construct the arrays based on scope
-    if annotation[:scope] == 'cluster'
+    case annotation[:scope]
```
devonbush, line 226: if/elsif can be clearer than `case` in situations like this where the number of  alternatives is small, because it can be challenging to remember how case/when behaves with respect to needing explicit break statements, since `case` behaves differently in different languages.

devonbush (approved):
that's a lot of nil-safing :)  This is good stuff -- should make our visualizations much more robust

app/lib/cluster_viz_service.rb
```diff
@@ -89,65 +89,31 @@ def self.load_image_options(study)
...
-        cluster_props[:z] = data[:z]
-        cluster_props[:textposition] = 'bottom right'
-      end
-
-      plot_data.push(cluster_props)
-    end
+    return nil if cluster.nil? || cluster.points < 100000
 
-    plot_data
+    ClusterGroup::SUBSAMPLE_THRESHOLDS.select { |val| val <= 100000 }.max
```
eweitz, line 103: Less hard-coding:

```suggestion
    ClusterGroup::SUBSAMPLE_THRESHOLDS.select { |val| val <= ClusterGroup::MAX_THRESHOLD }.max
```

eweitz (approved):

app/lib/cluster_viz_service.rb
bistline: Agreed, but rubocop was barking at me uniformly about converting all `if/elsif/else` blocks into case statements.

bistline:
Build failure addressed in SCP-4029

PR #1331, opened 2022-01-25
Fixing broken single-study bulk download, format curl command for all OS types (SCP-4001)

bistline:
This update fixes several new bugs with bulk download:

1. Removes `&&` in favor of `;` for command chaining as ampersands are not an allowed character in Windows PowerShell
2. Fixes a broken string interpolation that prevented single-study bulk download from opening the modal and rendering curl commands
3. Removes the `--compressed` directive for Windows machines as this is not supported by most installations of `curl.exe`

MANUAL TESTING
1. Sign in with any account
2. Run a search on the home page, progress through the bulk download UX, and validate that the `curl` command succeeds
3. Open any study that you do not have bucket permissions for (the easiest way is to sign in with a personal Gmail account and load any public study with files) and validate that the bulk download modal renders and commands work

This PR supports SCP-4001.

### DIFF
--- a/app/javascript/components/search/controls/download/DownloadCommand.js
+++ b/app/javascript/components/search/controls/download/DownloadCommand.js
@@ -109,7 +109,7 @@ function getDownloadCommand(authCode, downloadId) {
   // instead of the hostname
   const downloadCommand = (
     `${curlExec} "${url}" -${curlSecureFlag}o cfg.txt; ` +
-    `${curlExec} -K cfg.txt && rm cfg.txt` // Removes only if curl succeeds
+    `${curlExec} -K cfg.txt; rm cfg.txt` // Removes only if curl succeeds
   )
 
   return downloadCommand
--- a/app/javascript/components/search/controls/download/DownloadCommand.js
+++ b/app/javascript/components/search/controls/download/DownloadCommand.js
@@ -85,7 +85,8 @@ export default function DownloadCommand({ fileIds=[], azulFiles }) {
 function getDownloadCommand(authCode, downloadId) {
   // Get client os and determine correct curl invocation
   const clientOS = bardUtils.info.os()
-  const curlExec = clientOS.match(/Win/) ? 'curl.exe' : 'curl'
+  const isWindows = clientOS.match(/Win/)
+  const curlExec = isWindows ? 'curl.exe' : 'curl'
   const queryParams = {
     auth_code: authCode,
     download_id: downloadId,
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good.  How can folks with macOS laptops confirm this works on Windows?

jlchang (approved):
I'm able to curl download (where && has been replaced by ;) study files from my local dev environment. I see in the code that for my Mac I should expect --compressed in cfg.txt and it is only removed for Windows OS.

This passes the functional assessment of this PR.

app/javascript/components/search/controls/download/DownloadCommand.js
```diff
@@ -109,7 +109,7 @@ function getDownloadCommand(authCode, downloadId) {
   // instead of the hostname
   const downloadCommand = (
     `${curlExec} "${url}" -${curlSecureFlag}o cfg.txt; ` +
-    `${curlExec} -K cfg.txt && rm cfg.txt` // Removes only if curl succeeds
+    `${curlExec} -K cfg.txt; rm cfg.txt` // Removes only if curl succeeds
```
devonbush, line 112: BLOCKER: this changes the behavior, now we will delete cfg.txt regardless of success/fail.  I prefer the old behavior, since it makes troubleshooting much easier, but it's a bummer it's not supported in powershell.  Should we just tell users to use a cmd.exe shell instead?  Supporting two different windows download shells is going to be tricky.  

In any case, at the very least this needs the comment updated

devonbush (changes requested):
Nice fixups, would love to preserve the cfg.txt file if an error occurs though....

bistline:
> Code looks good. How can folks with macOS laptops confirm this works on Windows?

You won't be able to completely test that it works, but you can at least confirm that the Windows-specific code is triggering on the backend.  Make an edit to app/controllers/api/v1/bulk_download_controller.rb, line 303, enter os = 'Windows' to force Windows cURL configuration

app/javascript/components/search/controls/download/DownloadCommand.js
bistline: I agree that supporting multiple shells is not ideal.  After a lot of searching/testing, it appears we can do the following for PowerShell:

```
curl.exe -K cfg.txt ; if ($?) { rm cfg.txt }
```

I tested this and it worked.  I also did a whole bulk download request where I replaced `cfg.txt` in the middle command with `foo.txt`, and it correctly errored out and left `cfg.txt` in place.  I'm going to make this update then validate locally that the commands are generating properly.

devonbush (approved):

devonbush:
Thanks for digging into this -- glad there was a way to preserve the cfg file when the command fails that's multi-shell compliant!

PR #1333, opened 2022-01-25
Primary Human Data must have human as species (SCP-4030)

ehanna4:
If the file has been indicated to be a Primary Human Data sequence file the species should be human. Previously the user would be able to change the species even when having selected this. This will fix it so that a user cannot change the species when the Primary Human Data option has been selected. Please note a user is able to have the species be a human without having chosen the Primary Human Data.

Current/Previous Behavior:
Screen Shot 2022-01-25 at 1 57 31 PM

To test:
- Pull this branch
- go to the wizard to the sequence file upload section
- click back and forth between Yes and No on the Primary Human Data option and try changing the species

Video showing the new behavior plus printing the value in the "selectedSpecies" variable
https://user-images.githubusercontent.com/54322292/151044228-f6f58f43-8aa1-4dc3-8973-8fc03729ab8a.mov

### DIFF
--- a/app/javascript/components/upload/SequenceFileForm.js
+++ b/app/javascript/components/upload/SequenceFileForm.js
@@ -45,6 +45,7 @@ export default function SequenceFileForm({
     requiredFields = BAM_REQUIRED_FIELDS
   }
   const validationMessages = validateFile({ file, allFiles, requiredFields, allowedFileExts })
+  const humanTaxon = speciesOptions.find(opt => opt.label === 'human')
 
   return <div className="row top-margin" key={file._id}>
     <div className="col-md-12">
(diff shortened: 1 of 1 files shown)


ehanna4:
https://user-images.githubusercontent.com/54322292/151044481-226c9b2f-3e5e-44a1-8b03-3084bc736d98.mov

devonbush (approved):
Nicely done!

bistline (approved):

eweitz (commented):
Code looks good!  

Is this also relevant for sync?

ehanna4:
> Code looks good!
> 
> Is this also relevant for sync?

I don't know, I believe this behavior had existed in the previous wizard we just missed bringing it to the new one, so since sync wasn't changed if it was there it should still be there and if not we can add it with the new update :)

eweitz (approved):

PR #1335, opened 2022-01-26
Remove the incorrect validations from dense file validation (SCP-4034)

ehanna4:
parseDenseMatrix() was incorrectly and unnecessarily calling validateMetadataLabelMatches() and validateGroupColumnCounts(). This issue was brought to light when the file MIT_Milk_Study_Raw_counts.txt was attempted to be uploaded and the warnings were about issues that should not be applied to dense matrices.

Before:
Screen Shot 2022-01-26 at 9 55 30 AM

This removes those incorrect validations.
To test:
- Pull this branch
- Have a study you can upload a raw counts matrix to
- download MIT_Milk_Study_Raw_counts.txt.gz from SCP1671 in prod
- upload this file
- note it will take a couple of minutes
- observe no warning messages.

Success Now:
Screen Shot 2022-01-26 at 11 44 18 AM

Also updating an issue type to more precisely indicate its function - updated from 'format' to 'content' as the check is more accurately on the content of the file.

### DIFF
--- a/app/javascript/lib/validation/expression-matrices-validation.js
+++ b/app/javascript/lib/validation/expression-matrices-validation.js
@@ -3,8 +3,7 @@
 */
 
 import {
-  getParsedHeaderLines, parseLine,
-  validateUniqueCellNamesWithinFile, validateMetadataLabelMatches, validateGroupColumnCounts
+  getParsedHeaderLines, parseLine, validateUniqueCellNamesWithinFile
 } from './shared-validation'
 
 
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good!

Not blocking, but it might be worthwhile to note in the docstring for `validateGroupColumnCounts` that it only applies for metadata and cluster files.

devonbush (approved):

jlchang (approved):
confirmed the warning messages no longer appear.

was not able to trigger  "content:invalid-type:not-numeric" to confirm the errorType update but that may be related to the bug in SCP-4036. I've documented the issue in that ticket so it can be addressed at the appropriate time.

PR #1337, opened 2022-01-26
Adding custom colors for grouped scatter plots (SCP-2943)

devonbush:
SCP-2943.  This ended up going much faster than expected, so this is, IMHO, actually mergeable code, rather than the prototype suggested by the ticket.  It may not be fully-featured (e.g. we might want a 'reset all to default' button somewhere, or better help text and/or validation), but I would be comfortable releasing this in its current state to get the initial user feedback

If a user has edit permissions for a study, they will see a 'customize colors' option over the legend.  
image
Clicking customize colors puts the legend into color-editing mode (so you can no longer show/hide traces). 
image
Then clicking a legend entry brings up a color picker
image
The user can then hit 'Save colors' to save any colors they picked to the server.  Once saved, the colors will be visible to all users.

TO TEST:
1.  Open a study for which you do not have edit permissions
2. confirm scatter plot legend appears as before, and you cannot customize colors
3. Open a study for which you do have edit permissions
4. On the explore tab for a cluster, click 'customize colors'
5. Click one or more legend entries, and change the colors
6. click 'save colors'
7. refresh the page, confirm new colors still appear
8. sign out, and view the study, confirm new colors still appear


app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -116,6 +116,13 @@ def bam_file_info
           }
         end
 
+        def study_user_info
+          render json: {
+            annotations: AnnotationVizService.available_annotations(@study, cluster: nil, current_user: current_api_user),
+            canEdit: @study.can_edit?(current_api_user)
```
bistline, line 122: Non-blocking: we should use the Ruby convention for formatting (snake case) and call this `can_edit`.

bistline (changes requested):
I'm not sure why, but it looks as though `clusterFileId` is not being propagated properly.  I followed the manual testing instructions, and the results never saved.  Looking at the logs, I see this:
```
Started PATCH "/single_cell/api/v1/studies/SCP18/study_files/undefined" for 127.0.0.1 at 2022-01-26 14:24:45 -0500
...
Filter chain halted as :set_study_file rendered or redirected
Completed 404 Not Found in 1265ms (MongoDB: 0.6ms | Allocations: 55881)
```

ehanna4:
It seems that clicking `cancel` is giving me strange `return to default` _looking_ behavior (while i remain the user who did the change)

https://user-images.githubusercontent.com/54322292/151233760-bc66edc1-7baf-4b30-b086-45fdc84581fd.mov

_but_ it looks like the color doesn't actually return to default (as seen when i sign out the color stayed the updated color)
Screen Shot 2022-01-26 at 2 34 57 PM

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -203,6 +295,31 @@ export default function ScatterPlotLegend({
           </>
           }
         </div>
+        { enableColorPicking &&
+          <div>
+            { showColorControls && !isColorSaving &&
+              <>
+                <span>Click a label to select a new color</span><br/>
+                <div>
+                  <a role="button" data-analytics-name="legend-color-picking-save" onClick={saveColors}>
```
eweitz, line 304: I think analysts would more likely intuit it a bit differently.

```suggestion
                  <a role="button" data-analytics-name="legend-color-picker-save" onClick={saveColors}>
```

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -203,6 +295,31 @@ export default function ScatterPlotLegend({
...
+                    Save colors <FontAwesomeIcon icon={faSave}/>
+                  </a>
+                  <a role="button" className="pull-right" data-analytics-name="legend-color-picking-save" onClick={cancelColors}>
```
eweitz, line 307: Blocker: `cancel`, not `save`.

```suggestion
                  <a role="button" className="pull-right" data-analytics-name="legend-color-picker-cancel" onClick={cancelColors}>
```

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -203,6 +295,31 @@ export default function ScatterPlotLegend({
...
+                    Cancel
+                  </a>
+                </div>
+              </>
+            }
+            { showColorControls && isColorSaving &&
+              <LoadingSpinner/>
+            }
+            { !showColorControls &&
+              <a role="button" data-analytics-name="legend-color-picking-show" onClick={() => setShowColorControls(true)}>
```
eweitz, line 317:
```suggestion
              <a role="button" data-analytics-name="legend-color-picker-show" onClick={() => setShowColorControls(true)}>
```

app/models/cluster_file_info.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  # merges in a color update to the custom_colors hash
+  # all colors for the given annotations will be overwritten,
+  # but annotations not listed in update_colors will be unchanged
+  # e.g. if custom_colors = {
+  #  'annot1': {'label1': '#ee1166'},
+  #  'annot2': {'foo': '#336677'}
+  # }
+  # merge_color_updates(file, { 'annot1': { 'label2': '#334499'}}) will delete the prior custom color for label1
+  # but will leave the entry for 'annot2' unchanged.
```
eweitz, line 16: Thank you!

eweitz (changes requested):
Code looks good, except for one trivial blocker.

Beyond that, I suggest a handful of non-blocking refinements.

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -203,6 +290,32 @@ export default function ScatterPlotLegend({
...
+                <span>Click a label to select a new color</span><br/>
+                <div>
+                  <a role="button" data-analytics-name="legend-color-picking-save" onClick={saveColors}>
+                    Save colors <FontAwesomeIcon icon={faSave}/>
+                  </a>
+                  <a role="button" className="pull-right" data-analytics-name="legend-color-picking-save" onClick={cancelColors}>
+                    Cancel
+                  </a><br/>
+                  &nbsp;
+                  <a role="button" className="pull-right" data-analytics-name="legend-color-picking-save" onClick={resetColors}>
```
eweitz, line 306: Blocker: `reset`, not `save`.

```suggestion
                  <a role="button" className="pull-right" data-analytics-name="legend-color-picking-reset" onClick={resetColors}>
```

eweitz (changes requested):

app/controllers/api/v1/visualization/explore_controller.rb
devonbush: IIRC, we decided not to use snake case on endpoints explicitly designed for javascript consumption.  See, e.g. the `get_study_explore_info` method.  

I'd be ok revisiting that decision, since it tends to be easier to just assume "anything we get from the API is snake-cased" rather than have to know "models get returned snake_cased, custom APIs are camel".  For this PR, though, I think it's good practice to just match the other explore_controller methods

devonbush: added to engineering strategy rolling agenda

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -94,6 +101,34 @@ function RawScatterPlot({
     return scatter
   }
 
+  /** Save any changes to the legend colors */
+  async function saveCustomColors(newColors) {
+    const colorObj = {}
+    colorObj[annotation.name] = newColors
```
bistline, line 107: I've noted that if you're making an edit on the default annotation, this gets set to `""`, when it should either be `_default`, or ideally the actual name of the default annotation.

bistline (approved):
Looks good - one stumbling block for me was that I needed to clear the visualization cache for this to work properly the first time.  It was fine with caching off as well.  When we release this, we'll want a migration to completely clear the cache to avoid any visualization errors.

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -1,8 +1,14 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { log } from 'lib/metrics-api'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faPalette, faSave } from '@fortawesome/free-solid-svg-icons'
```
eweitz, line 4:
```suggestion
import { faPalette } from '@fortawesome/free-solid-svg-icons'
```

app/javascript/components/visualization/controls/ScatterPlotLegend.js
```diff
@@ -203,6 +291,32 @@ export default function ScatterPlotLegend({
...
+                <span>Click a label to select a new color</span><br/>
+                <div>
+                  <a role="button" data-analytics-name="legend-color-picker-save" onClick={saveColors}>
+                    Save colors
+                  </a>
+                  <a role="button" className="pull-right" data-analytics-name="legend-color-picker-cancel" onClick={cancelColors}>
+                    Cancel
+                  </a><br/>
+                  &nbsp;
+                  <a role="button" className="pull-right" data-analytics-name="legend-color-picking-reset" onClick={resetColors}>
```
eweitz, line 307: Blocker: sorry, my last suggestion missed this:

```suggestion
                  <a role="button" className="pull-right" data-analytics-name="legend-color-picker-reset" onClick={resetColors}>
```

eweitz (changes requested):
Kudos on leap-frogging the prototype phase, BTW -- this PR is 1337.  (Sorry, couldn't resist.)

I found another trivial blocker.  After that, this code looks good to merge.

eweitz (approved):

devonbush:
Test failure is unrelated -- in DataRepoClientTest#test_should_get_snapshot

PR #1339, opened 2022-01-26
Custom `exceptions_app` for handling/rendering errors (SCP-3974)

bistline:
Currently, SCP intercepts all instances of `Exception` (base class of all errors) and renders error views depending on environment/controller.  This is (apparently) ill-advised, as there are some low-level exceptions that when intercepted in this fashion can have serious consequences, or can otherwise lead to other unexpected errors such as `AbstractController::DoubleRenderError` or `LocalJumpError`, neither of which end up being handled correctly.  This is even called out explicitly in the [ActionController overview](https://guides.rubyonrails.org/action_controller_overview.html#rescue-from) from Ruby on Rails:

> Using rescue_from with Exception or StandardError would cause serious side-effects as it prevents Rails from handling exceptions properly. As such, it is not recommended to do so unless there is a strong reason.

This update undoes that paradigm and removes many of the `rescue_from` blocks and replaces them with two separate `ExceptionsController` instances that are added to the stack as the custom `exceptions_app`.  This allows for custom handling of all uncaught exceptions, similar to how `rescue_from` worked, but in a much more stable and extensible fashion.  This also allows for direct testing of our error handling now, whereas before this was much more narrow and difficult.  In addition, this update applies a default response format of `json` to all API responses. 

The main key is the configuration in `config/application.rb`, where `config.exceptions_app` is passed a `lambda` to allow inspection of the environment.  This allows us to look at the incoming request & application stack to decide where to route errors based off of the request path and current environment.  Under this new configuration:

1. All API-based uncaught exceptions are now handled by `Api::V1::ExceptionsController`, which will render a JSON response containing the error message, class name, and the first caller location from the backtrace
2. All non-API uncaught exceptions are now handled by the regular `ExceptionsController`, which will either render an HTML or JSON view of the exception, depending on the request format
3. In the `development` environment only, non-API exceptions will default back to the standard Rails error page, with debug information for developers

The difference in the two exceptions controllers is mostly for upstream error reporting and logging.

MANUAL TESTING
1. Boot as normal
2. In your IDE, add `raise 'this is an error'` to `app/controllers/api/v1/site_controller.rb`, line 48 (in `def studies`)
3. Open the Swagger API documentation, and go to the [SiteController#studies](https://localhost:3000/single_cell/api/v1#/Site/site_studies_path) endpoint
4. Execute the request, and validate the error is caught and rendered as below with a status code of `500`:
Screen Shot 2022-01-26 at 4 41 34 PM

This PR satisfies SCP-3974.


app/controllers/exceptions_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  layout 'application'
+
+  def render_error
+    @exception = request.env["action_dispatch.exception"]
+    MetricsService.report_error(@exception, request, current_user, @study)
+    ErrorTracker.report_exception(@exception, current_user, @study, params)
+    respond_to do |format|
+      format.html { render action: 'render_error', status: :internal_server_error }
+      format.json do
+        render json: {
```
devonbush, line 12: this method is very similar to the API method except for a couple of lines.  There might be room for a general `RequestUtils.handle_error(request, user, is_json)` method, or at least  extract the below lines out into a `RequestUtils.format_error_as_json(exception)` method so it's standardized between here and the API

config/application.rb
```diff
@@ -33,6 +33,19 @@ class Application < Rails::Application
 
     config.autoload_paths << Rails.root.join('lib')
 
+    # custom exceptions handling to render responses based on controller
+    # uncaught API errors will now render as JSON responses w/ 500 status
+    # normal controller errors will show 500 page, except in development environment (will show normal error page)
+    # lambda allows for inspecting environment, which will include request parameters
+    config.exceptions_app = lambda do |env|
```
devonbush, line 40: this is some gnarly stuff--nice work digging it all out!
It seems odd to have a switch based on environment in this file.  

If possible, I'd lean towards just redefining `config.exceptions_app` in `config/environments/development.rb` so that we're not mixing environments in config files

config/routes.rb
```diff
@@ -1,104 +1,106 @@
...
-        resources :reports, only: [:show], param: :report_name
-        resources :studies, only: [:index, :show, :create, :update, :destroy] do
-          post 'study_files/bundle', to: 'study_files#bundle', as: :study_files_bundle_files
-          resources :study_files, only: [:index, :show, :create, :update, :destroy] do
-            member do
-              post 'parse', to: 'study_files#parse'
-              patch 'chunk', to: 'study_files#chunk'
-            end
+    # API Routes, default all responses to JSON
+    defaults format: :json do
```
devonbush, line 8: this is nice -- I didn't know about this `defaults` method

devonbush (changes requested):
Good stuff.  Just a few cleanup suggestions

app/lib/request_utils.rb
```diff
@@ -154,4 +153,21 @@ def self.format_path_for_os(path, os = '')
       path
     end
   end
+
+  # handle upstream reporting/logging of errors in custom exceptions controllers
+  def self.log_exception(request, params, user: nil, study: nil)
+    @exception = request.env['action_dispatch.exception']
+    MetricsService.report_error(@exception, request, user, study)
+    ErrorTracker.report_exception(@exception, user, params)
+    Rails.logger.error ([@exception.message] + @exception.backtrace).join($/)
```
eweitz, line 162: Good stuff!

app/controllers/exceptions_controller.rb
bistline: 7f87f88

config/application.rb
bistline: 7f87f88

eweitz (approved):

app/controllers/api/v1/exceptions_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Api
+  module V1
+    # lightweight wrapper to intercept all uncaught API-based exceptions and render JSON response
+    class ExceptionsController < ApiBaseController
+      def render_error
+        ::RequestUtils.log_exception(request, params, user: current_api_user, study: @study)
+        render json: ::RequestUtils.exception_json(request.env['action_dispatch.exception']),
```
devonbush, line 7: nice, let's just pass in the full request instead of the `request.env['action_dispatch.exception']`, so that the signature better matches the signature of the `log_exception` method

app/controllers/exceptions_controller.rb
```diff
@@ -0,0 +1,15 @@
+# custom exceptions controller to render error views
+class ExceptionsController < ApplicationController
+  layout 'application'
+
+  def render_error
+    RequestUtils.log_exception(request, params, user: current_user, study: @study)
+    respond_to do |format|
+      format.html { render action: 'render_error', status: :internal_server_error }
+      format.json do
+        render json: RequestUtils.exception_json(request.env['action_dispatch.exception']),
```
devonbush, line 10: same as above

devonbush (approved):
great, a couple non-blocking suggestions for making method signatures match-up better

bistline:
Build failure seems to be due to Azul updating search indexes (expected results did not return, but subsequent builds failed with slightly different messages).

PR #1341, opened 2022-01-28
Consolidating metadata annotation processing (SCP-4018)

devonbush:
SCP-4018.  We are currently making 2 calls per metadata column on every study overview page load (and then we make the same calls again for the explore_info endpoint).   So for a convention compliant study, that's often 40+ calls.   Put together, these calls represented almost a third of DB calls for loading a convention-compliant study in the explore tab.  This consolidates that down to one call each.  There is certainly other work to be done on SCP-4018, but this seemed a big enough chunk that it was worth its own PR.

This also fixes a bug where annotations that were not shown due to having a corresponding ontology label were instead listed as having 'too many labels'

TO TEST:
 1. load a metadata-compliant study in explore tab.  
 2. confirm it renders correctly, and all annotations appear
 3. Confirm the default cluster&annotation menus in the study settings tab render and operate correctly
 3.   run `rails test test/integration/lib/annotation_viz_service_test.rb`

### DIFF
--- a/app/lib/annotation_viz_service.rb
+++ b/app/lib/annotation_viz_service.rb
@@ -111,7 +111,7 @@ class AnnotationVizService
   # returns a flat array of annotation objects, with name, scope, annotation_type, and values for each
   def self.available_annotations(study, cluster: nil, current_user: nil, annotation_type: nil)
     annotations = []
-    metadata_annots = available_metadata_annotations(study, annotation_type)
+    metadata_annots = available_metadata_annotations(study, annotation_type: annotation_type)
     annotations.concat(metadata_annots)
     cluster_annots = []
     if cluster.present?
@@ -139,8 +139,9 @@ class AnnotationVizService
     all_metadata = study.cell_metadata.to_a
     all_names = all_metadata.map(&:name)
     all_metadata.map do |annot|
-      # viewable if the type is numeric or there's no corresponding label
-      is_viewable = annot.annotation_type == 'numeric' || all_names.exclude?(annot.name + '__ontology_label')
+      # viewable if the type is numeric or there's no corresponding label and it's within the range of visualization values
+      is_viewable = annot.annotation_type == 'numeric' ||
+        all_names.exclude?(annot.name + '__ontology_label') && CellMetadatum::GROUP_VIZ_THRESHOLD === annot.values.size
       {
         name: annot.name,
         type: annot.annotation_type,
--- a/app/lib/annotation_viz_service.rb
+++ b/app/lib/annotation_viz_service.rb
@@ -111,17 +111,8 @@ class AnnotationVizService
   # returns a flat array of annotation objects, with name, scope, annotation_type, and values for each
   def self.available_annotations(study, cluster: nil, current_user: nil, annotation_type: nil)
     annotations = []
-    viewable = study.viewable_metadata
-    metadata = annotation_type.nil? ? viewable : viewable.select {|m| m.annotation_type == annotation_type}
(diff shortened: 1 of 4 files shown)


app/lib/annotation_viz_service.rb
```diff
@@ -111,17 +111,8 @@ def self.user_annot_to_annot(user_annotation, cluster)
...
-    metadata = metadata.map do |annot|
-      {
-        name: annot.name,
-        type: annot.annotation_type,
-        values: sanitize_values_array(annot.values, annot.annotation_type),
-        scope: 'study'
-      }
-    end
-    annotations.concat(metadata)
+    metadata_annots = available_metadata_annotations(study, annotation_type: annotation_type)
```
eweitz, line 114: I recently learned that in [Ruby 3.1](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/), values in hash literals and keyword arguments can be omitted.  So we could do `available_metadata_annotations(study, annotation_type:)`, pretty much like JavaScript.

With [Ruby 2.6 EOL in two months](https://endoflife.date/ruby), do you or @bistline happen to know of any dependencies that'd hold us back from upgrading to 3.1 as opposed to 2.7?

eweitz (approved):
Optimizations like this are always satisfying.  Code looks good!

I found no issues, just asked a tangential question on an EOL matter I just found out about.

devonbush:
Test failure is unrelated in `DataRepoClientTest#test_should_get_snapshot `

app/lib/annotation_viz_service.rb
bistline: Our Docker image uses [Phusion Passenger](https://github.com/phusion/passenger-docker) as the underlying layer.  They have yet to update to Ruby 3.1 (although they are at 3.0.3).  We'd need to wait for them to catch up, and then we'd also need to figure out the issues with updating to Ubuntu focal (20.04) and our base image build automation.  I tried to update the managed Google base image to 20.04, and I could not get it to build due to some really low-level weirdness with platform architecture.

app/lib/annotation_viz_service.rb
```diff
@@ -139,34 +130,27 @@ def self.available_annotations(study, cluster: nil, current_user: nil, annotatio
...
   end
 
+  # helper method to efficiently list out metadata annotations classed as valid/invalid for visualization
+  def self.available_metadata_annotations(study, annotation_type: nil)
+    # get all the metadata in a single query
+    all_metadata = study.cell_metadata.to_a
+    all_names = all_metadata.map(&:name)
+    all_metadata.map do |annot|
+      # viewable if the type is numeric or there's no corresponding label and it's within the range of visualization values
+      is_viewable = annot.annotation_type == 'numeric' ||
```
bistline, line 143: This can be massively simplified.  It does include another DB call, but doing `exists?` does not actually load any data, and is incredibly fast (sub-100ms)
```suggestion
      is_viewable = annot.can_visualize?
```

bistline (approved):
Non-blocking suggestion on simplifying some code that I don't think will impact performance in any meaningful way.

app/lib/annotation_viz_service.rb
devonbush: it's fast, but it would still require one DB call per metadata column, so for the typical convention-compliant study, that's 12ish DB calls, and then those twelve would repeat for the `explore` API call.  So we're saving ~24 DB calls.  At ~0.05sec per call (what I see locally for DB call time), that's a whole second

PR #1346, opened 2022-02-01
Adding feature_flag info to study file_info endpoint (SCP-3955)

bistline:
This update allows for `FeatureFlagOption` info to be retrieved from the `Api::V1::StudiesController#file_info` endpoint.  This now allows for the React upload wizard to check for user- or study-based exemptions for raw counts & metadata convention.  This simplifies the feature flag checking code, and allows for future refactoring to remove DOM-based rendering/reading of `FeatureFlagOption` values and pivot to an API-based model.  This update also swaggerizes the `Api::V1::StudiesController#file_info` endpoint for easier testing/debugging moving forward.

MANUAL TESTING
1. Boot as normal
2. In the rails console, set the `default_value` for `raw_counts_required_frontend` to `true`
```
FeatureFlag.find_by(name: 'raw_counts_required_frontend').update(default_value: true)
```
3. Back in the UI, sign in with an admin account and go to the upload wizard for any study that has no raw count expression files and click into the 'processed expression' tab
4. Validate that the upload form is disabled:
Screen Shot 2022-02-01 at 2 14 37 PM
5. In a separate tab, go to the 'Feature Flags' admin console by selecting 'Feature Flags' from the profile dropdown menu
6. Select `study` for `Object class`, and then enter the accession the study from above
Screen Shot 2022-02-01 at 2 25 34 PM
7. Set the value for `raw_counts_required_frontend` to `false` and save
8. Back in the upload wizard, refresh the page and validate that the form is now enabled
Screen Shot 2022-02-01 at 2 11 46 PM
9. (Optional) Back in the Feature Flag console, select `user` for `Object class` and enter the email of the current user
10. (Optional) Set `raw_counts_required_frontend` to `true` and save
11. (Optional) Back in the upload wizard, refresh the page and validate that the form is still enabled (study-based feature flags should take precedence over user-based ones in the upload wizard)

This PR satisfies SCP-3955.

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -85,10 +85,131 @@ module Api
       end
 
       # return JSON of the study, all study file objects, and any options values needed for the upload wizard
+      # also includes feature flags, with study flags taking precedence over user flags
+      swagger_path '/studies/{id}/file_info' do
+        operation :get do
+          key :tags, [
+            'Studies'
+          ]
+          key :summary, 'Get Study info'
+          key :description, 'Gets information about a single Study'
+          key :operationId, 'study_file_info_path'
+          parameter do
+            key :name, :id
+            key :in, :path
+            key :description, 'ID of Study to fetch'
+            key :required, true
+            key :type, :string
+          end
+          response 200 do
+            schema do
+              property :study do
+                key :description, 'Study object'
+                key :title, 'Study'
+                key :'$ref', :Study
+              end
+              property :files do
+                key :type, :array
+                key :title, 'Array<StudyFile>'
+                items do
+                  key :'$ref', :StudyFile
+                end
+              end
+              property :feature_flags do
+                key :type, :object
+                key :title, 'FeatureFlagOptions'
+                key :description, 'Hash of FeatureFlagOption values w/ defaults, e.g. flag_name => flag_value'
+              end
+              property :menu_options do
+                key :type, :object
+                key :title, 'MenuOptions'
+                key :description, 'Hash of select menu options for various views'
+                property :fonts do
+                  key :type, :array
+                  items do
+                    key :type, :string
+                  end
+                end
+                property :species do
+                  key :type, :array
+                  items do
+                    key :type, :object
+                    key :title, 'Species information'
+                    property :id do
+                      key :type, :string
+                    end
+                    property :common_name do
+                      key :type, :string
+                    end
+                  end
+                end
+                property :units do
+                  key :type, :array
+                  items do
+                    key :type, :string
+                    key :enum, ExpressionFileInfo::UNITS_VALUES
+                  end
+                end
+                property :library_preparation_protocol do
+                  key :type, :array
+                  items do
+                    key :type, :string
+                    key :enum, ExpressionFileInfo::LIBRARY_PREPARATION_VALUES
+                  end
+                end
+                property :modality do
+                  key :type, :array
+                  items do
+                    key :type, :string
+                    key :enum, ExpressionFileInfo::MODALITY_VALUES
+                  end
+                end
+                property :biosample_input_type do
+                  key :type, :array
+                  items do
+                    key :type, :string
+                    key :enum, ExpressionFileInfo::BIOSAMPLE_INPUT_TYPE_VALUES
+                  end
+                end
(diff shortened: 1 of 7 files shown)


devonbush:
Good stuff, I like the API updates.  Should we also update `application.html.erb` to use the `@study` if it's present.  e.g. line 234 would become
`FeatureFlaggable.feature_flags_for_instances(@selected_branding_group, current_user, @study)`

I believe this method is already nil-safed, and so would still work for pages that `@study` is not defined.

devonbush (approved):

app/controllers/api/v1/studies_controller.rb
```diff
@@ -85,10 +85,131 @@ def show
...
+                  items do
+                    key :type, :object
+                    key :title, 'GenomeAssembly'
+                    property :id do
+                      key :type, :string
+                    end
+                    property :name do
+                      key :type, :string
+                    end
+                    property :taxon_id do
```
eweitz, line 192: Is this a database ID, or an NCBI Taxonomy identifier like "9606"?

eweitz (approved):
Code looks good!  I suggest a few trivial non-blocking refinements.

app/controllers/api/v1/studies_controller.rb
bistline: Good question - this is a MongoDB BSON ID for the `taxon` class.

app/controllers/api/v1/studies_controller.rb
```diff
@@ -85,10 +85,132 @@ def show
...
+          key :tags, [
+            'Studies'
+          ]
+          key :summary, 'Get Study info'
+          key :description, 'Gets attribute, file, feature flag, and configuration information about a single Study'
+          key :operationId, 'study_file_info_path'
+          parameter do
+            key :name, :id
+            key :in, :path
+            key :description, 'ID of Study to fetch'
```
dmohs, line 100: `/studies/{id}/...` implies that this value is a study ID. Not sure if description is required, but if not, is this too much noise?

app/controllers/api/v1/studies_controller.rb
```diff
@@ -85,10 +85,132 @@ def show
...
+                    key :type, :object
+                    key :title, 'GenomeAssembly'
+                    property :id do
+                      key :type, :string
+                    end
+                    property :name do
+                      key :type, :string
+                    end
+                    # show ID of associated parent Taxon for storing belongs_to association
+                    property :taxon_id do
```
dmohs, line 193: Rather than a PR comment to answer Eric's question, could it be helpful to add a comment or rename the variable to add the type clarity there, e.g.
```suggestion
                    property :taxon_bson_id do
```

app/controllers/api/v1/studies_controller.rb
```diff
@@ -85,10 +85,132 @@ def show
...
+                key :type, :array
+                key :title, 'Array<StudyFile>'
+                items do
+                  key :'$ref', :StudyFile
+                end
+              end
+              property :feature_flags do
+                key :type, :object
+                key :title, 'FeatureFlagOptions'
+                key :description, 'Hash of FeatureFlagOption values with defaults, e.g. flag_name => flag_value'
```
dmohs, line 121: Is this right?
```suggestion
                key :description, 'Hash of FeatureFlagOption values applicable to the study with defaults, e.g. flag_name => flag_value'
```

dmohs (approved):
Sorry for the fly-by code review. I will emphasize that you are free to ignore any and all of my suggestions.

app/controllers/api/v1/studies_controller.rb
bistline: Technically, it will support either a Mongo BSON ID, or an accession.  See [here](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/controllers/api/v1/concerns/study_aware.rb#L8) for how studies are loaded in the API.

app/controllers/api/v1/studies_controller.rb
bistline: Good thought - though this attribute is generated automatically via the `has_many / belongs_to` relationship between the `Taxon` and `GenomeAssembly` classes.  We could overwrite it and pass in an attribute name to use for the accession, but that seems a little heavy-handed in this case.

PR #1347, opened 2022-02-01
Allow R format dense matrix (SCP-3971)

ehanna4:
Update the logic to parse and validate a dense matrix to allow uploading of an r-formatted file as defined in the ingest pipeline as:
An "R formatted" file can:
Not have GENE in the header and then:
- Have one less entry in the header than each successive row or
- Have "" as the last value in header.

This PR updates the dense matrix parsing to use a different delimiter determiner since `sniffdelimiter` is dependent on matching columns in rows and an r-formatted file might not match that.
Also adds the validation seen in ingestPipeline to allow an r-formatted file.
Also had to update some error messaging to be a bit more generic to suit the expanded allowance on dense matrices.

to test:
- pull this branch
- attempt to upload a variety of dense matrices

For example:
These would be valid file contents:

(generic non-r-formatted dense file)
`GENE	CELL_0001t	CELL_0002t`
`Itm2a	0	77`
`Isstm2a	0	77`

(r-formatted version 1)
`GENE	CELL_0001t`
`Itm2a	0	77`
`Isstm2a	0	77`

(r-formatted version 2)
`CELL_0001t	CELL_0002t`
`Itm2a	0	77`
`Isstm2a	0	77`

These would not be valid:
`CELL_0001t	CELL_0002t`
`Itm2a	0`
`Isstm2a	0	77`

`CELL_0001t	CELL_0002t	CELL_0003t`
`Itm2a	0	656`
`Isstm2a	0	77`

Updated error message:
Screen Shot 2022-02-01 at 4 30 57 PM


app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -138,7 +253,7 @@ function validateColumnNumber(line, isLastLine, headers, lineNum, dataObj) {
 
     const msg = `All rows must have the same number of columns. ` +
       `Please ensure the number of columns for ${rowText}: ${notedBadRows}, ` +
-      `matches the file-header-specified number of ${correctNumberOfColumns} columns-per-row.`
+      `${containText} the same number of columns-per-row.`
```
eweitz, line 256: Compound modifiers are hyphenated if they come before the noun they modify (e.g. "columns-per-row number"), but not after (like here).

```suggestion
      `${containText} the same number of columns per row.`
```

eweitz (changes requested):
Good stuff!  The `resetToFileStart` method is astute.

I found a trivial syntactic bug and noted it as a blocker.   Beyond that, I suggest a bunch of trivial non-blocking maintainability refinements.

More broadly, I'm also concerned the error message isn't clear enough for users.  I can imagine users will see a message like:

> Improperly formatted header row beginning with: "CELL_0001t"

And struggle to understand how it's improperly formatted.  `validateDenseHeader` has a few different conditions that cause it to add a validation error.  Rather than setting one generic message if any of those conditions fail as is done here, I think setting a specific message depending on _which_ condition failed would really help users fix issues in dense matrix headers.

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -71,17 +79,122 @@ export async function parseFeaturesFile(chunker, mimeType, fileOptions) {
...
+    } else {
+      properlyFormatted = false
+    }
+  } else {
+    properlyFormatted = true
+  }
+  if (!properlyFormatted) {
+    issues.push(['error', 'format:cap:missing-gene-column',
+    `Improperly formatted header row beginning with: '${header[0]}'`])
+  }
```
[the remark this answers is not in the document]
ehanna4: nice, applied this suggestion thanks

bistline (approved):
I think this is good, modulo Eric's comments.  The validation does perform as intended, though I agree the particular error message for this is somewhat unclear (though I don't have a good suggestion!)

app/javascript/lib/validation/chunked-line-reader.js
```diff
@@ -22,6 +22,18 @@ export default class ChunkedLineReader {
     this.chunkLines = []
     this.updateHasMoreChunks()
     this.updateHasMoreLines()
+    this.resetToFileStart()
+  }
+
+  /**
+   * Reset the reader to the start of the file
+   */
```
devonbush, line 30: great api addition

devonbush (approved):
Good stuff, agree with all of Eric's comments.

eweitz (approved):
Great refinements!

PR #1348, opened 2022-02-02
Link to sync survey (SCP-4015)

eweitz:
This adds a link to a Google Forms [survey about sync](https://forms.gle/V6uT94jbGq5pTeMG6).  Here's how it looks:

https://user-images.githubusercontent.com/1334561/152177135-e50d0041-7f88-4b61-a94c-298ec8bf0c53.mov

Two notes to clarify things that might stand out:
* Omitting `noreferrer` has been secure for almost 5 years; see [details](https://github.com/broadinstitute/single_cell_portal_core/pull/1328#discussion_r790874788)
* Given this is intended as brief, unique link, styling inline seems most maintainable

To test:
* Go to sync page
* Confirm link appears at top of page
* Click link
* Confirm it goes to survey

This satisfies SCP-4015.

### DIFF
--- a/app/views/studies/sync_study.html.erb
+++ b/app/views/studies/sync_study.html.erb
@@ -13,6 +13,20 @@
     </p>
   </div>
 </div>
+<a
+  href="https://forms.gle/V6uT94jbGq5pTeMG6"
+  class="btn btn-primary"
+  rel="noopener"
+  data-analytics-name="sync-survey-link"
+  data-toggle="tooltip"
+  data-original-title="Go to survey"
+  data-placement="right"
+  style="margin: 5px;"
+>
+  Take a 3-question survey to help improve sync
+</a>
+
+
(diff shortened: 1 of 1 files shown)


bistline (approved):

devonbush (changes requested):
You're right that there's nothing inherently wrong with using inline styling for temporary one-offs like this.  But if we can use our standard form architecture, we should.  I believe you can omit the need for inline styling here by wrapping this in a 
```
  <div class="row">
    <div class="col-md-12 bottom-pad">
```
And it will look like this:
image

devonbush:
Also, in the 'bikeshedding' category, IMO the button looks better if terra-secondary-btn is used instead of btn-primary, but that's just my opinion

image

eweitz:
Good suggestions!  Done in ace4c56.

devonbush (approved):

PR #1349, opened 2022-02-02
ensuring client-server sync on cluster colors (SCP-2943)

devonbush:
SCP-2943.  Clicking around, there were sometimes cases where the client and server would get out-of-sync as to the custom colors.  this change ensures the cluster_file_info and custom_colors are always included in the server response, and also adds better nil-safing for edge cases.

TO TEST:
 1. go to a study you have edit permissions
 2. click around a whole bunch customizing colors, cancelling, reset to defaults, set again etc...
 3. check that the plot, legend, and server color state are in sync


bistline (approved):
I am able to click around a bunch, and the UI stays in sync for the most part.  Occasionally I end up in a place where after I click "Reset", the re-rendered plot still has the updated colors, but then on refresh everything is back to normal.  I suspect this has something to do with front-end caching, and is not a major concern since the data is not actually being persisted server-side.

app/javascript/components/visualization/ScatterPlot.js
```diff
@@ -119,7 +119,7 @@ function RawScatterPlot({
       })
       store.addNotification(successNotification(`Colors saved successfully`))
       const newScatterData = Object.assign({}, scatterData, {
-        customColors: response.cluster_file_info.custom_colors[annotation.name]
+        customColors: response.cluster_file_info?.custom_colors[scatterData.annotParams.name] ?? {}
```
ehanna4, line 122: I'm not sure i've seen a double question mark operator before - is this essentially a check that the first part has something and if not assign it to be an empty object?

eweitz (approved):

app/javascript/components/visualization/ScatterPlot.js
devonbush: Yup, I just learned about it last week.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

PR #1358, opened 2022-02-04
adding feature flag to reference images (SCP-4067)

devonbush:
SCP-4067. Adding a feature flag so most users will not see reference images.  this feature will likely not be turned on and refined until further grant funding comes through.

This does fix a bug--previously image rendering was broken in upload wizard due to a missing token.

TO TEST:
1. run `rails db:migrate` in shell
2. Visit upload wizard -- confirm "Reference Images" does not appear in tabs
3. Visit sync tool -- confirm 'Image' is not an available file type
4. run `FeatureFlag.find_by(name: 'reference_image_upload').update!(default_value: true)` in console
5. visit upload wizard, confirm Reference Image capability appears

### DIFF
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -45,18 +45,27 @@ const STEPS = [
   ClusteringStep,
   SpatialStep,
   CoordinateLabelStep,
-  ImageStep,
   SequenceFileStep,
   GeneListStep,
   MiscellaneousStep
 ]
 
 const MAIN_STEPS = STEPS.slice(0, 4)
-const SUPPLEMENTAL_STEPS = STEPS.slice(4, 11)
+const SUPPLEMENTAL_STEPS = STEPS.slice(4, STEPS.length)
 
 
 /** shows the upload wizard */
 export function RawUploadWizard({ studyAccession, name }) {
+  const [serverState, setServerState] = useState(null)
+  const [formState, setFormState] = useState(null)
+
+  const allowReferenceImageUpload = serverState?.feature_flags?.reference_image_upload
+
+  if (allowReferenceImageUpload && !STEPS.includes(ImageStep)) {
+    STEPS.splice(5, 0, ImageStep)
+    SUPPLEMENTAL_STEPS.splice(1, 0, ImageStep)
+  }
+
   const routerLocation = useLocation()
   const queryParams = queryString.parse(routerLocation.search)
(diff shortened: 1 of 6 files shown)


app/javascript/components/upload/UploadWizard.js
```diff
@@ -45,18 +45,27 @@ const STEPS = [
...
   CoordinateLabelStep,
-  ImageStep,
   SequenceFileStep,
   GeneListStep,
   MiscellaneousStep
 ]
 
 const MAIN_STEPS = STEPS.slice(0, 4)
-const SUPPLEMENTAL_STEPS = STEPS.slice(4, 11)
+const SUPPLEMENTAL_STEPS = STEPS.slice(4, STEPS.length)
```
eweitz, line 54: Simpler:

```suggestion
const SUPPLEMENTAL_STEPS = STEPS.slice(4)
```

app/views/studies/_study_file_form.html.erb
```diff
@@ -1,3 +1,7 @@
+<%
+  allow_images = FeatureFlaggable.feature_flags_for_instances(@selected_branding_group, current_user, @study)['reference_image_upload']
+  file_type_hash = allow_images ? StudyFile::STUDY_FILE_TYPE_NAME_HASH :  StudyFile::STUDY_FILE_TYPE_NAME_HASH.except('Image')
```
eweitz, line 3:
```suggestion
  file_type_hash = allow_images ? StudyFile::STUDY_FILE_TYPE_NAME_HASH : StudyFile::STUDY_FILE_TYPE_NAME_HASH.except('Image')
```

app/views/studies/initialize_study.html.erb
```diff
@@ -417,6 +417,7 @@
   <div id="upload-wizard-target">
   </div>
   <script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+    window.SCP.readOnlyToken = '<%= get_read_access_token(@study, current_user) %>'
```
eweitz, line 420: Could you add a code comment on why `readOnlyToken` is needed in the upload wizard?

With this PR context in mind, I surmise it's needed for reference images (they're read directly from the bucket, right?), but I definitely won't recall that a few days from now.  Also, it's not totally clear to me now why this would be needed when we're _writing_ the image files, as my understanding is `readOnlyToken` is only needed to read them.

eweitz (approved):
Code looks good!  Below, I suggest some tweaks and ask for clarification.

PR #1360, opened 2022-02-05
Allow trailing whitespace on sparse matrix (SCP-4063)

ehanna4:
Scanpy and ingest pipeline don't care if a sparse matrix has trailing whitespace, they just ignore it so this updates the CSFV to also ignore that.

To test:
- pull this branch
- attempt to upload a sparse matrix that has trailing white spaces on any of the rows and there should be no error

Also updated a test file that had the incorrect number of lines listed.

### DIFF
--- a/app/javascript/lib/validation/expression-matrices-validation.js
+++ b/app/javascript/lib/validation/expression-matrices-validation.js
@@ -356,7 +356,7 @@ function validateValuesAreNumeric(line, isLastLine, lineNum, dataObj) {
 }
 
 /**
-   * Validate the first line in the sparse Matrix begins with '%%MatrixMarket'
+   * Validate the first line in the sparse matrix begins with '%%MatrixMarket'
    */
 function validateMTXHeaderLine(line) {
   const issues = []
--- a/app/javascript/lib/validation/expression-matrices-validation.js
+++ b/app/javascript/lib/validation/expression-matrices-validation.js
@@ -42,12 +42,15 @@ export async function parseSparseMatrixFile(chunker, mimeType, fileOptions) {
   await chunker.iterateLines(rawLine => {
     rawHeaderLine = rawLine
   }, 1)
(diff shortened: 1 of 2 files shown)


app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -353,13 +356,12 @@ function validateValuesAreNumeric(line, isLastLine, lineNum, dataObj) {
 }
 
 /**
-   * Validate the first line in the sparse Matrix begins with the string '%%MatrixMarket'
+   * Validate the first line in the sparse Matrix begins with '%%MatrixMarket'
```
eweitz, line 359:
```suggestion
   * Validate the first line in the sparse matrix begins with '%%MatrixMarket'
```

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -353,13 +356,12 @@ function validateValuesAreNumeric(line, isLastLine, lineNum, dataObj) {
...
    */
 function validateMTXHeaderLine(line) {
   const issues = []
-  const mtxHeader = line.slice(0, 14)
-  if (mtxHeader !== '%%MatrixMarket') {
-    const msg = `First line must begin with "%%MatrixMarket", not "${mtxHeader}"`
+  if (line[0] !== '%%MatrixMarket') {
+    const msg = `First line must begin with "%%MatrixMarket", not "${line}"`
```
eweitz, line 364: Blocker: we'll want to print the first element in `line`, not the whole array.

```suggestion
    const msg = `First line must begin with "%%MatrixMarket", not "${line[0]}"`
```

eweitz (changes requested):
Nice refinement!  I found a trivial blocker.

bistline (approved):

devonbush (approved):
If you can do it in <15 minutes, please add a js unit test for this case.  Otherwise, looks good.

eweitz (approved):

PR #1361, opened 2022-02-07
Unit test for (SCP-4063)

ehanna4:
I wrote the unit test in response to the PR feedback from Devon and forgot to push it up before merging the branch so here it is.

### DIFF
--- a/test/js/lib/validate-file-content.test.js
+++ b/test/js/lib/validate-file-content.test.js
@@ -161,6 +161,15 @@ describe('Client-side file validation', () => {
     expect(errors[0][1]).toEqual('format:empty-row')
   })
 
+  it('reports no error with sparse matrix file that has rows with trailing whitespace', async () => {
+    const file = createMockFile({
+      fileName: 'fo06.mtx',
+      content: '%%MatrixMarket matrix coordinate integer general\n4 8 9  \n4 8 9   \n4 1 0'
+    })
+    const { errors } = await validateFileContent(file, 'MM Coordinate Matrix')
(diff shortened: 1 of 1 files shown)


test/js/lib/validate-file-content.test.js
```diff
@@ -161,6 +161,15 @@ describe('Client-side file validation', () => {
     expect(errors[0][1]).toEqual('format:empty-row')
   })
 
+  it('reports no error with sparse matrix file that has rows with trailing whitespace', async () => {
+    const file = createMockFile({
+      fileName: 'fo06.mtx',
+      content: '%%MatrixMarket matrix coordinate integer general\n4 8 9  \n4 8 9   \n4 1 0'
```
eweitz, line 167: All [our example test full .mtx files](https://github.com/broadinstitute/single_cell_portal_core/search?q=%25%25MatrixMarket) have a two-line header block.  While not strictly required*, ideally our test data would usually be consistent in that regard.

\* The [NIST brief](https://math.nist.gov/MatrixMarket/formats.html) isn't explicit on the need for a second comment line, but the full NIST paper clearly notes it is optional.  It says:

> The second line is a comment. (Zero or more such lines may appear at this point in the file, each beginning with a %.)

eweitz (approved):
Looks good!  I note an esoteric detail maybe worth considering, nothing blocking.

bistline (approved):

ehanna4:
this PR is only a unit test, no functional code change and the failing test in the integration run is not the added test

PR #1362, opened 2022-02-07
Asynchronously loading the analysis tab (SCP-4071)

devonbush:
SCP-4071 Major thanks to Jon for helping me test this.  This adds 'canCompute' and 'analysisTabContent' to the user_study_info endpoint (which is also used to determine if the user can edit the study and load user annotations).  This endpoint is only fetched after the rest of the study loads. The analysis tab content is rendered using the existing template logic, so there is no change to the analysis UX.

I originally tried an alternate approach putting analysis-tab-check logic into a brand new endpoint, which was checked from a script inside `_study_tabs_nav.html.erb`.  This was nice in keeping the code isolated, but tricky from a timing perspective, as there was no easy way to ensure the async call would fire after the rest of the study async calls.  Also, this made the mental model of the page more complex, as now there were 4 different async calls being made, from two different places.  So I consolidated it inside the study_user_info endpoint with the advantage of code and page simplicity, but the drawback that user annotations and custom controls have to wait on the analysis tab.

TO TEST:
1. sign in as an admin
2. load a study
3. confirm that the other 4 tabs load first, and then the  analysis tab loads later (it may be ~15 seconds later in development)
4. confirm that user annotations and the "customize colors" still appear correctly within the explore tab
5. (If you have analyses configured in your environment).  Use the analysis tab to launch a workflow

### DIFF
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -140,9 +140,9 @@ module Api
           render json: {
             annotations: AnnotationVizService.available_annotations(@study, cluster: nil, current_user: current_api_user),
             canEdit: @study.can_edit?(current_api_user),
-            allow_firecloud_access: allow_firecloud_access,
-            can_compute: can_compute,
-            analysis_tab_content: analysis_tab_content
+            allowFirecloudAccess: allow_firecloud_access,
+            canCompute: can_compute,
+            analysisTabContent: analysis_tab_content
           }
         end
 
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -117,9 +117,32 @@ module Api
         end
 
         def study_user_info
+          allow_firecloud_access = AdminConfiguration.firecloud_access_enabled?
+          can_compute = false
+          analysis_tab_content = nil
+          if allow_firecloud_access
+            can_compute = TerraAnalysisService.user_can_compute?(@study, current_api_user)
+            if can_compute
+              submissions = TerraAnalysisService.list_submissions(@study)
+              workflows_list = AnalysisConfiguration.available_analyses
+              # The analysis tab code wasn't worth porting over to React/API due to how rarely it is used, so we just
+              # render the template as a string and pass it via API to get put on the page
+              # See https://marcqualie.com/2020/07/rendering-templates-in-rails-api-controllers for technical detail
+              analysis_tab_content = ActionController::Renderer.for(ApplicationController)
+                .render(partial: '/site/study_analysis', locals: {
+                  '@submissions': submissions,
+                  '@workflows_list': workflows_list,
+                  '@study': @study
+                })
+            end
+          end
+
           render json: {
             annotations: AnnotationVizService.available_annotations(@study, cluster: nil, current_user: current_api_user),
-            canEdit: @study.can_edit?(current_api_user)
+            canEdit: @study.can_edit?(current_api_user),
+            allow_firecloud_access: allow_firecloud_access,
+            can_compute: can_compute,
+            analysis_tab_content: analysis_tab_content
           }
         end
 
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -159,24 +159,6 @@ class SiteController < ApplicationController
     set_study_permissions(@study.detached?)
     set_study_default_options
     set_study_download_options
-
-    if @allow_firecloud_access && @user_can_compute
-      # load list of previous submissions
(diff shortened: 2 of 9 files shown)


app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -117,9 +117,32 @@ def bam_file_info
...
+                  '@study': @study
+                })
+            end
+          end
+
           render json: {
             annotations: AnnotationVizService.available_annotations(@study, cluster: nil, current_user: current_api_user),
-            canEdit: @study.can_edit?(current_api_user)
+            canEdit: @study.can_edit?(current_api_user),
+            allowFirecloudAccess: allow_firecloud_access,
```
eweitz, line 143: For anyone curious, Terra JS [also](https://github.com/search?l=JavaScript&q=org%3ADataBiosphere+firecloud&type=Code) writes "do FireCloud thing" as "doFirecloudThing" and not "doFireCloudThing".  Good intuition here!

app/controllers/api/v1/visualization/explore_controller.rb
```diff
@@ -117,9 +117,32 @@ def bam_file_info
...
+          can_compute = false
+          analysis_tab_content = nil
+          if allow_firecloud_access
+            can_compute = TerraAnalysisService.user_can_compute?(@study, current_api_user)
+            if can_compute
+              submissions = TerraAnalysisService.list_submissions(@study)
+              workflows_list = AnalysisConfiguration.available_analyses
+              # The analysis tab code wasn't worth porting over to React/API due to how rarely it is used, so we just
+              # render the template as a string and pass it via API to get put on the page
+              # See https://marcqualie.com/2020/07/rendering-templates-in-rails-api-controllers for technical detail
```
eweitz, line 130: Helpful!

eweitz (approved):
Code looks good!

If it's trivial, could you post some numbers on your local performance before and after this optimization?

app/lib/terra_analysis_service.rb
```diff
@@ -0,0 +1,52 @@
...
+  def self.list_submissions(study)
+    # load list of previous submissions
+    workspace = ApplicationController.firecloud_client.get_workspace(study.firecloud_project, study.firecloud_workspace)
+    submissions = ApplicationController.firecloud_client.get_workspace_submissions(study.firecloud_project, study.firecloud_workspace)
+
+    submissions.each do |submission|
+      update_analysis_submission(submission)
+    end
+    # remove deleted submissions from list of runs
+    if !workspace['workspace']['attributes']['deleted_submissions'].blank?
```
bistline, line 14: Non-blocking: I think re-writing these to use `dig` would be more robust:
```suggestion
    if !workspace.dig('workspace', 'attributes','deleted_submissions').blank?
      deleted_submissions = workspace.dig('workspace', 'attributes', 'deleted_submissions', 'items')
```

app/controllers/site_controller.rb
```diff
@@ -700,7 +671,6 @@ def set_firecloud_permissions(study_detached)
         buckets_ok = system_status.dig(FireCloudClient::BUCKETS_SERVICE, 'ok') == true
         @allow_downloads = buckets_ok
         @allow_edits = sam_ok && rawls_ok
-        @allow_computes = sam_ok && rawls_ok && agora_ok
```
bistline, line 703: While the risk is low given how few workflows have been launched from SCP, not having the status check here makes me somewhat uncomfortable.  I would be in favor of moving these checks into `TerraAnalysisService` and having a method in there for checking system status as a pre-flight check to rendering the analysis tab:

```
# check if computes are enabled & submissions can be accepted by orchestration
def self.computes_allowed?
  return false unless AdminConfiguration.firecloud_access_enabled?
  
  begin
    api_status = ApplicationController.firecloud_client.api_status
    sam_ok = system_status.dig(FireCloudClient::SAM_SERVICE, 'ok') == true
    agora_ok = system_status.dig(FireCloudClient::AGORA_SERVICE, 'ok') == true
    rawls_ok = system_status.dig(FireCloudClient::RAWLS_SERVICE, 'ok') == true
    sam_ok && rawls_ok && agora_ok
  rescue RuntimeError => error
    ErrorTracker.report_exception(error, nil)
    Rails.logger.error "Error checking compute status: #{e.message}"
    false  
  end
end
```

bistline (approved):
Some non-blocking suggestions for refinements.

app/lib/terra_analysis_service.rb
devonbush: concur, about the robustness, but I think I will file this under 'if it ain't broke, don't fix it'

app/controllers/site_controller.rb
devonbush: agree, updated the compute method to validate whether the services are up

devonbush:
Test failure is false positive -- is in `test/models/reviewer_access_test.rb:26`-- a date check error from the test running too close to midnight

PR #1364, opened 2022-02-07
removing 'replace' button for already-uploaded files (SCP-4081)

devonbush:
SCP-4081  This removes the 'Replace' option for files that have already been uploaded.  The feature was desirable, but was leading to bugs like https://broadinstitute.zendesk.com/agent/tickets/269187. 

I looked into updating the backend to treat 'replace' as 'delete and recreate', but the error handling edge cases were too many, and communicating to the UX that "there's now a file that should be treated the same as the old file, but that has a new id", would be problematic.

TO TEST:
1. open upload wizard for a study
2. for existing files, observe there is no 'Choose file' or 'Replace' button
3. click "add file"
4. observe there is a 'choose file' button
5. choose a file
6. observe a 'Replace' button replaces the 'Choose file' button

### DIFF
--- a/app/javascript/components/upload/FileUploadControl.js
+++ b/app/javascript/components/upload/FileUploadControl.js
@@ -93,9 +93,12 @@ export default function FileUploadControl({
       })
     }
   }
-  let buttonText = file.upload_file_name ? 'Replace' : 'Choose file'
+  const isFileChosen = !!file.upload_file_name
+  const isFileOnServer = file.status !== 'new'
+
+  let buttonText = isFileChosen ? 'Replace' : 'Choose file'
   let buttonClass = 'fileinput-button btn terra-tertiary-btn'
-  if (!file.upload_file_name && !file.uploadSelection) {
+  if (!isFileChosen && !file.uploadSelection) {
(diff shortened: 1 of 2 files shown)


bistline (approved):

app/javascript/components/upload/FileUploadControl.js
```diff
@@ -93,9 +93,12 @@ export default function FileUploadControl({
       })
     }
   }
-  let buttonText = file.upload_file_name ? 'Replace' : 'Choose file'
+  const isFileChosen = !!file.upload_file_name
```
ehanna4, line 96: I appreciate these const names as the data points are not as intuitive

ehanna4 (approved):
Tested as described and looks as described

One concern that isn't related to the "replace" button but is in the same vicinity so feels worth calling out here is the "save" button is clickable even when there are no changes to the form once a file has been uploaded. I am wondering if we ought to disable that to avoid confusion? It has different styling if you make a change but it seems confusing to allow saving if you haven't done anything.


Screen Shot 2022-02-08 at 10 36 22 AM

PR #1324, opened 2022-01-21
Refactor `window.SCP.readOnlyToken` into view for streaming files (SCP-4020)

bistline:
This update narrows the usage of `window.SCP.readOnlyToken` in an effort to mitigate the impact of the planned Terra terms of service updates.  Since all registered Terra users will need to re-accept the terms of service on or about 2022.02.15, any API requests made with user credentials who are not in compliance will be rejected.  Since generating pet service account tokens (the kind of token used in `window.SCP.readOnlyToken`) will fall into this category of API requests, this update only generates those tokens when explicitly needed.  This is in an effort to prevent the "Session timed out" modal from showing on every page load, as this would happen due to the `401/403` response from the Terra API.  The scenarios in which `window.SCP.readOnlyToken` is required are limited to visualization requests in private studies (public studies are unaffected), and include:

1. Viewing BAM files in IGV
2. Viewing inferCNV Ideogram annotations
3. Viewing Image files alongside scatter plots

**NOTE**: This update does not address concerns raised in SCP-3258 regarding the refactoring of genome visualization-specific code into the more accepted paradigm of JSON API + React.  This is purely a mitigation step to reduce incidence of spurious "Session timed out" modals on page loads for registered Terra users.

MANUAL TESTING
1. Boot as normal and sign in
2. In the Chrome Dev Tools JS console, validate that `window.SCP.readOnlyToken` is `undefined`
3. Load any study with BAM files, or upload `test/test_data/fixed_neurons_6days_2000_possorted_genome_bam_test_data.bam` and the associated `bai` file to an existing study (select `mouse` for the species and `GRCm38` for the assembly)
4. Open the Explore tab for the study, then click on the Genome tab, and confirm that IGV initializes and the BAM file loads without error
5. Back in the Chrome Dev Tools JS console, validate that `window.SCP.readOnlyToken` is now has a value starting with `ya29` 
6. If using the above BAM file, enter `chr1:3,043,159-3,043,318` into the IGV search bar, and validate the reads show as below:
Screen Shot 2022-01-21 at 10 57 49 AM
7. Go back to the home page, and validate that `window.SCP.readOnlyToken` is now back to `undefined`

This PR satisfies SCP-4020.

PR #1365, opened 2022-02-09
Restoring file downloads in upload wizard (SCP-4079)

bistline:
This fixes a regression where the file download links in the upload wizard stopped working due to the refactoring of `window.SCP.readOnlyToken` in #1324.  In addition, rather than streaming the blob of bytes through the app, this update restores the original functionality of generating a signed URL and opening a new tab directly on the resource in the bucket to let the client handle the download, rather that our server.  This also preserves the original filename, instead of the UUID that had been assigned before.  A newly created `StudyContext` React context can now be accessed anywhere in the upload wizard, which will have the entire study object loaded from the server.

MANUAL TESTING
1. Boot as normal and sign in
2. Load the upload wizard for any study that already has files
3. Click the file download links, and validate that the file opens in a new window (txt files will render in the browser, other types will download directly)

This PR satisfies SCP-4079.

### DIFF
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -67,10 +67,9 @@ export function RawUploadWizard({ studyAccession, name }) {
   const [formState, setFormState] = useState(null)
 
   // study attributes to pass to the StudyContext later for use throughout the RawUploadWizard component, if needed
-  const studyObj = {
-    studyAccession: serverState?.study?.accession,
-    studyName: serverState?.study?.url_safe_name
-  }
+  // use complete study object, rather than defaultStudyState so that any updates to study.rb will be reflected in
+  // this context
+  const studyObj = serverState?.study
 
   const allowReferenceImageUpload = serverState?.feature_flags?.reference_image_upload
 
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -15,7 +15,7 @@ import * as queryString from 'query-string'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faChevronLeft, faChevronRight } from '@fortawesome/free-solid-svg-icons'
 
-import { formatFileFromServer, formatFileForApi, newStudyFileObj } from './upload-utils'
+import { formatFileFromServer, formatFileForApi, newStudyFileObj, StudyContext } from './upload-utils'
 import {
   createStudyFile, updateStudyFile, deleteStudyFile,
   fetchStudyFileInfo, sendStudyFileChunk, RequestCanceller
@@ -53,14 +53,6 @@ const STEPS = [
 
 const MAIN_STEPS = STEPS.slice(0, 4)
 const SUPPLEMENTAL_STEPS = STEPS.slice(4)
-
-const defaultStudyState = {
-  accession: null,
-  url_safe_name: null
-}
-
-// context to pass through UploadWizard
-export const StudyContext = React.createContext(defaultStudyState)
 
 /** shows the upload wizard */
 export function RawUploadWizard({ studyAccession, name }) {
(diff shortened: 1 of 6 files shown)


app/javascript/components/upload/UploadWizard.js
```diff
@@ -54,12 +54,24 @@ const STEPS = [
 const MAIN_STEPS = STEPS.slice(0, 4)
 const SUPPLEMENTAL_STEPS = STEPS.slice(4)
 
+const defaultStudyState = {
+  accession: null,
+  url_safe_name: null
+}
+
+// context to pass through UploadWizard
+export const StudyContext = React.createContext(defaultStudyState)
```
devonbush, line 63: one thing to note here is that we now have a circular import chain from UploadWizard through all the intermediate components to FileDownloadControl, and back to UploadWizard.  This is generally fine, but can lead to weird instantiation timing issues.  So despite my original recommendation to do it all inline, let's move StudyContext and defaultStudyState to `upload-utils.js`

app/javascript/components/upload/UploadWizard.js
```diff
@@ -54,12 +54,24 @@ const STEPS = [
...
 
 /** shows the upload wizard */
 export function RawUploadWizard({ studyAccession, name }) {
   const [serverState, setServerState] = useState(null)
   const [formState, setFormState] = useState(null)
 
+  // study attributes to pass to the StudyContext later for use throughout the RawUploadWizard component, if needed
+  // use complete study object, rather than defaultStudyState so that any updates to study.rb will be reflected in
+  // this context
+  const studyObj = serverState?.study
```
devonbush, line 73: Yeah, originally I thought we'd want to limit the properties we pass to the context to just name and accession, but you're right that using the full study object as the context will cause a re-render of all the components that use the study context,  when attributes change.  And so if we ever want to have, e.g. a 'usage' tab in the upload wizard with a real-time updating view count, this will let us do it

app/javascript/components/upload/UploadWizard.js
devonbush: I think it's clearer to have the default study state just be `null` -- that way it's cleaner for components that rely on this to check 'is the study loaded?'  with `if (studyState)`. (and I see you already wisely nil-safed the property accesses in the download component)

config/environments/development.rb
```diff
@@ -63,7 +63,7 @@
     if env['action_dispatch.original_path']&.include?('api/v1')
       Api::V1::ExceptionsController.action(:render_error).call(env)
     else
-      ActionDispatch::PublicExceptions.new(Rails.public_path)
+      ActionDispatch::PublicExceptions.new(Rails.public_path).call(env)
```
devonbush: is this just a cleanup bugfix?

devonbush (changes requested):
Looks great, glad you had the good instinct not to update 30-odd files to pass the param all the way down.  Couple of changes on where the import should be and the default value, then this will be good to go.

app/javascript/components/upload/UploadWizard.js
devonbush: On thinking about this more, this setup will cause the components that rely on the studyContext to get rerendered every time we poll the server for study state (since the serverState object itself will be a different instance, even though it will have all the same property values.  So maybe not optimal, but totally fine for now given that we poll the server only every 10s, and not that many components rely on this context

config/environments/development.rb
bistline: Actually I backed this out because it doesn't do what I wanted it to.  There's weirdness with some errors (like the `ActionDispatch::RoutingError` we say in Sentry), but most other errors are still rendered with the nice development stack trace page.

app/javascript/components/upload/UploadWizard.js
bistline: [30f5269](https://github.com/broadinstitute/single_cell_portal_core/pull/1365/commits/30f526987b7cb411fcd90ab0c9635d3d13ed6c60)

devonbush (approved):

devonbush:
I think it's outside the scope of this ticket, and not worth a ticket on it's own, but we should take a mental note to start clearing out `bucketName` from the ~20 files that pass it as a param, and use the context instead

eweitz (approved):
Code looks good!  I suggest a trivial non-blocking readability refinement.

PR #1366, opened 2022-02-09
Add timeout on CSFV (SCP-4035)

ehanna4:
Adds a timeout to client side validation of 10 seconds. So if a user attempts to upload a super big file the CSFV will timeout after 10 seconds and print out a warning to the user letting them know that the validation will have to occur after file upload. Struggled to find good data on what the timeout time limit ought to be, went with 10 seconds as that felt acceptable anecdotally but again couldn't find any standards on this so input is welcome.

This adds a status type of 'timeout' for the mixpanel event as this isn't really a true "failure" or "success" and is presented as a "warning" to users in the UI.

also updates the iterateLines() function to take its params as an object for cleaner handling.
Screen Shot 2022-02-09 at 3 36 22 PM

To test:
- pull this branch
- attempt to upload a super big file, such as the `MIT_Milk_Study_Raw_counts.txt` file which you can obtain from the study SCP1671 in prod
- observe that after 10 seconds the file stops trying to validate and a warning appears and you can click save

### DIFF
--- a/app/javascript/lib/validation/chunked-line-reader.js
+++ b/app/javascript/lib/validation/chunked-line-reader.js
@@ -20,6 +20,7 @@ export default class ChunkedLineReader {
     this.linesRead = 0
     this.currentFragment = null // currentFragment stores the parts of lines that cross chunk boundaries
     this.chunkLines = []
+    this.startTime = new Date().getTime() // start the CSFV timer
     this.updateHasMoreChunks()
     this.updateHasMoreLines()
   }
@@ -41,7 +42,7 @@ export default class ChunkedLineReader {
    * maxBytesPerLine can be set to avoid reading the entire file into memory in the event the file is missing
    * proper newlines
   */
-  async iterateLines({ func, maxLines = Number.MAX_SAFE_INTEGER, maxBytesPerLine = 1000*1000*1024, startTime }) {
+  async iterateLines({ func, maxLines = Number.MAX_SAFE_INTEGER, maxBytesPerLine = 1000*1000*1024 }) {
     const prevLinesRead = this.linesRead
     while ((this.hasMoreChunks || this.chunkLines.length) &&
            !(this.linesRead === 0 && this.nextByteToRead > maxBytesPerLine) &&
--- a/app/javascript/lib/validation/chunked-line-reader.js
+++ b/app/javascript/lib/validation/chunked-line-reader.js
@@ -6,9 +6,9 @@ const newlineRegex = /\r?\n/
  * reads lines from a file in a chunked way, so that the file does not have to be in memory all at once
  * Basic usage:
  * const chunker = new ChunkedLineReader(file)
- * chunker.iterateLines((line, lineNum, isLastLine) => {
+ * chunker.iterateLines({func: (line, lineNum, isLastLine) => {
  *   // do stuff
- * }) {
+ * }}) {
  * }
  */
 export default class ChunkedLineReader {
@@ -22,20 +22,7 @@ export default class ChunkedLineReader {
     this.chunkLines = []
     this.updateHasMoreChunks()
     this.updateHasMoreLines()
-    this.resetToFileStart()
-    this.shouldStopReadingFile()
   }
-
-  /**
-   * Exception thrown for issues arrising during file reading
-   * @param {} key 
-   * @param {*} msg 
-   */
-  FileReadingException(key, msg) {
-    this.message = msg
-    this.key = key
-  }
-
 
   /**
    * Reset the reader to the start of the file
@@ -49,25 +36,12 @@ export default class ChunkedLineReader {
   }
 
   /**
-   * Calculate if should stop reading file lines
-   */
-  shouldStopReadingFile(timerStart) {
-    const max = timerStart + 10000 // timerStart is in milliseoconds so add 10,000 milliseconds
-    const currentTime = new Date().getTime()
-    if (currentTime > max) {
-      throw new this.FileReadingException('time-limit-exceeded',
-      `This file is too large for full client side validation, you can continue and save it and wait for the serverside validation to confirm validation.`)
-    }
-  }
-
-  /**
    * iterates over the remaining lines in the file, calling func(line, lineNum, isLastLine) for each line
    * (lineNum is 0-indexed).
    * maxBytesPerLine can be set to avoid reading the entire file into memory in the event the file is missing
    * proper newlines
   */
-  async iterateLines(func, maxLines=Number.MAX_SAFE_INTEGER, startTime) {
-    const maxBytesPerLine=1000*1000*1024
+  async iterateLines({ func, maxLines = Number.MAX_SAFE_INTEGER, maxBytesPerLine = 1000*1000*1024, startTime }) {
     const prevLinesRead = this.linesRead
     while ((this.hasMoreChunks || this.chunkLines.length) &&
            !(this.linesRead === 0 && this.nextByteToRead > maxBytesPerLine) &&
@@ -81,9 +55,18 @@ export default class ChunkedLineReader {
         this.updateHasMoreLines()
         func(line, this.linesRead - 1, !this.hasMoreLines)
       }
-      this.shouldStopReadingFile(startTime)
     }
   }
+
+  /**
+   * Reset the reader to the end of the file
+   */
+  setToFileEnd() {
+    this.chunkLines = []
+    this.hasMoreChunks = false
+    this.hasMoreLines = false
+  }
+
 
   /**
    * reads a file as lines in a set of 'chunks'  Each chunk is determined by chunkSize
--- a/app/javascript/lib/validation/chunked-line-reader.js
+++ b/app/javascript/lib/validation/chunked-line-reader.js
@@ -23,7 +23,19 @@ export default class ChunkedLineReader {
     this.updateHasMoreChunks()
     this.updateHasMoreLines()
     this.resetToFileStart()
+    this.shouldStopReadingFile()
   }
+
+  /**
+   * Exception thrown for issues arrising during file reading
+   * @param {} key 
+   * @param {*} msg 
+   */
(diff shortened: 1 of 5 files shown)


devonbush (changes requested):
This is quite clever -- adding the setFileToEnd method is a nice way of handling the problem of short-circuiting the line reader.   I also like how you switched iterateLines to named arguments for clarity.  Just one blocker around avoiding passing too many `startTime` parameters

Should we go ahead and add this timeout to all the other file types?  expression matrices are definitely where it's most needed, but it can also help if someone, e.g. accidentally selects the wrong file

app/javascript/lib/validation/shared-validation.js
```diff
@@ -186,3 +188,23 @@ export function validateGroupColumnCounts(headers, line, isLastLine, dataObj) {
...
+  const maxRealTime = startTime + maxTime
+  const currentTime = new Date().getTime()
+  const issues = []
+
+  if (currentTime > maxRealTime) {
+    // quit early by setting the file reader to the end of the file so it can't read anymore
+    chunker.setToFileEnd()
+    issues.push(['warn', 'timeout',
+      'Due to the size of the file, validation will occur after upload. Please be aware '+
+        'the absence of errors/warnings here is NOT an indication of the validity of the file.'])
```
eweitz, line 207: Briefer phrasing, with a worthwhile additional note, and a lower-volume "not".

```suggestion
      'Due this file\'s size, it will be fully validated after upload, and results will be emailed to you.  ' + 
      'The absence of issues here does not mean the file is valid.  ' +
```

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -39,19 +44,24 @@ export async function parseSparseMatrixFile(chunker, mimeType, fileOptions) {
...
+  })
   const header = rawHeaderLine.trim().split(whitespaceDelimiter)
 
   issues = validateMTXHeaderLine(header)
 
-  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
-    const line = rawLine.trim().split(whitespaceDelimiter)
+  await chunker.iterateLines({
+    func: (rawLine, lineNum, isLastLine) => {
+      issues = issues.concat(timeOutCSFV(chunker))
```
eweitz, line 58:
```suggestion
      if (lineNum % 100 === 0) {
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -75,9 +88,12 @@ export async function parseFeaturesFile(chunker, mimeType, fileOptions) {
   let issues = []
 
   const dataObj = {} // object to track multi-line validation concerns
-  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
-    issues = issues.concat(validateUniqueRowValuesWithinFile(rawLine, isLastLine, dataObj))
+  await chunker.iterateLines({
+    func: (rawLine, lineNum, isLastLine) => {
+      issues = issues.concat(timeOutCSFV(chunker))
```
eweitz, line 93:
```suggestion
      if (lineNum % 100 === 0) {
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -263,12 +263,16 @@ export async function parseMetadataFile(chunker, mimeType, fileOptions) {
...
 
   const dataObj = {} // object to track multi-line validation concerns
-  await chunker.iterateLines((rawline, lineNum, isLastLine) => {
-    const line = parseLine(rawline, delimiter)
-    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
-    issues = issues.concat(validateMetadataLabelMatches(headers, line, isLastLine, dataObj))
-    issues = issues.concat(validateGroupColumnCounts(headers, line, isLastLine, dataObj))
+  await chunker.iterateLines({
+    func: (rawline, lineNum, isLastLine) => {
+      issues = issues.concat(timeOutCSFV(chunker))
```
eweitz, line 268:
```suggestion
      if (lineNum % 100 === 0) {
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -281,11 +285,15 @@ export async function parseClusterFile(chunker, mimeType) {
   // add other header validations here
 
   const dataObj = {} // object to track multi-line validation concerns
-  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
-    const line = parseLine(rawLine, delimiter)
-    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
-    issues = issues.concat(validateGroupColumnCounts(headers, line, isLastLine, dataObj))
+  await chunker.iterateLines({
+    func: (rawLine, lineNum, isLastLine) => {
+      issues = issues.concat(timeOutCSFV(chunker))
```
eweitz, line 290:
```suggestion
      if (lineNum % 100) {
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -62,9 +72,12 @@ export async function parseBarcodesFile(chunker, mimeType, fileOptions) {
...
```
eweitz, line 77:
```suggestion
      if (lineNum % 100 === 0) {
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -3,32 +3,37 @@
...
-  await chunker.iterateLines((rawLine, lineNum, isLastLine) => {
-    const line = parseLine(rawLine, delimiter)
-    issues = issues.concat(validateValuesAreNumeric(line, isLastLine, lineNum, dataObj))
-    issues = issues.concat(validateColumnNumber(line, isLastLine, secondLineOfFile, lineNum, dataObj))
-    issues = issues.concat(validateUniqueCellNamesWithinFile(line, isLastLine, dataObj))
+  await chunker.iterateLines({
+    func: (rawLine, lineNum, isLastLine) => {
+    // check that it's still good time
+      issues = issues.concat(timeOutCSFV(chunker))
+
```
eweitz, line 30: Blocker: 100x more efficient, with negligible downside for UX:

```suggestion
  
      if (lineNum % 100) {
        // check that it's still good time
        issues = issues.concat(timeOutCSFV(chunker))
      }
```

app/javascript/lib/validation/shared-validation.js
```diff
@@ -186,3 +188,23 @@ export function validateGroupColumnCounts(headers, line, isLastLine, dataObj) {
...
 }
+
+/**
+ * Timeout the CSFV if taking longer than 10 seconds
+ *
+ */
+export function timeOutCSFV(chunker) {
+  const maxTime = 10000 // in milliseconds this equates to 10 seconds
+  const maxRealTime = chunker.startTime + maxTime
+  const currentTime = new Date().getTime()
```
eweitz, line 199: Blocker: [roughly 37% faster](https://jsbench.me/qqkzg2wy98/1), and functionally equivalent for this use case:

```suggestion
  const currentTime = Date.now()
```

app/javascript/lib/validation/chunked-line-reader.js
```diff
@@ -20,9 +20,9 @@ export default class ChunkedLineReader {
     this.linesRead = 0
     this.currentFragment = null // currentFragment stores the parts of lines that cross chunk boundaries
     this.chunkLines = []
+    this.startTime = new Date().getTime() // start the CSFV timer
```
eweitz, line 23: For consistency with faster timing method:

```suggestion
    this.startTime = Date.now() // start the CSFV timer
```

eweitz (changes requested):
Good stuff!  I think 10 seconds is a reasonable tradeoff on completeness vs. speed.

I note some blocking performance issues:
* Check time 100x less frequently (and that much faster) with negligible UX downside (now non-blocker, see [discussion](https://github.com/broadinstitute/single_cell_portal_core/pull/1366#discussion_r803105950))
* Use `Date.now()`, which is 25-40% [faster](https://jsbench.me/qqkzg2wy98/1) than `new Date().getTime()`

Beyond that, I suggest a non-blocking message refinement that seems more readable and actionable.

app/javascript/lib/validation/expression-matrices-validation.js
devonbush: I like the idea of not having the time check on every line, but this is going to defeat the purpose of the time limit on files with really long lines, or a misformatted file that doesn't have any newlines. Also, the time check operation should be much, much faster than actually parsing a line, so I don't think this will materially impact performace

eweitz: See [reply](https://github.com/broadinstitute/single_cell_portal_core/pull/1366#discussion_r803849935).

app/javascript/lib/validation/validate-file-content.js
devonbush: Yeah, I think you're right that this is an argument to move the time check to inside the chunker, but then that will require some API changes to chunker (chunker would raise a "TimeoutException" or somesuch), which are outside the scope of the PR

Yeah, definitely about tradeoffs, is a x% speedup worth a y% chance that it will end up taking longer on long runs because the validation read too many lines?

ehanna4:
I have updated the code to use the more performant `date.now()` but based on the discussion and the added complexity to find a better solution coupled with the fact that we hope to not keep this timeout code in the long run and that the performance hit is very minimal and only happening on the timeout code I am opting to not make any changes for this and will continue checking the time every line

PR #1368, opened 2022-02-10
Correct content header issues caused by gsutil and sync actions (SCP-3938)

bistline:
When a user pushes a gzipped file directly to a workspace bucket and then employs the "sync" feature, there is a corner case where ingest processes will fail if the user used `gsutil` to push the file and did not override the `Content-Type` header.  Instead, the original file `Content-Type` is stored, which causes the file to fail ingest as the subsequent file download is not correctly detected as gzipped.  While this is most often seen with sparse matrix bundles (mtx, 10x features, and 10x barcodes), it can happen with any parseable, text-based file that is gzipped and pushed to a workspace bucket with `gsutil`.

This update adds the `StudySyncService`, which is a stub library that will eventually house all sync-related methods (as originally outlined in SCP-3121).  For now, this class contains methods that will determine the correct content headers for any synced file, depending on how it was uploaded, and whether or not it is gzipped (only compressed files that were synced will have their `Content-Type` and `Content-Encoding` headers overridden.

Note: there is an unrelated issue where single-column gzipped text files do not decompress properly on MacOS.  In order to decompress these files, please use `gunzip [filename]` in the terminal.

MANUAL TESTING
Before beginning, if you have `gsutil` installed, you will want to run `gcloud auth login` to make sure you are authenticated to use `gsutil`
1. Boot as normal and sign in, making sure to run `rails jobs:work` in another console to start the job worker
2. Select a study from "My Studies" and click "Show Details"
3. Do one of the following:
    -  If you do not have `gsutil` installed, click the link to open the Storage Browser, and upload the file through the browser, then click "Edit Metadata" from the ellipsis menu, and set the `Content-Type` to `text/plain`
    - If you have `gsutil` installed, push the file `test/test_data/expression_matrix_example_gzipped.txt.gz` directly to the workspace bucket (you can get the bucket ID from the `Google Storage Bucket` line) with:
```
gsutil cp test/test_data/expression_matrix_example_gzipped.txt.gz gs://[bucket_id]
```
4. Back in SCP, click "Sync Workspace" for the selected study from the "My Studies" page
5. Sync the expression matrix and confirm that the success dialog shows
6. Back in the GCS browser, refresh the page and click "Edit Metadata" again, and confirm that `Content-Type` is now set to `application/gzip`
7. Wait for the expression matrix to finish parsing and confirm it completes without error

This PR satisfies SCP-3938

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -777,7 +777,7 @@ module Api
               # make sure file is not actually a folder by checking its size
               if file.size > 0
                 # fix content headers
-                fixed_file = StudySyncService.fix_file_content_headers(file)
+                fixed_file = ::StudySyncService.fix_file_content_headers(file)
                 # create a new entry
                 unsynced_file = StudyFile.new(study_id: @study.id, name: fixed_file.name, upload_file_name: fixed_file.name,
                                               upload_content_type: fixed_file.content_type,
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -776,8 +776,13 @@ module Api
             else
               # make sure file is not actually a folder by checking its size
               if file.size > 0
+                # fix content headers
+                fixed_file = StudySyncService.fix_file_content_headers(file)
                 # create a new entry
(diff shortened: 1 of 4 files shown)


devonbush (approved):
Nice!

eweitz (approved):

PR #1369, opened 2022-02-11
Client-side file validation for remote files in sync UI (SCP-4017)

eweitz:
This adds in-app file validation to the sync UI.  It enables users to fix problems in their files drastically faster.

Previously, when importing files from Terra workspaces into SCP, users had to wait at least 2-3 minutes to learn of any validation errors.  And those errors were emailed to the user.  Now, users learn of such sync file validation problems in < 10 seconds and right in the web app.

Sync client-side file validation (CSFV) extends similar functionality recently introduced to the upload UI.  The validations are the same.  Other aspects of the sync UI still use the legacy design, as a larger sync UI revamp is being contemplated.  The main technical difference between upload and sync CSFV is that the latter uses [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) to read the first part of the file.  That's because sync files are remote, existing in a Google Cloud Storage bucket, unlike upload files which exist on a user's local machine.  Relevant new Mixpanel analytics events are also recorded, e.g. whether the full file was fetched and how much time that took.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/153594323-bb45dcf8-cfd4-4e89-94b2-47f9998cc766.mov

To test:
* Add files to a study's Google bucket, including valid and invalid files, both larger and smaller than 50 MiB
* Go to Sync page for study
* Select file type from each file's dropdown menu
* Confirm no issues are reported for small valid files
* Confirm errors are reported for small invalid files
* Confirm only "incomplete" warning is reported for large valid files, e.g. [Dense_matrix_204_MB_good.txt](https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-public/test/Dense_matrix_204_MB_good.txt)
* Confirm errors and "incomplete" warning are reported for large invalid files, e.g. [Dense_matrix_204_MB_bad_GENE_header.txt](https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-public/test/Dense_matrix_204_MB_bad_GENE_header.txt)
* Confirm click "Sync" button still works

This satisfies SCP-4017.


app/javascript/lib/validation/validate-file-content.js
```diff
@@ -22,7 +22,8 @@ import {
 
 import getSCPContext from 'providers/SCPContextProvider'
 
-// from lib/assets/metadata_schemas/alexandria_convention_schema.json (which in turn is from scp-ingest-pipeline/schemas)
+// from lib/assets/metadata_schemas/alexandria_convention_schema.json
+// (which in turn is from scp-ingest-pipeline/schemas)
 export const REQUIRED_CONVENTION_COLUMNS = [
```
bistline, line 27: Just thinking out loud here - we have an existing API endpoint at https://singlecell.broadinstitute.org/single_cell/api/v1/metadata_schemas/alexandria_convention/latest/json that will serve up the current version of the Alexandria convention.  There's a property at the end of the JSON schema file called `required` that we could simply consume, rather than hard-coding.  Just a thought...

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,77 @@
+/**
+* @fileoverview Client-side file validation (CSFV) for sync
+*/
+
+import { validateFileContent } from './validate-file-content'
+import { fetchBucketFile } from 'lib/scp-api'
+
+/**
+ * 50 MiB, max number of bytes to read and validate from remote file for sync.
```
bistline, line 9: Very nice detail.

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -1,9 +1,81 @@
 // shared JS functions for sync forms
 
-$('#study-file-<%= study_file.id %>').on('change', '.file-type', function() {
+// Used for sync CSFV
+if (typeof window.SCP.readOnlyToken === 'undefined') {
```
bistline, line 4: Blocker: we will have an issue with this feature not working in private studies where the user is not registered for Terra.  Since this is a "streaming" scenario, we can't simply use the user's regular OAuth token from `window.SCP.userAccessToken`, because if the user isn't registered for Terra, that token will not work.  However, `get_read_access_token` will return `nil` if the study is private and the user is not registered for Terra, and all fetch calls will fail with `401` and display the "Your session may have timed out" modal.

I'm not completely sure what the right thing to do here is, but my feeling is that we should use `window.SCP.userAccessToken` since this is already registered in the DOM, and we don't need to interpolate it like we do with `window.SCP.readOnlyToken` in this case.  If the user is not registered for Terra, we should display a notice that client-side validation will not work due to the incomplete Terra profile, and direct them to `app.terra.bio` to complete their registration.  We would then disable CSFV on the sync page.

I can also see just checking for `window.SCP.readOnlyToken` to be present, and the doing the same as above, but I think using the regular token is slightly cleaner, and we'd just do a check on `current_user.registered_for_firecloud`.

bistline (changes requested):
See comment on "Blocker".  TL;DR - this feature will break on private studies where the user is not registered for Terra.  This comprises ~11% of all studies right now (124 of 1152), which I consider large enough to warrant some kind of remediation or workaround.

Other than that, looks good, and no other suggestions.

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,77 @@
...
+
+  return file
+}
+
+/**
+* Validate a file in a GCS bucket; used for sync
+*/
+export async function validateRemoteFileContent(
+  bucketName, fileName, fileType, fileOptions
+) {
```
devonbush, line 56: The perftime measurement and other features will be simplified if this function is a very thin wrapper around `validateFileContent`.  All it needs to do is call validateFileContent with the bucketName added to fileOptions.

Then put most of the code below into a new function `{file, fileInfo} = createFileFromBucket(fileName, bucketName, maxBytes)` that returns the file object and `fileInfo`, which is  whatever metainfo (e.g. perfTimes) you need.  That way the everything lives inside the main timing loop, and `createFileFromBucket` is a more independently testable/useful method.  Assuming `createFileFromBucket` gets called from inside `parseFile`, this also opens the door for easy optimizations like only fetching the first few bytes of non-parseable file types, since all we want to do is validate the gzip encoding

app/views/studies/_shared_sync_functions.js.erb
devonbush, line 4: why is this `if` needed?   Wouldn't we want the latest version of the token even if it's already defined?

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -1,9 +1,81 @@
...
+  window.SCP.readOnlyToken = '<%= get_read_access_token(@study, current_user) %>'
+}
+
+/**
+* Transform validation results to HTML that can be shown to user.
+*
+* This function is a temporary backport of the issue messaging from
+* FileUploadControl.js.  It should be replaced when sync code migrates to
+* React.
+*/
```
devonbush, line 14: BLOCKER: Instead of duplicating all this, let's extract lines 151-164 of FileUploadControl into a new component `<ValidationMessages fileValidation={fileValidation}/>`. Then make a helper method `window.SCP.renderValidationMessages(target, fileValidation)` (similar to the `UploadWizard.renderUploadWizard` helper method).  You're right this may be temporary, but that render helper pattern is an important one for us to master as we're steadily migrating functionality over to React

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -95,6 +167,31 @@ $('#study-file-<%= study_file.id %>').on('change', '.file-type', function() {
...
     }
+
+    // FileUploadControl updates this more elegantly via React,
+    // but this direct approach is easier and as effective here.
+    let useConvention = false
+    if (fileType === 'Metadata') {
+      const conventionSelector =
+        '#study-file-<%= study_file.id %> ' +
+        '#study_file_use_metadata_convention_true'
+      useConvention = ($(conventionSelector)[0].checked === true)
```
devonbush, line 178: I *think* you can simplify the above code (and avoid jquery) as
`const conventionSelector = '#study-file-<%= study_file.id %> #study_file_use_metadata_convention_true'
`const useConvention = !! document.querySelector(conventionSelector)?.checked`

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -95,6 +167,31 @@ $('#study-file-<%= study_file.id %>').on('change', '.file-type', function() {
...
+    const fileOptions = {
+      use_metadata_convention: useConvention
+    }
+
+    const bucketName = '<%= @study.bucket_id %>'
+    const results = await window.SCP.validateRemoteFileContent(
+      bucketName, fileName, fileType, fileOptions
+    );
+
+    const issuesHtml = getIssuesHtml(results, fileName, fileId)
```
devonbush, line 189: as mentioned above, put an <div id="validation-results-{fileId}></div> on the form, and then just use a render helper method to put it on the form

test/js/lib/validate-remote-file-content.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+  it('sends range request for remote file', async () => {
+    const fakeLog = jest.spyOn(global, 'fetch')
+    fakeLog.mockImplementation(() => { })
+
+    window.SCP = {
+      readOnlyToken: 'test'
+    }
+
+    const bucketName = 'broad-singlecellportal-public'
```
devonbush, line 29: pull this into a constant at the top of the file -- we'll want it for other tests soon probably

devonbush (changes requested):
An impressive technical achievement!  Some moving around of methods and signatures will clean up some interfaces and avoid logic duplication.  

On the more trivial side, you split quite a few existing lines that were well within the 120 character .eslintc, and (IMHO) reasonably readable (e.g. the import in validate-file-content).  Was that intentional and, if so, should we revisit our line-length guidelines?  I don't want to get into a situation where code that is otherwise not touched gets auto-changed based on whoever's IDE saved it last.

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -1,9 +1,81 @@
...
+    // Tell user how to get in touch
+    const supportEmailAddress = 'scp-support@broadinstitute.zendesk.com'
+    const supportEmailLink =
+      `<a href="mailto:${supportEmailAddress}">` +
+        supportEmailAddress +
+      '</a>'
+    const supportMessage =
+      '<div>' +
+        `If you need help, let us know at ${supportEmailLink}.  ` +
+        'Please include the above errors, your file, and study accession.' +
```
devonbush, line 31: I like this inclusion of a help link right by the errors, let's add this to the new ValidationMessages component.  Since there's already likely to be a lot of text on the screen, looking for a way to shorten this might be helpful
`Need help?  Send us an email at {email} with the errors, your accession (${studyAccession}), and a description/attachment of your file`

devonbush:
One more thought -- how will this handle Jonah's case where, once he's got his pipeline working, he's going to want to add 200+ files at once, and not going to want to wait for validation?  It might be easy for him to just add setTimeouts in his script to give time for the validation to complete, or we might want to consider a way he can turn this off

eweitz:
Thanks for the thorough review, @devonbush!

> how will this handle Jonah's case

@jlchang might have insight on how the SCP CLI plus `gsutil` could best suit that programmatic use case.

If that's infeasible and if we suspect Jonah setting a `setTimeout` would be too finicky, then could you open a separate ticket to implement and verify a mechanism to disable sync CSFV?  That strikes me as best scoped, whether done via our pre-existing exemption architecture or otherwise.

devonbush:
yup, agree that it's outside the scope of this ticket.  If we have time, we'll ask Jonah about it tomorrow

app/javascript/lib/validation/validate-file-content.js
eweitz: Agreed, I think there's consensus it'd be cleaner.  Refactoring along those things seems most opportune once CSFV handles ontology validation (which I think is feasible to do quite performantly).

app/views/studies/_shared_sync_functions.js.erb
eweitz: This is a great technical find, but I don't think it should block merge.  Your data up top elevates the discussion:

> this feature will break on private studies where the user is not registered for Terra. This comprises ~11% of all studies right now (124 of 1152)

A more precise way to assess impact is "_How many users_ not registered for Terra synced for private studies?"  Over the last year, [only 1 user not registered for Terra has visited the Sync page](https://mixpanel.com/s/2ghU35), and it's unclear if that was even a private study.

That's < 0.01% of our users.  The very low figure makes sense, since sync is premised on transferring data from a Terra workspace, and having a user who isn't registered for Terra do that seems like a very rare use case.

Given the vanishingly small impact, I think this refinement makes sense to consider as a separate task.

app/javascript/lib/scp-api.js
```diff
@@ -324,13 +324,18 @@ export async function deleteStudyFile(studyAccession, fileId, mock=false) {
  * @param {String} bucketName bucket name
  * @param {String} fileName file name
 */
-export async function fetchBucketFile(bucketName, fileName, mock=false) {
+export async function fetchBucketFile(bucketName, fileName, maxBytes=null, mock=false) {
```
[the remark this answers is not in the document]
eweitz: Plausible!  Streaming parallel validation would likely reduce sync CSFV times by 2-5x.

That might be a good case for doubling `MAX_SYNC_CSFV_BYTES` as you suggest, but we'd still want some maximum download size.  The percentiles for files by size drop off sharply beyond 50 MB, yet a significant portion of files are much bigger.   E.g. [> 20% of dense matrices are > 1 GB](https://mixpanel.com/s/4klxkX).

Streaming parallel validation eliminates file size as a concern for upload CSFV, but not sync CSFV.  That's because bytes get sent up from the client _regardless of CSFV_ in the upload UI, but bytes get sent down to the client _only for CSFV_ in the sync UI.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -367,16 +366,29 @@ async function parseFile(file, fileType, fileOptions={}) {
   return parseResult
 }
 
-/** Validate a local file, return { errors, warnings, summary } object, where errors is an array of errors, and summary
- * is a message like "Your file had 2 errors"
+/**
+ * Validate a File object, return { errors, warnings, summary } object, where:
+ *   - errors is an array of errors,
+ *   - warnings is an array of warnings, and
+ *   - summary is a message like "Your file had 2 errors"
```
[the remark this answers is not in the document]
eweitz: Done!

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -409,7 +421,7 @@ function getValidationTrigger() {
 }
 
 /** Get properties about this validation run to log to Mixpanel */
-function getLogProps(fileInfo, errorObj, perfTime) {
+function getLogProps(fileInfo, errorObj, perfTime, syncProps) {
```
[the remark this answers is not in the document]
eweitz: Computing properties only intended for logging seems appropriate for a function named `getLogProps`.  Comments in my other reply beginning "Let's revisit" might relate here, too.

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -425,8 +437,21 @@ function getLogProps(fileInfo, errorObj, perfTime) {
...
+  }
+
+  if (syncProps) {
+    Object.assign(defaultProps, syncProps)
+
+    // Top-level perfTime is duration for the whole event.
+    // The 'perfTime:foo' construct is also used in "plot:" events.
+    Object.keys(syncProps).forEach(key => {
+      if (key.startsWith('perfTime:')) {
+        defaultProps.perfTime += syncProps[key]
```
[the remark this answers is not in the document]
eweitz: Per my other reply (beginning "Let's revisit"), such a refactor might end up being misleading and differently kludgy, so somewhat of a wash.  But it's worth considering as part of later work.

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,77 @@
...
+  return syncProps
+}
+
+/** Construct a File object from superficially-parsed range response data */
+export function getFileFromRangeData(content, fileName, contentType, syncProps) {
+  // If Range request didn't fetch the full file, then truncate the last
+  // line, which is almost certainly incomplete and thus invalid.
+  let cleanLineContent = content
+  if (!syncProps.fetchedCompleteFile) {
+    cleanLineContent = content.split('\n').slice(0, -1).join('\n')
```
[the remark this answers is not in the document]
eweitz: Done, including an optimization to check on each chunk instead of each line.

Out of curiosity, afterwards I benchmarked the original performance:

```
const newLineRegex = /\r?\n/
const text60MB =('foo'.repeat(100) + '\n').repeat(200_000)
const numBytes = (new TextEncoder().encode(text60MB)).length
console.log('Bytes in string: ' + numBytes)
const t0 = Date.now()
const cleanText = text60MB.split(newLineRegex).slice(0, -1).join(newLineRegex)
const t1 = Date.now()
console.log('Milliseconds to split and rejoin 60 MB: ' + (t1 - t0))

Bytes in string: 60200000
Milliseconds to split and rejoin 60 MB: 113
```

So the original way here is at worst ~2% slower (113 ms/5000 ms vs. ?/5000), but less complex and more self-contained (across 2 simple functions in 1 module vs. across 3 functions, one quite complex, in 3 modules).  If the update per suggestion is problematic I'd go with the original approach, but export `newLineRegex` from over there and use it here.

I also noticed (after refactoring, of course!) that the ticket notes refining this to use ChunkedLineReader is out of scope.

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,77 @@
...
+ * per https://www.speedtest.net/global-index/united-states.  > 80% of ingests
+ * are for files <= 50 MiB per https://mixpanel.com/s/3EbFMj.
+ *
+ * So 50 MiB means sync CSFV fully scans > 80% files, usually in < 5 seconds.
+ * Local tests gave 4.6 s (3.4 s remote read, 1.1 s validate; 138 Mbps down).
+ */
+export const MAX_SYNC_CSFV_BYTES = 50 * 1024 * 1024
+
+/** Add file-size-related entries to syncProps */
+export function updateSyncProps(contentRange, contentLength, syncProps) {
```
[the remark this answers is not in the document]
eweitz: Done!

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,77 @@
...
+    fileSizeFetched,
+    fileSizeTotal,
+    fetchedCompleteFile
+  })
+
+  return syncProps
+}
+
+/** Construct a File object from superficially-parsed range response data */
+export function getFileFromRangeData(content, fileName, contentType, syncProps) {
```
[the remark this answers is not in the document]
eweitz: I like the rename, but ended up inlining this as a side-effect of incorporating your other suggestion.

app/javascript/lib/validation/validate-remote-file-content.js
eweitz: Let's revisit this in a separate task, e.g. auto-selecting core file types for sync (SCP-4065).

A nice aspect of the original approach here is that validating remote files is structurally similar to validating local files.  In both local and remote cases, the `File` object is created outside `validateFileContent` (and thus outside `parseFile`).  Beyond the parallel structure, surely that's clearer than _creating_ a Foo in a function named `parseFoo`.

I like your point about validating gzip encoding.  I could see fetching the first ~500 kB instead of 3-4 bytes, then validating both magic numbers for gzip _and_ (assuming that succeeds) infer SCP file study type from the first few lines.  Perhaps both could be part of a pre-emptive tier of validation done before the user selects the file type.  In any case, I think such work is best scoped for another ticket.

app/views/studies/_shared_sync_functions.js.erb
eweitz: I'm actually not sure!  I naively reused the approach from `_study_visualize.html.erb`.   f6ecfa6 removes it.

app/views/studies/_shared_sync_functions.js.erb
eweitz: This seems like an eye-of-the-beholder scenario -- to me, the existing approach looks simpler!

app/views/studies/_shared_sync_functions.js.erb
eweitz: This makes sense to do as part of a separate task for `ValidationMessage`.

test/js/lib/validate-remote-file-content.test.js
eweitz: Good point, done!

app/views/studies/_shared_sync_functions.js.erb
eweitz: Nicely trimmed, and reasonable to scope as part of a separate task for `ValidationMessage`.

eweitz:
> On the more trivial side, this splits quite a few existing lines that were well within the 120 character .eslintc, and (IMHO) reasonably readable (e.g. the import in validate-file-content). Was that intentional and, if so, should we revisit our line-length guidelines? I don't want to get into a situation where code that is otherwise not touched gets auto-changed based on whoever's IDE saved it last.

That was a manual change by me, not an IDE auto-change.  I had been futzing with various imports, and wouldn't be surprised if I exceeded 120 characters at some point, manually split lines, then removed some experimental imports without going back to the old line length.

I similarly fixed a few line-length lint errors that existed before I started these changes, so it's also plausible that I pared back some technically conformant lines as part of that too!  I find <= 80-character lines most readable, but anything <= 120-characters is reasonable.

app/views/studies/_shared_sync_functions.js.erb
devonbush: Sorry for the confusion due to the vague ticket.  In the ticket, 'legacy UI' just meant that we were working inside existing legacy UI, instead of upgrading the whole thing to react,  and so would be reliant on legacy ways of, e.g. listening for menu changes.  We don't want to create *new* legacy UI structures that are duplicates of React components unless we absolutely have to, and here we don't have to.  Especially since this validation messaging is going to be under active development over the next months, tracking it in two places will make that work more difficult.

And I can't emphasize enough how much the render helper pattern will be important to migrating more of SCP to react/API patterns in an efficient way.  Right now, I'm assuming you did it the way you did because it was faster and easier for you.  We need to get to the point where it's faster and easier for you to say 'oh, I'll just reuse our existing component for that'.  I'm happy to help out any way I can -- grabbing an existing react component and rendering on a legacy UI template shouldn't take more than 10 minutes

app/views/studies/_shared_sync_functions.js.erb
devonbush: If you prefer your text, keep it (any text refinement suggestions from me should always be taken with grains of salt).  But otherwise, at least make the text change here and now so we remember and can iterate on it later with users having seen it

devonbush (changes requested):
Let's chat -- I don't think any of the requested changes should take too long.  I'm happy to help pair/tackle them with you if you want

app/views/studies/_shared_sync_functions.js.erb
eweitz: No worries, its difficult assess how much debt tech is reasonable for a functional MVP, especially when the ticket can be reasonably read a few different ways on the matter!  

I do like the pattern you introduced, but I do think most engineers with experience in both jQuery and React would find it faster and easier to do basic templating in jQuery when working in code that's in jQuery.

Per our chat I went ahead and implemented that new React component.  I also replaced up the redundant vestigial component and updated its test to apply to the new component.

app/views/studies/_shared_sync_functions.js.erb
eweitz: Done!  I included it in `ValidationMessage`.

app/javascript/lib/validation/validate-remote-file-content.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * 50 MiB, max number of bytes to read and validate from remote file for sync.
+ * Median fixed US bandwidth is 16 MiB/s (17 MB/s, 136 Mbps) as of 2021-12
+ * per https://www.speedtest.net/global-index/united-states.  > 80% of ingests
+ * are for files <= 50 MiB per https://mixpanel.com/s/3EbFMj.
+ *
+ * So 50 MiB means sync CSFV fully scans > 80% files, usually in < 5 seconds.
+ * Local tests gave 4.6 s (3.4 s remote read, 1.1 s validate; 138 Mbps down).
+ */
+export const MAX_SYNC_CSFV_BYTES = 50 * 1024 * 1024
```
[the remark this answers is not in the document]
eweitz: My understanding of `maxBytesPerLine` is that setting the default at 1 GiB ensures that very long lines can be _read in from memory_, while giving some protection that pathologically long lines don't freeze the UI.  (@devonbush, please correct me if that's wrong!)

`MAX_SYNC_CSFV_BYTES`, on the other hand, is a crude way to help ensure we _don't spend too much time downloading from network`_.  This value is much lower simply because network reads are so much slower than memory reads.

app/javascript/components/validation/ValidationMessage.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+/**
+ * Displays result of file validation issue for sync UI
+ *
+ * TODO (SCP-4119): Integrate this component into FileUploadControl.js
+ */
+export default function ValidationMessage({
+  studyAccession, errorMsgs, warningMsgs, fileName, isSync=false
```
devonbush, line 42: Renaming `isSync` to `showRefreshLink` would be clearer about what the parameter does

app/javascript/lib/validation/validate-remote-file-content.js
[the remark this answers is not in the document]
devonbush: That's exactly right -- the chunker has to hold all of a single line in memory at once, even though the line may be split into chunks.  Having a max line prevents the browser from running out of RAM trying to read a file that doesn't have newlines

app/javascript/lib/validation/validate-remote-file-content.js
devonbush: nice, very much appreciate the new refactoring!

app/views/studies/_shared_sync_functions.js.erb
devonbush: but *clearly* the suggested approach is *much* better.  jkjkjk, yup, either is fine

devonbush (approved):
Looks good.  One non-blocking code clarity suggestion, and then just the request to, if possible, update FileUploadComponent to use ValidationMessage (or you could do that in your message-trimming PR)

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -357,12 +357,23 @@ async function parseFile(file, fileType, fileOptions={}) {
...
+      if (syncProps?.fetchedCompleteFile === false) {
+        ignoreLastLine = true
+        const msg =
+          'Due to this file\'s size, it will be fully validated after sync, ' +
+          'and any errors will be emailed to you.'
+
+        parseResult.issues.push(['warn', 'incomplete:range-request', msg])
+      }
+
+      const chunker = new ChunkedLineReader(file, undefined, ignoreLastLine)
```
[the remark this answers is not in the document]
eweitz: I've reordered parameters to make `undefined` unnecessary, which seems unwonky while still succinct.

PR #1370, opened 2022-02-11
Adding condensed error summary to test logs (SCP-3191)

devonbush:
This adds a condensed error report to the end of our test runs, so that you don't have to scroll through to find the UI and Rails test results separately.   Next step will be to get this condensed report into the PR comments themselves.

The summary looks like this:
```
*** Yarn test failures **
  FAIL test/js/lib/github-reports.test.js
      ✕ fails when asserts fail (4 ms)
      ✕ fails when errors happen (1 ms)

*** Rails test failures **

  GithubReportsTest#test_should_fail_on_assert_failure [/Users/dbush/code/scp/single_cell_portal_core/test/lib/github_reports_test.rb:14]:
  Expected: 1

  GithubReportsTest#test_should_fail_on_exception:
  RuntimeError: Crazy test error

  GithubReportsTest2#test_should_fail_on_assert_failure_2_ [/Users/dbush/code/scp/single_cell_portal_core/test/lib/github_reports_test2.rb:14]:
  Expected: 1

  GithubReportsTest2#test_should_fail_on_exception_2:
  RuntimeError: Crazy test error

```

TO TEST:
1. manually edit some rails or js tests to fail
2. delete lines 1-120 and lines 153-166 of bin/run_tests.sh so you can run `bin/run_tests.sh` quickly locally
3. run `TEST_FILEPATH=test/lib/whatever_test_you_wanted bin/run_tests.sh`
4. Observe the console output now contains the summary log


devonbush:
It's a great question Jon -- We could expand `-A1` in the grep to `-A2` and get single line assertion fails, exceptions, and the first part of 'Expected: <x>' messages.  Making sure we got the full message (but ignoring the full stack trace) would be trickier--then we'd be doing advanced sed'ing to get all the lines not starting with certain amount of whitespace, and before a blank line.

devonbush:
done, updated example in root PR comment with the change -- good suggestion

eweitz (approved):
Useful!  I hadn't known about `pipeline` or `tee`.

PR #1376, opened 2022-02-15
Minor fixes to Upload Wizard (SCP-4105)

ehanna4:
This is the minimum changes from my dev choice project - still working on cooler/more dynamic fixes but wanted to at minimum put up code to fix some of the strange behavior that's on prod.

Fixes:
 - Bug that Devon noted where if a study has a very long name it used to cause an overlap, now the study name will be truncated
 - Previously if you sized the window small the content in the right side of the wizard would cover over the wizard steps panel. Now the two parts of the wizard content are more fixed in size/position so that resizing won't lose content and so that the panel doesn't get unnecessarily wide allowing for more real-estate to go towards the main content of the page.

To test:
- pull this branch
- open up a study to the upload wizard
- try moving around the window and observe that the panel and right-side content both stay accessible.

*Note
- there is a bug that is not addressed where the navbar doesn't dynamically resize in time so sometimes the "Help & resources" "create study" "profile" buttons swing down into the main page content. This exists in Prod and is not addressed in this PR

https://user-images.githubusercontent.com/54322292/154144396-6cd2e39f-d3cb-4639-bd07-745bfa6280fd.mov

### DIFF
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -340,20 +340,20 @@ export function RawUploadWizard({ studyAccession, name }) {
     <StudyContext.Provider value={studyObj}>
       <div className="upload-wizard-react">
         <div className="row">
-          <div className="col-md-12 wizard-top-bar">
-            <span>{serverState?.study?.name}</span> / &nbsp;
-            <a href={`/single_cell/study/${studyAccession}`}>View study</a>
+          <div className="col-md-12 wizard-top-bar no-wrap-ellipsis">
+          <a href={`/single_cell/study/${studyAccession}`}>View study</a> / &nbsp;
+            <span >{serverState?.study?.name}</span>
           </div>
         </div>
         <div className="row wizard-content">
-          <div className="col-md-3">
+          <div>
             <WizardNavPanel {...{
               formState, serverState, currentStep, setCurrentStep, studyAccession, mainSteps: MAIN_STEPS,
               supplementalSteps: SUPPLEMENTAL_STEPS, studyName: name
             }} />
           </div>
-          <div className="col-md-9">
-            <div className="flexbox-align-center top-margin">
+          <div>
+            <div className="flexbox-align-center top-margin margin-left">
               <h4>{currentStep.header}</h4>
               <div className="prev-next-buttons">
                 { prevStep && <button
--- a/app/javascript/components/upload/UploadWizard.js
+++ b/app/javascript/components/upload/UploadWizard.js
@@ -342,7 +342,7 @@ export function RawUploadWizard({ studyAccession, name }) {
         <div className="row">
           <div className="col-md-12 wizard-top-bar no-wrap-ellipsis">
           <a href={`/single_cell/study/${studyAccession}`}>View study</a> / &nbsp;
-            <span >{serverState?.study?.name}</span>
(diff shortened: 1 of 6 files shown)


app/javascript/styles/_uploadWizard.scss
```diff
@@ -147,3 +146,7 @@
 .prev-next-buttons {
   margin-left: auto;
 }
+
+.margin-left {
+  margin-left: 15px;
+}
```
eweitz, line 152:
```suggestion
}

```

eweitz: Ending files with newlines is an SCP code convention.  I believe there's a VS Code setting for it.

app/javascript/components/upload/UploadWizard.js
```diff
@@ -340,20 +340,20 @@ export function RawUploadWizard({ studyAccession, name }) {
...
+          <div>
             <WizardNavPanel {...{
               formState, serverState, currentStep, setCurrentStep, studyAccession, mainSteps: MAIN_STEPS,
               supplementalSteps: SUPPLEMENTAL_STEPS, studyName: name
             }} />
           </div>
-          <div className="col-md-9">
-            <div className="flexbox-align-center top-margin">
+          <div>
+            <div className="flexbox-align-center top-margin margin-left">
```
eweitz, line 356: For consistency with `top-margin`.

```suggestion
            <div className="flexbox-align-center top-margin left-margin">
```

Since `top-margin` already exists, I wouldn't be surprised if `left-margin` does too.  If so, we'll want to name this selector something like `foo-left-margin`, to match the general naming pattern while being clearly more specific.

app/javascript/styles/_uploadWizard.scss
```diff
@@ -147,3 +146,7 @@
 .prev-next-buttons {
   margin-left: auto;
 }
+
+.margin-left {
```
eweitz, line 150: For consistency with e.g.  `top-margin`.

```suggestion
.left-margin {
```

app/javascript/components/upload/UploadWizard.js
```diff
@@ -340,20 +340,20 @@ export function RawUploadWizard({ studyAccession, name }) {
     <StudyContext.Provider value={studyObj}>
       <div className="upload-wizard-react">
         <div className="row">
-          <div className="col-md-12 wizard-top-bar">
-            <span>{serverState?.study?.name}</span> / &nbsp;
-            <a href={`/single_cell/study/${studyAccession}`}>View study</a>
+          <div className="col-md-12 wizard-top-bar no-wrap-ellipsis">
+          <a href={`/single_cell/study/${studyAccession}`}>View study</a> / &nbsp;
+            <span >{serverState?.study?.name}</span>
```
eweitz, line 345: Helpful for accessibility:

```suggestion
            <span title="{serverState?.study?.name}">{serverState?.study?.name}</span>
```

eweitz (approved):
Code looks generally good!  I suggest accessibility and selector refinements, no blockers.

app/javascript/components/upload/UploadWizard.js
ehanna4: I will double check, i had looked and didn't see a `left-margin` but i certainly could've missed it

bistline (approved):

PR #1377, opened 2022-02-16
Update library preparation protocol dropdown (SCP-4082)

jlchang:
Long-time SCP study owner, Amanda Kedaigle let us know she has an upcoming study with SHARE-seq data (Zendesk 270215). She suggested adding both SHARE-seq and 10x multome (a comparable 10x protocol) to our library preparation protocol dropdown.

The dropdown had been organized by assay type but that organization was in comments only visible to developers. To make protocols more easily findable by study owners, the list has been re-organized to present protocols alphabetically.

PR adds the following terms the list of options:
'10x feature barcode/cell surface protein', # targeted proteomic
'10x feature barcode/CRISPR', # expression perturbation
'10x feature barcode/multiplexing', # scRNAseq
'10x Ig enrichment', # targeted transcriptomic
'10x multiome', # multiomic ATAC-seq
'10x TCR enrichment', # targeted transcriptomic
'SHARE-seq', # multiomic ATAC-seq
'Slide-seq', # spatial transcriptomic
'Slide-seqV2', # spatial transcriptomic

Also updated "scATAC-seq/10x" to "10x scATAC-seq" for uniformity
(studies with "scATAC-seq/10x" stored were [updated](https://broadinstitute.slack.com/archives/G01HY2DJTFY/p1645028901167319))

To test:
in your local instance, verify that the list is now organized alphabetically and the new terms above now appears in the dropdown list for library preparation protocol

This PR supports SCP-4082.

### DIFF
--- a/app/models/expression_file_info.rb
+++ b/app/models/expression_file_info.rb
@@ -44,6 +44,7 @@ class ExpressionFileInfo
                                 '10x feature barcode/multiplexing', # scRNAseq
                                 '10x Ig enrichment', # targeted transcriptomic
                                 '10x multiome', # multiomic ATAC-seq
+                                '10x scATAC-seq', # scATAC-seq
                                 '10x TCR enrichment', # targeted transcriptomic
                                 '10x Visium', # spatial transcriptomic
                                 'CEL-seq2', # scRNAseq
@@ -55,7 +56,6 @@ class ExpressionFileInfo
                                 'MARS-seq', # scRNAseq
                                 'MERFISH', # spatial transcriptomic
                                 'osmFISH', # spatial transcriptomic
-                                'scATAC-seq/10x', # scATAC-seq
                                 'scATAC-seq/Fluidigm', # scATAC-seq
                                 'sci-ATAC-seq', # scATAC-seq
                                 'sci-RNA-seq', # scRNAseq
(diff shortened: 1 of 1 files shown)


app/models/expression_file_info.rb
```diff
@@ -32,41 +32,47 @@ class ExpressionFileInfo
...
-                                '10x Visium',
-                                'MERFISH',
-                                'osmFISH',
-                                'SeqFISH+',
-                                'Slide-seq',
-                                'smFISH',
-                                'STARmap', # Note WIP for EFO term: https://broadinstitute.zendesk.com/agent/tickets/139334
-                                # single cell ChIP-seq assays
-                                'Drop-ChIP'].freeze
+  LIBRARY_PREPARATION_VALUES = [
```
bistline, line 35: Just a note about this - making updates to any constants that are used in validation can be problematic if we remove values (which thankfully we are not doing here).  If we ever do remove/deprecate values, we will need to add migrations to update any existing documents and replace the bad values with a known good.  Otherwise, this will prevent users from updating existing matrix files due to the validation failure.  Not a concern here, but one I think we should be aware of.

bistline (approved):

app/models/expression_file_info.rb
jlchang: You're right, the "scATAC-seq/10x" to "10x scATAC-seq" change would fall into that category. I've made manual updates (because there were very few files with "scATAC-seq/10x". So a migration for this rare case is not necessary.

ehanna4 (approved):
lgtm - the comments next to the constants to help provide context is a nice touch!

devonbush (approved):

PR #1380, opened 2022-02-17
Remove old wiz & flag (SCP-4076)

ehanna4:
Remove the old upload wizard code and the feature flag

To Test:
- Pull this branch
- Run the migration script via `bin/rails db:migrate`
- Go to the Feature Flags page and see that the react upload wizard flag is no longer there
- Go to the upload wizard and confirm all works.
Screen Shot 2022-02-17 at 11 22 36 AM


eweitz (approved):
Code looks good!  I see `currentStudyFiles` in 9 other seemingly old and unused files.  Can any more of those files be safely deleted?

devonbush:
To Eric's question, I believe most, if not all, of the other instances of currentStudyFiles are implicated in sync tool.  After we migrate sync tool to API usage is when we'll be able to clean out a TON of stuff

PR #1382, opened 2022-02-18
Improve CSFV logging robustness (SCP-4089)

eweitz:
This prevents Bard / Mixpanel errors caused by logging too many issues, or too big issues, from client-side file validation.

Previously, especially before [CSFV timeout support](https://github.com/broadinstitute/single_cell_portal_core/pull/1366), it was possible for Mixpanel logging to silently fail when the user encountered too many validation issues.  This could happen when e.g. selecting the wrong file type, or any time a particular validation issue occurred very many times.  That impacts observability, like debugging or triage or OKR assessment.

Now, lower-level protection against such Mixpanel errors is in place.  It does this by truncating the number of raw logged issues to <= 20, and each raw error message to <= 200 characters.  The full _number_ of errors and warnings is still reported, but their content is trimmed.  This also enables richer querying by logging `numErrorTypes` and `numWarningTypes`.

This specific prevention will endure even if we eventually stop using CSFV timeout or drastically speed up validation.  If this proves successful, it might be worth generalizing it to all string or array properties for all events logged to Mixpanel.

To test:
* Disable CSFV timeout, e.g. comment out the line with `timeoutOutCSFV` in `parseMetadataFile`.
* Create a study, or go to a study without a metadata file
* Open Network panel in Chrome DevTools
* Select a big (> 50 MB) file that isn't a metadata file, e.g. [Dense_matrix_204_MB_bad_GENE_header.txt](https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-public/test/Dense_matrix_204_MB_bad_GENE_header.txt)
* Wait for validation to complete
* In Network tab, verify it sent a `file-validation` request that got a 200 OK response status

This satisfies SCP-4089.


app/javascript/lib/validation/validate-file-content.js
```diff
@@ -417,8 +417,15 @@ function getValidationTrigger() {
   }
 }
 
+/** Trim messages logged to Mixpanel, to prevent HTTP 413 errors */
+function getTrimmedIssueMessages(issues) {
+  return issues.map(columns => {
+    return columns[2].slice(0, 200) // Show <= 200 characters for each message
```
devonbush, line 423: super nit-picky, but unless there are a large number of array-based callers, it's generally better to have functions accept singular entities, and let the callee handle iteration.  That way the logic is as isolated as possible for easy unit-testing

devonbush (approved):
good stuff -- I vote this gets included in the off-cycle release next week

app/javascript/lib/validation/validate-file-content.js
eweitz: Per Friday chat, we agree this is well-factored, and I've clarified the docstring in 67ad6f4.

PR #1387, opened 2022-02-22
adding HCA ToS to bulk download dialog

devonbush:
Adds a link to HCA Data Use Agreement.  It might look nicer below the table, but keeping it above the table makes sure it is visible even when results are long

image

TO TEST:
1. Make sure XDSS is enabled in your environment
2. do a search, e.g. "lymph" which returns SCP and HCA studies
3. click 'Download'
4. confirm text appears, and link takes you to HCA agreement in a new tab

### DIFF
--- a/app/javascript/components/search/controls/download/DownloadSelectionModal.js
+++ b/app/javascript/components/search/controls/download/DownloadSelectionModal.js
@@ -118,6 +118,9 @@ export default function DownloadSelectionModal({ studyAccessions, show, setShow 
             { showAzulSelectionPane &&
               <div>
                 <h4>Human Cell Atlas studies</h4>
+                <span className="detail">HCA studies are vailable under
+                  their <a href="https://data.humancellatlas.org/about/data-use-agreement" target="_blank" rel="noopener noreferrer">Data Use Agreement</a>
+                </span>
                 <DownloadSelectionTable
                   isLoading={isLoadingAzul}
                   downloadInfo={downloadInfoAzul}
--- a/app/javascript/styles/_search.scss
+++ b/app/javascript/styles/_search.scss
@@ -327,7 +327,7 @@ nav.search-links, #search-panel {
 }
 
 .download-table-container {
-  margin: 2em 0em 1em 0em;
+  margin: 1em 0em 1em 0em;
   overflow-x: hidden;
 }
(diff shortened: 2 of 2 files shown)


app/javascript/components/search/controls/download/DownloadSelectionModal.js
```diff
@@ -118,6 +118,9 @@ export default function DownloadSelectionModal({ studyAccessions, show, setShow
             { showAzulSelectionPane &&
               <div>
                 <h4>Human Cell Atlas studies</h4>
+                <span className="detail">Available under
+                  the <a href="https://data.humancellatlas.org/about/data-use-agreement" target="_blank" rel="noopener noreferrer">HCA Data Use Agreement</a>
```
eweitz, line 122: Let's omit `rel="noopener noreferrer"`.  It was a security measure for browsers that are no longer supported.

[`target=_blank` has implied `rel=noopener` by default](https://chromestatus.com/feature/6140064063029248) since Chrome 88, Firefox 79, and Safari 12.  The last browser version without this feature was EOL'd last year.  

I imagine IE 11 isn't up to par here, and although it's not EOL'd upstream until June 15, 2022, it's already not supported on Terra, so we're not obliged to support it either.  [< 0.01%](https://mixpanel.com/s/2OIWfe) of SCP users are on Internet Explorer.

`noreferrer` was a security patch for poor `noopener` support, so these points mean it's also unnecessary.

```suggestion
                  the <a href="https://data.humancellatlas.org/about/data-use-agreement" target="_blank">HCA Data Use Agreement</a>
```

eweitz (approved):
Looks good!  I suggest we omit `rel="noopener noreferrer"` -- details below.

bistline (approved):

app/javascript/components/search/controls/download/DownloadSelectionModal.js
devonbush: agree with the analysis.  I'll leave this in for now because it's part of our eslint (my IDE added it automatically), and we can make a separate PR to strip these from everywhere

eweitz: Good idea, I've opened SCP-4133.

PR #1393, opened 2022-02-25
Add Authors to keyword search (SCP-4124)

ehanna4:
Add authors to be searchable by keyword and phrase search.
MVP functionality:
- keyword search will search on first_name and last_name for author so if you have an author named "John Doe" a keyword search on jo or john or do or john do or john d will result with this study in the list.
- phrase search is very basic for the mvp and so a phrase will only match an author on either first or last name not both aka "john" or "doe" but not "john doe"

There is also no highlighting at this time for results, probably worth another ticket? Curious others thoughts as well on that.

Updated help modal:
Screen Shot 2022-02-28 at 4 34 14 PM

Updated search title and placeholder text:
Screen Shot 2022-02-28 at 4 34 24 PM

Results that include a study that has an author with last name Doe
Screen Shot 2022-02-28 at 4 34 51 PM

To Test:
- pull this branch
- go to a study you have edit permissions in and add an author
- go back to the home page
- search the author name using keyword, phrase or both
-


test/api/search_controller_test.rb
```diff
@@ -23,7 +23,7 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
     @other_user = FactoryBot.create(:api_user, test_array: @@users_to_clean)
     @study = FactoryBot.create(:detached_study,
                                name_prefix: "Search Testing Study #{@random_seed}",
-                               public: true,
+                               public: true,
```
devonbush, line 26: extra whitespace?

test/api/search_controller_test.rb
```diff
@@ -204,6 +204,19 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'testing', json['studies'].first['term_matches'].first
     assert_equal 'API Test Study', json['studies'].last['term_matches'].first
 
+    # test author keyword search
+    @author = FactoryBot.create(:author, study: @study, last_name: 'Doe', first_name: 'john')
```
devonbush, line 208: lets move this author creation into the 'before(:all)' block so that they get created at a predictable time

devonbush (approved):
Looks great -- this accomplishes a lot with very little extra code -- my favorite sort of change :).  Just a few code cleanup suggestions, and I'm sure there will be lots of good feedback on the search text modal from demo.

devonbush:
With respect to text highlighting, I suspect that will require some UX to display the author information alongside the description (which sounds like something useful to do anyway -- people care a lot about it).  You're right that if we do it, it should be in another ticket

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -22,10 +22,15 @@ export default function KeywordSearch({ keywordPrompt }) {
...
-    those that contain any of the specified terms. You can quote (&quot;) phrases to find matches that contain the entire
-    phrase, and you can combine both single terms and quoted phrases in your search.
+    <h4 className="text-center">Title, author, and description search</h4>
+    <p>Use the seach box below to perform a text-based search on study titles, decriptions and authors.</p>
+    <ul>
+      <li>A search without (&quot;)s will return any study that contains any of the specified terms.</li>
+      <li>To search for an exact phrase within a study title or description use (&quot;)s around the entire phrase like:</li>
+      <p>&quot;single cell&quot;</p>
+      <li> To search for studies that contain an exact phrase and/or the other search term/s, combine single terms and quoted phrase like:</li>
+      <p>&quot;single cell&quot; Smith</p>
```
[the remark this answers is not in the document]
ehanna4: oh great point - I will make that update

app/models/study.rb
```diff
@@ -968,7 +968,8 @@ def search_weight(terms)
         terms: {}
     }
     terms.each do |term|
-      text_blob = "#{self.name} #{self.description}"
+      author_names = authors.pluck(:first_name, :last_name).flatten.join(' ')
+      text_blob = "#{self.name} #{self.description} #{author_names}"
```
bistline, line 972: If we're indexing `institution` (from `app/models/author.rb#L35`), then we also need to add it here, otherwise the end-to-end integration on matching won't work correctly.

bistline (approved):
Looks good, and works as advertised.  I have some non-blocking suggestions that would be good to integrate.

I do have concerns about the reporting on this, however: given this proposed integration, there's not going to be any way for us to known when we hit the OKR for `at least 8 users launch searches based on author information`.  Since this is integrated directly with keyword-based search (which I agree is the correct way), then we don't know when a keyword is in fact an author match.  I don't have an idea off the top of my head for how we can achieve this reporting, but it is something we should think about - otherwise we need to rework that OKR into something we can measure and report on.

Also, I could see some downstream UI work to show author matches as badges, in the instance where the author name does not appear in the title/description, so that it is more obvious what the search result matched on.

ehanna4:
Leaving this here: https://broadworkbench.atlassian.net/browse/SCP-4147 so can reference later as the UX ticket for matched studies

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -13,7 +13,7 @@ import { closeModal } from 'components/search/controls/SearchPanel'
  * optionally takes a 'keywordValue' prop with the initial value for the field
  */
 export default function KeywordSearch({ keywordPrompt }) {
-  const placeholder = keywordPrompt ? keywordPrompt : 'Search title and description text'
+  const placeholder = keywordPrompt ? keywordPrompt : 'Search...'
```
eweitz, line 16: Ellipses are good for telling users to wait.  However, as seen in the [Terra style guide](https://www.sketch.com/s/cca3a834-985b-4165-bbae-21f615aebd2d) and almost all UIs, using them in text input or drop-down selection placeholders is unconventional these days.

```suggestion
  const placeholder = keywordPrompt ? keywordPrompt : 'Search'
```

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -22,15 +22,21 @@ export default function KeywordSearch({ keywordPrompt }) {
   const [showTextSearchHelpModal, setTextShowSearchHelpModal] = useState(false)
 
   const textSearchModalContent = (<div>
-    <h4 className="text-center">Title and description search</h4><br/>
-    Using the search box below will perform a text-based search of all study titles and descriptions, and will return
-    those that contain any of the specified terms. You can quote (&quot;) phrases to find matches that contain the entire
-    phrase, and you can combine both single terms and quoted phrases in your search.
+    <h4 className="text-center">Title, author, and description search</h4>
+    <p>Use the seach box below to perform a text-based search on study titles, decriptions and authors*.</p>
```
eweitz, line 26: Blocker: typos

```suggestion
    <p>Use the search box below to perform a text-based search on study titles, descriptions and authors*.</p>
```

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -22,15 +22,21 @@ export default function KeywordSearch({ keywordPrompt }) {
...
+    <ul>
+      <li>A search without (&quot;)s will return any study that contains any of the specified terms.</li>
+      <li>To search for an exact phrase within a study title or description use (&quot;)s around the entire phrase like:</li>
```
eweitz, line 29: Enclosing quote literals in parentheses and pluralizing them looks a bit awkward.  Spelling out quotes as a word and clarifying with quote-literals seems more readable.

```suggestion
      <li>A search without quotes (i.e. not wrapped in &quot;) will return any study that contains any of the specified terms.</li>
      <li>To search for an exact phrase within a study title or description use quotes around the entire phrase like:</li>
```

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -22,15 +22,21 @@ export default function KeywordSearch({ keywordPrompt }) {
...
+      <p>&quot;single cell&quot;</p>
+      <li> To search for studies that contain an exact phrase and/or the other search term/s, combine single terms and quoted phrase like:</li>
```
eweitz, line 31: Using "term(s)" would be more conventional than "term/s".  But I think such exacting techniques are more distracting than necessary here, so I'd opt to simply pluralize it.

```suggestion
      <li> To search for studies that contain an exact phrase and/or the other search terms, combine single terms and quoted phrases like:</li>
```

app/javascript/components/search/controls/KeywordSearch.js
```diff
@@ -22,15 +22,21 @@ export default function KeywordSearch({ keywordPrompt }) {
...
+      <p>&quot;single cell&quot; Smith</p>
+    </ul>
+    <p>*Structured data for authors is new in SCP, and many studies lack it, so author search results may be limited.</p>
```
eweitz, line 34: I'd omit the asterisk, but if we keep it let's space it properly.

```suggestion
    <p>* Structured data for authors is new in SCP, and many studies lack it, so author search results may be limited.</p>
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -516,10 +516,12 @@ def self.escape_terms_for_regex(term_list:)
       def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, query_context:)
         case query_context
         when :keyword
-          base_studies.any_of({:$text => {:$search => terms}}, {:accession.in => accessions})
+          author_match_study_ids = Author.where(:$text => {:$search => terms}).pluck(:study_id)
+          base_studies.any_of({:$text => {:$search => terms}}, {:accession.in => accessions}, {:id.in => author_match_study_ids})
```
eweitz, line 520: Blocker: for analytics on author search usefulness, and whether we hit the OKR, let's log to Mixpanel the studies returned by each query clause in `generate_mongo_query_by_context`.

Something like:
```suggestion
          by_text = {:$text => {:$search => terms}}
          by_accession = {:accession.in => accessions}
          by_author = {:id.in => author_match_study_ids}
          MetricsService.log('search-results', {
            numByAccession: by_accession.length
            numByText: by_text.length,
            numByAuthor: by_author.length
          }
          base_studies.any_of(by_text, by_accession, by_author)
```

app/controllers/api/v1/search_controller.rb
```diff
@@ -516,10 +516,12 @@ def self.escape_terms_for_regex(term_list:)
...
         when :phrase
           study_regex = escape_terms_for_regex(term_list: terms)
-          base_studies.any_of({name: study_regex}, {description: study_regex}, {:accession.in => accessions})
+          author_match_study_ids = Author.any_of({first_name: study_regex}, {last_name: study_regex}, {institution: study_regex}).pluck(:study_id)
+          base_studies.any_of({name: study_regex}, {description: study_regex}, {:accession.in => accessions}, {:id.in => author_match_study_ids})
```
eweitz, line 524: Blocker: let's log analytics similar to the suggestion for the other clause.

eweitz (changes requested):
Good stuff!  I note a trivial and a maybe non-trivial blocker, and suggest some non-blocking refinements.

I usually wouldn't block on analytics, but given we believe it's important to measure usage per the OKR and hitting the OKR seems unlikely this quarter otherwise and it seems straightforward to do here, I think it's worth adding now.

eweitz (approved):
Reviewing the [design doc](https://docs.google.com/document/d/1x_3qMjdZaa4aCAWDKJkBZ57hEHxxFPyPhvuSpqYTLco/edit), I see analytics is marked as optional, and a manual approach to measuring the OKR is described.  

Given the workaround, and the fact that the design doc notes it is optional, I no longer consider analytics a blocker here -- although it'd certainly help!

PR #1399, opened 2022-03-01
Suggest sync when large file is selected for upload (SCP-4013)

eweitz:
This suggests using sync upon selecting a >= 1 GiB for upload.  It improves discoverability of a way to add data that's sometimes much faster than SCP's default.

Syncing is often faster than the SCP upload UI for large files in these cases, ordered by descending fitness:
1. Files are already in Terra workspace
2. Files are already in another Google bucket, and user is comfortable on command line
3. Files are not yet in bucket, and user has a very fast upload connection, is comfortable on command line, and sometimes mistakenly navigates away from upload page or closes browser tab while uploading

This also links to relevant docs from atop the sync UI.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/156242169-40124298-0c8c-48d7-bacd-83be69f371f4.mov

For now, this suggestion is limited to users who are registered for Terra.  Refining the UX and docs for Terra-unregistered users is tracked in another ticket.  Documentation refinements to emphasize copying from another bucket (instead of uploading from a local machine) are also tracked separately.

This also refactors a fair amount of CSFV code to improve cohesion and remove obsolete code.  Previously, very similar code to start validating local and remote files was spread over three files.  Now it's in one file, `validate-file.js`.  Code for CSFV logging is now also centralized, separately, in `log-validation.js`.  This reduces duplication, helps maintainability, and eases future refinements and consolidations.

To test:
* Go to upload UI
* Select a file that is > 1 GiB in size
* Confirm sync suggestion appears
* Click suggested link
* Confirm it opens the sync UI in another tab
* Click "What is sync?" and "Adding data with gsutil" links
* Confirm they open related docs in another tab

This satisfies SCP-4013.


app/javascript/lib/validation/validate-file.js
```diff
@@ -0,0 +1,167 @@
...
+ * @param file {File} the File object from the input
+ * @param studyFile {StudyFile} the JS object corresponding to the StudyFile
+ * @param allStudyFiles {StudyFile[]} the array of all files for the study, used for name uniqueness checks
+ * @param allowedFileExts { String[] } array of allowable extensions, ['*'] for all
+ */
+export async function validateLocalFile(file, studyFile, allStudyFiles=[], allowedFileExts=['*']) {
+  const nameIssues = validateFileName(file, studyFile, allStudyFiles, allowedFileExts)
+
+  let issuesObj
+  console.log('nameIssues', nameIssues)
```
bistline, line 55: Debug statement

bistline (approved):
Looks good - especially appreciate the sensible refactoring.  I found one debug `console.log()` statement, otherwise everything looks ready to go.

app/javascript/lib/validation/validate-file.js
eweitz: Removed in 1abfbcf!

jlchang (approved):
Functional reivew - large file detected and sync suggested. Following link to sync page finds nice links to explain page functionality.

Hoping we see more sync usage! Especially after we confirm that Terra registration is not necessary and can expand the messaging to more users.

app/javascript/lib/validation/io.js
```diff
@@ -9,7 +9,10 @@
  * [1] https://stackoverflow.com/a/47976589
 */
 
-export const DEFAULT_CHUNK_SIZE = 1024 * 1024 // 1 MiB
+export const MiB = 1024 * 1024 // 1 MiB
+
+export const GiB = MiB * 1000 // 1 GiB
```
devonbush, line 14: I found these export labels confusing (but that might just be that I've never actually used Mebibytes before, instead of MB).  I might have named these ONE_MiB or 1_MiB.  Is there a reason we are using the much more obscure Mebibyte than a MB?
Also, the things I'm looking at online say that a GiB is 1024^3 instead of MiB * 1000.

app/javascript/lib/validation/validate-file.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    logFileValidation(fileInfo, issuesObj, perfTimes)
+  } else {
+    issuesObj = formatIssues(nameIssues)
+  }
+
+  if (file.size >= GiB) {
+    issuesObj.infos = [['info', 'size:large', '']]
+  }
```
devonbush, line 77: being able to add this here so cleanly indicates to me that you did a nice job refactoring the local/remote validation logic

devonbush (approved):
Nice refactor--super clean.  The usage of MiB may be unnecessarily technical, but I don't think it's a big deal either way.

app/javascript/lib/validation/io.js
eweitz: Great spot with GiB.  Fixed.

It might be worth reevaluating, but I recall we began using binary prefixes (e.g. MiB) instead of decimal prefixes (e.g. MB) because that's what's most often used in network transfers, file storage, and/or quotas.  

However, those binary prefixes are actually _displayed_ in UIs as decimal prefixes, which can be very confusing.  So we opted to use the correct (and admittedly rather technical) binary prefix in our code.

app/javascript/lib/validation/validate-file.js
eweitz: Thanks!  The refactor largely centered around this specific line.

PR #1407, opened 2022-03-04
Base64 encode custom color annotation/label names to avoid MongoDB (SCP-4151)

bistline:
This update addresses an issue in our version of MongoDB where BSON documents cannot have key names with either periods or dollar signs (`.`, `$`).  This constraint is due to both being control characters that have special function (periods are used for hash key lookup notation, and dollar signs denote clauses/actions in queries).  This is not an issue in other parts of the application where we control the keys being saved, but for custom color objects in `ClusterFileInfo`, the key names map to annotation and label names, all of which are user-generated.

To get around this issue, this update now base64 encodes hash key names for `ClusterFileInfo.custom_colors` documents.  A helper method called `transform_custom_colors` is used to both encode and decode nested hashes, and a convenience method of `decoded_custom_colors` can be used in the UI for automatically decoding values at runtime.  In addition, a reversible database migration has been added to covert existing data in the database.

MANUAL TESTING
1. Boot as normal and sign in
2. In a separate terminal window, ensure caching is turned off with `bin/rails dev:cache` and look for the message `Development mode is no longer being cached.`
3. Load any study with existing clusters, and update the colors for one annotation and save
4. Upload the file `test_data/validation/cluster_example_with_periods.txt` to this same study
5. Back in the other terminal window, run the database migration with `bin/rails db:migrate` (doing this after saving a custom color object will validate the migration transform works)
6. Reload the study, and validate the custom colors you saved earlier still work
7. Load the `cluster_example_with_periods.txt` cluster, and set custom colors for the `annotation.with.periods` annotation
8. Confirm that colors save and render correctly

This PR satisfies SCP-4151.


eweitz (approved):

app/models/cluster_file_info.rb
```diff
@@ -5,6 +5,22 @@ class ClusterFileInfo
 
   field :custom_colors, type: Hash, default: {}
 
+  # decode base64 annotation/label names back into human-readable text
+  def decoded_custom_colors
+    self.class.transform_custom_colors(custom_colors, :decode64)
+  end
+
+  # helper to transform a nested custom color hash to/from base64
+  def self.transform_custom_colors(hash, transform = :encode64)
```
devonbush, line 14: why would we want to call this function with a different transform?  Wouldn't that cause data problems if some clusters were encoded with different transforms?  I think if :encode64 and :decode64 are declared at the top of the file as COLOR_ENCODE_TRANSFORM, it communicates more clearly what we're doing

devonbush (approved):
Looks good, one question about parameters

app/models/cluster_file_info.rb
bistline: I was initially going to do that, but thought that adding an extra constant just for that seemed overkill.  But your point holds - we only want to allow `:encode64` and `:decode64`, so I'll change that.

PR #1408, opened 2022-03-04
Update title to Loom (SCP-4095)

ehanna4:
Update the download modal to use a label of "Loom" rather than "Analysis"

This will only impact the mixpanel event `click:input-checkbox` 's property of `label` if a user choose no Loom files during download. I didn't find a way in mixpanel to link two properties in an event but it doesn't look like this is a metric we used much yet so doesn't seem like much of an issue.

To test:
- pull this branch
- search for some azul data (i.e. pulmonary) while signed in
- click to download and you will see the section label for loom files is Loom now

Screen Shot 2022-03-04 at 3 26 33 PM

### DIFF
--- a/app/javascript/components/search/controls/download/DownloadSelectionTable.js
+++ b/app/javascript/components/search/controls/download/DownloadSelectionTable.js
@@ -178,9 +178,9 @@ const COLUMNS = {
     info: 'List of available project files and associated project-level metadata.',
     default: true
   },
-  loom: {
+  analysis: {
     title: 'Loom',
-    types: ['loom_file'],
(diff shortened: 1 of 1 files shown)


eweitz (approved):

devonbush:
I don't understand how the filter will work if you're changing the file_types to 'loom_file' -- azul_search_service is still using 'analyis_file' to categorize these files.

devonbush:
I don't understand how the filter will work if you're changing the file_types to 'loom_file' -- azul_search_service is still using 'analyis_file' to categorize these files.

devonbush (approved):

PR #1412, opened 2022-03-08
Fix so JS or Ruby test fails the CI check (SCP-4165)

ehanna4:
Fixes a bug where the RETURN_CODE had been getting re-set to the result of the Ruby tests always (see comment) now if either the JS or Ruby tests fail the test failure will result in a true failure at the `SCP Continuous Integration / Run-All-Tests (pull_request)` level.

[Slack Discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1646757615557049)

### DIFF
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -125,7 +125,10 @@ set -o pipefail
 if [[ "$TEST_FILEPATH" == "" ]]; then
   # "2>&1 | tee yarn_test.log" puts the output of the tests both into stdout and the yarn_test.log file
   yarn ui-test 2>&1 | tee yarn_test.log
-  RETURN_CODE=$? # immediately capture exit code to prevent this from getting clobbered
+  code=$? # immediately capture exit code to prevent this from getting clobbered
+  if [[ $code -ne 0 ]]; then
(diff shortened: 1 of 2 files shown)


bin/run_tests.sh
```diff
@@ -151,7 +151,7 @@ grep -A2 -B1 ")\sFailure:\|)\sError:" rails_test.log | \
     grep -v ")\sFailure:\|)\sError:\|--" >> test_summary.txt
 
 # the return code will be 1 if either the js or rails tests exited with 1
-(( RETURN_CODE = exit_status || $code ))
+(( RETURN_CODE = RETURN_CODE || $code ))
```
ehanna4, line 154: Since `exit_status` wasn't actually getting defined and then `$code` gets overwritten with the exit status of the Ruby test runs, previously the JS test run results were getting ignored. Now the JS results get stored in `RETURN_CODE` so if either fails we will get the correct failed returned code

eweitz (approved):
I see this works, as [JS test failures now cause build failures](https://github.com/broadinstitute/single_cell_portal_core/runs/5469978946?check_suite_focus=true#step:4:4686), and I appreciate the note!

PR #1414, opened 2022-03-09
removing analysis tab and related workflow/submission views (SCP-4132)

devonbush:
SCP-4132. This removes the analysis tabs and related workflow/submission views from SCP.  This does not remove the backend models and functionality, as we may still want to preserve records of past workflows run.

TO TEST:
1) sign in, and open javascript console
2) load a study
3) confirm the analysis tab does not appear, and no errors in the JS console
4) go to the study details for a study, confirm 'submissions' table is not there, and again no JS errors in console
5) go to 'Reports' page, and confirm submissions report does not appear


eweitz (approved):
Looks good!  Removing so much unused code nicely improves maintainability.

Do you anticipate any performance benefit?

devonbush:
Not particularly, since we'd already async'ed most of this code.  But it will speed up work on user annotations and custom colors, since those had to wait for the same async call.

ehanna4 (approved):
Looks good 👍

PR #1415, opened 2022-03-11
Add tracking for user clicking next with Azul data to download (SCP-4157)

ehanna4:
Add tracking of the number of SCP and number of Azul files that have been chosen for download when a user clicks the 'Next' button in the download modal which triggers the `logDownloadAuthorization` event.

To test
- pull this branch
- log in to SCP
- perform a search that will have Azul results such as 'male'
- click the download button
- Check out[ this report in Mixpanel ](https://mixpanel.com/s/viTZ8)to see an event come through tracking that your user profile clicked next

Screen Shot 2022-03-11 at 11 08 37 AM
Screen Shot 2022-03-11 at 11 08 47 AM

### DIFF
--- a/app/javascript/lib/scp-api-metrics.js
+++ b/app/javascript/lib/scp-api-metrics.js
@@ -217,11 +217,25 @@ export function getLogPlotProps() {
 }
 
 /**
+ * Extract the number of files that have been chosen for download
+ */
+function getNumAzulFiles(azulFiles) {
+  let numAzulFilesChosen = 0
+  for (const val of Object.entries(azulFiles)) {
+    if (val[1].length > 0) {
+      numAzulFilesChosen += val[1].length
+    }
+  }
+  return numAzulFilesChosen
+}
+
+/**
  * Log when a download is authorized.
  * This is our best web-client-side methodology for measuring downloads.
  */
-export function logDownloadAuthorization(perfTimes) {
-  const props = { perfTimes }
+export function logDownloadAuthorization(perfTimes, fileIds, azulFiles) {
+  const numAzulFilesChosen = getNumAzulFiles(azulFiles)
+  const props = { perfTimes, 'numSCPFiles': fileIds.length, 'numAzulFiles': numAzulFilesChosen }
   log('download-authorization', props)
   ga('send', 'event', 'advanced-search', 'download-authorization') // eslint-disable-line no-undef, max-len
 }
--- a/app/javascript/lib/scp-api-metrics.js
+++ b/app/javascript/lib/scp-api-metrics.js
@@ -217,7 +217,7 @@ export function getLogPlotProps() {
 }
 
 /**
- * Extract the number of files that have been chosen for download
+ * Extract the number of Azul files that have been chosen for download
  */
 function getNumAzulFiles(azulFiles) {
   let numAzulFilesChosen = 0
@@ -230,7 +230,7 @@ function getNumAzulFiles(azulFiles) {
 }
 
 /**
- * Log when a download is authorized.
+ * Log when a download is authorized and the number of files from each source that are being downloaded.
  * This is our best web-client-side methodology for measuring downloads.
  */
 export function logDownloadAuthorization(perfTimes, fileIds, azulFiles) {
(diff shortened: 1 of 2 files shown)


app/javascript/lib/scp-api-metrics.js
```diff
@@ -217,11 +217,25 @@ export function getLogPlotProps() {
...
+ * Extract the number of Azul files that have been chosen for download
+ */
+function getNumAzulFiles(azulFiles) {
+  let numAzulFilesChosen = 0
+  for (const val of Object.entries(azulFiles)) {
+    if (val[1].length > 0) {
+      numAzulFilesChosen += val[1].length
+    }
+  }
+  return numAzulFilesChosen
```
devonbush, line 229: I'm not sure why you need the 'if' in this method, if the length is 0 it's okay to add it.  Also, array destructuring here would help as implicit documentation to clarify what the structure of the `azulFiles` object is.
If you want to get fancy, you can also use reduce to avoid the initial set to zero line

```suggestion
  return Object.entries(azulFiles).reduce((totalNum, [azulStudyId, fileList]) =>  {
     return totalNum += fileList.length
  }, 0)
```

app/javascript/lib/scp-api-metrics.js
```diff
@@ -217,11 +217,25 @@ export function getLogPlotProps() {
...
+}
+
+/**
+ * Log when a download is authorized and the number of files from each source that are being downloaded.
```
devonbush, line 233: please add a comment about what the fileIds and azulFiles arguments are

app/javascript/lib/scp-api.js
```diff
@@ -90,7 +90,7 @@ export async function fetchAuthCode(fileIds, azulFiles, mock=false) {
   })
   const [authCode, perfTimes] = await scpApi('/bulk_download/auth_code', init, mock)
 
-  logDownloadAuthorization(perfTimes)
+  logDownloadAuthorization(perfTimes, fileIds, azulFiles)
```
devonbush: adding @param documentation for the fileIds and azulFiles to fetchAuthCode would also be helpful

devonbush (changes requested):
Looks good, I appreciate the helper function you pulled out for the file number computation.  Just some stylistic refinements and comments to add.

app/javascript/lib/scp-api-metrics.js
ehanna4: I made an update based on your feedback, please let me know what you think.
I removed the extraneous if-statement and added an array destructure so the format of the object is more obvious.

I am hesitant to use reduce here though since the for-loop works and is a more generally recognizable pattern rather than `reduce` which is more explicitly JS-y and doesn't add too much in this case.

app/javascript/lib/scp-api-metrics.js
```diff
@@ -217,11 +217,28 @@ export function getLogPlotProps() {
...
+ * fileIds is an array containing the file ids for SCP sourced files
+ * azulFiles is an object with file info sourced from HCA files
+ *
  * This is our best web-client-side methodology for measuring downloads.
  */
-export function logDownloadAuthorization(perfTimes) {
-  const props = { perfTimes }
+export function logDownloadAuthorization(perfTimes, fileIds, azulFiles) {
+  const numAzulFilesChosen = getNumAzulFiles(azulFiles)
+  const props = { perfTimes, 'numSCPFiles': fileIds.length, 'numAzulFiles': numAzulFilesChosen }
```
eweitz, line 241: More concisely:

```suggestion
  const numAzulFiles = getNumAzulFiles(azulFiles)
  const props = { perfTimes, 'numSCPFiles': fileIds.length, numAzulFiles }
```

app/javascript/lib/scp-api.jsx
```diff
@@ -75,6 +75,8 @@ function defaultPostInit(mock=false) {
  * Docs: https:///singlecell.broadinstitute.org/single_cell/api/swagger_docs/v1#!/Search/search_auth_code_path
  *
  * @param {Boolean} mock If using mock data.  Helps development, tests.
+ * @param {Array} fileIds The ids of the SCP sourced files selected for download
+ * @param {Object} azulFiles Object with arrays of file info for the Azul sourced files selected for download
```
eweitz, line 79: Use same order as function signature:

```suggestion
 * @param {Array} fileIds The ids of the SCP sourced files selected for download
 * @param {Object} azulFiles Object with arrays of file info for the Azul sourced files selected for download
 * @param {Boolean} mock If using mock data.  Helps development, tests.
```

eweitz (approved):
Code looks good!  I suggest a few trivial non-blocking refinements.

PR #1416, opened 2022-03-11
greying and translucent unspecified label points (SCP-3782)

devonbush:
SCP-3782.  This grays and makes more translucent points with unspecified labels.  
image

See the [design doc](https://docs.google.com/document/d/100mDxhnJgQpFjBDVxTHKRnsxdSpsY9stl1oycmlmkDg/edit#) for more details about why this implementation path was chosen.

TO TEST:
1. Load a study with a number of 'unspecified' points.  (e.g. cluster2 of the human_lymph synthetic study)
2. confirm the points appear as gray
3. increase the point size of the study so that the points overlap (for the human_lymph study, increasing the point size to ~20 does the trick
4. confirm the points under the unspecified points appear.


ehanna4 (approved):
Looks good, tested with the suggested study and all looks as described.

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
```diff
@@ -194,6 +193,9 @@ function getShowHideEnabled(hiddenTraces, countsByLabel) {
  * Get color for the label, which can be applied to e.g. the icon or the trace
  */
 function getColorForLabel(label, customColors={}, editedCustomColors={}, i) {
+  if (label === '--Unspecified--' && !editedCustomColors[label] && !customColors[label]) {
+    return 'rgba(80, 80, 80, 0.4)'
```
eweitz, line 197: This has a [1.96 contrast ratio](https://www.joedolson.com/tools/color-contrast.php?type=rgb&color=80%2C80%2C80&color2=255%2C255%2C255&alpha=.4), which is inaccessible for some users.  Bumping alpha to 0.8 makes it [accessible](https://www.joedolson.com/tools/color-contrast.php?type=rgb&color=80%2C80%2C80&color2=255%2C255%2C255&alpha=.8).  I can imagine an argument that "Unspecified" cells are of minimal interest and peripheral to the graphic, so I don't necessarily consider this a blocker.

```suggestion
    return 'rgba(80, 80, 80, 0.8)'
```

eweitz (approved):
Code looks good!  I suggest a non-blocking accessibility refinement.

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
devonbush: Yeah, that's a good point about accessibility.  My sense is that you're right, we're trying to make it clear that these points are peripheral, and so making them *very* different from the specified points is actually a net benefit to visually impaired users, even if it means they might not see them at all.

PR #1418, opened 2022-03-15
Add file name and trigger to Rails ingest event Mixpanel logging (SCP-4062)

eweitz:
This adds `fileName` and `trigger` properties to Rails `ingest` event Mixpanel logging.  It helps analyze product questions.  

For example, knowing if the trigger was "sync" or "upload" can reveal gaps in client-side file validation in cases where Ingest Pipeline logging (unknowingly) lacks coverage.  That's part of two Q1 OKRs.  Leveraging file names can also help rough analyses of e.g. performance of various file extensions, and correlate ingest debugging across our technology stack.

Here's how it looks in Mixpanel:

Trigger_and_fileName_properties_in_ingest_Mixpanel_log_event__SCP_2022-03-15

This satisfies SCP-4062.


app/models/ingest_job.rb
```diff
@@ -537,14 +537,23 @@ def read_parse_logfile(filepath, delete_on_read: true, range: nil)
   #   - (Hash) => Hash of job statistics to use with IngestJob#log_to_mixpanel
   def get_job_analytics
     file_type = self.study_file.file_type
+
+    if self.study_file.remote_location.present?
+      trigger = 'sync'
+    else
+      trigger = 'upload'
+    end
```
devonbush, line 545: declaring variables inside an if block that later get used outside the 'if' block can make code tougher to read as finding the source of a variable is more difficult (especially depending on code folding in the IDE).  It also risks harder-to-catch typos if one of the  declarations inside the if branch is misspelled.  Avoiding it will keep our codebase more generally legible
```suggestion
    trigger = self.study_file.remote_location.present? ?  'sync' : 'upload'
```

devonbush (approved):
simple enough -- one non-blocking suggestion for code clarity.

app/models/ingest_job.rb
eweitz: Concise, too -- I like it.

PR #1420, opened 2022-03-16
improving performance of large data array loading (SCP-3756)

devonbush:
In performance testing large plot rendering, I noticed some suboptimal loading of data arrays.  This preserves doing the sort in rails (necessary to avoid mongo sort limits), but now fetches the data in a single query, and does not unnecessarily duplicate arrays on assembly.  Note that in Ruby [1] + [2] will create a new array, while [1].concat([2]) will just append the result to the original, and so is much more performant for stitching together multiple large arrays.  See https://stackoverflow.com/questions/40510579/how-to-efficiently-concatenate-multiple-arrays-in-ruby 

This resulted in a 10-20% speedup in my local instance (~100sec vs ~90sec to load a 1.3M cell cluster).  I suspect this will have negligible impact in production, but the development speedup will be worth it (and we can check our production perf numbers just in case this has an impact :) )

TO TEST:
1. Load a large study
2. confirm the data displays correctly

### DIFF
--- a/app/models/concerns/concatenatable.rb
+++ b/app/models/concerns/concatenatable.rb
@@ -7,7 +7,8 @@ module Concatenatable
   # concatenate the values together into a single contiguous array
   def concatenate_arrays(query)
     arrays = where(query)
-    ids = arrays.pluck(:id, :array_index).sort_by(&:last).map(&:first)
-    ids.map { |id| find(id).values }.reduce([], :+)
+    arr_vals = arrays.pluck(:array_index, :values)
+    arr_vals = arr_vals.sort_by(&:first)
+    arr_vals.reduce([]) { |aggregate, arr_val|  aggregate.concat(arr_val.last) }
   end
 end


app/models/concerns/concatenatable.rb
```diff
@@ -7,7 +7,8 @@ module Concatenatable
   # concatenate the values together into a single contiguous array
   def concatenate_arrays(query)
     arrays = where(query)
-    ids = arrays.pluck(:id, :array_index).sort_by(&:last).map(&:first)
-    ids.map { |id| find(id).values }.reduce([], :+)
+    arr_vals = arrays.pluck(:array_index, :values)
+    arr_vals = arr_vals.sort_by(&:first)
+    arr_vals.reduce([]) { |aggregate, arr_val|  aggregate.concat(arr_val.last) }
```
eweitz, line 12: In JavaScript, [`reduce` is > 3x slower than `for` loop](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arrayreduce-vs-for-and-forof), so perhaps more primitive iteration would optimize performance here in Ruby, too.

eweitz (approved):
10-20% faster is a solid improvement.  Would we expect to see the speed improvement mostly in gene expression plots, since cluster plots are cached?

Below, I speculate on another optimization.

eweitz:
The CI [build failure](https://github.com/broadinstitute/single_cell_portal_core/runs/5563556362?check_suite_focus=true#step:4:4625) seems to be a false positive.  It concerns permissions, which seem unchanged by this code.

```
### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  ApiSiteControllerTest#before_all:
  RuntimeError: Wrapped undumpable exception for: RuntimeError: single-cell-portal-test/api-site-controller-study-pyq04 does not exist or you do not have permission to use it
Exiting with code: 1
```

ehanna4 (approved):
Saw a few seconds speed up locally (but only tested once and on a cluster plot as that's all I have that's large locally). LGTM update makes sense and everything still works :)

app/models/concerns/concatenatable.rb
devonbush: I think that is mainly true where the number of iterations is very large, and so saving a function call overhead is important.  Here, the number of iterations is typically <100 (a 1MM cell study will have ~100 data arrays),  so reduce does better since ruby can anticipate the array assembly.

PR #1429, opened 2022-03-22
Add metrics to capture the match type for SCP studies (SCP-4155)

ehanna4:
For our OKR on adding author info to studies adding this metric will enable us to more easily see how many studies match the search based on term, phrase or author match. 

This PR will only add tracking for SCP studies (aka not Azul studies), since the OKR is around author search which isn't supported for Azul at this time.

There is definitely potential follow up work to add Azul match data and more study result info. However getting this bit out for OKR tracking seemed most important so I am planning to file a follow on ticket for that other work as long as there is no strong opposition.

To test:
- Pull this branch and run it locally
- Search from the home page using different terms and/or phrases
- Check out the network tab for events titles 'search' and check out the property scpStudiesMatchData
- Or check out mixpanel dev for the same

Screen Shot 2022-03-22 at 3 33 04 PM

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -329,16 +329,16 @@ module Api
           if facets_to_keywords.any?
             @inferred_terms = facets_to_keywords.values.flatten
             logger.info "Running inferred search using #{facets_to_keywords}"
-            search_match_obj = inferred_studies = self.class.generate_mongo_query_by_context(terms: facets_to_keywords,
+            search_match_obj = self.class.generate_mongo_query_by_context(terms: facets_to_keywords,
                                                                           base_studies: @viewable,
                                                                           accessions: @matching_accessions,
                                                                           query_context: :inferred)
-            @studies = search_match_obj[:studies]
+            @inferred_studies = search_match_obj[:studies]
             @match_by_data = search_match_obj[:results_matched_by_data]
-            @inferred_accessions = inferred_studies.pluck(:accession)
+            @inferred_accessions = @inferred_studies.pluck(:accession)
             logger.info "Found #{@inferred_accessions.count} inferred matches: #{@inferred_accessions}"
             @matching_accessions += @inferred_accessions
-            @studies += inferred_studies.sort_by { |study| -study.search_weight(@inferred_terms)[:total] }
+            @studies += @inferred_studies.sort_by { |study| -study.search_weight(@inferred_terms)[:total] }
           end
         end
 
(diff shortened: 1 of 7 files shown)


devonbush (changes requested):
Code looks good.  Please update 'should return search results using keywords' in search_controller_test.rb to test your new metrics object.

Also, it looks like if a combined keyword+facet search is done, this will return the # and types of keyword matches, regardless of whether they were filtered out by the facet results.  Is that intended? (this is probably a rare enough case that it doesn't really matter)

eweitz (changes requested):
Good stuff!  I hope this is the first in a string of analytics enhancements for search.

I note one blocker: Mixpanel log properties should be flattened per SCP convention.  Even if nested properties are searchable (I suspect they aren't, or at least not nearly as easily), being consistent in format with other log properties is important for analyst UX.

Here's how nested properties look in Mixpanel -- not glanceable, and likely problematic for easy queryability:

Nested_property_in_Mixpanel__SCP_2022-03-23


I also suggest several non-blocking refinements on log property nomenclature ("matches" -> "results") and segmentation (e.g. `"scpStudiesMatchData: {'numMatchesByAuthor': 2}"` -> `"numResults:scp:author": 2`).

ehanna4:
> Also, it looks like if a combined keyword+facet search is done, this will return the # and types of keyword matches, regardless of whether they were filtered out by the facet results. Is that intended? (this is probably a rare enough case that it doesn't really matter)

Is a combined keyword+facet search the inferred search? If so this will set the `matchData` to an empty {} so it will report nothing rather than incorrect data

devonbush:
I should have been clearer -- by a combined keyword + facet search I meant if someone searches for the facet `Species: rattus norvegicus` and the text 'Smith', that will return no results to the user, since there are no studies that have both that facet and the text 'Smith'.  However, the text match "Smith" will find two studies.  I believe your change means that we will record the search as having two text matches, even though it returned none to the user.  I *think* that's desirable or at least "good enough for now", but just wanted to clarify

app/controllers/api/v1/search_controller.rb
```diff
@@ -515,22 +522,59 @@ def self.escape_terms_for_regex(term_list:)
       def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, query_context:)
         case query_context
         when :keyword
+          puts('in keyword')
```
eweitz, line 525:
```suggestion
```

eweitz (approved):
Thanks for addressing that issue, and incorporating the suggestions!

ehanna4:
@devonbush  oh okay I think actually it will report nothing for match data if the facet filter is used, regardless if there's a match or not

only "pure" text searches have match data

Screen Shot 2022-03-24 at 9 44 36 AM
Screen Shot 2022-03-24 at 9 43 31 AM
Screen Shot 2022-03-24 at 9 44 07 AM

devonbush:
thanks for clarifying!

devonbush (approved):

PR #1433, opened 2022-03-28
fixing auth code analytics capture (SCP-4207)

devonbush:
I think the recent extra instrumentation added to azul files in search bulk download created this bug in single-study bulk download.  I've increased the amount of nil-safing in the analytics methods, and also updated the call to generateAuthCode from the single-study page to include the list of file ids

TO TEST:
1. sign in as a user without direct google bucket access (e.g. not the study owner), so e.g. your personal gmail account
2. open a study, and go to download
3. press 'bulk download'
4. follow instructions, and confirm you can download the study files locally.

### DIFF
--- a/app/javascript/lib/scp-api-metrics.js
+++ b/app/javascript/lib/scp-api-metrics.js
@@ -220,7 +220,7 @@ export function getLogPlotProps() {
 /**
  * Extract the number of Azul files that have been chosen for download
  */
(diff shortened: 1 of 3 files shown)


ehanna4 (approved):
Thanks for catching this! I tested with development and saw the error on bulk download and then tried with this branch and there was no error 👍

PR #1436, opened 2022-03-29
Add quotes around items in error messages (SCP-4198)

ehanna4:
As seen here: https://sentry.io/organizations/broad-institute/issues/2685546838/?project=1424198&query=is%3Aunresolved

Screen Shot 2022-03-29 at 2 47 14 PM

The missing annotation is an empty string but it's unclear by the current message. This PR adds quotes around the value so that it is more clear what the missing annotation is. This also adds quotes around other values that could potentially be blank or would generally benefit from the indication that quotes adds.


lib/generic_profiler.rb
```diff
@@ -205,21 +205,21 @@ def self.profile_expression(study_accession, genes: nil, plot_type: 'violin', co
   # validate arguments for :profile_clustering and :profile_expression
   def self.validate_cluster_args(study_accession)
     study = Study.find_by(accession: study_accession)
-    raise ArgumentError.new("#{study_accession} is not valid") if study.nil?
+    raise ArgumentError.new("'#{study_accession}' is not valid") if study.nil?
     cluster = study.default_cluster
     annotation = study.default_annotation
-    raise ArgumentError.new("#{study_accession} cannot visualize clusters") if (cluster.nil? || annotation.nil?)
+    raise ArgumentError.new("'#{study_accession}' cannot visualize clusters") if (cluster.nil? || annotation.nil?)
```
eweitz, line 211:
```suggestion
    raise ArgumentError.new("\"#{study_accession}\" cannot visualize clusters") if (cluster.nil? || annotation.nil?)
```

lib/generic_profiler.rb
```diff
@@ -205,21 +205,21 @@ def self.profile_expression(study_accession, genes: nil, plot_type: 'violin', co
...
   end
 
   # validate arguments for :profile_expression
   def self.validate_expression_args(study, genes, plot_type, consensus)
-    raise ArgumentError.new("#{study.accession} has no gene expression data") if study.genes.empty?
-    raise ArgumentError.new("#{plot_type} is not a valid plot type") if !EXPRESSION_PLOT_TYPES.include?(plot_type)
-    raise ArgumentError.new("#{consensus} is not a valid consensus") if !CONSENSUS_TYPES.include?(consensus)
+    raise ArgumentError.new("'#{study.accession}' has no gene expression data") if study.genes.empty?
+    raise ArgumentError.new("'#{plot_type}' is not a valid plot type") if !EXPRESSION_PLOT_TYPES.include?(plot_type)
+    raise ArgumentError.new("'#{consensus}' is not a valid consensus") if !CONSENSUS_TYPES.include?(consensus)
```
eweitz, line 219:
```suggestion
    raise ArgumentError.new("\"#{study.accession}\" has no gene expression data") if study.genes.empty?
    raise ArgumentError.new("\"#{plot_type}\" is not a valid plot type") if !EXPRESSION_PLOT_TYPES.include?(plot_type)
    raise ArgumentError.new("\"#{consensus}\" is not a valid consensus") if !CONSENSUS_TYPES.include?(consensus)
```

lib/generic_profiler.rb
```diff
@@ -205,21 +205,21 @@ def self.profile_expression(study_accession, genes: nil, plot_type: 'violin', co
...
     if genes.present?
       names = extract_gene_names(genes)
-      raise ArgumentError.new("#{genes} not present in #{study.accession}") if study.genes.where(:name.in => names).empty?
+      raise ArgumentError.new("'#{genes}' not present in '#{study.accession}'") if study.genes.where(:name.in => names).empty?
```
eweitz, line 222: Quoting `study.accession` strikes me as more awkward than worthwhile here.

```suggestion
      raise ArgumentError.new("\"#{genes}\" not present in #{study.accession}") if study.genes.where(:name.in => names).empty?
```

test/integration/study_validation_test.rb
```diff
@@ -342,7 +342,7 @@ class StudyValidationTest < ActionDispatch::IntegrationTest
     until  ['parsed', 'failed'].include? uploaded_matrix.parse_status  do
       puts "After #{seconds_slept} seconds, #{new_matrix} is #{uploaded_matrix.parse_status}."
       if seconds_slept >= max_seconds_to_sleep
-        raise "Sleep timeout after #{seconds_slept} seconds when waiting for parse of #{new_matrix}."
+        raise "Sleep timeout after #{seconds_slept} seconds when waiting for parse of '#{new_matrix}'."
```
eweitz, line 345:
```suggestion
        raise "Sleep timeout after #{seconds_slept} seconds when waiting for parse of \"#{new_matrix}\"."
```

eweitz (approved):
I like the more ambitious scope here! 

Non-blocking suggestions:

* Use double-quotes instead of single-quotes in user-facing text.  They're less ambiguous, e.g. an empty string enclosed by two single quotes can be hard or (depending on font) impossible to visually distinguish from one double quote.  In other words, '' can be very easily confused with ", whereas "" is visually unambiguous.  Single-quotes are easier to write in interpolated Ruby and easier to read for engineers, but I think readability for users is much more important.

* When a value is quoted, don't also precede it with a colon.  For example, messages like `Invalid accession: Foo` and `Invalid accession "Foo"` are preferable to `Invalid accession: "Foo"` -- using either quotes or the colon is adequate, but using both is extraneous.

* Omit quotes when the value is unlikely to be missing or invalid in form.  Quotes in such cases seem more awkward than clarifying.

These things are somewhat a matter of stylistic taste, so I defer to you.  If you find the suggestions persuasive, I've tried to make them trivial to accept via the "Add suggestion to batch" button.

devonbush:
Test failures are unrelated Azul failures (See SCP-4218)

PR #1437, opened 2022-03-30
fixing azul test suite to reflect updated fields (SCP-4218)

devonbush:
I'm guessing that HCA removed the 'source' field from their responses (which we were not using anyway).  We use the 'fileSource' field to determine whether a study file is an analysis file, so I've updated the test to check for that field

TO TEST: 

1. run `rails test test/integration/hca_azul_client_test.rb`
2. confirm passage


devonbush:
Yeah, I don't see any version numbers in the base URL in `hca_azul_client` not sure if they make versions available (and if they did, if we could trust that they would work)

PR #1435, opened 2022-03-29
Fix single-gene Explore view when default annotation is numeric (SCP-4200)

eweitz:
This fixes a broken "Distribution" display for the single-gene Explore view when default annotation is numeric.  Showing distributions like violin plots only makes sense for categorical annotations.  Now for this case, we show an annotated scatter plot as expected, and no distribution visualization.

To test:
* Sign in to your local SCP
* Go to a study that has numeric annotations, e.g.  the synthetic study "Male mouse brain"
* Go to Settings, then click "View options", change default annotation to "Intensity", click "Update settings"
* Refresh page, go to Explore tab, search a gene -- e.g. Adcy5

Expected view:

Expected_single-gene_Explore_view_when_default_annotation_is_numeric__SCP_2022-03-29

This satisfies part of SCP-4200.   A fix for the related situation in global gene search will follow in another PR.

PR #1440, opened 2022-03-31
Fix global gene search for studies with default numeric annotation (SCP-4200)

eweitz:
This fixes the display of results for global gene search when the study has a default annotation of type "numeric".

Previously, likely for at least 1-2 years, an error was displayed for such results.  Now, an annotated scatter plot is shown, similar to what's shown in gene search for the Explore tab in Study Overview (#1435).

Here's how it looks:

https://user-images.githubusercontent.com/1334561/161144590-a0d6c4f2-6c27-4f77-9566-69766fb9f524.mov

This satisfies SCP-4200.

### DIFF
--- a/app/javascript/components/search/genes/StudyGeneExpressions.jsx
+++ b/app/javascript/components/search/genes/StudyGeneExpressions.jsx
@@ -22,7 +22,7 @@ import {
  */
 function getBasicAnnotations(study) {
   let basicAnnotationList = null
-  if (study.annotation_list) {
+  if (study.annotation_list?.annotations.length > 0) {
     basicAnnotationList = study.annotation_list
     basicAnnotationList.annotations = basicAnnotationList.annotations.filter(a => a.scope !== 'user')
   }
--- a/app/javascript/components/search/genes/StudyGeneExpressions.jsx
+++ b/app/javascript/components/search/genes/StudyGeneExpressions.jsx
@@ -20,7 +20,8 @@ export default function StudyGeneExpressions({ study }) {
   const [clusterParams, setClusterParams] = useState(_clone(emptyDataParams))
   const [annotationList, setAnnotationList] = useState(study.annotation_list ?? null)
   let controlClusterParams = _clone(clusterParams)
-
+const defaultAnnotation = getAnnotationFromIdentifier(study.default_annotation)
+const defaultAnnotationIsNumeric = defaultAnnotation?.type === 'numeric'
   if (annotationList && !clusterParams.cluster) {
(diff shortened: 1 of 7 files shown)


app/javascript/components/search/genes/StudyGeneExpressions.jsx
```diff
@@ -17,7 +18,7 @@ import { getDefaultClusterParams, getAnnotationValues, emptyDataParams } from '~
   */
 export default function StudyGeneExpressions({ study }) {
   const [clusterParams, setClusterParams] = useState(_clone(emptyDataParams))
-  const [annotationList, setAnnotationList] = useState(null)
+  const [annotationList, setAnnotationList] = useState(study.annotation_list ?? null)
   let controlClusterParams = _clone(clusterParams)
```
devonbush, line 23: Add a `getAnnotationFromIdentifier` helper method to cluster-utils.js that mirrors the existing `getIdentifierFromAnnotation`, and then you can do
```suggestion
const defaultAnnotation = getAnnotationFromIdentifier(study.default_annotation)
const defaultAnnotationIsNumeric = defaultAnnotation?.type === 'numeric'
```
For bonus points, update line 80 of ExploreTabRouter.jsx to use the new helper function

devonbush (approved):
I'd lean toward sending the whole default annotation identifier just because it seems cleaner than extracting a single boolean, given that we'll probably want to know other things about the default annotation later.  But I don't have my heart set on it :)

eweitz:
To give adequate review time, I'll keep this as originally scoped and open a separate PR when the other fix is ready.

ehanna4 (changes requested):
I think this change might have broken a previous fix for another type of error.
When I run a gene search on `development` 2 studies do not display a graph but instead display a small message explaining why no graph is shown. However, when I run this branch now I see a 'Something went wrong.' red error message
Before (on development branch):
Screen Shot 2022-04-08 at 10 42 59 AM
With this branch:
Screen Shot 2022-04-08 at 10 44 12 AM
Screen Shot 2022-04-08 at 10 41 55 AM

eweitz:
Good catch, @ehanna4!  Fixed in 8eb26f2.  The fix now accounts for studies that lack visualizations.

jlchang (approved):
Functional review: 
- confirmed annotated scatter generated for studies with default annotation of TYPE numeric 
- confirmed no plot generated for studies with gene expression that have no cluster file and instead the appropriate message "This study does not have cluster data to support visualization in the portal" is conveyed.

ehanna4 (approved):
Nice, confirmed previous bug is gone and no new issues 👍

PR #1441, opened 2022-04-06
hiding sec-ch-ua header (SCP-4249)

devonbush:
this is confirmed to work in production, where this change was made to the config file live.
This was done after several failed attempts to use `proxy_set_header`, but then https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_set_header indicated that since we are using the passenger nginx module, `passenger_set_header` was instead the correct parameter.

TO TEST:
 1. go to production SCP in chrome
 2. click around, confirm random 403s are not recieved

### DIFF
--- a/webapp.conf
+++ b/webapp.conf
@@ -34,6 +34,9 @@ server {
     proxy_set_header  X-Forwarded-Proto $scheme;
     proxy_set_header  Host $host;
     proxy_set_header  X-Forwarded-Host $host;
+    # This address repeated bugs we've found with Chrome using non-standard quote escaping in this parameter
+    # see e.g. https://docs.google.com/document/d/1TTIGgu82qLJqGt130DlxWN5lsP7Wzo6FwuJSn5O_J_I
+    passenger_set_header sec-ch-ua "";
 
     # The following deploys your Ruby/Python/Node.js/Meteor app on Passenger.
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Validated in production on 4/5 that this change resolved the issue where Chrome users would  hit "Authentication Failed" and/or primitive 403 page.

Kudos to Devon for finding a setting that would filter out the problematic header!

ehanna4 (approved):
Nice! And appreciate the comment

eweitz (approved):
Great find!  I confirm this works, and we also see the fix reflected in [a related Mixpanel report](https://mixpanel.com/s/2xnApB).  This minimal test case also passes -- no `status: 403 Forbidden` results:

```
for i in `seq 1 10`; do curl -sI 'https://singlecell.broadinstitute.org/single_cell/api/v1/search/facets' -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="99", "Google Chrome";v="99"' | grep status && sleep 1; done
```

Naively, it seems like peculiar API design that Passenger would discard upstream Nginx directives that don't necessarily conflict with downstream, lower-level Passenger directives.

My biggest outstanding question: why did this bug trigger _inconsistently_?

PR #1442, opened 2022-04-06
HOTFIX - fix cluster renaming bug (SCP-4255)

devonbush:
Renaming a study file after creation would succeed, but not rename the underlying clusterGroup, which means the UI will not be updated.  I would like to merge/release this tonight to unblock Stefano's study creation.  The bug was on line 364
`cluster = ClusterGroup.find_by(study_file_id: safe_file_params[:_id])`
This never found the existing cluster group, since earlier in the method we delete the safe_file_params[:_id] for security.  

I've also included some general cleanup of the API method to make it not use instance variables

TO TEST:
1. upload a cluster file
2. view the cluster in the UX
3. in the upload wizard, change the name of the cluster
4. view the cluster in the UX, and confirm the name change propagated

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -296,6 +296,7 @@ module Api
       # if the save is successful and includes a complete upload, will send it to firecloud
       # and kick off a parse job if requested
       def perform_update(study_file)
+        study = study_file.study
         safe_file_params = study_file_params
         is_chunked = false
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Functional review: was able to confirm propagation of new cluster name (did not upload cluster file as Terra was having issues).

eweitz (approved):
Code looks good.

devonbush:
Build failure is in unrelated test due to Terra workspace deletion being unavailable

PR #1444, opened 2022-04-07
Fix BSON::String::IllegalKey for bulkdownload Azul files (SCP-4201)

ehanna4:
This fixes a bug where an Azul study name contained a '.' which is an illegal key value for a hash in MongoDB. This change will update the azul files hash to be encoded to a JSON string before persisting to the DB thereby avoiding the hash key constraints of MongoDB entirely. This is closely modeled after the fix that was done to correct this issue for custom color annotations.

To test:
- Pull this branch
- Sign into SCP
- Do a search for 'Interneuron' (this will give a result of an Azul study that has a period at the end of it's name)
- click download
- confirm you are able to run the curl command to download this study.

### DIFF
--- a/app/models/download_request.rb
+++ b/app/models/download_request.rb
@@ -7,7 +7,40 @@ class DownloadRequest
 
   field :auth_code, type: String
   field :file_ids # Mongo ids of study files to download
-  field :tdr_files, type: Hash, default: {} # Hash of TDR project shortnames to arrays of access urls
-  field :azul_files, type: Hash, default: {} # Hash of Azul file summaries (Project Manifests, analysis_files, etc)
+  field :tdr_files, type: String # Hash of TDR project shortnames to arrays of access urls
+  field :azul_files, type: String # Hash of Azul file summaries (Project Manifests, analysis_files, etc)
   field :user_id # User making the request
+
+  before_save :sanitize_hashes
+
+  ALLOWED_TRANSFORMS = %i[encode decode]
+
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded
+  # in the way the DB can accept
+  def sanitize_hashes 
+    if self.azul_files.is_a?(Hash)
+    self.tdr_files =  self.class.transform_files(tdr_files, :encode)
+      self.azul_files = self.class.transform_files(azul_files, :encode)
+    end
+  end
+  
+  # decode JSON-ified tdr files back into their original hash format
+  def decoded_tdr_files
+    self.class.transform_files(tdr_files, :decode)
+  end
+
+  # decode JSON-ified azul files back into their original hash format
+  def decoded_azul_files
+    self.class.transform_files(azul_files, :decode)
+  end
+
+  # helper to transform a hash to/from JSON for storage in the DB
+  # transformation options are encode which will result in JSON-ifing the hash or
+  # decode which will return the JSON-ified files back to thier original hash format
+  # This is necessary to work around MongoDBs constraints on '.' and '$' in hash keys
+  def self.transform_files(file_data, transform = :encode)  
+    raise ArgumentError, "#{transform} is not a valid transform" unless ALLOWED_TRANSFORMS.include?(transform)
+    file_data = transform.eql?(:encode) ? ActiveSupport::JSON.encode(file_data) : ActiveSupport::JSON.decode(file_data.gsub('=>', ':'))
+    return file_data
+    end
 end
--- a/db/migrate/20220401103323_encode_azul_files_names.rb
+++ b/db/migrate/20220401103323_encode_azul_files_names.rb
@@ -1,0 +1,18 @@
+class EncodeAzulFilesNames < Mongoid::Migration
+    def self.up
+      download_reqs = DownloadRequest.where(:azul_files.ne => nil)
+      download_reqs.each do |req|
+        encoded_azul_files = DownloadRequest.transform_files(req.azul_files, :encode)
+        req.update(azul_files: encoded_azul_files)
+      end
+    end
+  
+    def self.down
+    download_reqs = DownloadRequest.where(:azul_files.ne => nil)
+    download_reqs.each do |req|
+        decoded_azul_files = DownloadRequest.transform_files( req.azul_files, :decode)
+        req.update(azul_files: decoded_azul_files)
+      end
+    end
+  end
+  
--- a/test/models/download_request_test.rb
+++ b/test/models/download_request_test.rb
@@ -1,0 +1,33 @@
+require 'test_helper'
+
+class DownloadRequestTest < ActiveSupport::TestCase
+
+  before(:all) do
+    @azul_hash = {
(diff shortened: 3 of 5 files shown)


app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+
+  before_save :sanitize_hashes
+
+  ALLOWED_TRANSFORMS = %i[encode decode]
+
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded
+  # in the way the DB can accept
+  def sanitize_hashes 
+    if self.azul_files.is_a?(Hash)
+    self.tdr_files =  self.class.transform_files(tdr_files, :encode)
```
eweitz, line 22: Indent, remove extra space:

```suggestion
      self.tdr_files = self.class.transform_files(tdr_files, :encode)
```

app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+    self.class.transform_files(tdr_files, :decode)
+  end
+
+  # decode JSON-ified azul files back into their original hash format
+  def decoded_azul_files
+    self.class.transform_files(azul_files, :decode)
+  end
+
+  # helper to transform a hash to/from JSON for storage in the DB
+  # transformation options are encode which will result in JSON-ifing the hash or
```
eweitz, line 38:
```suggestion
  # transformation options are encode which will result in JSON-ifying the hash or
```

db/migrate/20220401103323_encode_azul_files_names.rb
```diff
@@ -0,0 +1,18 @@
...
+      download_reqs.each do |req|
+        encoded_azul_files = DownloadRequest.transform_files(req.azul_files, :encode)
+        req.update(azul_files: encoded_azul_files)
+      end
+    end
+  
+    def self.down
+    download_reqs = DownloadRequest.where(:azul_files.ne => nil)
+    download_reqs.each do |req|
+        decoded_azul_files = DownloadRequest.transform_files( req.azul_files, :decode)
```
eweitz, line 13:
```suggestion
        decoded_azul_files = DownloadRequest.transform_files(req.azul_files, :decode)
```

app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded
+  # in the way the DB can accept
+  def sanitize_hashes 
+    if self.azul_files.is_a?(Hash)
+    self.tdr_files =  self.class.transform_files(tdr_files, :encode)
+      self.azul_files = self.class.transform_files(azul_files, :encode)
+    end
+  end
+  
+  # decode JSON-ified tdr files back into their original hash format
```
eweitz, line 27:
```suggestion
  # decode JSON-ified TDR files back into their original hash format
```

eweitz (approved):
Code looks good!  I note some spacing nits, no blockers.

jlchang (approved):
Functional review: confirmed able to generate curl command and curl command successfully downloads file (project manifest)

app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+  field :azul_files, type: String # String representation of a hash of Azul file summaries (Project Manifests, analysis_files, etc)
   field :user_id # User making the request
+
+  before_save :sanitize_hashes
+
+  ALLOWED_TRANSFORMS = %i[encode decode]
+
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded
+  # in the way the DB can accept
+  def sanitize_hashes
```
devonbush, line 20:
```suggestion
  def stringify_hashes 
```

app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+
+  # decode JSON-ified azul files back into their original hash format
+  def decoded_azul_files
+    self.class.transform_files(azul_files, :decode)
+  end
+
+  # helper to transform a hash to/from JSON for storage in the DB
+  # transformation options are encode which will result in JSON-ifying the hash or
+  # decode which will return the JSON-ified files back to thier original hash format
+  # This is necessary to work around MongoDBs constraints on '.' and '$' in hash keys
```
devonbush, line 40: This transform method is overkill now that we're just stringifying.  The `azul_files_as_hash` methods can just call `ActiveSupport::JSON.decode` directly, and `stringify_hashes` can call `ActiveSupport::JSON.encode` directly

app/models/download_request.rb
```diff
@@ -7,7 +7,40 @@ class DownloadRequest
...
+
+  before_save :sanitize_hashes
+
+  ALLOWED_TRANSFORMS = %i[encode decode]
+
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded
+  # in the way the DB can accept
+  def sanitize_hashes 
+    if self.azul_files.is_a?(Hash)
+      self.tdr_files = self.class.transform_files(tdr_files, :encode)
```
devonbush, line 22: BLOCKER: the tdr_files needs a separate `if` check to make sure it doesn't get double-encoded in a case where a DownloadRequest gets updated

test/models/download_request_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'test_helper'
+
+class DownloadRequestTest < ActiveSupport::TestCase
+
+  before(:all) do
```
devonbush, line 5: no need to have this is a before(:all) block since it is only used in one test, and does not save anything to the DB.  Might as well keep the tests self-contained and move this setup to inside the test itself

test/models/download_request_test.rb
```diff
@@ -0,0 +1,33 @@
...
+            'project_id' => 'hca_project_id'
+          }
+        ],
+        'Fake.HcaS.tu.d.y.2' => []
+      }
+      
+    @encoded_azul_hash = "{\"Fake.HCA.Study.1\":[{\"source\":\"hca\",\"count\":1,\"upload_file_size\":10485760,\"file_format\":\"loom\",\"file_type\":\"analysis_file\",\"accession\":\"FakeHCAStudy1\",\"project_id\":\"hca_project_id\"}],\"Fake.HcaS.tu.d.y.2\":[]}"
+
+  end
+
```
devonbush, line 24: BLOCKER: This test will be much more robust if it confirms a DownloadRequest can be saved to the database, and then read back

e.g.
download_req = DownloadRequest.new({ ..., azul_files: { <<hash >>}})
download_req.save!
assert_equal encoded_azul_hash, download_req.azul_files_as_hash

devonbush: (and yup, I realize ClusterFileInfoTest doesn't have such a test -- that was an oversight by me :P )

devonbush (changes requested):

devonbush (approved):
Nice!

PR #1445, opened 2022-04-09
adding exception for sync tool directory listings (SCP-4239)

devonbush:
Previously, any group of 10/20+ files with the same extension would get made into a 'block' in sync tool, which means they could not be ingested, but only bulk uploaded as misc/other.  This adds an exception to that treatment for non-sequence data files in the root directory.  This will be needed for Jonah to upload his 100+ clusters via sync tool.  For now, this is definitely in the 'quick fix' category.  Once Jon returns, who knows more about the use cases for DirectoryListing, we can revisit this and expand/refine the exception, and add tests around it.

TO TEST:
1. for an existing study, add 20+ cluster files to the workspace bucket (they can all be copies of the same file with different names)
2. Visit sync tool for that study
3. confirm the files appear individually, and are able to be specified as cluster files

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -769,7 +769,9 @@ module Api
           file_type = DirectoryListing.file_type_from_extension(file.name)
           directory_name = DirectoryListing.get_folder_name(file.name)
           if @file_extension_map.has_key?(directory_name) && !@file_extension_map.dig(directory_name, file_type).nil? &&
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good, though duplicative.  How feasible would it be to consolidate it?

jlchang (approved):
Functional review: 
Confirmed upload of 20 csv files were listed in sync indivdually and could be assigned file type of cluster.

ehanna4 (approved):
I'm unable to reproduce the issue that triggered this fix, but I am able to successfully upload 21 cluster files through sync individually.

PR #1446, opened 2022-04-11
Log search result selection for global study search (SCP-4238)

eweitz:
This logs when the user selects a result from global study search.  It aims to help improve search through analytics.

Previously, when the user selected a search result, we would merely log a `click:link` event.  This had sparse information that impeded quantitative UX research on how users interact with our search interface.

Now, a much richer event is logged: `select-search-result`.  It naturally contains data on which result was selected -- including its rank in the list of search results, how many searches were done before selecting the result, and whether it was sourced from SCP or Azul / Human Cell Atlas, among other properties about the selection.  The new event also logs data about the query and larger set of results.  

The number of searches done before selection is part of click-through rate (CTR), a general search product quality indicator.

Further refinements will help.  "To dos" include making facet query data easier to analyze, and eliminating somewhat duplicative logging in both this and related pre-existing instrumentation.

To test:
* Go to home page
* Open Network panel in Chrome DevTools
* At upper-left in panel, filter requests to `bard method:POST`
* Search for a term by typing into the keyword text input
* Click on an SCP study result _and immediately press the "esc" key_
* Confirm in Network panel that a `select-search-result` event is logged to Bard / Mixpanel
* Click on an HCA / Azul result (do another keyword search if needed)
* Confirm in Network panel that a `select-search-result` event is logged
* Execute a faceted search
* Click on a result
* Confirm `select-search-result` is logged

This satisfies SCP-4238.


app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+ * properties about other aspects of the search -- e.g. the full result set
+ * and query -- by prefixing the latter with a "namespace", i.e. `results:`.
+ *
+ * @param {Object} study Selected result
+ * @param {Object} logProps Other properties to log
+ *
+ * TODO (SCP-4256): Refine logged properties for search result selection
+ * TODO (SCP-4257): Eliminate duplicate default/custom search event logging
+ */
+export function logSelectSearchResult(study, logProps={}) {
```
ehanna4, line 191: I wonder if this mixpanel-y specific function fits best in this `StudySearchResult` file or if it would make more sense to put it in the metrics file?

app/javascript/lib/scp-api-metrics.js
```diff
@@ -71,12 +74,34 @@ function getFriendlyFilterListByFacet(facets) {
   return filterListByFacet
 }
 
+/** Format parts of query entered by the user to ease analysis in Mixpanel */
+export function formatQueryUserInput(searchedTerms, searchedGenes, searchedFacets) {
```
devonbush, line 78: `getLogPropsForSearchParams()` might be a more descriptive name.  At first I thought this was going to generate a url/query string

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+  if (logProps['results:matchByData'] !== null) {
+    Object.entries(logProps['results:matchByData']).forEach(([key, value]) => {
+      logProps[`results:${key}`] = value
+    })
+  }
+  delete logProps['results:matchByData']
+
+  let refinedLogProps = {}
+  Object.entries(logProps).forEach(([key, value]) => {
+    // Counts of "foo" in Mixpanel are conventionally structured as `numFoos`.
```
devonbush, line 233: there's only 3 of these properties that I see, handling them manually will make this code less error-prone.

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+    if (key.slice(-6) === '_count') {
+      const namespace = key.includes('results:') ? 'results:' : ''
+      // Convert e.g. `gene_count` to `numGenes`
+      key = `${namespace}num${key[0].toUpperCase() + key.slice(1).replace('_count', '') }s`
+    }
+    if (key.slice(0, 13) === 'results:total') {
```
devonbush, line 239: I understand how `_count` isn't really JS-ified, and so changing those might be worth the effort.  But we called these properties `totalX` and not `numX` for a reason--it distinguished that they were the total available, rather than just the number returned.  I'm not sure what benefit we get by renaming them, and why that's worth the possible confusion.  I guess we could call them `numTotalPages` if it makes navigating mixpanel easier to have all the count-y variables alphabetically close? 
And again, there's not many of them, I'd lean towards writing them out

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+      // Convert e.g. `totalPages` to `numPages`
+      key = `results:num${key.replace('results:total', '')}`
+    }
+    refinedLogProps[key] = value
+  })
+  refinedLogProps = camelcaseKeys(refinedLogProps)
+
+  const formattedQueryUserInput = formatQueryUserInput(
```
devonbush, line 247: `formatted` to me implies 'string'.  `queryMixpanelProps` might be a more specific name

app/javascript/components/search/results/StudySearchResult.jsx
devonbush: Concur.  We now have a few hundred lines of search-specific analytics code, so it probably deserves it's own  file

devonbush (approved):
this is some intense logging code!  I made some suggestions, but this is firmly in your domain, so I defer to you on all the style matters.

jlchang (approved):
Performed functional review and confirmed I can see "select-search-result" events for both SCP and Azul search results.

Thanks Eric for walking me through interpreting the properties for this new event - the data is, indeed, very rich!

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: I like the idea here; I went with `searchQueryLogProps` in 8d57e81.

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: Yup, analysis UX consistency and simplicity are the main benefit of renaming these.  I've gone with `numTotalFoo` in cases of `totalFoo` as you suggest, and now write them out; 8d57e81.

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: Done in 8d57e81.  It feels less robust, but more understandable.

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+    {
+      studyAccession: study.accession,
+      context: 'study',
+      scope: 'global'
+    },
+    study,
+    logProps
+  )
+  delete logProps.accession // Redundant with more-conventional studyAccession
+
```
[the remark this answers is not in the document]
eweitz: This seems perhaps subjective, yet definitely worth a deeper look for consistency with other logging patterns.  I've added a re-evaluation along the lines you suggest as part of a downstream ticket.

app/javascript/lib/scp-api-metrics.js
eweitz: Good point, I've updated this to `getSearchQueryLogProps` in 8d57e81.

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: Agreed.  High-level reorganization makes me nervous so close to freeze, so I've added this to a downstream ticket.

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -170,17 +175,109 @@ function studyTypeBadge(study) {
...
+    }
+    logProps[`results:${key}`] = value
+  })
+  delete logProps.results // Remove nested property; flattened above
+
+  // Number of searches done before this result was selected.
+  // This is needed to easily analyze changes in click-through rate (CTR).
+  logProps.numSearches = numSearches
+
+  // We don't log study name, like Terra doesn't log workspace name, per
```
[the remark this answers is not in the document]
eweitz: Both good suggestions!  Done in 8d57e81.

eweitz:
For reference, after discussing these log properties, my sense is that grouping query-related properties into `results:` doesn't adequately distinguish them.  I've changed those to be prefixed with `query:` instead, for better clarity.

PR #1451, opened 2022-04-13
fixing legend counts for hidden traces (SCP-4259)

devonbush:
Hiding a trace in the legend no longer makes the number of points displayed go to zero.  this also opens up substantial performance optimizations, as using the visibility attribute rather than cell removal means that we can show/hide traces without doing a full recompute of all traces.  Likewise, we can reorder traces without doing a recompute.  Those optimizations are not part of this PR, as we want to get this out before release.

TO TEST:
  1. load study explore tab with cluster
  2. click show/hide traces
  3. confirm the number of points next to each legend label does not change when they are shown/hidden

### DIFF
--- a/app/javascript/lib/plot.js
+++ b/app/javascript/lib/plot.js
@@ -114,35 +114,24 @@ PlotUtils.filterTrace = function({
     expFilterMax = expMin + totalRange * expressionFilter[1]
   }
 
-  const labelNameHash = {}
-  if (isHidingByLabel) {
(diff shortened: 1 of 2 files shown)


ehanna4 (approved):
👍 tested locally and saw how in development the hideall sets the annotations to 0 and with this branch it does not. Code change looks reasonable and casual, overall looks good to me!

eweitz (approved):

PR #1452, opened 2022-04-13
scatter plot performance improvements (SCP-4259)

devonbush:
No functional changes.  These are the performance improvements unlocked by switching from cell-based hiding to trace-based hiding in  #1451 .  Instead of recomputing everything when a label is hidden or brought to the front, now we just reorder or modify the traces in-place.  I think this also makes the code more legible, as now we have sorting and hiding split out into separate functions

For a 1MM cell study, in my local environment, this reduced the time to hide a trace from ~0.8sec to ~0.4sec.  Likewise bringing traces to the front via hovering got faster as well. 

TO TEST:
1. load a cluster scatter plot
2. click around, hovering, hiding, showing traces
3. confirm behavior is as expected.

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -242,8 +242,6 @@ function RawScatterPlot({
       PlotUtils.sortTraces(plotlyTraces, activeTraceLabel)
       Plotly.react(graphElementId, plotlyTraces, scatterData.layout)
     }
-    // look for updates of individual properties, so that we don't rerender if the containing array
-    // happens to be a different instance
   }, [activeTraceLabel])
 
   // Handles window resizing
@@ -253,8 +251,6 @@ function RawScatterPlot({
     if (scatterData && !isLoading) {
       resizePlot()
     }
-    // look for updates of individual properties, so that we don't rerender if the containing array
-    // happens to be a different instance
   }, [widthAndHeight.height, widthAndHeight.width])
 
   // Handles Plotly `data` updates, e.g. changes in color profile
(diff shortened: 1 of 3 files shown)


app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -221,8 +221,30 @@ function RawScatterPlot({
     }
     // look for updates of individual properties, so that we don't rerender if the containing array
     // happens to be a different instance
-  }, [hiddenTraces.join(','), Object.values(editedCustomColors).join(','),
-    Object.values(customColors).join(','), activeTraceLabel, expressionFilter.join(',')])
+  }, [Object.values(editedCustomColors).join(','),
+    Object.values(customColors).join(','), expressionFilter.join(',')])
```
eweitz, line 225: Easier reading:

```suggestion
  }, [
    Object.values(editedCustomColors).join(','),
    Object.values(customColors).join(','), 
    expressionFilter.join(',')
  ])
```

eweitz (approved):
Good stuff!  I tested this locally and things work, and seem snappy.

`perfTime` instrumentation for trace hiding / showing would be wonderful to help track performance in this over time -- I can't easily tell if it's faster or slower -- but I don't feel strongly about it.

app/javascript/lib/plot.js
```diff
@@ -209,18 +228,7 @@ PlotUtils.getColorForLabel = function(label, customColors={}, editedCustomColors
 
 /** Returns an array of labels, sorted in the order in which they should be plotted (last is 'on top') */
 PlotUtils.getPlotSortedLabels = function(countsByLabel, activeTraceLabel) {
```
bistline, line 230: Any reason why we need to now keep an empty function?

bistline (approved):
Looks good, though curious why `PlotUtils.getPlotSortedLabels` is now empty, and apparently unused.

app/javascript/lib/plot.js
devonbush: good catch, refactoring remnant, will remove

devonbush:
Test failure was Unrelated:  firecloud_client test -- failure to delete a workspace.

PR #1455, opened 2022-04-13
 Make study usage information visible (SCP-4268)

devonbush:
As shown in demo, a page, linked from Study Settings, that shows basic usage graphs for the given study
image

image

This pulls the data via Mixpanel API.  Accordingly, a development and production service account were created, and the corresponding secrets were stored in Vault.  Once this is approved, and prior to merge, I'll handle updating everyone's vault secrets accordingly.

This also adds some functionality to the LoadingSpinner component so it can be used in a more streamlined way in some components--it optionally takes children and an `isLoading` parameter.  The LoadingSpinner component then takes care of the show/hide logic.

TO TEST:
1. You'll need the new mixpanel secret in your environment -- (this is just until we've put it in everyone's vault config):
             a.  run `vault read secret/kdux/scp/development/dbush/scp_config.json`
             b.  note the value for MIXPANEL_SECRET
             c.  Stop your rails server
             d.  run `export MIXPANEL_SECRET=<<the noted value>>`
             e.  restart rails server in the terminal where you ran the export command
          
2. go to the 'Settings' tab for a given study
3. click 'Usage stats' in the left. menu
4. Observe the usage stats appear

### DIFF
--- a/app/javascript/components/my-studies/StudyUsageInfo.jsx
+++ b/app/javascript/components/my-studies/StudyUsageInfo.jsx
@@ -5,11 +5,15 @@ import Plotly from 'plotly.js-dist'
 import { fetchStudyUsage } from '~/lib/scp-api'
 import LoadingSpinner from '~/lib/LoadingSpinner'
 import { supportEmailLink } from '~/lib/error-utils'
+import InfoPopup from '~/lib/InfoPopup'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faChevronUp, faChevronDown } from '@fortawesome/free-solid-svg-icons'
 
 /** display mixpanel stats for a given study */
 export default function StudyUsageInfo({ study }) {
   const [usageInfo, setUsageInfo] = useState(null)
   const [isLoading, setIsLoading] = useState(true)
+  const [showTable, setShowTable] = useState(false)
   const [viewsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
   const [downloadsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
   const studyAccession = study.accession
@@ -59,7 +63,7 @@ export default function StudyUsageInfo({ study }) {
     <h3>Study statistics</h3>
     {studyAccession}: <a href={`/single_cell/study/${studyAccession}`}>{ study.name }</a><br/>
     <div className="text-center">
-      Users per month viewing this study.<br/>
+      Users per month viewing this study. <InfoPopup content="Unique users per month visiting the study overview page"/><br/>
       <LoadingSpinner isLoading={isLoading}/>
     </div>
     <div
@@ -69,7 +73,7 @@ export default function StudyUsageInfo({ study }) {
     ></div>
     <br/><br/>
     <div className="text-center">
-      File download events<br/>
+      File download events <InfoPopup content="Clicks on the single-file download button.  Bulk download stats are tracked, but not available in this view"/><br/>
       <LoadingSpinner isLoading={isLoading}/>
     </div>
     <div
@@ -79,7 +83,12 @@ export default function StudyUsageInfo({ study }) {
     ></div>
     <br/><br/>
     <div className="form-terra">
-      { !isLoading && <table className="table-terra compressed">
+      <h5 onClick={() => setShowTable(!showTable)}>
+        <span className="action">
+          { showTable ? <FontAwesomeIcon icon={faChevronUp}/> : <FontAwesomeIcon icon={faChevronDown}/> }
+        </span> Tabular Data
+      </h5>
+      { !isLoading && showTable && <table className="table-terra compressed">
         <thead>
           <tr>
             <td>Month</td>
@@ -97,7 +106,7 @@ export default function StudyUsageInfo({ study }) {
       </table> }
     </div> <br/>
     <div className="text-center">
-      If you would like more information about views, downloads, or other statistics for your studies, contact us at { supportEmailLink }
+      If you would like more information about views, downloads, or other statistics for your study, contact us at { supportEmailLink }
     </div>
   </div>
 }
--- a/app/javascript/components/my-studies/StudyUsageInfo.jsx
+++ b/app/javascript/components/my-studies/StudyUsageInfo.jsx
@@ -1,0 +1,107 @@
+import React, { useEffect, useState } from 'react'
+import _uniqueId from 'lodash/uniqueId'
+import Plotly from 'plotly.js-dist'
+
+import { fetchStudyUsage } from '~/lib/scp-api'
+import LoadingSpinner from '~/lib/LoadingSpinner'
+import { supportEmailLink } from '~/lib/error-utils'
+
+/** display mixpanel stats for a given study */
+export default function StudyUsageInfo({ study }) {
+  const [usageInfo, setUsageInfo] = useState(null)
+  const [isLoading, setIsLoading] = useState(true)
+  const [viewsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
+  const [downloadsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
+  const studyAccession = study.accession
+  useEffect(() => {
+    fetchStudyUsage(study._id.$oid).then(response => {
+      setUsageInfo(response)
+      setIsLoading(false)
+    })
+  }, [study.accession])
+
+  const tableData = []
+
+  if (usageInfo) {
+    const viewData = usageInfo.pageViews.data
+    const viewsTrace = {
+      type: 'bar',
+      x: viewData.series,
+      y: viewData.series.map(x => {
+        return viewData.values[x] ? viewData.values[x] : 0
+      })
+    }
+    const viewsLayout = {
+      xaxis: { title: 'Month' },
+      yaxis: { title: 'Unique users' }
+    }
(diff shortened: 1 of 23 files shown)


test/js/study-usage.test.js
```diff
@@ -0,0 +1,56 @@
...
+    const mockPlotlyReact = jest.spyOn(Plotly, 'react')
+    mockPlotlyReact.mockImplementation(() => {})
+    const study = {
+      accession: 'SCP12',
+      name: 'Test study 12',
+      _id: { $oid: 'fakeId4' }
+    }
+    render(<StudyUsageInfo study={study}/>)
+    await waitForElementToBeRemoved(() => screen.getByTestId('study-usage-spinner'))
+    expect(fetchUsageInfo).toHaveBeenLastCalledWith('fakeId4')
```
eweitz, line 45: Deprecating MongoDB object identifiers from API endpoints simplifies tests, among those other benefits:

```suggestion
    }
    render(<StudyUsageInfo study={study}/>)
    await waitForElementToBeRemoved(() => screen.getByTestId('study-usage-spinner'))
    expect(fetchUsageInfo).toHaveBeenLastCalledWith('SCP12')
```

eweitz (changes requested):
Very neat, useful integration of Mixpanel for study owners and editors --  kudos!

I note a blocker around API identifiers, to implement a prior engineering strategy takeaway and help debuggability.  

It should be orthogonal, but if the app URL needing to be parameterized by MongoDB object ID instead of accession was a key part of that decision, then I've confirmed updating `set_study` in `app/controllers/studies_controller.rb` to the following would resolve that:

```
  def set_study
    @study = Study.find(params[:id])
    if !@study
      @study = Study.find_by(accession: params[:id])
    end
  end
```

We'd update the "Usage stats" link with the accession too in that case, of course.

app/javascript/components/my-studies/StudyUsageInfo.jsx
```diff
@@ -0,0 +1,129 @@
...
+  const [usageInfo, setUsageInfo] = useState(null)
+  const [isLoading, setIsLoading] = useState(true)
+  const [showTable, setShowTable] = useState(false)
+  const [viewsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
+  const [downloadsGraphElementId] = useState(_uniqueId('study-usage-graph-'))
+  const studyAccession = study.accession
+
+  /** get the usage information via API call */
+  useEffect(() => {
+    fetchStudyUsage(study._id.$oid).then(response => {
```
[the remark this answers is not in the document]
bistline: While I agree with Eric's point, there is somewhat of a gotcha here - because accessions are not _globally_ unique (they're only unique to a given instance), you can get pollution of stats across dev/staging instances.  I confirmed this by booting my local instance and looking at the stats for a clone of SCP1 from production (sNuc-seq study, Naomi's data).  In my local instance, this is SCP13.  When I looked at the stats, it registered views & downloads for April, which obviously haven't happened as I only just booted my server again today.

That being said, I think it's only an issue in local dev & staging, and should not affect production as this uses a different Mixpanel instance.  But we should at least be aware of the issue.

app/javascript/components/my-studies/StudyUsageInfo.jsx
```diff
@@ -0,0 +1,129 @@
...
+  const tableData = []
+
+  if (usageInfo) {
+    // generate the trace information and table information
+    const viewData = usageInfo.pageViews.data
+    const viewsTrace = {
+      type: 'bar',
+      x: viewData.series,
+      y: viewData.series.map(x => {
+        return viewData.values[x] ? viewData.values[x] : 0
```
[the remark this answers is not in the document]
bistline: Oh that's cool

app/javascript/components/my-studies/StudyUsageInfo.jsx
```diff
@@ -0,0 +1,129 @@
...
+    tableData.push(viewsTrace.y)
+
+    // sum together the single file clicks and the bulk downloads for the graph
+    const downloadData = usageInfo.fileDownloads.data
+    const downloadTableData = []
+    const bulkDownloadData = usageInfo.bulkDownloads.data
+    const bulkDownloadTableData = []
+    const downloadsTrace = {
+      type: 'bar',
+      x: downloadData.series,
```
bistline, line 56: Non-blocking: it would be nice if in the Plotly graph for the hover tootlips `Month` axis, we omit the day.  For instance, instead of saying `March 1, 2022`, it just said `March 2022`.  That might be because we're getting an actual date from Mixpanel for the `Month` axis.  There might be an opportunity to parse that out into a string and omit the day.  Just a thought.

app/javascript/components/my-studies/StudyUsageInfo.jsx
```diff
@@ -0,0 +1,129 @@
...
+    <div
+      className="scatter-graph"
+      id={downloadsGraphElementId}
+      data-testid={viewsGraphElementId}
+    ></div>
+    <br/><br/>
+    <div className="form-terra">
+      <h5 onClick={() => setShowTable(!showTable)}>
+        <span className="action">
+          { showTable ? <FontAwesomeIcon icon={faChevronUp}/> : <FontAwesomeIcon icon={faChevronDown}/> }
```
bistline, line 100: Non-blocking: Traditionally, I think the icons for "closed" is `faChevronRight` and "open" is `faChevronDown`.  I found the "up" chevron confusing there.

app/lib/mixpanel_client.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    response = RestClient::Request.execute(request)
+    result = JSON.parse(response.body)
+    if on_string.empty?
+      # simplify the response so that the consumer doesn't have to dig by the event name
+      result['data']['values'] = result['data']['values'][event]
+    end
+    result
+  end
+
```
bistline, line 51: Some blank space here, and we're missing a terminal newline.

config/application.rb
```diff
@@ -33,6 +33,10 @@ class Application < Rails::Application
 
     config.autoload_paths << Rails.root.join('lib')
 
+    # for all non-prod environments, use the development mixpanel API
+    config.mixpanel_service_account = 'scp_terra_dev.f25a4f.mp-service-account'
```
bistline, line 37: Just to confirm - these are not sensitive secrets?

bistline (approved):
Really like this - I'm sure study owners will appreciate it greatly.  Some small non-blocking suggestions, and a request for clarification on the Mixpanel additions to our configs.

app/javascript/components/my-studies/StudyUsageInfo.jsx
devonbush: Good point -- I'll leave it as is because I think there's some complexity around how partial months get handled.  And we'll definitely keep this in mind if we decide to put some more polish behind this page in the future

app/javascript/components/my-studies/StudyUsageInfo.jsx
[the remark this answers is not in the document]
eweitz: That's a keen observation!  We can distinguish staging and development by adding an `env` property with value `staging` or `development` to our logs, but your points stands at least _within_ different development environments.

We don't log the study's MongoDB object ID to Mixpanel for these events, IIRC, so using accession in this SCP API wouldn't affect that matter.  Less persuasively, but maybe also worth noting: which study property is queried in Mixpanel also seems somewhat orthogonal to the type of identifier used in the SCP API.

If we deem intra-development environment Mixpanel log pollution worth fixing, I'd be inclined to do so by adding a new property, perhaps something like `instance`, to all logged event.  This would help distinguish all events across instances within an environment, not just those for a specific study.

app/javascript/components/my-studies/StudyUsageInfo.jsx
devonbush: I don't have a strong opinion on this, and there's nothing in the Terra Style guide (https://www.sketch.com/s/cca3a834-985b-4165-bbae-21f615aebd2d) about it.  I just pulled this from the Upload Wizard design from Jerome.  Totally open to changing both, as long as we're consistent

config/application.rb
devonbush: correct, that's just the service account name.  the correspond  secret/password is in Vault

app/javascript/components/my-studies/StudyUsageInfo.jsx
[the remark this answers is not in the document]
devonbush: Inter-dev-environment pollution can also be addressed by breaking results out by Mixpanel distinct user id, leveraging the fact that we don't use each other's instances

devonbush:
@eweitz  your provided code totally works, and thanks for bringing up the issue.  I think the correct solution is to update this controller to use the StudyAware concern so we're not duplicating that logic.  I spent a few minutes on that, and it's going to be good, but non-trivial.  So that can wait until next month's developer days :)

eweitz (approved):
Reasonable!  

I'd opt to ship 4 weeks sooner with these working API refinements then refactor more, or even ship without these suggestions if the refinements are prioritized this quarter (as it seems they'd be, thus my approval here), but I can also see deferring shipping until a more robust solution is ready.

In any scenario, users will be fans of this new feature.

devonbush:
Ok, will merge, and then we can talk about the best strategy for upgrading the API to be accessionable

PR #1456, opened 2022-04-14
Update 20220401103323_encode_azul_files_names.rb (SCP-4276)

ehanna4:
Fix the migration script - final PR feedback fix involved removing the newly unnecessary function `transform_files` but missed doing that to the migration so making that fix here

### DIFF
--- a/db/migrate/20220401103323_encode_azul_files_names.rb
+++ b/db/migrate/20220401103323_encode_azul_files_names.rb
@@ -8,10 +8,10 @@ class EncodeAzulFilesNames < Mongoid::Migration
     end
   
     def self.down
-    download_reqs = DownloadRequest.where(:azul_files.ne => nil)
-    download_reqs.each do |req|
-      azul_files_as_hash = ActiveSupport::JSON.decode(req.azul_files.gsub('=>', ':'))
-      req.update(azul_files: azul_files_as_hash)
+      download_reqs = DownloadRequest.where(:azul_files.ne => nil)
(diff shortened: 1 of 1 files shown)


db/migrate/20220401103323_encode_azul_files_names.rb
```diff
@@ -2,16 +2,16 @@ class EncodeAzulFilesNames < Mongoid::Migration
...
       download_reqs = DownloadRequest.where(:azul_files.ne => nil)
       download_reqs.each do |req|
-        encoded_azul_files = DownloadRequest.transform_files(req.azul_files, :encode)
+        encoded_azul_files = ActiveSupport::JSON.encode(req.azul_files)
         req.update(azul_files: encoded_azul_files)
       end
     end
   
     def self.down
     download_reqs = DownloadRequest.where(:azul_files.ne => nil)
```
bistline, line 11: Non-blocking: these four lines should be indented, as per normal Ruby style:

```
def self.down
  download_reqs = DownloadRequest.where(:azul_files.ne => nil)
  download_reqs.each do |req|
    azul_files_as_hash = ActiveSupport::JSON.decode(req.azul_files.gsub('=>', ':'))
    req.update(azul_files: azul_files_as_hash)
  end
end
```

bistline (approved):
Just a style nit - fix works as advertized.

eweitz (approved):

devonbush (approved):

PR #1459, opened 2022-04-18
Revert "improving performance of large data array loading" (SCP-4281)

devonbush:
This reverts commit 8ebc3eeb3e069cd10d859207c17a229548a876c2.  It was done to improve performance (mostly in development) of loading large studies.  But the result was a couple of mongo connection issues, so we're reverting it here.  See https://sentry.io/organizations/broad-institute/issues/3193612380/?project=1424198&query=is%3Aunresolved for an example error

TO TEST:
1. load a cluster view of a large study (with > 100,000 cells)
2. observe that the study loads


eweitz (approved):
Code looks good; I confirm it reverts #1420.

PR #1462, opened 2022-04-20
Updating to latest version of `singlecellportal/rails-baseimage` (SCP-4290)

bistline:
This updates our Dockerfile to use the latest release of `singlecellportal/rails-baseimage:2.0.0`.  This changes the underlying `ubuntu` image from `18.04` to `20.04`, as well as updating to the latest release of the `phusion/passenger-docker` image.  In addition, this update changes our default Ruby version to `2.6.9`, and will allow us to update to `3.1.x` in the future if we so choose.

This also fixes the link to the CI badge that has been broken since we migrated to Github Actions.

MANUAL TESTING
To ensure that this image builds correctly:
1. Pull this branch
2. Authenticate into vault, and ensure Docker is running on your machine
3. Run either one of the following commands in the project root directory:
```
### only build the Docker image
docker build -t single_cell_docker:latest -f Dockerfile .

### build the Docker image and boot the portal container (will take dramatically longer)
bin/load_env_secrets.sh -p secret/kdux/scp/development/$(whoami)/scp_config.json \
                        -r secret/kdux/scp/development/$(whoami)/read_only_service_account.json \
                        -s secret/kdux/scp/development/$(whoami)/scp_service_account.json
```

This PR satisfies SCP-4290

PR #1463, opened 2022-04-21
adding corrupted study to list not to delete

devonbush:
This was delaying build results for Jon's PR #1462


bistline (approved):

devonbush:
As far as I know, yes.  The next time we get one of these, we might want to research how to make the delete workspace call directly to SAM, rather than through orchestration.  Per Matt, Bemis the problem is that orchestration can't handle these calls, but SAM can.

PR #1464, opened 2022-04-21
Array label splitting in cluster plots (SCP-4269)

devonbush:
As shown in demo, this enables array-based columns to be split out into individual labels, and then multiple points per bead are plotted.  The backend change is a new field in `ClusterFileInfo` that is a hash of annotation names to boolean values of whether the annotation should be split by default.

The frontend reads that default to determine whether to initially split the arrays, and then also displays a control above the legend for toggling the display.  the 'jitter' will plot points in a 3x3 grid around the center point, and can therefore handle up to 9 distinct points per cell/bead.  
image

For performance, this has almost no impact on regular plots.  We do add an iteration over all labels to check for '|' characters to determine whether to show the split control (which takes ~0.5sec per 1MM labels).  But that iteration happens after the Plot renders.  Rendering the split plot is determined by how many additional points are created (so, e.g. a 100K cluster split into 400K points will take about as long to render as a 400K cluster, minus server loading time)

This currently has no way for users to control the default setting.  Jonah seemed ambivalent about how he'd want it set, and he has so many clusters we'll probably want to update his study in the console anyway.

This also updates custom_colors to use the technique Emily developed for DownloadRequest of storing hashes with potentially forbidden key characters as Strings, rather than base64 encoding.  Storing them as strings is simpler (especially given nested hash/array objects) than base64 encoding keys, and also makes the in-db content more human readable.  If we have more fields like these in the future, we might want to consider making a custom Mongoid `hash_as_string` field type

TO TEST:
1.  Load a study with array-based labels.  You can run `SyntheticStudyPopulator.populate('array_mouse_eye')` for an example
2.  view the cluster
3.  select the array-based annotation ('cell_type_array') for the 'array_mouse_eye' study
4. Confirm the cluster plot appears normally, and with the link to 'Split array labels'
5. Click the 'Split array labels' link, confirm the labels are split into individual types.  You may have to zoom in/out to see the individual points.


app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -186,6 +188,7 @@ function RawScatterPlot({
         }
       })
     }
+    scatter.hasArrayLabels = scatter.data.annotations.some(annot => annot.includes('|'))
```
eweitz, line 191: I'm curious how often vertical pipes -- i.e. | -- occur in pre-existing group labels / annotation labels.  I imagine it's infrequent, but in cases where it doesn't denote an array this would be a bug.

app/javascript/lib/plot.js
```diff
@@ -153,6 +148,62 @@ PlotUtils.filterTrace = function({
...
+  // The value of 400 was picked because assuming that point size
+  // is ~3px, and the graph is rendered ~800px tall/wide, a displacement of 800/400 = ~2px is enough to push
+  // the points far enough apart to be seen individually, but still be associated with each other.
+  // this may be made more sophisticated later.
+  const jitterFraction = 400
+  // jitter will place cells in a 3x3 grid whose center is the actual coordinate
+  const xJitterMods = [0, -1, 1, 1, -1, -1, 0, 1, 0]
+  const yJitterMods = [0, -1, -1, 1, 1, 0, -1, 0, 1]
+  let xJitter = 0
+  let yJitter = 0
```
eweitz, line 179: Slick!

eweitz (approved):
Code looks good!

app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -235,8 +235,10 @@ def self.get_cluster_viz_data(study, cluster, url_params)
           }
 
           coordinate_labels = ClusterVizService.load_cluster_group_coordinate_labels(cluster)
-          custom_colors = cluster.study_file.cluster_file_info&.decoded_custom_colors || {}
+          custom_colors = cluster.study_file.cluster_file_info&.custom_colors_as_hash || {}
           custom_annotation_colors = custom_colors[annotation[:name]] || {}
+          annotation_split_defaults = cluster.study_file.cluster_file_info&.annotation_split_defaults_as_hash || {}
+          split_label_arrays = annotation_split_defaults[annotation[:name]] || false
```
bistline, line 241: Since `nil` and `false` both evaluate as "falsey":
```suggestion
          split_label_arrays = annotation_split_defaults[annotation[:name]].present?
```

app/models/cluster_file_info.rb
```diff
@@ -3,24 +3,36 @@ class ClusterFileInfo
...
+  end
 
-    hash.map do |key, nested_hash|
-      {
-        Base64.send(transform, key) => nested_hash.map { |k, v| { Base64.send(transform, k) => v } }.reduce({}, :merge)
-      }
-    end.reduce({}, :merge)
+  # This is called before the data is saved to the DB to ensure the file hashes are encoded as JSON strings
+  # This is necessary to work around MongoDBs constraints on '.' and '$' in hash keys
+  def stringify_hashes
```
bistline, line 25: Very slick - neatly sidesteps the original issue, and maintains a "human readable" form in the DB.

bistline (approved):

app/javascript/components/visualization/ScatterPlot.jsx
devonbush: I can't run the analysis for cluster-specific annotations, but all study-wide annotations in production that have `|` characters in labels are clearly intended as arrays

app/controllers/api/v1/visualization/clusters_controller.rb
devonbush: Oh that is interesting -- I did not realize `false.present?` was `false`!  The code shortening is a little too much dark magic for me, but I'm very glad I know that fact now

PR #1465, opened 2022-04-21
adding last_public and last_initialized to study report (SCP-4253)

devonbush:
This adds the most recent time a study was made public and the most recent time it was initialized to the study report.  This also adds a couple of helper methods to `study.rb` for computing them for individual studies (although they are non-trivial DB queries, so for performance reasons likely should be used sparingly).  The report does not use the helper methods, and instead uses a bulk query to fetch the data for all studies at once.

Note that this does not instrument any specific mixpanel events with these times -- the goal is, per Leyla's recent demo to use lookup tables to provide that capability

TO TEST:
1. toggle some studies public/private in your local instance
2. go to 'Reports' in the main SCP menu, and then export the 'Study Data' report
3. confirm that for each study, last_public and last_initialized reflect the correct times

### DIFF
--- a/app/lib/reports_service.rb
+++ b/app/lib/reports_service.rb
@@ -63,7 +63,8 @@ class ReportsService
 
     history_hash = ReportsService.study_histories
     study_hash.keys.each do |study_id|
-      study_hash[study_id][:last_public] = history_hash[study_id] ? history_hash[study_id][:last_public] : nil
+      study_hash[study_id][:last_public] = history_hash.dig(study_id, :last_public)
+      study_hash[study_id][:last_initialized] = history_hash.dig(study_id, :last_initialized)
       study_hash[study_id][:last_initialized] = history_hash[study_id] ? history_hash[study_id][:last_initialized] : nil
     end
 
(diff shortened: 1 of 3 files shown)


app/lib/reports_service.rb
```diff
@@ -60,6 +60,29 @@ def self.study_data
         study_hash[study_id][:has_raw_counts] ||= is_raw_counts.present? ? is_raw_counts['is_raw_counts'] : false
       end
     end
+
+    history_hash = ReportsService.study_histories
+    study_hash.keys.each do |study_id|
+      study_hash[study_id][:last_public] = history_hash[study_id] ? history_hash[study_id][:last_public] : nil
```
bistline, line 66: A simpler way to do both of these - it will gracefully handle the non-existent key and default to `nil`:
```suggestion
      study_hash[study_id][:last_public] = history_hash.dig(study_id, :last_public)
      study_hash[study_id][:last_initialized] = history_hash.dig(study_id, :last_initialized)
```

bistline (approved):
Non-blocking suggestions.

eweitz (approved):
Looks good!

Would the forthcoming Mixpanel reports replace the "Studies created per month" graphs in our [analytics spreadsheet](https://docs.google.com/spreadsheets/d/1FSpP2XTrG9FqAqD9X-BHxkCZae9vxZA3cQLow8mn-bk/edit#gid=0)?  If so, would extending this approach replace the whole spreadsheet?

devonbush:
Possibly!  The goal here is just to export data in as general-purpose a form as possible.  Whether we plug it into mixpanel or a spreadsheet or something else is up to us.

PR #1466, opened 2022-04-22
Count result selections per search, for refined CTR metrics (SCP-4291, SCP-4303)

eweitz:
This tracks the number of times search results were clicked _per search_.  It refines UX analytics for global study search.

Previously, if a user searched, say, "human" and clicked 6 results for a given search, e.g. `command` key + click, we could get peculiar average click-through rate (CTR) metrics like "600%".  Here's an example: https://mixpanel.com/s/UpXjw.  Original CTR analytics were instrumented in #1446.

Now, we'll be able to restrict Mixpanel queries by e.g. `numSelections = 1`, giving us more intuitive CTR metrics.  There's a fair chance what we call "CTR" will actually be "first CTR" -- i.e. the rate at which searches were proceeded by selection of _any_ search result.  Having `numSelections` be an integer count rather than a Boolean (e.g. `firstSelection = true`) lets us analyze multi-result-selection UX more deeply in the future, if we want.

Per feedback, this update also consolidates search-related logging.

To test:
1. Open Network panel in DevTools, filter requests to "bard method:POST", go to SCP home page
2. Click a result in the list such that it opens in a new tab, e.g. via `command` key + click on an SCP study title
3. Confirm an event `select-search-result` contains `numSearches: 0` and `numSearchSelections: 1`
4. Do step 2 again
5. Confirm an event `select-search-result` contains `numSearches: 0` and `numSearchSelections: 2`
6. Search "human"
7. Do step 2 again
8. Confirm an event `select-search-result` contains `numSearches: 1` and `numSearchSelections: 1`
9. Confirm an event `select-search-result` contains `numSearches: 1` and `numSearchSelections: 2`

This satisfies SCP-4291 and SCP-4303.


devonbush (commented):
I'm not sure I understand how this is supposed to work.  The comment says it will track multiple clicks (e.g. via "Open in new tab"), but if I do a search and then open results in new tabs, as far as I can tell those aren't tracked outside of 'contextmenu' events.  Then if I click on a result after opening others in new tabs, the `numSearchSelections` is still just 1

devonbush:
Also, if I open a result in a new tab by using middle-click (which I almost never use), those don't seem to be tracked at all.  Not even a contextmenu event

eweitz:
Oh, I'm actually using `command` key + click to open in a new tab.  I assumed selecting via "Open in new tab" would be the same event, but I confirm what you see that it's not.  My mistake!  I've updated the description.

As you note, opening via context menu and "Open in new tab" is not tracked, so those aren't contributing here.  That's likely worth addressing as part of the search OKR but seems separate from this task.  I've opened a ticket for it.

Looking closer at analytics, it seems > 100% CTR users are command-clicking (or otherwise trackably clicking) to open SCP studies in a new tab, or simply clicking HCA project results.  So I think these updates resolve the original issue, and you've uncovered an additional issue that likely affects the OKR (and broader link click tracking).

devonbush:
I looks like there are about ~100 contextmenu events per day on the homepage, and I'd expect most of those are opening search links in new tabs.  So given that we see ~1000 study searches in a given day, yeah, we'll want to add a fudge factor of ~10% on any click-through rates

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -212,6 +215,16 @@ export function logSelectSearchResult(study, logProps={}) {
   // This is needed to easily analyze changes in click-through rate (CTR).
   logProps.numSearches = numSearches
 
+  // Number of selections done for the current search.  Users can click
+  // multiple results for a given search, e.g. via "Open in new tab", and this
```
ehanna4, line 219: Should this comment be updated to reflect that this is only tracked with keyboard commands not the mouse ones for new tab? (if i've understood the rest of the conversation in this PR correctly that is)

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -212,6 +215,16 @@ export function logSelectSearchResult(study, logProps={}) {
...
+  // property lets us account for that when analyzing CTR and related  metrics.
+  if (numSearches != prevNumSearches) {
```
ehanna4, line 221: Let's use the strict equals check here:
```suggestion
  if (numSearches !== prevNumSearches) {
```

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: Good catch!  Done in 7695326.

bistline (approved):
This works, although in the initial example of the empty search, when I `CMD` + click on a study entry, the `numSearchSelections` parameter is always `0`, not `1`.  After I run a search, it is then `1`.  Not sure if this is intended or not.

app/javascript/components/search/results/StudySearchResult.jsx
```diff
@@ -174,6 +174,9 @@ function studyTypeBadge(study) {
   }
 }
 
+let prevNumSearches = -1
+let numSearchSelections = 0
+
 /**
```
devonbush, line 180: It will be much clearer if these properties are tracked inside scp-api-metrics along with `numSearches`--these are much more about global page state than any particular search result component.    Accordingly, let's move these into `scp-api-metrics` alongside `numSearches` and also move logSelectSearchResult there too (or another suitable `[X]-metrics.js` file.)

devonbush (changes requested):

app/javascript/components/search/results/StudySearchResult.jsx
eweitz: Agreed!  Done in b176314.  The refactor had been scoped elsewhere, but it makes sense to do with this update.

app/javascript/lib/search-metrics.js
```diff
@@ -0,0 +1,275 @@
...
+    // two searches.
+    //
+    // To consider: integrate a way to determine which *interaction* triggered
+    // search.  This was considered very early for separate reasons, but
+    // abandoned as it was invasive.  The clearly-brittle nature of preventing
+    // these artifactual searches shifts that cost-benefit, somewhat.
+    return
+  }
+
+  numSearches += 1 // Number of searches insofar as usage analytics is concerned
```
devonbush, line 140:
```suggestion
  numSearches += 1 // Number of searches insofar as usage analytics is concerned
  numSearchSelections = 0 // Number of selected results for the current search
```

app/javascript/lib/search-metrics.js
```diff
@@ -0,0 +1,275 @@
+import camelcaseKeys from 'camelcase-keys'
+
+import { getNumFacetsAndFilters } from '~/providers/StudySearchProvider'
+import { log } from '~/lib/metrics-api'
+
+let prevNumSearches = -1
```
devonbush, line 6:
```suggestion
```
I think you can get rid of `prevNumSearches` if instead you reset `numSearchSelections` inside `logSearch`

app/javascript/lib/search-metrics.js
```diff
@@ -0,0 +1,275 @@
...
+  logProps.numSearches = numSearches
+
+  // Number of selections done for the current search.  Users can click
+  // multiple results for a given search, e.g. via command key + click, and
+  // this property lets us account for that when analyzing CTR and related
+  // metrics.
+  if (numSearches !== prevNumSearches) {
+    numSearchSelections = 1
+    prevNumSearches = numSearches
+  }
```
devonbush, line 223:
```suggestion
  if (numSearches !== prevNumSearches) {
    numSearchSelections = 1
    prevNumSearches = numSearches
  }
```
I think you can delete this if block if numSearchSelections is reset inside `logSearch`

devonbush (approved):
Thanks for the refactor -- much easier to get a handle on the logging setup now.  One suggestion for tracking numSearchSelections in an easier-to-follow way.

bistline (approved):
Works as advertised now!

app/javascript/lib/search-metrics.js
eweitz: Elegant, confirmed locally (with a minor adjustment, ad0ad70).

PR #1467, opened 2022-04-22
Expanding keyword search to study metadata (SCP-4275)

bistline:
Currently, keyword searches in SCP only look at study titles/descriptions, and also author information.  This update extends that functionality to now look through stored search facet entries to determine if a provided keyword matches a given facet filter.  Upon matching, this will trigger a metadata search in MongoDB (faster than BigQuery, and contains the same data) to find all possible hits.  These additional studies are then rolled into the normal keyword-based matches, and a combined weight of text + metadata hits is calculated to appropriately weight results.

For instance, searching for terms like `blood` and `tuberculosis` will hit on the `organ` and `disease` facets, and then append any studies that have those values present in their metadata files.  Furthermore, since this is a keyword rather than faceted search, partial matches (whole words, not substrings) are also included.  This means that `tuberculosis` will match on filters such as `esophageal tuberculosis`, `intestinal tuberculosis`, and `oral tuberculosis`.

Screenshot showing new keyword metadata matches of the above scenario:
Screen Shot 2022-04-22 at 3 34 44 PM

In addition, this update refactors some of the logic in `Api::V1::SearchController` into the `StudySearchService`, with the intent being better test coverage.  As such, new test suites have been added for `StudySearchResultsObjects` and `StudySearchService`.

MANUAL TESTING
2. Search for `blood tuberculosis` in the text search box
3. Confirm that the synthetic studies `Tuberculosis subtypes in human male, blood study with arrays` and `Human HIV and Tuberculosis, blood study` both show up first, and have both highlighted text matches and metadata badges

This PR satisfies SCP-4275.


app/controllers/api/v1/search_controller.rb
```diff
@@ -557,24 +565,48 @@ def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, que
           results_matched_by_data['numResults:scp'] = studies.length # Total number of SCP results
           # Azul study results to be added with SCP-4202
 
-          return {:studies => studies, :results_matched_by_data => results_matched_by_data}
```
ehanna4, line 560: What's the difference between doing` {:variable => value}` vs `{variable: value}` and how did you know to use which one where lol

bistline: They're functionally equivalent - `{variable: value}` is a linting suggestion from Rubocop.  It's just fewer characters is all.

app/controllers/api/v1/search_controller.rb
```diff
@@ -762,6 +794,26 @@ def self.find_matching_filters(facet:, filter_values:)
         matching_filters
       end
 
+      # take a term and match to a possible search facet/filter
+      # will return a single hash with keys as facet names, and values as an array of filter matches
+      def self.match_facet_filters_from_terms(term_list)
+        terms = term_list || []
+        filters_by_facet = {}
+        terms.each do |term|
+          facets = SearchFacet.find_facets_from_term(term)
```
devonbush, line 803: I suspect this operation is relatively cheap, but it's worth noting that we'll duplicate this operation in `format_facet_query_from_keyword` in hca_azul_client, so it might be nice if we could consolidate

app/controllers/api/v1/search_controller.rb
```diff
@@ -557,24 +565,48 @@ def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, que
...
+      end
+
+      # take the output from :match_facet_filters_from_keywords and match to any possible CellMetadatum in the database
+      # will return a hash of accessions to matched metadata entries
+      def self.get_studies_from_term_conversion(terms)
+        accessions_to_filters = {}
+        match_facet_filters_from_terms(terms).each do |identifier, values|
+          # we need to look in '__ontology_label' entries as well as these will include plain text entries
+          metadata_names = [identifier, "#{identifier}__ontology_label"]
+          matches = CellMetadatum.where(:name.in => metadata_names, :values.in => values)
```
devonbush, line 595: Wow, I was a little worried if this query would be sluggish in production, so I tested it:  `Benchmark.measure { CellMetadatum.where(:name.in => ['cell_type','cell_type__ontology_label'],  :values.in => ['amacrine cell']) }`  shows 0.0002 seconds real-time.

devonbush (changes requested):
Cool stuff - big step forward for search capability, and kudos for the high-perofrmance solution.  This absolutely needs some dedicated test cases though -- our search logic is getting very complex (and probably will get more complex soon), and so we need to keep our coverage high.

app/controllers/api/v1/search_controller.rb
```diff
@@ -522,29 +525,34 @@ def self.escape_terms_for_regex(term_list:)
       def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, query_context:)
         case query_context
         when :keyword
-          author_match_study_ids = Author.where(:$text => {:$search => terms}).pluck(:study_id)
+          author_match_study_ids = Author.where(:$text => { :$search => terms }).pluck(:study_id)
```
devonbush, line 528: This method is getting more complex -- let's move `generate_mongo_query_by_context` to `app/lib/study_search_service`

app/controllers/api/v1/search_controller.rb
```diff
@@ -557,24 +565,48 @@ def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, que
...
-    
+          studies = base_studies.any_of({ :$text => { :$search => terms } }, { :accession.in => accessions })
+          { studies: studies, results_matched_by_data: {}, metadata_matches: {} }
+        end
+      end
+
+      # take the output from :match_facet_filters_from_keywords and match to any possible CellMetadatum in the database
+      # will return a hash of accessions to matched metadata entries
+      def self.get_studies_from_term_conversion(terms)
+        accessions_to_filters = {}
```
devonbush, line 591: This method also should go in `app/lib/study_search_service.rb` and we should have independent test cases for it

app/controllers/api/v1/search_controller.rb
```diff
@@ -762,6 +794,26 @@ def self.find_matching_filters(facet:, filter_values:)
         matching_filters
       end
 
+      # take a term and match to a possible search facet/filter
+      # will return a single hash with keys as facet names, and values as an array of filter matches
+      def self.match_facet_filters_from_terms(term_list)
```
devonbush, line 799: same, move to `app/lib/study_search_service` and write independent tests for it

bistline:
> Combined facted and keyword search does not generate match indicator for faceted match.
> 
> Functional testing: Searched for "cell" in keyword search AND faceted search for 'bipolar neuron'
> 
> Search matches synthetic study "Male Mouse brain"
> 
> Expected [Metadata match for cell_type "B cell, amacrine cell, retinal cone cell"] AND the term "biopolar neuron" added to the metadata match bubble OR possibly a separate [Metadata match for cell_type "bipolar neuron"] bubble
> 
> Is it possible in this PR to have the AND highlighted in red for combined keyword and faceted search?

I think so, but I'm not sure.  The combination of both keywords & facets gets hairy in terms of matching/ordering.

app/controllers/api/v1/search_controller.rb
bistline: Yeah, any query in Mongo really only bogs down in the serialization of data.  Running the queries is very, very fast.  And in this case, we're only pulling out small amounts of data.

app/lib/study_search_service.rb
```diff
@@ -55,4 +55,124 @@ def self.sanitize_gene_params(genes)
...
+  # :accessions based on the :query_context (included by default, but excluded in :inferred to avoid duplicates)
+  def self.generate_mongo_query_by_context(terms:, base_studies:, accessions:, query_context:)
+    case query_context
+    when :keyword
+      author_match_study_ids = Author.where(:$text => { :$search => terms }).pluck(:study_id)
+
+      matches_by_text = base_studies.where({ :$text => { :$search => terms } })
+      metadata_matches = get_studies_from_term_conversion(terms.split)
+      all_accessions = accessions + metadata_matches.keys
+      matches_by_accession = base_studies.where({ :accession.in => all_accessions })
```
ehanna4, line 72: `matches_by_accession` looks like it's over-reporting the things that have matched by accession, or at least something is funny with the math I'm seeing - see attached screenshots for where `matches_by_accession` has the same number as the total results but other match properties also claim results. 

Screen Shot 2022-04-25 at 4 06 37 PM
Screen Shot 2022-04-25 at 4 10 10 PM

It is not happening when results are purely text or author based
Screen Shot 2022-04-25 at 4 14 39 PM
Screen Shot 2022-04-25 at 4 15 02 PM

eweitz (changes requested):
Swift work on a significant enhancement!  I manually confirmed the test steps.

Blocker: please add `numResults:scp:metadata` with an appropriate count wherever `@match_by_data` is assigned.  

Like the other `numResults:scp:foo` properties (assigned in `generate_mongo_query_by_context`), `numResults:scp:metadata` would let us log those search result count by "backend match type" in both `search` and `selectSearchResults` events, and thereby help us assess how much various backend search mechanics are leveraged, and how much they contribute to various metrics of our search product quality.

bistline:
> Swift work on a significant enhancement! I manually confirmed the test steps.
> 
> Blocker: please add `numResults:scp:metadata` with an appropriate count wherever `@match_by_data` is assigned.
> 
> Like the other `numResults:scp:foo` properties (assigned in `generate_mongo_query_by_context`), `numResults:scp:metadata` would let us log those search result count by "backend match type" in both `search` and `selectSearchResults` events, and thereby help us assess how much various backend search mechanics are leveraged, and how much they contribute to various metrics of our search product quality.

Ah, that explains what those are better. I'll add that property.

app/lib/study_search_service.rb
bistline: @ehanna4 you are correct - it was.  With Eric's suggestion of adding the `numResults:scp:metadata` property, this is now resolved.  The underlying query still finds the studies via their accession (after doing the metadata match), but we are able to report them discretely.  Now, only studies queried for using their accessions (SCPXXXX) will roll up on `numResults:scp:accession`.

eweitz (approved):
Thanks for addressing those metrics -- looks good!

I confirmed that I get more results now than I did before.  For example, the synthetic study "Human all genes" is returned upon keyword-searching `blood tuberculosis` now, and wasn't before.  

A few issues remain.  I don't consider them blockers.

First is likely just a nit.  When I keyword-search `blood tuberculosis`, I get `numResults:scp:metadata: 17` in the related Bard / Mixpanel `search` event, but only 10 results are actually returned by the search.  Quick review suggests the 7 metadata-matched-but-unreturned studies are queued for deletion.  I imagine this isn't a significant issue on production, but I can see it confusing staff when examining things on their development environment.  It'd be nice if deletion-queued studies were excluded from results matched by metadata.

Second is a pair of related UI issues.  The keyword search input caption and query summary are now both less accurate, and presumably confusing for their intended audience -- i.e., users who have an advanced mental model of SCP keyword search mechanics.  

The keyword input caption says "Title, author, and description search".  That's now incomplete, as it now also includes metadata.  Of course, the filter caption says "Metadata search", making "Title, author, description, and metadata search" somewhat unclear (and rather verbose).  I might suggest recaptioning the keyword search to instead say "Search by keywords", and the filter search to "Search by filters".

The query summary --  e.g. "Text contains (blood OR tuberculosis)" -- is also now less accurate than it was before.  Title, description and author seem reasonable to construe as "text"; metadata less so.  I think "Result contains (blood OR tuberculosis)" would be more accurate.  

This screenshot shows the metrics nit, and those two UI issues.

Screen Shot 2022-04-26 at 10 21 46 AM

For users who pay close attention to search logic, or precisely what's in their result set and why, my hunch is they'd be better served by a discreet link to documentation, or an icon-with-popup.  Few will engage with those resources, and I think that's likely roughly commensurate with the size of the audience interested in such details.

I don't consider these things blocking, but I thought I'd surface them sooner than later.

test/integration/lib/study_search_service_test.rb
```diff
@@ -0,0 +1,126 @@
...
+                                           ]
+                                         }
+                                       ])
+    @name_study = FactoryBot.create(:detached_study,
+                                    name_prefix: 'Cell Name Test',
+                                    public: false,
+                                    user: @user,
+                                    test_array: @@studies_to_clean)
+    @base_studies = Study.where(user: @user)
+    @filters = [
```
devonbush, line 30: can you add a filter to this list that doesn't match `cell` so that we confirm non-matching filters don't end up in the results?

devonbush (commented):
Thanks for the tests, and the refactor -- big wins on readability and modularity.  One small change to increase test coverage, and then this is good to go

devonbush (approved):

PR #1469, opened 2022-04-27
Precomputed heatmaps upgrade (SCP-4267)

devonbush:
Putting in draft until after demo.  right now there's a lot of flux on naming things

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import _clone from 'lodash/clone'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faLink, faArrowLeft, faCog, faTimes, faDna, faUndo } from '@fortawesome/free-solid-svg-icons'
@@ -22,6 +22,7 @@ import ScatterPlot from '~/components/visualization/ScatterPlot'
 import StudyViolinPlot from '~/components/visualization/StudyViolinPlot'
 import DotPlot from '~/components/visualization/DotPlot'
 import Heatmap from '~/components/visualization/Heatmap'
+import GeneListHeatmap from '~/components/visualization/GeneListHeatmap'
 import GenomeView from './GenomeView'
 import ImageTab from './ImageTab'
 import { getAnnotationValues, getDefaultSpatialGroupsForCluster } from '~/lib/cluster-utils'
@@ -34,11 +35,12 @@ import { log } from '~/lib/metrics-api'
 const tabList = [
   { key: 'loading', label: 'loading...' },
   { key: 'scatter', label: 'Scatter' },
-  { key: 'annotatedScatter', label: 'Annotated Scatter' },
+  { key: 'annotatedScatter', label: 'Annotated scatter' },
   { key: 'correlatedScatter', label: 'Correlation' },
   { key: 'distribution', label: 'Distribution' },
   { key: 'dotplot', label: 'Dot Plot' },
   { key: 'heatmap', label: 'Heatmap' },
+  { key: 'geneListHeatmap', label: 'Precomputed heatmap' },
   { key: 'spatial', label: 'Spatial' },
   { key: 'genome', label: 'Genome' },
   { key: 'infercnv-genome', label: 'Genome (inferCNV)' },
@@ -114,7 +116,7 @@ export default function ExploreDisplayTabs({
 
   const annotationList = exploreInfo ? exploreInfo.annotationList : null
   // hide the cluster controls if we're on a genome/image tab, or if there aren't clusters to choose
-  const showClusterControls = !['genome', 'infercnv-genome', 'images'].includes(shownTab) &&
+  const showClusterControls = !['genome', 'infercnv-genome', 'images', 'geneListHeatmap'].includes(shownTab) &&
                                 annotationList?.clusters?.length
 
   let hasSpatialGroups = false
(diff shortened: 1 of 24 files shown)


app/javascript/components/upload/upload-utils.js
```diff
@@ -79,6 +80,14 @@ export function formatFileFromServer(file) {
       raw_counts_associations: []
     }
   }
+  if (file.file_type === 'Gene List' && !file.heatmap_file_info) {
+    // some legacy studies will not have supplemental heatmap file info
+    file.heatmap_file_info = {
+      custom_coloring: false,
+      color_min: -1,
+      color_max: 1
```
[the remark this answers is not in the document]
devonbush: my understanding is that placeholders != defaults.   Placeholders, if not interacted with, do not actually populate the form.  (e.g. the `Genes e.g. "PTEN NF2"` placeholder in global gene search does not actually search that string if you hit enter).  These are default values, not placeholders, so they should be in black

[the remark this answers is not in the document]
devonbush: I see your point, though that it's not obvious where we decide to actually populate default values, versus we use empty strings as the default.  Here, I thought since these values directly control the scientific visualization, it would be good to populate them explicitly, so if we ever change the defaults, existing visualizations are not changed

ehanna4:
I don't know if this is overkill but I was able to set a min that was bigger than my max and it saved and the map worked (just basically flipped blue and red as being the high/low) - just wanted to mention it 
Screen Shot 2022-04-29 at 2 13 20 PM

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -22,23 +22,26 @@ import ScatterPlot from '~/components/visualization/ScatterPlot'
...
 
 
 const tabList = [
   { key: 'loading', label: 'loading...' },
   { key: 'scatter', label: 'Scatter' },
-  { key: 'annotatedScatter', label: 'Annotated Scatter' },
+  { key: 'annotatedScatter', label: 'Annotated scatter' },
   { key: 'correlatedScatter', label: 'Correlation' },
   { key: 'distribution', label: 'Distribution' },
   { key: 'dotplot', label: 'Dot Plot' },
```
ehanna4, line 42: Since you updated Annotated scatter to use sentence case should Dot Plot get the same treatment?

app/javascript/components/explore/ExploreView.jsx
```diff
@@ -43,15 +43,6 @@ function RoutableExploreTab({ studyAccession }) {
 
   useEffect(() => {loadStudyData()}, [studyAccession])
 
-  useEffect(() => {
-    // if the user hasn't selected anything, and there are genelists to view, but no clusters
-    // default to the first gene list
-    if ((exploreInfo && exploreInfo.annotationList.clusters.length === 0 &&
```
ehanna4, line 49: Do we need to be careful with undefined annotionLists or clusters here? aka there be '?' checks? aka `exploreInfo.annotationList?.clusters?.length` ?

app/javascript/components/upload/GeneListStep.jsx
```diff
@@ -39,7 +44,7 @@ function GeneListForm({
       <div className="col-md-12">
         <div className="form-terra">
           <p>
-            A list of genes and any computed expression values (mean, median, etc.) across any clusters.
+            A list of genes and any computed expression values (mean, median, etc.) across any clusters, for visualization as a heatmap
```
ehanna4, line 47: Does this need a period?

ehanna4 (approved):
Small nits plus the min/max question but overall functional review looks good and high level code review looks good as well

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -22,23 +22,26 @@ import ScatterPlot from '~/components/visualization/ScatterPlot'
...
   { key: 'heatmap', label: 'Heatmap' },
+  { key: 'geneListHeatmap', label: 'Precomputed heatmap' },
```
bistline, line 44: The only thing that I find confusing about this is that we give the user the option to customize the name of the dropdown menu, but the tab name does not inherit it.  I feel like it should, but not necessarily enough to warrant a change request.

app/javascript/components/upload/form-components.jsx
```diff
@@ -18,15 +18,27 @@ export function AddFileButton({ newFileTemplate, addNewFile }) {
 /** renders a basic label->value text field in a bootstrap form control */
 export function TextFormField({ label, fieldName, file, updateFile }) {
   const fieldId = `${fieldName}-input-${file._id}`
+  let value = file[fieldName] ? file[fieldName] : ''
```
bistline, line 21: Simpler:
```suggestion
  let value = file[fieldName] ?? ''
```

app/javascript/components/upload/form-components.jsx
```diff
@@ -18,15 +18,27 @@ export function AddFileButton({ newFileTemplate, addNewFile }) {
...
+  const [objName, nestedPropName] = fieldName.split('.')
+  if (nestedPropName) {
+    // handle a nested property like 'heatmap_file_info.custom_scaling'
+    value = file[objName][nestedPropName] ? file[objName][nestedPropName] : ''
```
bistline, line 25: Same as above:
```suggestion
    value = file[objName][nestedPropName] ?? ''
```

app/views/site/_study_settings_viz.html.erb
```diff
@@ -40,6 +40,12 @@
           <%= opts.text_field :expression_label, value: @study.default_options[:expression_label], placeholder: @study.default_expression_label, class: 'form-control' %>
         </div>
       </div>
+      <div class="form-group row">
```
bistline, line 43: Blocker: One thing I noted in this form that is not a result of a change you've made, but rather a lurking bug - if you have a study w/o clustering data that does have gene lists, we cannot render this form.  We may need to abstract out these various sections into partials and render them iteratively depending on what kind of data the study has.

bistline (changes requested):
Only one blocker - studies w/o clustering data don't render the view options settings panel, meaning users cannot configure the dropdown name until they add clustering & metadata.  There is also a question about the tab name that I don't consider blocking.

app/javascript/components/upload/GeneListStep.jsx
devonbush: Hah, I could go either way on this, Eric recommended the period be removed

app/javascript/components/explore/ExploreDisplayTabs.jsx
devonbush: ah, great point.  Yeah, the tab names right now are relatively hardcoded, but that's worth keeping in mind as this feature gets more usage

app/views/site/_study_settings_viz.html.erb
devonbush: Good catch -- will address

app/javascript/components/upload/upload-utils.js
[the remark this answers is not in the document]
eweitz: Thanks, this clarifies things nicely for me.

PR #1471, opened 2022-04-27
fixing auth on user_annotations controller (SCP-4311)

devonbush:
I spent a few minutes attempting to add a positive auth case, but the setup was complex enough that it wasn't worth it for this ticket given the relatively low usage of user annotations.  

TO TEST:
1. run `rails test test/api/user_annotations_controller_test.rb `
2. (if desired) Sign in to local instance
3. open a study in the explore tab and click 'create annotation'
4. wait 30mins until you get unauthed (or sign yourself out by byebugging an API controller method, and then doing `sign_out current_api_user` from the resulting console)
5. select an annotation, and then hit 'save'
6. Confirm the error appears as an auth error, not a 500

### DIFF
--- a/app/controllers/api/v1/user_annotations_controller.rb
+++ b/app/controllers/api/v1/user_annotations_controller.rb
@@ -4,7 +4,7 @@ module Api
     class UserAnnotationsController < ApiBaseController
       include Concerns::ApiCaching
 
-      before_action :set_current_api_user!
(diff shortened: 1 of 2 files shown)


bistline (approved):

devonbush:
Build failure in study_file_controller -- likely Terra instability

PR #1473, opened 2022-04-28
Better error for hidden char/unable to decode file (SCP-4309)

ehanna4:
This will report out an error when a file contains hidden or special characters that cause the decoding to fail. 

I was unable to add a test as the file reading done for the tests is mocked and encodes/decodes the content so even passing in content to the test that fails correctly in a real case doesn't for the test.

To test
- Pull this branch
- Go to the upload wizard for a study
- Choose 'expression matrix' dense to upload
- try to upload this file: [lines1-4.csv](https://github.com/broadinstitute/single_cell_portal_core/files/8585614/lines1-4.csv)  or add 'п»ї' to a dense matrix file you have like so:` п»їGENE	CELL_0001...`
- You should see an error like: "File cannot be uploaded in it\'s current state. Please reach out for assistance"

Screen Shot 2022-04-28 at 3 05 16 PM

### DIFF
--- a/app/javascript/lib/validation/validate-file-content.js
+++ b/app/javascript/lib/validation/validate-file-content.js
@@ -426,8 +426,8 @@ async function parseFile(file, fileType, fileOptions={}, sizeProps={}) {
     if (error instanceof ParseException) {
       parseResult.issues.push(['error', error.key, error.message])
     } else if (error instanceof TypeError) {
-      const msg = 'File cannot be uploaded in it\'s current state. Please reach out for assistance'
-      parseResult.issues.push(['error', 'type-error', msg])
+      const msg = 'File cannot be uploaded in its current state. Please reach out for assistance.'
+      parseResult.issues.push(['error', 'parse:unhandled:js-typeerror', msg])
     } else {
       parseResult.issues.push(['error', 'parse:unhandled', error.message])
     }


app/javascript/lib/validation/validate-file-content.js
```diff
@@ -425,6 +425,9 @@ async function parseFile(file, fileType, fileOptions={}, sizeProps={}) {
     // get any unhandled or deliberate short-circuits
     if (error instanceof ParseException) {
       parseResult.issues.push(['error', error.key, error.message])
+    } else if (error instanceof TypeError) {
+      const msg = 'File cannot be uploaded in it\'s current state. Please reach out for assistance'
```
bistline, line 429: I think this is great that we're catching/handling these now.  Non-blocking: If possible, it would be amazing if we could get more info on the file rather than "cannot be uploaded", like an offending line number or character string.

bistline (approved):

app/javascript/lib/validation/validate-file-content.js
jlchang: Is it accurate to say that "non UTF-8" (or maybe non-standard?) characters were detected in the file? 

It might be worth making the error message specific enough that we'll remember what the root cause of the issue is. (Because it's really hard to see/find the hidden characters).

devonbush: Yup, I like the idea of a `File cannot be parsed: non-UTF-8 characters are present.` message

devonbush (approved):
code fix looks good, I assume the wordsmithing will result in something good :)

app/javascript/lib/validation/validate-file-content.js
eweitz: It seems incorrect to tell the user this failed because non-UTF-8 characters were in the file, because the three characters in the test case -- [п](https://util.unicode.org/UnicodeJsps/character.jsp?a=041F) and [»](https://util.unicode.org/UnicodeJsps/character.jsp?a=00BB) and [ї](https://util.unicode.org/UnicodeJsps/character.jsp?a=0457) -- are all valid UTF-8.  

If we essentially keep the current generic message, let's use "[its](https://www.merriam-webster.com/words-at-play/when-to-use-its-vs-its)" and end the sentence with a period.

To Jon's comment, to my understanding we won't know line number, but I'd be curious if we'd have any more specific error information (whether offending string as he wonders, or anything else).

app/javascript/lib/validation/validate-file-content.js
```diff
@@ -425,6 +425,9 @@ async function parseFile(file, fileType, fileOptions={}, sizeProps={}) {
...
+      parseResult.issues.push(['error', 'type-error', msg])
```
eweitz, line 430: Blocker: let's classify this as a kind of unhandled parse error, to make related analytics easier and more extensible.

```suggestion
      parseResult.issues.push(['error', 'parsed:unhandled:js-type', msg])
```

eweitz (changes requested):
Nice CSFV refinement!  A note a trivial blocker, and add to the error message discussion.

Also, for auditability, please add Jira ticket ID to PR title.

eweitz:
Non-blocking: how feasible would it be to have an automated test for `lines1-4.csv` in `validate-file-content.js`?

app/javascript/lib/validation/validate-file-content.js
ehanna4: I'm didn't use 'unhandled' because isn't this being handled by having it output a user-readable error message?

ehanna4: I also think since 'type-error' is the actual error type it is useful to have as the error type in the mixpanel tracking. Does "js-type" have a meaning I am missing though? I might not be aware of what the significance is there and am open to using that since we do track the explicit error type elsewhere

ehanna4:
> Non-blocking: how feasible would it be to have an automated test for `lines1-4.csv` in `validate-file-content.js`?

I mentioned this in the PR description though not in super great detail. I was severely struggling to add a test for this as the error is coming from these lines in io.js:
`const arrayBuffer = await blob.arrayBuffer()

  const enc = new TextDecoder('utf-8')
  const stringContent = enc.decode(arrayBuffer)`

and with the tests in validate-file-content.js we are mocking out this logic - (from `validate-file-content.js`) "We have to mock readFileBytes since blob.arrayBuffer() is not supported in nodejs" 
`if (mockIO) {
    const readFileSpy = jest.spyOn(Io, 'readFileBytes')
    readFileSpy.mockImplementation((_file, startByte, chunkSize=Io.DEFAULT_CHUNK_SIZE) => {
      return content.slice(startByte, startByte + chunkSize)
    })`

so the test files are not facing the same issues that real files are with the io.

app/javascript/lib/validation/validate-file-content.js
ehanna4: unfortunately yeah we won't have the line number since it's occurring within io.js and all that has is chunk start and chunk end which don't seem super helpful to tell the user. 

I also kept this message more generic because the error type of "TypeError" could be caused by other things than just hidden/special characters and the error message that comes with this error is:  "Failed to execute 'decode' on 'TextDecoder': The encoded data was not valid.". So it's just about an decodable file so I didn't want to make the message incorrectly specific.

app/javascript/lib/validation/validate-file-content.js
eweitz: Thanks for helping me understand the original rationale!  Here's the rationale behind my suggestions.

**Why start with `parse:unhandled`?**
`parse:unhandled` errors do show users a readable error message, although it's generated by a built-in JS error object.  See e.g. https://mixpanel.com/s/3T3eal: the user saw the (admittedly cryptic) error message: "The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired."  It turns out that's the generic message for a certain JavaScript error object that's built into browser APIs.  

I'm unaware of a prior definition for `parse:unhandled`, but I'd suggest we define it as "when an error occurs that is not due to a specific validation issue in their file, and where we do not provide users guidance on how to fix it themselves".  Here, we can't parse their file, and the current error message doesn't indicate how the user can fix it without contacting us.

**Why include `js-type`?**
Simply `type-error` strikes me as easy to misinterpret.  If I'm looking at [CSFV errors by type](https://mixpanel.com/s/3PstMY), I think I'd be more likely to confuse an error typed as `type-error` with one of our validation errors, e.g. something like a group-vs-numeric annotation type validation error, or a validation error where a column header says content is numeric but the value isn't actually numeric.

Having `js-type` seems likelier to help the analyst intuit it's a _JavaScript_ TypeError, not a validation error like a format error or content error.  I initially thought having `-error` was redundant -- it's already classified as an error -- but I see your point that it helps map things.  

Maybe `parse:unhandled:js-typeerror` or `parse:unhandled:typeerror` would be best?  I've updated my suggestion to hopefully incorporate your rationale.

In any case, I appreciate you outlining your reasoning.  If my case above isn't persuasive, that's OK, I no longer consider this blocking!

app/javascript/lib/validation/validate-file-content.js
eweitz, line 429: Fixing trivial typo and adding period:

```suggestion
      const msg = 'File cannot be uploaded in its current state.  Please reach out for assistance.'
```

eweitz (approved):
After your helpful explanation, I'm less concerned about the logged `errorType`.  I offer more thoughts on the matter, but I defer to you!  I also suggest some trivial copyediting polish.

PR #1475, opened 2022-04-28
fixing sorting bugs, adding `:created_at` and `:view_count for` HCA entries (SCP-4287)

bistline:
This update fixes some lingering bugs with sorting results from cross-dataset search requests, where results are either instances of the `Study` class, or `Hash` entries representing HCA projects.  Now, any requested search & sort order will support a mix of both result classes.  In addition, HCA entries now include their original `submissionDate` as a proxy for `created_at`, and a default `view_count` of `0`, as we have no data on views from Azul.

This update also changes the default logging level for all Google API calls in the `test` environment to `:info`, from `:debug`.

MANUAL TESTING
2. Open the Swagger API documentation, and go to the search endpoint
3. Run a search that will return a mix of SCP & HCA entries, such as:
    * facets: `species:NCBITaxon_9606`
    * terms: `HIV`
4. Note that results return in a specific order by looking at the `matching_accessions` attribute in the response
5. Change the `order` parameter to both `recent` and `popular`, and note that the request completes and changes the order accordingly

This PR satisfies SCP-4287.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -308,24 +308,29 @@ module Api
           @studies = @studies.sort_by { |study| @accession_list.index(study.accession) }
         when :facet
           @studies = @studies.sort_by do |study|
-            accession = self.class.get_study_accession(study)
+            accession = self.class.get_study_attribute(study, :accession)
             metadata_weight = @metadata_matches.present? ?
-                                @metadata_matches.dig(study.accession, :facet_search_weight).to_i : 0
+                                @metadata_matches.dig(accession, :facet_search_weight).to_i : 0
             -(@studies_by_facet[accession][:facet_search_weight] + metadata_weight)
           end
         when :recent
-          @studies = @studies.sort_by(&:created_at).reverse
+          @studies = @studies.sort_by { |study| self.class.get_study_attribute(study, :created_at) }.reverse
         when :popular
-          @studies = @studies.sort_by(&:view_count).reverse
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -490,9 +495,13 @@ def sanitize_search_values(terms)
         end
       end
 
-      # extract study accession, accounting for source (SCP vs. external)
-      def self.get_study_accession(search_result)
-        search_result.try(:accession) || search_result[:accession]
+      # extract study attribute (like accession) accounting for source (SCP vs. external)
```
devonbush, line 498: this is nice

devonbush (changes requested):
This is some nice elegant code.  Please allow me to be a broken record and say that search code should have test cases :).   Since these sort params aren't supported in our UX, the test case doesn't need to be extensive, copying the beginning from  test 'should return search results using facets' ` from line 135 of search_controller_test.rb, adding a sort param, and then just confirming it doesn't break is probably sufficient

eweitz (approved):

bistline:
> This is some nice elegant code. Please allow me to be a broken record and say that search code should have test cases :). Since these sort params aren't supported in our UX, the test case doesn't need to be extensive, copying the beginning from test 'should return search results using facets' ` from line 135 of search_controller_test.rb, adding a sort param, and then just confirming it doesn't break is probably sufficient

@devonbush See [da3cba5](https://github.com/broadinstitute/single_cell_portal_core/pull/1475/commits/da3cba55d59816969bf44efc648959578894d619)

ehanna4 (approved):

PR #1476, opened 2022-04-28
Do not return HCA projects in collection searches (SCP-4295)

bistline:
This update adds a clause to skip running cross-dataset searches if a user is searching inside a collection.  This is to avoid any confusion as to the provenance of the data, as no officially hosted HCA projects are a part of any SCP collections (though there are some studies based off of the data, but not the original source projects).

MANUAL TESTING
1. Boot as normal
2. Run a search request that will return HCA results, such as `species:Homo sapiens`
3. If you do not have any collections in your local instance, run `SyntheticBrandingGroupPopulator.populate_all` to load some collections
4. Sign in, and select a collection from the `Browse collections` page
5. Re-run the same search as above, and validate that no HCA projects are returned

This PR satisfies SCP-4295.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -265,7 +265,8 @@ module Api
 
         # perform Azul search if there are facets/terms provided by user
         # run this before inferred search so that they are weighted and sorted correctly
-        if @facets.present? || @term_list.present?
+        # skip if user is searching inside a collection
+        if (@facets.present? || @term_list.present?) && @selected_branding_group.nil?
           begin
             @studies, @studies_by_facet = ::AzulSearchService.append_results_to_studies(@studies,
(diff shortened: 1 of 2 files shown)


devonbush (approved):
If it doesn't take too long to write, this is probably worth a test case.  Less because this `if` might not work, and more because then `search_controller_test.rb` serves as useful documentation of intended behavior.   What might be most effective (since the exact azul matches returned from a query could change unexpectedly) is to just duplicate the   `test 'should return search results using facets'` from line 135 of `search_controller_test.rb` but add a branding group to the query.  And then use a stub to assert that `AzulSearchService.append_results_to_studies` is not called.  (and then maybe add a parallel assert to the regular facet test that it *is* called)

ehanna4 (approved):
👍

PR #1481, opened 2022-05-02
Adding metrics for Azul search results, fixing existing count omissions (SCP-4202)

bistline:
This update adds two new Mixpanel metrics for the `search` event: `numResults:azul` and `numResults:total`.  These are for tracking the number of projects found in Azul, and the total number of results found across SCP & Azul, respectively.  In addition, this also fixes an omission where metadata matches from BigQuery are not being reported.  Now, these matches will be folded in with any results that were found via the `CellMetadata` collection, ensuring that accessions are not double counted.

Example of new metrics in Mixpanel:
Screen Shot 2022-05-02 at 11 37 35 AM

MANUAL TESTING
1. Boot as normal, and open Chrome DevTools > Network
2. Run a search that will return both SCP & Azul results, such as `species:Mus musculus`
3. Find the `search` event in the Network tab, and confirm that the new metrics should up, and the counts match what is shown in the UI

This PR satisfies SCP-4202.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -244,7 +244,15 @@ module Api
           # build up map of study matches by facet & filter value (for adding labels in UI)
           @studies_by_facet = self.class.match_studies_by_facet(query_results, @facets)
           # uniquify result list as one study may match multiple facets/filters
-          @convention_accessions = query_results.map {|match| match[:study_accession]}.uniq
+          @convention_accessions = query_results.map { |match| match[:study_accession] }.uniq
+          # report on matches for metadata, ensuring we don't double-count some accessions
+          # this can happen if we get a term converstion to metadata match
+          @match_by_data ||= {}
+          existing_metadata_matches = @metadata_matches.try(:keys) || []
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/search_controller.rb
```diff
@@ -244,7 +244,15 @@ def index
           # build up map of study matches by facet & filter value (for adding labels in UI)
           @studies_by_facet = self.class.match_studies_by_facet(query_results, @facets)
           # uniquify result list as one study may match multiple facets/filters
-          @convention_accessions = query_results.map {|match| match[:study_accession]}.uniq
+          @convention_accessions = query_results.map { |match| match[:study_accession] }.uniq
+          # report on matches for metadata, ensuring we don't double-count some accessions
+          # this can happen if we get a term converstion to metadata match
```
devonbush, line 249:
```suggestion
          # this can happen if we get a term conversion to metadata match
```

devonbush (approved):
looks good, one typo in comment

eweitz (approved):
Useful instrumentation refinement.  Code looks good, and I confirmed functionality locally.

PR #1482, opened 2022-05-02
Add Sentry logging for JS ErrorBoundary, Google storage, Bard (SCP-4096)

ehanna4:
This adds Sentry logging for application errors, Google Storage fetch errors and Bard request errors coming from the JS side and sets the `logger:app-frontend` to enable differentiating between these JS errors and backend errors.

To test:
- Pull this branch
- bundle install and yarn install to ensure the react-sentry stuff is fully added
- Open this file `app/javascript/lib/sentry-logging.js` and comment out lines 45-47
- To see Bard errors I find it simplest to open up `metrics.api` and change line 426 to fetch something like `${bardDomain}/api-this-will-not-work/event/`
- Then just load up your local and check out [Sentry](https://sentry.io/organizations/broad-institute/issues/?environment=development&project=1424198&query=&statsPeriod=24h) where you should see an error posted similar to:
Screen Shot 2022-05-02 at 11 37 50 AM

to see a different sort of error I found it easiest to:
- open up `StudyResults.jsx` and change line 77 to something like: `<StudyComponentIdonotExist`
- then load up local host and watch Sentry to see an error come through like:
Screen Shot 2022-05-02 at 4 16 55 PM

The logger tag in Sentry will be set to 'app-frontend'

Screen Shot 2022-05-02 at 11 37 56 AM
Screen Shot 2022-05-02 at 11 38 10 AM
Screen Shot 2022-05-02 at 11 38 19 AM


bistline (changes requested):
I can't get this to work correctly, I keep seeing this.  I ran `bundle install` and `yarn install`, both of which completed w/o error:

Screen Shot 2022-05-03 at 1 01 49 PM

ehanna4:
> I can't get this to work correctly, I keep seeing this. I ran `bundle install` and `yarn install`, both of which completed w/o error:
> 

Oops yep forgot a step in the setup - sorry! I will add that, although I am seeing a bunch of test failures as well so might move this back to draft while I try to fix that up..

app/javascript/lib/metrics-api.js
```diff
@@ -422,7 +423,15 @@ export function log(name, props = {}) {
...
   if ('SCP' in window || metricsApiMock) {
-    fetch(`${bardDomain}/api/event`, init)
+    fetch(`${bardDomain}/api/event/`, init).then(response => {
+      // log failed attempts to connect with Bard to Sentry
+      if (!response.ok) {
+        logJSFetchExceptionToSentry(response, 'failed to log event to Bard')
+      }
+    // log errored attempts to connect with Bard to Sentry
+    }).catch(error => {
+      logJSFetchErrorToSentry(error, 'failed to log event to Bard')
```
bistline, line 433: Non-blocking: It might be better to have a different log message here, like `error in logging event to Bard` to denote that the request threw an error, rather than a complete round-trip request that had a non-OK status code.

app/javascript/lib/scp-api.jsx
```diff
@@ -340,7 +341,15 @@ export async function fetchBucketFile(bucketName, fileName, maxBytes=null, mock=
...
 
-  const response = await fetch(url, init).catch(error => error)
+  const response = await fetch(url, init).then(response => {
+    // log failed attempts to access google storage to Sentry
+    if (!response.ok) {
+      logJSFetchExceptionToSentry(response, 'failed to connect to google storage')
+    }
+  // log errored attempts to access google storage to Sentry
+  }).catch(error => {
+    logJSFetchErrorToSentry(error, 'failed to connect to google storage')
```
bistline, line 351: Same comment as above - a slightly different log message might be nice.

app/javascript/lib/sentry-logging.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Determine if logging should occur based on environment
+ */
+function shouldLog() {
+  const env = getSCPContext().environment
+
+  // do not log for development or test environments
+  // to test locally comment out lines 45-47
+  if (['development', 'test'].includes(env)) {
```
bistline, line 45: I understand that you did this to make it easier to test, this is a little cleaner:
```suggestion
  return ['development', 'test'].includes(env)
```
For testing, you could simply write `return true` above that line to skip the `env` check.

On a related note, this would mean that we're logging errors from `staging`, which we may or may not want to do.  Typically, we only use services like Sentry in production.  If we want to keep that behavior, you can rewrite line 45 as:
```
return env === 'production'
```

bistline (approved):
Some non-blocking style comments, and a question about logging errors from `staging` instead of only `production`.

app/javascript/lib/metrics-api.js
ehanna4: oh good call - i'll update that (and the one below)

app/javascript/lib/sentry-logging.js
ehanna4: I can definitely make that change if that would be better - I was following what I had seen `ErrorTracker.rb` where it has the logic:

>     # only report to Sentry if configured and not in dev or test environment
>     context = format_extra_context(*context_objects)
>     if %w(development test).include?(Rails.env) || ENV['SENTRY_DSN'].nil?
>       Rails.logger.error "Suppressing error reporting to Sentry: #{exception.class.name}:#{exception.message}, context: #{context}"
>     else
>       Raven.capture_exception(exception, user: {identifer: extract_user_identifier(user)}, extra: context)
>     end

so figured I should match that but let me know if you think I should update this to only do Prod

devonbush (changes requested):
Looks great.  Only suggestion is to consolidate as much of the sentry logic as we can

app/javascript/lib/sentry-logging.js
bistline: No, that actually makes sense - let's leave it.

app/javascript/lib/scp-api.jsx
```diff
@@ -340,7 +341,15 @@ export async function fetchBucketFile(bucketName, fileName, maxBytes=null, mock=
   init.headers = new Headers(init.headers)
   const url = `https://storage.googleapis.com/download/storage/v1/b/${bucketName}/o/${fileName}?alt=media`
 
-  const response = await fetch(url, init).catch(error => error)
+  const response = await fetch(url, init).then(response => {
+    // log failed attempts to access google storage to Sentry
+    if (!response.ok) {
+      logJSFetchExceptionToSentry(response, 'error in connecting to google storage')
```
eweitz, line 347: Refining per other comments:

```suggestion
      logJSFetchExceptionToSentry(response, 'Error in fetch response when connecting to Google storage')
```

app/javascript/lib/scp-api.jsx
```diff
@@ -340,7 +341,15 @@ export async function fetchBucketFile(bucketName, fileName, maxBytes=null, mock=
...
+    }
+  // log errored attempts to access google storage to Sentry
+  }).catch(error => {
+    logJSFetchErrorToSentry(error, 'failed to connect to google storage')
```
eweitz, line 351:
```suggestion
    logJSFetchErrorToSentry(error, 'Error in JavaScript when to connecting to Google storage')
```

app/javascript/lib/sentry-logging.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @fileoverview Functions for logging JS errors to Sentry
+ *
+ */
+import * as Sentry from '@sentry/react'
+import getSCPContext from '~/providers/SCPContextProvider'
+
+/**
+ * Log an exception to Sentry for bad response JS fetch executions
+ * i.e. executions that results in a 404 response
```
eweitz, line 10: Blocker: with i.e. meaning "that is", this currently implies _only_ 404's are logged, which is incorrect.  

I'd go with e.g. ("for example"), but I defer to you as long as it's not i.e.

```suggestion
 * e.g. requests that result in a 404 response
```

app/javascript/vite/application.js
```diff
@@ -24,6 +27,31 @@ import * as ScpApi from '~/lib/scp-api'
 
 window.SCP = window.SCP ? window.SCP : {}
 
+// Initialize Sentry to enable logging JS errors to Sentry
+Sentry.init({
+  dsn: 'https://a713dcf8bbce4a26aa1fe3bf19008d26@o54426.ingest.sentry.io/1424198',
+  integrations: [new BrowserTracing()],
+
+  // send 100% of the transactions to Sentry since they will only occur on errors
+  tracesSampleRate: 1.0
```
eweitz, line 36: Out of curiosity, is there a setting that would let us do something like "Don't log more than 500 errors per user per day?"  DSP Sentry clients sometimes overlog, risking disabling Sentry for all other DSP clients until quota refresh, and we often only become aware of such incidents retroactively.

It'd be wonderful if Sentry clients like this had an easy way to cap or throttle their own logging _after_ reaching some client-defined threshold, to avoid such systemic risks for observability.

eweitz (changes requested):
This is a big observability improvement!

I note two blockers, both concerning documentation.  One is a subtle typo that could lead to substantive misunderstanding, and another is a proactive measure to avoid undue false-positive alarms regarding secrets.

Beyond that, I suggest some non-blocking messaging refinements, and wonder about Sentry log throttling.

ehanna4:
Re-running tests but the failure is in Rails which I did not alter in this PR so am pretty confident it is a false positive

app/javascript/lib/sentry-logging.js
```diff
@@ -0,0 +1,81 @@
...
+    return false
+  }
+
+  return true
+}
+
+/**
+ *  Initialize Sentry to enable logging JS errors to Sentry
+ */
+export function setUpSentry() {
```
devonbush, line 56:
```suggestion
export function setupSentry() {
```
I think 'setup' is more commonly used as a single word when in this context.  at least a case-sensitive search in our code yields a lot more matches for 'setup' than 'setUp'

devonbush (approved):
Good stuff

eweitz (approved):
Nice work!

PR #1483, opened 2022-05-02
Integration for differential expression PAPI jobs (SCP-4194)

bistline:
This update adds basic command-line integration of launching differential expression jobs from the Rails console using the `PapiClient` class.  This will now take the various inputs required and submit a job via the Pipelines API, which will localize files from the study bucket, run the calculations, and the delocalize outputs back to the bucket.

In addition, this adds test coverage for the `PapiClient` class to ensure against future updates breaking our PAPI submission infrastructure.  Also, extra parameters for differential expression jobs are now managed by the `DifferentialExpressionParameters` class, which will handle validating and setting CLI options for PAPI.

MANUAL TESTING
To fully validate that you can run a DE job on actual data, this will require creating a new study using the following example data (taken from [this](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP294/de-test-2) study on staging).  The example data can be downloaded from here:

[https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/SCP294_DE](https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/SCP294_DE?project=broad-singlecellportal-staging&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)

You will need to ingest at least the metadata, clustering, and raw counts matrix for this example to work.
1. Pull this branch, boot as normal, and start `Delayed::Job` workers
2. Set your `Ingest Pipeline Docker Image` to `gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_delocalize_de:dce674337152`
3. Create a new study using the above example data, and ingest required files (this should take 5-10 minutes total)
   *  You can get expression matrix information from `file_supplemental_info.tsv`, though it is not required for this to succeed
4. Once the data is ingested successfully, open up the "Study Details" page for this study and click the link to open the bucket in a browser
5. In a separate rails console session, run the following snippet to assemble the required parameters for the job, interpolating in the accession for your new study:
```
study = Study.find_by(accession: <new study accession>)
cluster_file = study.cluster_ordinations_files.first
metadata_file = study.metadata_file
user = study.user
action = :differential_expression
matrix_file = study.expression_matrix_files.first
options = {
  annotation_name: 'cell_type__ontology_label',
  annotation_type: 'group', 
  annotation_scope: 'study',
  annotation_file: metadata_file.gs_url,
  cluster_file: cluster_file.gs_url,
  cluster_name: cluster_file.name,
  matrix_file_path: matrix_file.gs_url,
  matrix_file_type: 'dense'
}
de_params = DifferentialExpressionParams.new(options)
``` 
6. Submit the job to PAPI:
```
client = ApplicationController.papi_client
job = client.run_pipeline(study_file: cluster_file, user: user, action: action, params_object: de_params)
```
7. To poll for completion, you can re-run the following snippet to ensure that the job completes:
```
pipeline = client.get_pipeline(name: job.name)
pipeline.done?
pipeline.error
```
8. Once `pipeline.done?` returns `true`, confirm that `pipeline.error` returns `nil`
9. Go back to the storage browser tab and validate that the DE output files appear in the bucket:
Screen Shot 2022-05-04 at 3 39 26 PM

This PR satisfies SCP-4194.

### DIFF
--- a/app/models/differential_expression_parameters.rb
+++ b/app/models/differential_expression_parameters.rb
@@ -1,0 +1,67 @@
+# class to hold parameters specific to differential expression jobs in PAPI
+class DifferentialExpressionParameters
+  include ActiveModel::Model
+
+  # regular expression to validate GS url format
+  GS_URL_REGEXP = %r{\Ags://}.freeze
+
+  # annotation_name: name of annotation to use for DE
+  # annotation_type: type of annotation, must be 'group'
+  # annotation_scope: scope of annotation (study, cluster)
+  # annotation_file: source file for above annotation
+  # cluster_file: clustering file with cells to use as control list for DE
+  # cluster_name: name of associated ClusterGroup object
+  # matrix_file_path: raw counts matrix with source expression data
+  # matrix_file_type: type of raw counts matrix (dense, sparse)
+  # gene_file (optional): genes/features file for sparse matrix
+  # barcode_file (optional): barcodes file for sparse matrix
+  attr_accessor :annotation_name, :annotation_type, :annotation_scope, :annotation_file, :cluster_file,
+                :cluster_name, :matrix_file_path, :matrix_file_type, :gene_file, :barcode_file
+
+  validates :annotation_name, :annotation_type, :annotation_scope, :annotation_file, :cluster_file,
+            :cluster_name, :matrix_file_path, :matrix_file_type, presence: true
+  validates :annotation_file, :cluster_file, :matrix_file_path,
+            format: { with: GS_URL_REGEXP, message: 'is not a valid GS url' }
+  validates :annotation_type, inclusion: %w[group]
+  validates :annotation_scope, inclusion: %w[cluster study]
+  validates :matrix_file_type, inclusion: %w[dense sparse]
+  validates :gene_file, :barcode_file,
+            presence: true,
+            format: {
+              with: GS_URL_REGEXP,
+              message: 'is not a valid GS url'
+            },
+            if: -> { matrix_file_type == 'sparse' }
+
+  # convert attribute name into CLI-formatted option
+  def self.to_cli_opt(param_name)
+    "--#{param_name.to_s.gsub(/_/, '-')}"
+  end
+
+  # default attributes hash
+  def attributes
+    {
+      annotation_name: annotation_name,
+      annotation_type: annotation_type,
+      annotation_scope: annotation_scope,
+      annotation_file: annotation_file,
+      cluster_file: cluster_file,
+      cluster_name: cluster_name,
+      matrix_file_path: matrix_file_path,
+      matrix_file_type: matrix_file_type,
+      gene_file: gene_file,
+      barcode_file: barcode_file
+    }.with_indifferent_access
+  end
+
+  # return array of all initialized attributes in Python CLI form, e.g. annotation_name => --annotation-name
+  # will also append --differential-expression at the end
+  def to_options_array
+    options_array = []
+    attributes.each do |attr_name, value|
+      options_array += [self.class.to_cli_opt(attr_name), value] if value.present?
+    end
+    options_array << '--differential-expression'
+    options_array
+  end
+end
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -23,20 +23,17 @@ class IngestJob
   attr_accessor :reparse
   # Boolean indication of whether or not to delete file from bucket on parse failure
   attr_accessor :persist_on_fail
-  # extra key/value storage of attributes
-  attr_accessor :options
 
   # number of tries to push a file to a study bucket
   MAX_ATTEMPTS = 3
 
   # Mappings between actions & models (for cleaning up data on re-parses)
   MODELS_BY_ACTION = {
-      ingest_expression: Gene,
-      ingest_cluster: ClusterGroup,
-      ingest_cell_metadata: CellMetadatum,
-      subsample: ClusterGroup,
-      differential_expression: nil # data only lives in files in the bucket, has special handling
-  }.freeze
+    ingest_expression: Gene,
+    ingest_cluster: ClusterGroup,
+    ingest_cell_metadata: CellMetadatum,
+    subsample: ClusterGroup
+  }
 
   # Push a file to a workspace bucket in the background and then launch an ingest run and queue polling
   # Can also clear out existing data if necessary (in case of a re-parse)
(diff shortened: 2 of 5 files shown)


app/models/ingest_job.rb
```diff
@@ -23,6 +23,8 @@ class IngestJob
   attr_accessor :reparse
   # Boolean indication of whether or not to delete file from bucket on parse failure
   attr_accessor :persist_on_fail
+  # extra key/value storage of attributes
+  attr_accessor :options
```
devonbush, line 27: I know this is just a draft, but I started reviewing it without realizing that first :).  Instead of creating a generic 'options' hash, consider creating a class `differential_expression_params` and including one here (sort of similar to how we embed `expression_file_info` inside StudyFile.  That way all the documentation, parsing, and validation logic for those params can live in a more easily findable place

bistline:
Build failure is strange, there are no failed/errored tests listed at the end but some `MockExpectationError` were thrown during the run that I've not seen before.  Local testing could not reproduce test failures.  I'm rerunning to see if it happens again.

app/models/papi_client.rb
```diff
@@ -9,18 +9,30 @@
...
+    ingest_subsample: ['Cluster'],
+    differential_expression: ['Cluster']
+  }.freeze
+
+  # list of parameter names allowed for differential_expression calls
+  DE_PARAMETER_NAMES = %w[
+    annotation_name annotation_type annotation_scope annotation_file
+    cluster_file cluster_name matrix_file_path matrix_file_type
+  ].freeze
+
```
devonbush, line 33: I think this might be a case where a small class (possibly embeddable if we ever need to store it along with an ingest_job), akin to expression_file_info.rb, to store these parameters will really help clarify them, since then it's easier to inline descriptions of what each parameter does.  Likewise, then instead of having a generic `extra_options` parameter which requires some code-sleuthing to figure out what it accepts, the `get_command_line` function can just take an optional `differential_expression_params` object.

app/models/papi_client.rb
```diff
@@ -324,6 +351,54 @@ def get_command_line_options(study_file, action)
...
+  #
+  # * *params*
+  #   - +options+ (Hash) => Hash of DE-specific parameters, whose names must include all values from DE_PARAMETER_NAMES
+  #
+  # * *returns*
+  #   - (Array) => Array of DE-specific command line arguments to pass to Docker in "exec" form
+  #
+  # * *raises*
+  #   - (ArgumentError) => if any required parameters from DE_PARAMETER_NAMES are missing or malformed
+  def get_de_parameters(options)
```
devonbush, line 364: yeah, if we have a `DifferentialExpressionParams` class, this method will get much simpler if it gets to deal with an object with fields rather than having to match keys/names with `options.detect`. That `DifferentialExpressionParams` class can have it's own `valid?` or `validate` method, which will have the advantage of moving the DE-specific validation logic into one place, and this method can just focus on python command line generation

app/models/papi_client.rb
```diff
@@ -324,6 +351,54 @@ def get_command_line_options(study_file, action)
...
+
+  # transform a parameter from Hash key or String into a representation that can be called in the Python CLI
+  # converts underscores (_) to dashes (-) and prepends double dashes (--)
+  #
+  # * *params*
+  #   - +param_name+ (Symbol, String) => Name of parameter
+  #
+  # * *returns*
+  #   - (String) => New parameter name, encoded as kwarg for Python command line
+  def to_cli_opt(param_name)
```
devonbush, line 398: For easier testing without having to instantiate the class
```suggestion
  def self.to_cli_opt(param_name)
```

test/models/papi_client_test.rb
```diff
@@ -0,0 +1,254 @@
...
+    assert_equal actions, pipeline_request.pipeline.actions
+    assert_equal environment, pipeline_request.pipeline.environment
+    assert_equal resources, pipeline_request.pipeline.resources
+    assert_equal exp_cmd, pipeline_request.pipeline.actions.commands
+  end
+
+  test 'converts keys into cli options' do
+    assert_equal '--annotation-name', @client.to_cli_opt(:annotation_name)
+    assert_equal '--foo', @client.to_cli_opt('foo')
+  end
```
devonbush, line 253: These are all extremely readable tests, which is impressive given the somewhat arcane nature of what the code accomplishes.  Nice work!

devonbush (changes requested):
Good stuff--kudos on the exceptionally straightforward tests, they show both test-writing skill and that the code you've written is reasonably modular.  I'm willing to be convinced otherwise, but I strongly suspect a class for the differential expression params will make this code simpler and easier to read.

app/models/papi_client.rb
bistline: Yeah ok I'll bite on that - I can see a non-persisted `ActiveModel` based class for this, akin to `FireCloudProfile` (something that only lives in memory but still acts like a model).

app/models/papi_client.rb
bistline: [8f7fef6](https://github.com/broadinstitute/single_cell_portal_core/pull/1483/commits/8f7fef66d1357c74e4dea90ce08a41f8fb81ae98)

app/models/papi_client.rb
bistline: [8f7fef6](https://github.com/broadinstitute/single_cell_portal_core/pull/1483/commits/8f7fef66d1357c74e4dea90ce08a41f8fb81ae98)

app/models/differential_expression_parameters.rb
```diff
@@ -0,0 +1,67 @@
+# class to hold parameters specific to differential expression jobs in PAPI
+class DifferentialExpressionParameters
+  include ActiveModel::Model
+
+  # regular expression to validate GS url format
+  GS_URL_REGEXP = %r{\Ags://}.freeze
```
devonbush, line 6: This constant might be more at home in `GoogleServiceClient`

devonbush (approved):
Awesome, great use of model-validates hooks.  One optional suggestion

app/models/differential_expression_parameters.rb
bistline: Maybe, but that class is very service account-specific.  This is certainly related, but really just deals with string values - it's not actually validating that the file is there.  If we end up with more parameters classes like this then I think it makes sense to abstract into a module like this.

eweitz (approved):
Code looks good!  

I suggest a few PAPI argument name refinements that I think would help maintainability and debuggability.   @jlchang, what do you think?  (I don't consider the suggestions blocking.)

app/models/differential_expression_parameters.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  # annotation_name: name of annotation to use for DE
+  # annotation_type: type of annotation, must be 'group'
+  # annotation_scope: scope of annotation (study, cluster)
+  # annotation_file: source file for above annotation
+  # cluster_file: clustering file with cells to use as control list for DE
+  # cluster_name: name of associated ClusterGroup object
+  # matrix_file_path: raw counts matrix with source expression data
+  # matrix_file_type: type of raw counts matrix (dense, sparse)
```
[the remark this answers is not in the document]
bistline: These are what the parameters are named in `scp-ingest-pipeline`, so I'm trying to maintain a 1:1 mapping to avoid confusion.

app/models/differential_expression_parameters.rb
```diff
@@ -0,0 +1,67 @@
+# class to hold parameters specific to differential expression jobs in PAPI
+class DifferentialExpressionParameters
+  include ActiveModel::Model
+
+  # regular expression to validate GS url format
+  GS_URL_REGEXP = %r{\Ags://}.freeze
+
+  # annotation_name: name of annotation to use for DE
+  # annotation_type: type of annotation, must be 'group'
```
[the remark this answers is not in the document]
bistline: That's a fair point - and you're right in that it is basically only a validation to avoid calling on a numeric annotation.  Mostly, I want to maintain parity between the interface here and in `scp-ingest-pipeline`, and have parameter names line up.  It seems somewhat strange to me to have an upstream validation on a parameter value when the rest are validated here.  I dunno - I could go either way, and this implementation has the benefit of being internally consistent and very compact.  But I agree it's weird to only allow a single value as an input.

PR #1484, opened 2022-05-03
fixing typos in boot_docker

devonbush:
Bunch of argument ordering typos when I was adding MIXPANEL_SECRET to boot_docker.  This also adds the MIXPANEL secret to the cron env, even though it is not needed there yet, it probably makes sense to keep that environment at parity for simplicity 

TO TEST:
 build dockerized SCP, confirm it works (alternatively, and probably faster/better, merge to staging and check there)


eweitz (approved):

ehanna4 (approved):

devonbush:
merging prior to test completion as staging is down, and this script is not used in tests

PR #1489, opened 2022-05-06
adding all scp_config environment variables to cron (SCP-4286)

devonbush:
This will, in part, help prevent outages like the ingest outage we had a couple weeks ago, by ensuring that cron has access to the same set of variables our regular production environment does.  
I briefly experimented with converting this to a Ruby script, but the value-add seemed low -- because the logic is very simple, the script wasn't especially more comprehensible, and the cost of retesting everything would be high.  So I kept it as-is.  

TO TEST: (not recommended locally, this is *much* easier and more accurate to test in staging)
1. run portal in dockerized mode
2. shell into the container
3. add a file `delayed_job.9.pid` to the `/home/app/webapp/tmp/pids` directory with a non-existent process number (e.g. 99596)
4. wait until the cron job restarts the delayed job process
5. confirm file ingests either work, or at least give an email on failure


rails_startup.bash
```diff
@@ -40,15 +40,17 @@ elif [[ $PASSENGER_APP_ENV = "development" ]]; then
...
-echo "export PROD_DATABASE_PASSWORD='$PROD_DATABASE_PASSWORD'" >| /home/app/.cron_env
-echo "export SENDGRID_USERNAME='$SENDGRID_USERNAME'" >> /home/app/.cron_env
-echo "export SENDGRID_PASSWORD='$SENDGRID_PASSWORD'" >> /home/app/.cron_env
-echo "export MONGO_LOCALHOST='$MONGO_LOCALHOST'" >> /home/app/.cron_env
-echo "export MONGO_INTERNAL_IP='$MONGO_INTERNAL_IP'" >> /home/app/.cron_env
-echo "export SECRET_KEY_BASE='$SECRET_KEY_BASE'" >> /home/app/.cron_env
-echo "export GOOGLE_CLOUD_PROJECT='$GOOGLE_CLOUD_PROJECT'" >> /home/app/.cron_env
-echo "export MIXPANEL_SECRET='$MIXPANEL_SECRET'" >> /home/app/.cron_env
+echo "# cron environment setup" >| /home/app/.cron_env
+for ENV_VAR in GA_TRACKING_ID GCP_NETWORK_NAME GCP_SUB_NETWORK_NAME LOGSTASH_HOST MIXPANEL_SECRET MONGODB_ADMIN_PASSWORD \
```
bistline, line 46: Non-blocking: While this works fine for the scope of this PR, I think this is going to lead to the exact same issue as before - we'll add an environment variable, then forget to add it here, and `cron` will be missing required data.

What might be more robust (though totally out of scope as it would be a large refactoring) would be to prepend all environment variables with `SCP_`.  That way, we could simply have a regular expression to match all SCP-related secrets from the environment, and then automatically add them to `.cron.env`.

bistline (approved):

eweitz (approved):
Code looks good overall -- nice refinements.

I suggest some maintainability improvements, no blockers.

rails_startup.bash
```diff
@@ -40,15 +40,17 @@ elif [[ $PASSENGER_APP_ENV = "development" ]]; then
...
+  GOOGLE_CLOUD_KEYFILE_JSON GOOGLE_PRIVATE_KEY GOOGLE_CLIENT_EMAIL GOOGLE_CLIENT_ID GOOGLE_CLOUD_PROJECT \
+  MONGODB_ADMIN_USER MONGO_INTERNAL_IP MONGO_LOCALHOST NEWRELIC_AGENT_ID OAUTH_CLIENT_ID OAUTH_CLIENT_SECRET PORTAL_NAMESPACE \
+  PROD_DATABASE_PASSWORD PROD_HOSTNAME SECRET_KEY_BASE SENDGRID_PASSWORD SENDGRID_USERNAME TCELL_AGENT_API_KEY \
+  TCELL_AGENT_APP_ID T_CELL_SERVER_AGENT_API_KEY
```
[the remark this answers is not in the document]
devonbush: I'll add a comment.  Some of these come from scp_config, others come from other places.  I don't think we have access to scp_config.json here -- this is running inside the container.  scp_config.json is never sent to the server, it only is downloaded to the jenkins machine

rails_startup.bash
devonbush: yeah, I like that idea -- we should keep that in mind

PR #1496, opened 2022-05-17
"My studies" refactor (SCP-4284)

devonbush:
This updates the My Studies dashboard to use react/API infrastructure.  The backend also no longer has an N+1 query issue, so it should load much faster for users with a large number of studies (previously, a separate DB query was made for every study to get the email address of the author)
(There are also two additional igv.css.map fixes added to routes.rb to reduce those errors in development)

image

TO TEST:
1. load the "My studies" page
2. confirm table loads, sorts, and filters
3. confirm study actions work as before.


eweitz:
Could you add a description and screenshots?

app/javascript/components/my-studies/MyStudiesPage.jsx
```diff
@@ -0,0 +1,276 @@
...
+  sortType: (rowA, rowB) => {
+    return rowA.original.name.localeCompare(rowB.original.name, 'en', { numeric: true, ignorePunctuation: true })
+  },
+  disableFilters: true
+}, {
+  Header: 'Created ',
+  accessor: 'createdAt',
+  Cell: ({ value }) => {
+    const date = new Date(value)
+    return `${date.getFullYear()}-${date.getMonth()}-${date.getDate()}`
```
eweitz, line 66: Blocker: For dates, let's use Terra’s style — <3-letter month> <unpadded day>, <year> — or ISO-8601.  The screenshot’s unpadded ISO-8601 style is unfamiliar to both technical folks (ISO-8601 is padded, e.g. 2022-05-18 instead of 2022-5-18) and non-technical folks.

```suggestion
    return `${date.toISOString().split('T')[0]}`
```

eweitz (changes requested):
Good stuff!  I'm glad this page will be faster.

I note a trivial user-facing blocker; we should use a more conventional date style.

app/javascript/components/my-studies/MyStudiesPage.jsx
devonbush: done.  Also, I upgraded the text ellipsis styling/cutoff per your comments from demo

ehanna4:
Screen Shot 2022-05-19 at 4 37 02 PM

Can somebody else check if the table looks right on their local? I'm confused why it looks so janky on mine since I know at demo Devon showed it and it looked good...

devonbush:
Good catch, Emily, I'd made a tweak to the truncation style, and it looks like it broke for certain description sizes.  I've fixed it now

ehanna4 (changes requested):
The double-arrow buttons do not appear to be working correctly

https://user-images.githubusercontent.com/54322292/169556647-65333f43-8486-4d3f-99d0-44238e578e94.mov

app/controllers/api/v1/studies_controller.rb
```diff
@@ -48,6 +48,14 @@ def firecloud_independent_methods
       # GET /single_cell/api/v1/studies
       def index
         @studies = Study.editable(current_api_user)
+        study_owner_ids = @studies.pluck(:id, :user_id)
+        user_info = User.where(:id.in => study_owner_ids.map{ |a| a[1] }).pluck(:id, :email)
```
ehanna4, line 52: It might be helpful for later to name `a` something a bit more descriptive

ehanna4 (commented):
small nit and blocking on the arrows working but otherwise looks good !

ehanna4 (approved):
👍 looks good thanks!

app/views/studies/index.html.erb
```diff
@@ -1,3 +1,8 @@
+<div id="my-studies-content"></div>
+<script type="text/javascript" nonce="<%= content_security_policy_script_nonce %>">
+  window.SCP.renderComponent('my-studies-content', 'MyStudiesPage')
+</script>
+<!--
```
bistline, line 5: Blocker: despite this being commented out, the code is still executing.  You can prove this by going into `app/controllers/studies_controller.rb#L39` and commenting out line `@studies = Study.accessible(current_user).to_a`.  This will then throw an error on page reload.

Furthermore, that query is still happening in the background, causing the slowdown to still occur in loading all studies (especially for admins).  I think this new view is great, and there's no reason to keep around the legacy code.  It can still be found on older commits if we really need it, which I doubt.

bistline (approved):
I'm approving preemptively, modulo my blocker, as it only asks for code to be deleted, not refactored.  We need to remove the legacy view/controller code so that it does not hinder performance when loading this page.  If we don't do that, it will actually render slower than it did before, which seems to defeat the purpose altogether.

app/views/studies/index.html.erb
devonbush: Great catch!

PR #1497, opened 2022-05-17
Moving studies_controller to StudyAware (SCP-3684)

devonbush:
This updates api/v1/studies_controller to be STudyAware compatible, which means it can take either accessions or ids as the url parameter.  This then updates the new study usage page to use accession rather than id.

TO TEST:
0. open the chrome network tab 
1. go to the study settings page for a study.  Click "usage stats"
2. in the network tab, observe that a request went out for `/single_cell/api/studies/SCPXXX/usage_stats`  (as opposed to  `/single_cell/api/studies/[[study_id]]/usage_stats` 
3. Load the explore tab for a study, and confirm that clusters still load

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -8,9 +8,11 @@ module Api
         [:index, :show, :file_info]
       end
 
+      before_action :set_study_id_param
       before_action :authenticate_api_user!
       before_action :set_study, except: [:index, :create]
-      before_action :check_study_permission, except: [:index, :create, :generate_manifest]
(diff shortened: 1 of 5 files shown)


eweitz (approved):
Thanks for following through!   This will help maintainability.

It might not be in scope here, but does anything come to mind that would prevent using accessions instead of MongoDB object IDs in the URLs for e.g. the study details page, upload wizard, and sync page?

eweitz:
(For auditability, please add Jira ticket to PR title.)

devonbush:
Not that I know of -- I think the solution will be similar, where the controllers will have to do some param copy magic

ehanna4 (approved):

PR #1498, opened 2022-05-17
updating default color scale behavior (SCP-4365)

devonbush:
This repurposes the study default color profile to apply to all continuous data scatter plots, and not just the default (numeric) annotation.  It also cleans up some cruft in clusters_controller and ScatterPlot -- there was still code floating around as if we were returning the colorscale to use along with the plot itself.  That's not something we do anymore (and we don't want to -- we want the color scale to be owned by the frontend).

TO TEST:
1. load a study, click around expression and numeric cluster plots and confirm the default 'Reds' colorscale is used
2. go to Study Settings, and change the default color profile to something else
3. Reload the page (take care that there are no leftover extra URL params)
4. Confirm expression and numeric annotation plots load with the selected color scale

### DIFF
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -167,14 +167,6 @@ module Api
           include_annotation = data_fields.include?('annotation')
           include_cells = data_fields.include?('cells')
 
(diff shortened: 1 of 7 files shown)


jlchang (approved):
Functional review - behaves as described.

PR #1504, opened 2022-05-24
adding override to annotation viz limit values (SCP-4372)

devonbush:
This adds a new study default options field 'override_viz_limit_annotations' which is an array of annotations to show regardless of whether they exceed the viz limit threshold.  This will be needed for the U19 study.  Right now, this property is not accessible via the UX -- it's complex and rare enough that I think it makes the most sense to make it require SCP support to set.  Obviously, if this becomes a frequent request, we can look at the best place to expose it in our UX, but for now I don't think it would be worth either the effort or the extra complexity this would bring to the upload UX.  

TO TEST:
1. populate the new synthetic study `SyntheticStudyPopulator.populate('mouse_multilabel')`
2.  Observe that the 'manylabel_field' is not visualizable in the explore tab
3. In the rails console, run 
 ```
    s = Study.find_by(accession: '<<accession of the new study')
    s.add_override_viz_limit_annotation('manylabel_field')
```
4. After that operation completes, confirm that the 'manylabel_field' is now visualizable in the explore tab

### CODE REVIEW

File: app/models/study.rb [hunk 1]
```diff
@@ -1192,6 +1192,28 @@ def default_color_profile
...
+    self.default_options[:override_viz_limit_annotations] || []
+  end
+
+  # make an annotation visualizable despite exceeding the default values limit
+  def add_override_viz_limit_annotation(annotation_name)
+    cell_metadatum = self.cell_metadata.find_by(name: annotation_name)
+    if cell_metadatum
+      # we need to populate the 'values' array, since that will not have been done at ingest
+      uniq_vals = cell_metadatum.concatenate_data_arrays(annotation_name, 'annotations').uniq
+      cell_metadatum.update!(values: uniq_vals)
```
bistline, line 1206: Blocker: the reason we stopped doing this is because some annotations (like `biosample_id`) can have 100K+ values, and can eventually blow out the MongoDB 16MB document size limit.  We saw this with Eran's data, which was when we introduced that change.  Unfortunately, we're going to simply have to read these out of the database every time and not store them in the `values` array.  Given the potentially low amount of times this will be used, I think that tradeoff is acceptable.

bistline (changes requested):
Unfortunately, this implementation is prone to a MongoDB exception that we cannot get around.  We'll need to refactor this to not attempt to store the unique `values` data and simply read it out of the database at render time.

app/models/study.rb
devonbush: that's a valid concern.  The problem with reading them every time is that for Jonah's study, that's reading a 10MM value array into memory.  Which, based on what we've seen from Jean's testing, is a ~45-second operation.  And that's during the initial fetch of explore data, so that's a 45-second wait before a user gets to even start waiting for their first graph to load.   So, I think our options are (1) assuming that any column that we run this on will fit within the 16MB limit (we'll only be running it on columns specifically requested by the study author, so we don't have to worry about, e.g. biosample_id) or (2) finding some alternate place to store these arrays -- maybe just park it in a file somewhere?    I'd lean towards (1), at least until we confirm it won't work for Jonah's study, and then switch to (2) if it doesn't

devonbush:
Just added the dynamic construction fallback and better error-handling we talked about.  Fun fact that I discovered while testing this:  AnnotationVizService.available_metadata_annotations gets called 3 seperate times when a study gets rendered on the explore tab.  First to render the html for the Settings tab, then for the explore API response, and then lastly when attempting to fetch any user-specific annotations.  So computing the values dynamically definitely should be a last resort

eweitz (approved):
Code and functionality look good!

Screen Shot 2022-05-31 at 11 20 53 AM

app/models/study.rb
```diff
@@ -1196,6 +1196,33 @@ def default_color_profile
...
+  # make an annotation visualizable despite exceeding the default values limit
+  def add_override_viz_limit_annotation(annotation_name)
+    cell_metadatum = self.cell_metadata.find_by(name: annotation_name)
+    if cell_metadatum
+      # we need to populate the 'values' array, since that will not have been done at ingest
+      begin
+        uniq_vals = cell_metadatum.concatenate_data_arrays(annotation_name, 'annotations').uniq
+        cell_metadatum.update!(values: uniq_vals)
+      rescue => e
+        Rails.logger.error "Could not cache unique annotation values: #{e.message}"
```
bistline, line 1213: Non-blocking: since this will be run from the Rails console, it would be more helpful to log these with `puts` so that the user can see the error.

bistline (approved):
Looks good, minor logging suggestion.

PR #1502, opened 2022-05-23
Differential expression MVP (SCP-4321)

eweitz:
This matures the prototype differential expression (DE) UX from #1486 into a minimally viable product.

Previously, the prototype supported exactly one annotation and had a slightly clunky UX.  Now the DE feature supports multiple annotations that have manually-validated DE analyses: users can pick from 1 of 6 such annotations.  Beyond the expanded data, the UX now has lower interaction cost, greater accessibility, more intuitive and robust layout, richer analytics, more refined code, and an automated test.  These help towards [previously outlined](https://github.com/broadinstitute/single_cell_portal_core/pull/1486) scientific product goals.

Here's how it looks.

https://user-images.githubusercontent.com/1334561/169854738-bfa4c08c-9a5e-47db-bbbd-a68bfd897e13.mov

If you haven't run manual tests from #1486:
1.  Load [human genome annotation in species metadata](https://broadinstitute.slack.com/archives/CBEHTH601/p1651603466458739?thread_ts=1651603449.817109&cid=CBEHTH601), if you haven't already
2.  Start delayed job runner: execute `rails jobs:work` in a terminal
3.  Run migration: `rake db:migrate`
4.  Load public test DE study: `echo "SyntheticStudyPopulator.populate('human_milk_de_49k_cell')" | bundle exec rails console -e development`.
5.  In output of above, copy the `bucket_id` value
6.  Continue in steps below, but omit the first

If you have run manual tests from #1486:
1.  Get the bucket ID for "Human milk - differential expression" study in your local SCP
2.  Remove old prototype DE data: `rm -r gs://<your local DE study's bucket_id>/_scp_internal*` (it's OK if this errs)
3. Copy the reference public test DE study's `_scp_internal` directory to your new study's bucket: `gsutil cp -r "gs://broad-singlecellportal-public/test/studies/SCP1671/_scp_internal" "gs://<your local DE study's bucket_id>"`
4.  Sign in to your local SCP, open menu at top-right of page, select "Feature flags"
9.  Find your local DE study, set `differential_expression_frontend` feature flag to `true`
10. Sign out
9.  Go to Study Overview for your test DE study, "Human milk - differential expression"
10. Go to annotation "cell_type__ontology_label"
11.  Go to Explore tab, refresh, click "Differential expression" button
13.  In resulting modal, select group "epithelial cell"
14. Confirm that a table of genes is shown in right panel
15. Click gene "CSN2" in DE panel
16. Confirm CSN2 gene is searched
[...]

PR #1505, opened 2022-05-24
Pin DE pilot to default home page study list (SCP-4366)

eweitz:
This pins our differential expression pilot to the home page, to help assess user interest in differential expression.

The DE pilot -- [SCP1671](https://singlecell.broadinstitute.org/single_cell/study/SCP1671), "Cellular and transcriptional diversity over the course of human lactation" -- will be the second study in the study list shown upon loading the home page.  It will only be pinned if the user hasn't searched.  

By making the study more prominent, we hope to increase its traffic and thus faster measure if users engage with the new differential expression MVP (#1502).  Otherwise, we risk getting a false-positive lack of signal, or especially noisy engagement rates due to low-cardinality data.

To test:
1. Go to default home page, without searching anything
2. Go to last page, get accession for a study in it -- we'll call this the DE pilot here
3. In `app/controllers/api/v1/search_controller.rb` line 373, comment out `Rails.env.production? && `
4. Replace SCP1671 with accession from step 2
5. Go to default home page, without searching anything
6. Confirm study originally on last page is now on first page
7. Do a keyword search
8. Confirm "DE pilot" study is not 2nd result
9. Do a faceted search
10. Confirm "DE pilot" study is not 2nd result

This satisfies SCP-4366.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -368,6 +368,15 @@ module Api
           end
         end
 
+        # Pin public DE pilot study to second row of home page default study list
+        # TODO (SCP-4374): Remove this block once frontend uses is_differential_expression_enabled
+        if Rails.env.production? && !params[:terms].present? && !@facets.any?
+          de_accession = "SCP1671"
+          differential_expression_study = @studies.find {|study| study.accession == de_accession}
+          @studies.delete_if {|study| study.accession == de_accession}
+          @studies.insert(1, differential_expression_study)
+        end
+
         @matching_accessions = @studies.map { |study| self.class.get_study_attribute(study, :accession) }
 
         logger.info "Final list of matching studies: #{@matching_accessions}"


app/controllers/api/v1/search_controller.rb
```diff
@@ -368,6 +368,15 @@ def index
           end
         end
 
+        # Pin public DE pilot study to second row of home page default study list
+        # TODO (SCP-4374): Remove this block once frontend uses is_differential_expression_enabled
+        if Rails.env.production? && !params[:terms].present? && !@facets.any?
```
bistline, line 373: Looks good, just some linting suggestions:
```suggestion
        if Rails.env.production? && params[:terms].blank? && @facets.empty?
          de_accession = "SCP1671"
          differential_expression_study = @studies.find { |study| study.accession == de_accession }
          @studies.delete_if { |study| study.accession == de_accession }
          @studies.insert(1, differential_expression_study)
        end
```

bistline (approved):
Looks fine, and works as advertised.  Not crazy about the hard-coded accession, but given the turnaround time and measurement needs it seems appropriate.  

It would be interesting to have a general "pinning" feature for future feature development.  We used to do this somewhat with the `view_order` study attribute, but a more generalized and extensible feature could be nice for instances when we want to roll out a sandboxed feature and get metrics on usage before making it more generally available.

app/controllers/api/v1/search_controller.rb
eweitz: Are these linting suggestions you get in RubyMine, @bistline?  If so, I can tinker with my VS Code Ruby linter to hopefully iron out such wrinkles before my next review.  (I pulled this syntax from close above.)

PR #1508, opened 2022-05-31
Hotfix: Ignore pinning DE study if viewing collection (SCP-4388)

bistline:
This is a hotfix to ignore pinning the differential expression example study to empty search results when viewing a collection.  The bug is that this study does not exist in any collection other than Alexandria on production, and as such will insert a `nil` object into the array of studies, which then errors out when attempting to correctly sort/weight results.

MANUAL TESTING
1. Pull branch and boot as normal
2. If you do not have any collections, run `SyntheticBrandingGroupPopulator.populate_all` (you will need DelayedJob running for this)
3. In `app/controllers/api/v1/search_controller.rb`, line 373, remove `Rails.env.production? &&`
4. Click "Browse collections" and select any collection
5. Confirm that the home page loads studies w/o error

This PR satisfies SCP-4388.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -370,7 +370,7 @@ module Api
 
         # Pin public DE pilot study to second row of home page default study list
         # TODO (SCP-4374): Remove this block once frontend uses is_differential_expression_enabled
-        if Rails.env.production? && params[:terms].blank? && @facets.empty?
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Looks good, I confirmed fix locally.

jlchang (approved):
Functional review: verified that hotfix resolves the inability to load studies in the collection.

PR #1510, opened 2022-06-01
Filtering most expression-based visualizations by cells appearing in matrix (SCP-2486)

bistline:
Currently, Single Cell Portal only stores non-zero expression values when ingesting a dense expression matrix.  All zeros are discarded along with their associated cell names for that particular gene, and the zeros are then inferred at render time by using the list of cells from the given cluster, and comparing that against the retrieved expression data.  If the cell name does not appear in the list, then its expression is inferred as `0`.  While this is very often true, it is also possible that the given cell was not observed at all in the expression matrix, in which case the real expression value is `null` or `NaN`.  This has the unintended consequence of skewing distribution-based plots by over-inflating the true amount of `0` expression values, causing median and quartile values to be lower than they normally would be.

This update adds filtering to most (but not all) expression-based plots to account for this.  Now, when a given gene/cluster/annotation is requested, we take the intersection of cells from the matrix and cluster, and only provide expression values for that intersection.  Cells not appearing in the matrix are discarded from distribution-based plots, and for normal expression scatter plots, the value is set to `null`.

So far, this covers all distribution-based plots, such as violin, dot plots, and heatmaps, as well as any multi-gene query where the user collapses by a summary statistic.  2D annotation-based scatter plots (for numeric annotations vs. gene expression) and correlation plots are currently not covered in this feature due to the difficulty in maintaining associations between coordinate, expression, and annotation data based on how the data is assembled for those visualizations.  It would be possible to do, but likely have larger performance impacts than the above scenarios, so it is being exempted for now.

SCREENSHOTS

Normal single-gene expression scatter:
Screen Shot 2022-06-01 at 3 17 31 PM

Single-gene violin plot (3rd donor ID has been completed removed, compared to reference plot above):
Screen Shot 2022-06-01 at 3 17 40 PM

Multi-gene dot plot (3rd donor ID has been completely removed, compared to reference plot above):
Screen Shot 2022-06-01 at 3 17 56 PM

Multi-gene heatmap (3rd donor ID has been completed removed, compared to reference plot above):
Screen Shot 2022-06-01 at 3 18 02 PM

MANUAL TESTING
1. Pull branch, boot as normal, and start DelayedJob workers
2. In a rails console session, run `SyntheticStudyPopulator.populate('exp_cell_filtering')`
3. Once example files have ingested, load the explore tab for the above study and select `donor_id` as the annotation
4. Query for the gene `Foo`, and confirm the expression-based scatter and violin plots match above screenshots
5. Additionally query for `Bar`, and confirm the dotplot and heatmap match above screenshots

This PR satisfies SCP-2486


app/lib/annotation_viz_service.rb
```diff
@@ -83,6 +83,24 @@ def self.populate_annotation_by_class(source:, scope:, type:)
...
+  # this is required for all expression-based scenarios to allow filtering of data by observed cells in matrices
+  # without the annotation map, array index position becomes unreliable and annotation assignments will drift
+  def self.get_annotation_as_hash(study, cluster, annotation, subsample_threshold = nil)
+    subsample_annotation = "#{annotation[:name]}--#{annotation[:type]}--#{annotation[:scope]}"
+    case annotation[:scope]
+    when 'study'
+      study.cell_metadata.by_name_and_type(annotation[:name], annotation[:type])&.cell_annotations || {}
+    else
+      # cluster- and user-based annotations behave the same in this case
+      cells = cluster.concatenate_data_arrays('text', 'cells', subsample_threshold, subsample_annotation)
```
devonbush, line 96: BLOCKER:  improve performance of the method by taking a (possibly optional) cluster_cells argument to this `get_annotation_as_hash` method. In the places where this method is called, it looks like we've already run the cluster.concatenate_data_arrays('text', 'cells', subsample_threshold, subsample_annotation) query, so we can save running it twice

app/models/gene.rb
```diff
@@ -62,11 +58,6 @@ def species
   # CLASS INSTANCE METHODS
   ##
 
-  # find if a study has a given gene quickly for search gating
-  def self.study_has_gene?(study_id:, expr_matrix_ids:, gene_name:)
-    Gene.where(study_id: study_id, :study_file_id.in => expr_matrix_ids).any_of({name: gene_name},{searchable_name: gene_name.downcase}).exists?
-  end
-
   # calculate a mean value for a given gene based on merged expression scores hash
```
devonbush, line 61: nice removal of the unused study_has_gene? method

app/models/study.rb
```diff
@@ -1354,6 +1354,19 @@ def all_expression_matrix_cells
     all_cells.uniq # account for raw counts & processed matrix files repeating cell names
   end
 
+  # array of all cell names from non-raw count matrix files
+  # used in filtering distribution-based expression visualizations (violin/dot plots)
+  def processed_matrix_cells
```
devonbush, line 1359: it's not strictly related to visualization, but I would consider moving this method to `expression_viz_service`.  I could see it going either way

app/models/study.rb
```diff
@@ -1433,6 +1446,11 @@ def expression_matrices
     self.study_files.where(:file_type.in => ['Expression Matrix', 'MM Coordinate Matrix'])
   end
 
+  # get all processed (i.e. non-raw count) matrix files
```
devonbush, line 1449: This is a great study method to add -- nice

test/integration/lib/expression_viz_service_test.rb
```diff
@@ -352,4 +352,23 @@ def load_all_genes(study)
...
+                              name_prefix: 'Filter test',
+                              user: @user,
+                              test_array: @@studies_to_clean)
+    matrix = FactoryBot.create(:study_file,
+                               name: 'dense.txt',
+                               file_type: 'Expression Matrix',
+                               study: study)
+    # we have to manually create the "all cells" array as we don't have a Factory for this yet
+    DataArray.create!(name: 'dense.txt Cells', array_type: 'cells', linear_data_type: 'Study', linear_data_id: study.id,
+                      values: %w[A b c], study_id: study.id, study_file_id: matrix.id, cluster_name: 'dense.txt Cells',
```
devonbush: changing 'b' and 'c' in the `values` array to 'f' and 'g' will make it a little clearer what's going on, and that those cells are not meant to match.

test/integration/lib/expression_viz_service_test.rb
```diff
@@ -352,4 +352,23 @@ def load_all_genes(study)
...
+                      cluster_group_id: nil, array_index: 0, subsample_annotation: nil, subsample_threshold: nil)
+
+    study.reload
+    filtered_cells = ExpressionVizService.filter_cells_for_plot(study, %w[A B C])
+    assert_equal %w[A], filtered_cells
```
devonbush, line 370: BLOCKER: This is a great test of the partial match capability.  Let's extend this to cover the full match case and the no-match case.  e.g. check that    
`ExpressionVizService.filter_cells_for_plot(study, %w[A b c])`
and 
`ExpressionVizService.filter_cells_for_plot(study, %w[q r s])`
return what you expect

devonbush (commented):
Nice--overall approach is great.  Couple of blockers around avoiding double-fetching and expanding test coverage, and then some scattered optional style suggestions.

devonbush (changes requested):
see comments above--clicked the wrong radio button :)

app/lib/annotation_viz_service.rb
bistline: That's a great idea - will do.

test/integration/lib/expression_viz_service_test.rb
bistline: [faab2ba](https://github.com/broadinstitute/single_cell_portal_core/pull/1510/commits/faab2ba9bfe2a5ba0477186b58a10246d2f0964a)

test/integration/lib/expression_viz_service_test.rb
bistline: [faab2ba](https://github.com/broadinstitute/single_cell_portal_core/pull/1510/commits/faab2ba9bfe2a5ba0477186b58a10246d2f0964a)

db/seed/example_studies/exp_cell_filtering/study_info.json
```diff
@@ -0,0 +1,36 @@
+{
+  "study": {
+    "name": "Expression Matrix Cell Filtering",
+    "description": "<p>Example study showing effects of filtering distribution-based plots for cells only present in processed expression matrices.</p>",
```
eweitz, line 4: Enclosing this in `p` tags seems odd, but I imagine it's merely a vestige of previous synthetic studies.

```suggestion
    "description": "<p>Example study showing effects of filtering distribution-based plots for cells only present in processed expression matrices.</p>",
```

test/integration/cache_management_test.rb
```diff
@@ -15,28 +15,29 @@ class CacheManagementTest < ActionDispatch::IntegrationTest
...
+    @study_exp_file = FactoryBot.create(:expression_file,
                                         name: 'dense.txt',
+                                        cell_input: %w[A B C],
                                         file_type: 'Expression Matrix',
                                         study: @study)
 
     @study_metadata_file = FactoryBot.create(:metadata_file,
                                              name: 'metadata.txt', study: @study,
-                                             cell_input: ['A', 'B', 'C'],
+                                             cell_input: %w[A B C],
```
eweitz, line 37: Can we change Robocop configuration to do the opposite?  This impedes reading flow for folks much more familiar with JavaScript or Python than Ruby.

eweitz (approved):
Code looks good overall, and my manual local test verifies this works as described!

I note two nits, neither blocking.

db/seed/example_studies/exp_cell_filtering/study_info.json
bistline: Agreed, that's a copy/paste.

test/integration/cache_management_test.rb
bistline: To be fair... we're writing Ruby here, not Python or Javascript.  Both Python and Javascript have ways of notating things (like list comprehensions, or the `...` notation in ES6) that initially can be difficult to understand, but are considered idiomatic to those languages.  However, once I can grasp the context, their utility becomes obvious.  This is the same - Ruby has a [quite a few literals](https://docs.ruby-lang.org/en/2.6.0/syntax/literals_rdoc.html#label-Percent+Strings) for creating objects, like arrays of strings/symbols, regular expressions, and so on.  The edited syntax has dramatically fewer characters, and is _sooo_ much easier to type, rather than constantly opening/closing quotes.

I guess the TL;DR is "I'd prefer not to, because this is how the Ruby community indicates it to be written".  While not all linter rules make the same amount of sense, I think that this is a very good one.  For context - we've had the same discussion about `unless` vs. negative `if`.  I think idiomatic shortcuts like this help to increase our familiarity with languages we may not be as facile in.

eweitz: That's fair!  I think this comes down to taste; differing prioritization of idiomaticness vs. broad reading ease.

PR #1512, opened 2022-06-01
Adding convenience methods to backfill differential expression data (SCP-4196)

bistline:
This update adds two convenience methods to `DifferentialExpressionService` for backfilling data on existing studies:

1. `run_differential_expression_on_default`: for a specified study, run a DE job on the default annotation, if valid
2. `run_differential_expression_on_all`: for a specified study, run a DE job all valid study- and cluster-based annotations

Both methods accept a study accession as input, and an optional `User` object for handling email delivery (otherwise will default to the study owner).

MANUAL TESTING
1. Boot as normal, and start DelayedJob
2. Complete step 3 from #1483 if you have not already added the testing differential expression study to your local instance
3. Once ingested, load the study from above in a Rails console session and set the default annotation to `disease__ontology_label`:
```
accession = <accession of DE study>
study = Study.find_by(accession: accession)
study.default_options[:annotation] = 'disease__ontology_label--group--study'
study.save
```
4. Run a DE job on the new default annotation, and confirm the output is `true`
```
DifferentialExpressionService.run_differential_expression_on_default(accession)
```
5. (Optional) Once the job completes, run DE jobs for all remaining annotations (there should be 5 or 6, depending on what has already been computed):
```
DifferentialExpressionService.run_differential_expression_on_all(accession)
```

This PR satisfies SCP-4196

### DIFF
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -17,6 +17,9 @@ class DifferentialExpressionService
   def self.run_differential_expression_on_default(study_accession, user: nil)
     study = Study.find_by(accession: study_accession)
     validate_study(study)
+    raise ArgumentError, "#{study.accession} has no default cluster" if study.default_cluster.blank?
+    raise ArgumentError, "#{study.accession} has no default annotation" if study.default_annotation.blank?
+
     annotation_name, annotation_type, annotation_scope = study.default_annotation.split('--')
     annotation = {
       annotation_name: annotation_name,
@@ -36,6 +39,9 @@ class DifferentialExpressionService
   #
   # * *yields*
   #   - (IngestJob) => Differential expression job in PAPI for each valid cluster/annotation combination
+  #
+  # * *returns*
+  #   - (Integer) => Number of DE jobs yielded
   #
   # * *raises*
   #   - (ArgumentError) => if requested study cannot run any DE jobs
@@ -89,7 +95,7 @@ class DifferentialExpressionService
 
           annotation_params = annotation.deep_dup # make a copy so we don't lose the association next time we check
           annotation_params.delete(:cluster_file_id)
-          job_identifier = "#{study_accession}: #{cluster_file.name} (#{annotation_params})"
+          job_identifier = "#{study_accession}: #{cluster_file.name} (#{annotation_params.values.join('--')})"
           log_message "Checking DE job for #{job_identifier}"
           DifferentialExpressionService.run_differential_expression_job(
             cluster_file, study, requested_user, **annotation_params
@@ -102,6 +108,7 @@ class DifferentialExpressionService
       end
     end
     log_message "#{study_accession} yielded #{job_count} differential expression jobs"
+    job_count
   end
 
   # handle setting up and launching a differential expression job
@@ -204,8 +211,6 @@ class DifferentialExpressionService
     raise ArgumentError, 'Requested study does not exist' if study.nil?
     raise ArgumentError, "#{study.accession} is not public" unless study.public?
     raise ArgumentError, "#{study.accession} is not initialized" unless study.initialized?
-    raise ArgumentError, "#{study.accession} has no default cluster" if study.default_cluster.blank?
-    raise ArgumentError, "#{study.accession} has no default annotation" if study.default_annotation.blank?
   end
 
   # shortcut to log to STDOUT and Rails log simultaneously
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -1,5 +1,109 @@
 # handle launching differential expression ingest jobs
 class DifferentialExpressionService
+  # run a differential expression job for a given study on the default cluster/annotation
+  #
+  # * *params*
+  #   - +study_accession+ (String) => Accession of study to use
+  #   - +user+ (User) => Corresponding user, will default to study owner
+  #
+  # * *yields*
+  #   - (IngestJob) => Differential expression job in PAPI
+  #
+  # * *returns*
+  #   - (Boolean) => True if job queues successfully
+  #
+  # * *raises*
+  #   - (ArgumentError) => if requested parameters do not validate
+  def self.run_differential_expression_on_default(study_accession, user: nil)
+    study = Study.find_by(accession: study_accession)
+    validate_study(study)
+    annotation_name, annotation_type, annotation_scope = study.default_annotation.split('--')
+    annotation = {
+      annotation_name: annotation_name,
+      annotation_type: annotation_type,
+      annotation_scope: annotation_scope
+    }
+    cluster_file = study.default_cluster.study_file
+    requested_user = user || study.user
+    run_differential_expression_job(cluster_file, study, requested_user, **annotation)
+  end
+
+  # same as above method, except runs differential expression job on all valid annotations
+  #
+  # * *params*
+  #   - +study_accession+ (String) => Accession of study to use
+  #   - +user+ (User) => Corresponding user, will default to study owner
+  #
+  # * *yields*
+  #   - (IngestJob) => Differential expression job in PAPI for each valid cluster/annotation combination
+  #
+  # * *raises*
+  #   - (ArgumentError) => if requested study cannot run any DE jobs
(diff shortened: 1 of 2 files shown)


app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+  #   - (Integer) => Number of DE jobs yielded
+  #
+  # * *raises*
+  #   - (ArgumentError) => if requested study cannot run any DE jobs
+  def self.run_differential_expression_on_all(study_accession, user: nil)
+    study = Study.find_by(accession: study_accession)
+    validate_study(study)
+    available_annotations = []
+
+    metadata = study.cell_metadata.where(annotation_type: 'group', is_differential_expression_enabled: false)
```
devonbush, line 53: What would happen if we reran DE on an annotation we've already run it on?  I think that might be very useful in some circumstances, e.g. if we tweak the output format, or the user uploads a fresh expression matrix.  Maybe that should be an argument to this method whether we rerun and overwrite existing DE?  (possibly out of scope for this PR)

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+                    .select(&:can_visualize?)
+    available_annotations += metadata.map do |meta|
+      { annotation_name: meta.name, annotation_type: meta.annotation_type, annotation_scope: 'study' }
+    end
+
+    cell_annotations = []
+    study.cluster_groups.each do |cluster|
+      next if cluster.cell_annotations.empty?
```
devonbush, line 61:
```suggestion
    groups_to_process = study.cluster_groups.reject {|cg| cg.cell_annotations.empty?}
    groups_to_process.map do |cluster|
```
Filtering the list outside the each will make debugging/troubleshooting easier because you can quickly identify which and how many groups will be processed.  channeling Mohs, you could then replace `each` with `map`

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+
+    log_message "#{study_accession} has data available for DE; validating inputs"
+    requested_user = user || study.user
+
+    job_count = 0
+    study.cluster_ordinations_files.each do |cluster_file|
+      available_annotations.each do |annotation|
+        begin
+          # skip if this is a cluster-based annotation and is not available on this cluster file
+          next if annotation[:scope] == 'cluster' && annotation[:cluster_file_id] != cluster_file.id
```
devonbush, line 94: same as above, I think this will be clearer and more debuggable if the filter is done outside the `each`

devonbush (approved):
Nice, the approach is solid.   Some style recommendations, and one question about how we should handle re-running DE for studies.

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+  # * *raises*
+  #   - (ArgumentError) => if requested study cannot run any DE jobs
+  def self.run_differential_expression_on_all(study_accession, user: nil)
+    study = Study.find_by(accession: study_accession)
+    validate_study(study)
+    available_annotations = []
+
+    metadata = study.cell_metadata.where(annotation_type: 'group', is_differential_expression_enabled: false)
+                    .select(&:can_visualize?)
+    available_annotations += metadata.map do |meta|
```
eweitz, line 55: I like Jean's term "eligible" here, as I think it better connotes that we're applying more selection criteria for these annotations.  Ideally we'd change how we talk of these things or change the code to align with how we talk.

```suggestion
    eligible_annotations += metadata.map do |meta|
```

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+          annotation_params.delete(:cluster_file_id)
+          job_identifier = "#{study_accession}: #{cluster_file.name} (#{annotation_params.values.join('--')})"
+          log_message "Checking DE job for #{job_identifier}"
+          DifferentialExpressionService.run_differential_expression_job(
+            cluster_file, study, requested_user, **annotation_params
+          )
+          log_message "DE job for #{job_identifier} successfully launched"
+          job_count += 1
+        rescue ArgumentError => e
+          log_message "Skipping DE job for #{job_identifier} due to: #{e.message}"
```
eweitz, line 106: Is this error worth logging to Sentry?

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+      {
+        annotation_name: annot[:name],
+        annotation_type: annot[:type],
+        annotation_scope: 'cluster',
+        cluster_file_id: annot[:cluster_file_id]
+      }
+    end
+    raise ArgumentError, "#{study_accession} does not have any valid annotations" if available_annotations.empty?
+
+    log_message "#{study_accession} has data available for DE; validating inputs"
```
eweitz, line 86: If we switch to "eligible" above.

```suggestion
    raise ArgumentError, "#{study_accession} does not have any eligible annotations" if available_annotations.empty?

    log_message "#{study_accession} has annotations eligible for DE; validating inputs"
```

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+      annotation_name: annotation_name,
+      annotation_type: annotation_type,
+      annotation_scope: annotation_scope
+    }
+    cluster_file = study.default_cluster.study_file
+    requested_user = user || study.user
+    run_differential_expression_job(cluster_file, study, requested_user, **annotation)
+  end
+
+  # same as above method, except runs differential expression job on all valid annotations
```
eweitz, line 34:
```suggestion
  # same as above method, except runs differential expression job on all eligible annotations
```

app/lib/differential_expression_service.rb
```diff
@@ -1,5 +1,116 @@
...
+  def self.run_differential_expression_on_default(study_accession, user: nil)
+    study = Study.find_by(accession: study_accession)
+    validate_study(study)
+    raise ArgumentError, "#{study.accession} has no default cluster" if study.default_cluster.blank?
+    raise ArgumentError, "#{study.accession} has no default annotation" if study.default_annotation.blank?
+
+    annotation_name, annotation_type, annotation_scope = study.default_annotation.split('--')
+    annotation = {
+      annotation_name: annotation_name,
+      annotation_type: annotation_type,
```
eweitz, line 26: Any objections to removing this parameter from the DE pipeline, @jlchang?  It'd reduce cognitive load for maintainers.  Context [here](https://github.com/broadinstitute/single_cell_portal_core/pull/1483#discussion_r868173916).

eweitz (approved):
Looks good!  

I suggest terminology refinements, and nudge an unresolved prior discussion; no blockers.

app/lib/differential_expression_service.rb
bistline: That's a great point... you would still be able to overwrite any existing results using the `run_differential_expression_job` method, as that doesn't check the status of `is_differential_expression_enabled`.  It will simply overwrite outputs in the bucket, and do the normal state setting once it is done.

app/lib/differential_expression_service.rb
bistline: This one is a little trickier as it requires the `each` since we're comparing the annotation against the current clustering file.  I don't think the above approach will work.

app/lib/differential_expression_service.rb
bistline: No, because it's not actually an "error" that we care about - it's a validation error as opposed to something failing.  The way the code is written, it either submits a job or emits and `ArgumentError` because the requested parameters do not meet the requirements for a valid DE job.  I wrote it this way so that it was more explicit when a job can't run, rather than exiting silently.

app/lib/differential_expression_service.rb
bistline: I have no arguments, but that would require upstream edits in `scp-ingest-pipeline`, as you noted.  I think that can be done later and doesn't necessarily need to be done right now.

app/lib/differential_expression_service.rb
bistline: Agreed, though I think I'll invert it from `reject|empty?` to `select|any?` as that seems a little more appropriate in this case.

PR #1515, opened 2022-06-03
Improve error handling on gene search in a study (SCP-4373)

ehanna4:
Previously if a user searched for a gene that didn't exist in a study there would be a 404 error thrown from rails but it was not being handled so it left the page with no results and a never-ending spinning loading icon.

https://user-images.githubusercontent.com/54322292/171872110-4b23522e-018e-4f90-9500-d011fb19c0a1.mov

Now if a user enters a gene that doesn't exist in the study first a modal will pop up to try to guide them to remove the offending gene from their search. However, if they persist and do the search anyway the error is handled and is shown to the user. Popping up the modal will reduce the failing network calls but even if they do it anyway the error will be made obvious.

Screen Shot 2022-06-03 at 10 13 05 AM

https://user-images.githubusercontent.com/54322292/171885553-56140668-73d3-4430-a814-c499b2adc68d.mov

To test:
- pull this branch
- open single cell portal and go to a study
- try searching for a variety of genes - in particular try misspelled ones, and combos of bad and good ones, plus try deleting and adding different genes.
- Observe in the console that there will be a 404 error but the error should be also displayed to the user in the plot region


app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -201,7 +201,7 @@ def self.get_cluster_viz_data(study, cluster, url_params)
               }
             elsif is_correlated_scatter
               if genes.count != 2
-                raise ArgumentError, "Correlated scatter plots require specifying 2 genes"
+                raise ArgumentError, "Correlated scatter plots require specifying 2 valid genes"
```
devonbush, line 204: nice :)

app/javascript/components/explore/StudyGeneField.jsx
```diff
@@ -30,14 +30,29 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
     * an array of already entered genes (geneArray),
     * and the current text the user is typing (inputText) */
   const [geneArray, setGeneArray] = useState(enteredGeneArray)
-
   const [showEmptySearchModal, setShowEmptySearchModal] = useState(false)
 
+  const [badGeneSet, setBadGeneSet] = useState(new Set([]))
```
devonbush, line 35: can we be more precise than 'bad'?  'missingGenes'? 'notPresentGenes'?

app/javascript/components/explore/StudyGeneField.jsx
```diff
@@ -30,14 +30,29 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
...
   function handleSearch(event) {
     event.preventDefault()
     const newGeneArray = syncGeneArrayToInputText()
-    if (newGeneArray && newGeneArray.length) {
+
+    let shouldSearch = true
+    if (newGeneArray) {
+      newGeneArray.forEach(gene => {
+        // if an entered gene is not in the valid gene options for the study
+        if (geneOptions.length > 0 && !geneOptions.find(geneOpt => geneOpt.label === gene.label) && !badGeneSet?.has(gene.label)) {
```
devonbush, line 47: BLOCKER: how does this handle case sensitivity?  We don't require the user to have the correct casing (e.g. they can type 'pTen' and it will match 'Pten'), but this looks like it's requiring exact match.

devonbush (changes requested):
Looks good, minus some case sensitivity issues.  And let's pair Monday on writing JS tests for this

app/javascript/components/explore/StudyGeneField.jsx
ehanna4: oo interesting, good catch, the autofill kept fixing the casing when I was testing so I didn't think about that - will update this

bistline (approved):
Approving, modulo Devon's comment about case sensitivity.  I like the new UX - the gentle suggestion is good.

devonbush (approved):

PR #1516, opened 2022-06-03
Clientside throttling for Sentry error logging (SCP-4335)

eweitz:
This adds frontend sampling controls to Sentry logs, improving signal-to-noise ratio for error observability.

Sentry error logging from the browser was introduced in #1482.  This refines that by throttling certain high-volume events:

* TypeError: Failed to fetch: Error in JavaScript when logging event to Bard
* TypeError: NetworkError when attempting to fetch resource.: Error in JavaScript when logging event to Bard

Only 5% of those will now be logged.  In the past 14 days while logging nearly 100% of them, we've accrued over 6000 such events.  That's too much of our nominal quota.  The new 0.05 sample rate means we expect roughly 300 in the next two weeks.  That lower-but-meaningful number helps mitigate Sentry quota exhaustion while not suppressing all such events.  Full suppression might make sense once we further investigate causes and ticket work.

Other noisy Sentry logs are now fully suppressed:
* Sentry errors and traces in `development` and `test` environments (only some such events were suppressed before)
* The high-volume, low-impact error described in SCP-4360

Like Ruby Sentry logging, JavaScript Sentry logging also now prints to the console when suppressed.  Automated tests verify (and document by example) complex new logic in `shouldLog`.

To manually test:
1. Open DevTools in browser
2. Load home page
4. Open `StudyResults.jsx` and change line 77 to `<StudyComponentIdonotExist`
5. Confirm warning message in console "Suppressing error report to Sentry: in an unlogged environment"
6. Go to Network tab
7. Filter to "sentry"
8. Confirm only 1 request is sent to Sentry, with POST body including `{"type":"client_report"}`
9. Edit line 63 in `sentry-logging.js` to return `false`
10. In reloaded page, confirm multiple requests are sent to Sentry
11. Undo steps 4 and 9

This satisfies SCP-4335.


app/javascript/lib/sentry-logging.js
```diff
@@ -30,20 +44,41 @@ export function logJSFetchExceptionToSentry(response, titleInfo = '') {
...
+    Sentry.captureException(new Error(`${error}: ${titleInfo}`))
+  }
+}
+
+/** Print suppression warning message to the console */
+function printSuppression(reason, response) {
+  if (isSuppressedEnv) {
+    reason = 'in an unlogged environment'
+  }
+  const message = `Suppressing error reporting to Sentry, because it is ${reason}.`
```
devonbush, line 60:
```suggestion
  const message = `Suppressing error report to Sentry: ${reason}.`
```

app/javascript/lib/sentry-logging.js
```diff
@@ -30,20 +44,41 @@ export function logJSFetchExceptionToSentry(response, titleInfo = '') {
...
 /**
  * Determine if logging should occur based on environment
+ * @param {Boolean} useThrottle - whether to apply clientside rate limit throttling. Default false.
+ * @param {Object} response - Fetch response object.  Default null.
+ * @param {Number} sampleRate - % of events to log, only applied if `useThrottle = true`
+ *  1 = log all events, 0 = log no events, default = 0.05 (i.e. log 5% of events)
  */
-function shouldLog() {
-  const env = getSCPContext().environment
+export function shouldLog(useThrottle = false, response = null, sampleRate = 0.05) {
```
devonbush, line 76: BLOCKER: this seems to mix concerns, as this function does more than just make the determination, it also handles logging to console.  And it also has some logic that seems particular to only one sort of error with how it logs response errors.

  Let's rescope this function to  `logToSentry(errorMessage, useThrottle=false, throttleRate=0.05)`

Then this function can handle invoking `Sentry.captureException(errorMessage)` or logging that it was suppressed.  And it can return true/false on whether it was logged.  Then the callee (e.g. `logJSFetchExceptionToSentry`) can log anything else it wants (e.g. the response) if it finds the error was surpressed.  (or in this case, we might just want to log the response to console all the time)

test/js/metrics-api.test.js
```diff
@@ -2,10 +2,12 @@
 /* eslint-disable*/
 
 const fetch = require('node-fetch')
-import {logClick, logClickLink, logMenuChange, setMetricsApiMockFlag} from 'lib/metrics-api'
 import { screen, render, fireEvent } from '@testing-library/react'
 import React from 'react';
 
+import {logClick, logClickLink, logMenuChange, setMetricsApiMockFlag} from 'lib/metrics-api'
+import {shouldLog, setIsSuppressedEnv} from 'lib/sentry-logging'
```
devonbush, line 9: as mentioned above, let's replace this with mocking getSPContext().  Assuming we're going to expand the number of environment-specific things we do as time goes on, we'll want to become more adept at mocking the environment, rather than always having to build backdoor setters.

devonbush (changes requested):
Solid approach, could just use some refactoring for clarity and testability.

app/javascript/lib/sentry-logging.js
```diff
@@ -54,12 +89,17 @@ function shouldLog() {
  *  Initialize Sentry to enable logging JS errors to Sentry
  */
 export function setupSentry() {
+  if (isSuppressedEnv) {
```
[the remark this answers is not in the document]
eweitz: A common kind of error gets suppressed, but without per-error warnings.  This covers that.

Three code paths logged from the browser to Sentry:
1. logJSFetchExceptionToSentry (when Response object has error)
2. logJSFetchErrorToSentry (like above, but passing a built-in Error object)
3. other React JS errors

When the 3rd kind of error occurs, per-error warnings become very noisy and duplicative, much more so than the other two kinds.  The idea here is to warn the user once that such errors aren't logged to Sentry, which seems valuable and is much less clutter.

eweitz:
@devonbush, seeking re-review as your reply in [this thread](https://github.com/broadinstitute/single_cell_portal_core/pull/1516#discussion_r889247005) could likely change my strategy for incorporating your other feedback.

eweitz:
> example instructions are incorrect

Fixed, thanks!

app/javascript/lib/sentry-logging.js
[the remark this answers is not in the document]
devonbush: so you're saying that for (3), we won't even log those to the console?  that seems odd that we think there are errors worth logging to sentry, but that we won't even show in the console in development?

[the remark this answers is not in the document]
devonbush: In general, I think a statement that prints to the console every time in development is a bad idea--it will very quickly get tuned out and scrolled out of view by other messages.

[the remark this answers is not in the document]
eweitz: Thanks for chatting just now.  To summarize, (3) does get logged to console as an error, but is not proceeded by a warning that it's suppressed from Sentry logging.  That's because `console.warn` messages in such cases include a duplicative stack trace, making it too noisy.  To lower noise while somewhat reusing our other Sentry logging patterns, I opted to print that message once, when the page loads.

You rightly note we don't want to set a precedent of browser-console logging a bunch of environmental defaults, which seems like the strongest argument to block.

So, per our discussion, I'll plan to use per-error console messages for (3), just like the backend, and:
1.  See if `console.warn` has a parameter to omit stack traces in such cases, and if not:
2. See if `console.info` or `console.log` would omit it, and if not:
3. Abandon that pattern reuse, and drop such messages altogether

app/javascript/lib/sentry-logging.js
```diff
@@ -54,12 +89,17 @@ function shouldLog() {
...
+    const reason = 'in an unlogged environment'
+    printSuppression(reason)
```
ehanna4, line 94: Couldn't you just call `printSuppression()` (without the if statement here) since it has the same logic to check if it's a `isSuppressedEnv` ?

eweitz: Good catch!  Will do.

test/js/metrics-api.test.js
```diff
@@ -102,4 +106,46 @@ describe('Library for client-side usage analytics', () => {
...
     })
   })
+
+  it('appropriately determines whether to log to Sentry', () => {
+
+    console.log = jest.fn();
+
+    jest
+      .spyOn(SCPContextProvider, 'getSCPContext')
+      .mockReturnValue({environment: 'logged-env'})
```
devonbush, line 116: is there any significance to this `logged-env` string?  If not, let's replace it with either something that clearly indicates it's meaningless (e.g. 'something' or maybe 'anything-other-than-test'), or with either 'production' or 'development'.

test/js/explore/explore-display-tabs.test.js
```diff
@@ -18,6 +18,16 @@ jest.mock('components/visualization/InferCNVIdeogram', () => {
...
+import React from 'react'
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import * as Reach from '@reach/router'
+import jquery from 'jquery'
+
+import MockRouter from '../lib/MockRouter'
+import * as ScpApi from 'lib/scp-api'
+import ExploreView from 'components/explore/ExploreView'
+import { mockPerformance } from '../mock-performance'
```
devonbush, line 30: BLOCKER: we can remove all these imports

devonbush (changes requested):
Thanks for the code refactoring and the great 'to test' instructions.  Two code blockers, and also this doesn't seem to actually suppress logging.  When I check out your branch and follow your test instructions, I see a 'suppressing' message in the log, but I still see a POST request made to Sentry in the network tab.  Also, if I change line 158 in `scp-api.jsx` to have the path `search/facetsblahblah`, I get a 500 error that is logged to Sentry, with no suppression message at all, which suggests that's a class of error that is not being handled by this framework?

test/js/explore/explore-display-tabs.test.js
eweitz: Ironically, I [suggested](https://github.com/broadinstitute/single_cell_portal_core/pull/1525#discussion_r892394449) the same.  Removed!

test/js/metrics-api.test.js
eweitz: Yes, the significance is that it conveys the environment would be a logged one, without coupling it to implementation details.

This strikes me as a matter of taste; I'd love to see precedent in other code either way.  Your suggested change is fortunately trivial here, so I've gone ahead and implemented it.

eweitz:
> I see a 'suppressing' message in the log, but I still see a POST request made to Sentry in the network tab

My original implementation suppressed all Sentry POSTs in development, but a side-effect of [an earlier suggestion](https://github.com/broadinstitute/single_cell_portal_core/pull/1516/#discussion_r889247005) is that at least some Sentry events get logged there.  

I've refined the implementation of suppression to better account for that, as well as those testing instructions.

devonbush (approved):

eweitz:
Impact follow-up: analytics show this [decreases Sentry error log rate by 3-5x](https://sentry.io/organizations/broad-institute/projects/single-cell-portal/?end=2022-06-28T13%3A38%3A23&environment=production&project=1424198&start=2022-05-29T13%3A38%3A23), with negligible decrease in observability value.

PR #1518, opened 2022-06-06
persist hidden traces (SCP-4380)

devonbush:
To support browsing spatial clusters while keeping certain cell types hidden, this persists the hidden legend entries, as well as the 'splitLabelArrays' param.  This puts the hidden traces in the URL.  This may cause issues for some extremely large number of hidden traces, but Chrome's URL limit is 2MB, so even hiding 10000 traces that are each 100 characters long should work.  

TO TEST:
 1. load a study that has two clusters with a common annotation (e.g. a study-wide annotation).  For example, the `chicken_raw_counts` synthetic study
 2. View a cluster, select the study-wide annotation, and then hide one or more traces
 3. Switch to the other cluster, confirm the traces remain hidden
 4. Switch to a different annotation, and then switch back, confirm traces are no longer hidden
 5. Load a study with pipe-delimited array labels (e.g. `array_mouse_eye` synthetic study). 
 6. Confirm that you can split/merge array labels

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -208,6 +208,11 @@ export default function ExploreDisplayTabs({
     ) {
       updateParams.tab = 'annotatedScatter'
     }
+    // if the user changes annotation, unset hiddenTraces
+    if (newParams.annotation && newParams.annotation.name !== exploreParams.annotation.name) {
+      updateParams.hiddenTraces = []
+    }
+
     updateExploreParams(updateParams)
   }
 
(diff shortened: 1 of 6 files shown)


bistline (approved):

eweitz (changes requested):
The persisting traces are a nice touch!  I verified they work locally.  But I found a related bug.

Blocker: the default view in the "Murine vision and HIV with arrays" lacks a label legend in my local test, but [equivalent study on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP298/murine-vision-and-hiv-with-arrays#study-visualize) has a legend.

Screen Shot 2022-06-07 at 4 23 12 PM

devonbush:
Nice catch Eric--that bug happened if the default annotation for a study was cluster, and the user didn't change an annotation quickly (which I almost always did)

devonbush:
Build failure is Azul facet search - unrelated

PR #1519, opened 2022-06-06
holding zoom fixed on legend show/hide (SCP-3878)

devonbush:
Showing/hiding a legend entry no longer auto-resizes the graph. 

TO TEST:
1. Load a cluster plot for a study (e.g. the "mouse_brain" study)
2. hide a trace with points at the edge of the graph (e.g. for the "Category" annotation of the"mouse_brain" study, hide the "A" trace)
3. Observe that the graph axes stay fixed
4. Load a study with two different clusters with different ranges (e.g. the human_lymph study)
5. Switch between the clusters, showing and hiding traces, and confirm the axes change on cluster change, but remain constant for showing/hiding traces


ehanna4 (approved):
@jlchang and I reviewed code and tested locally - works as advertised

PR #1524, opened 2022-06-07
Address 400 Bad Request from Azul (SCP-4348)

bistline:
(I'm trying a new format for my PRs that I think may help give people more high-level info faster, and let them gloss over implementation details if they so chose - let me know what you think!)

**BACKGROUND**

Certain study search queries are returning `400 Bad Request` from Azul when issued.  This results in that particular search request not getting any results from Azul, when there may have been HCA projects that matched the search.  Specifically, the term `organoid` was resulting in a query in Azul that was invalid.

**CHANGES**

The underlying issue is in how we convert metadata names from the Alexandria schema to the names that Azul uses as facets.  Specifically, our search facet of `sample_type` did not have a mapping listed.  This is a combination of an oversight, as the term `sampleEntityType` in Azul is a direct analog, and also that this facet was superseded in newer versions of the Alexandria convention by `biosample_type`.  Adding that mapping in `FacetNameConverter` for `biosample_type` fixes this particular problem. Moving forward, and mappings that do not exist will now be discarded to prevent future facet conversion issues.  This way, we can leave the existing `sample_type` facet in place on production (since there is data in BigQuery for this), but it will be ignored now in XDSS requests.

It is worth noting that there is an associated issue where certain keyword-based search requests result in queries that _should_ find data in Azul, but do not given how their search API is constructed, and how we translate keyword-based into facet-based searches.  Solving this seems intractable, as it would require complicated logic to split out search requests into individual calls, parallelizing them, and then unifying the results after all requests complete.  Such logic could result in 200+ requests for a generic term like "cell", and as such the more optimal course is to contact the Azul team and inquire about them reinstating keyword-based searching via their API.

**MANUAL TESTING**
1. If you do not have the `biosample_type` facet in your local instance, create it with the following command in the Rails console:
```
SearchFacet.create!(name: "biosample type", identifier: "biosample_type", filters: [], is_ontology_based: false, ontology_urls: [], data_type: "string", is_array_based: false, big_query_id_column: "biosample_type", big_query_name_column: "biosample_type", big_query_conversion_column: nil, convention_name: "alexandria metadata convention", convention_version: "latest", unit: nil, min: nil, max: nil, visible: true)
```
2. Run `SearchFacet.update_all_facet_filters` to update all locally stored facets with values from Azul
3. Sign into the home page on your local instance and run a keyword search for `organoid`
4. Confirm the search request completes and finds results from Azul, such as `Kidney organoid reproducibility across multiple human iPSC lines and diminished off target cells after transplantation revealed by single cell transcriptomics`

This PR satisfies portions of SCP-4348.

### DIFF
--- a/app/models/hca_azul_client.rb
+++ b/app/models/hca_azul_client.rb
@@ -269,6 +269,8 @@ class HcaAzulClient
     facets.each do |facet|
       safe_facet = facet.with_indifferent_access
       hca_term = FacetNameConverter.convert_schema_column(:alexandria, :azul, safe_facet[:id])
+      next if hca_term.nil?
+
       scp_facet = facet[:db_facet]
       if scp_facet.is_numeric?
         min = safe_facet.dig(:filters, :min)
--- a/lib/facet_name_converter.rb
+++ b/lib/facet_name_converter.rb
@@ -48,6 +48,7 @@ class FacetNameConverter
     'organ_region' => 'organPart',
     'organism_age' => 'organismAge',
     'preservation_method' => 'preservationMethod',
+    'sample_type' => 'sampleEntityType',
     'sex' => 'biologicalSex',
     'species' => 'genusSpecies',
     'study_accession' => 'projectShortname',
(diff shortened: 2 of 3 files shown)


devonbush (approved):
I like it, and appreciate the new PR format.  I think the breakdown into sections is good.  I might suggest "BACKGROUND". "SUMMARY" and "TO TEST" as the names for the sections, just to make them shorter and easier to quickly identify.

ehanna4 (approved):
@jlchang and I reviewed code and tested locally and works as advertised

bistline:
Build failure looks like a blip in Azul - running the failed test locally and on staging both pass.  It's possible we got an incomplete list of facets from Azul.

PR #1525, opened 2022-06-07
disabling gene search bar if study has no genes (SCP-4363)

devonbush:
Disables the search bar and removes gene set upload button (removing it was easier and IMHO cleaner-looking than greying it) if the study has no genes

image

TO TEST:
1. load a study in the explore tab with genes
2. confirm gene search bar appears as normal
3. load a study with clusters but no genes
4. confirm search bar is disabled


bistline (approved):

test/js/explore/explore-display-tabs.test.js
```diff
@@ -18,6 +18,16 @@ jest.mock('components/visualization/InferCNVIdeogram', () => {
...
+import React from 'react'
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import * as Reach from '@reach/router'
+import jquery from 'jquery'
+
+import MockRouter from '../lib/MockRouter'
+import * as ScpApi from 'lib/scp-api'
+import ExploreView from 'components/explore/ExploreView'
+import { mockPerformance } from '../mock-performance'
```
eweitz, line 30: VS Code shows only one of these imports is used, and 

```
yarn test test/js/explore/explore-display-tabs.test.js # Run a particular test suite
```

passes with the suggested change to remove the unused code.

```suggestion
import '@testing-library/jest-dom/extend-expect'
```

eweitz (approved):
Code looks good overall.  Manual testing instructions would benefit from refinement:

> load a study with clusters but no genes

Is there a synthetic study that has this readily available?  In my local SCP, I see the study shown in your screenshot, "Mouse stomach cells and GERD" (`mouse_stomach`), has only clusters with genes.

PR #1526, opened 2022-06-07
Widen availability of differential expression frontend (SCP-4374)

eweitz:
This expands the DE pilot MVP (#1502), letting users explore differential expression for more studies as enabled upstream.

Now, users can view differential expression for any annotation that has DE outputs available from our initial backfill of 10 [valuable studies](https://docs.google.com/spreadsheets/d/1f09NOe7d1LCqU7pvFMCIJGcNrhO_7Y0eaqSq8OQJZv4/edit#gid=0).  This is designed to gather larger sample sizes for [DE usage analytics](https://mixpanel.com/project/2120588/view/19059/app/dashboards#id=3292575) that inform future work, and more fully technically explore the differential expression MVP.  The implementation removes hard-coding for our DE pilot and generalizes when the "Differential expression" button is shown.

To test:
1. If you haven't already, populate the DE test study and its DE outputs in your local SCP, as described in #1502
2. Ensure the `differential_expression_frontend` feature flag is enabled for that study, as described there
3. Enter your Rails console:  `rails console -e development`
4. Manually enable DE on the default annotation: `Study.find_by(accession: 'SCP138').cell_metadata.find_by(name: 'General_Celltype').update(is_differential_expression_enabled: true)`
5. Navigate to the DE test study: "Human milk - differential expression"
6. Confirm "Differential expression" button appears in the default annotation (i.e., "General_Celltype")
7. Switch to another annotation
8. Confirm "Differential expression" button does not appear

N.B.: That DE test study fails to run in the current DE pipeline, due to an out-of-memory error thrown in the PAPI job.  I've also verified this works with a higher-fidelity example -- a local copy of staging SCP294 (production SCP1539) -- with DE outputs populated using the DE backfill script.

This satisfies SCP-4374.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -48,7 +48,7 @@ const tabList = [
   { key: 'images', label: 'Images' }
 ]
 
-/** Determine if currently selected annotation has differential outputs available */
+/** Determine if currently selected annotation has differential expression outputs available */
 function annotHasDe(exploreInfo, exploreParams) {
   const flags = getFeatureFlagsWithDefaults()
   if (!exploreInfo || flags?.differential_expression_frontend) {
@@ -104,16 +104,14 @@ export default function ExploreDisplayTabs({
   const [currentPointsSelected, setCurrentPointsSelected] = useState(null)
 
   // Differential expression settings
-  // TODO (SCP-4374): Integrate is_differential_expression_enabled from API
-  const [showDeGroupPicker, setShowDeGroupPicker] = useState(false)
+  const isDifferentialExpressionEnabled = annotHasDe(exploreInfo, exploreParams)
+  const [, setShowDeGroupPicker] = useState(false)
   const [deGenes, setDeGenes] = useState(null)
   const [deGroup, setDeGroup] = useState(null)
   const [showDifferentialExpressionPanel, setShowDifferentialExpressionPanel] = useState(deGenes !== null)
 
   // Hash of trace label names to the number of points in that trace
   const [countsByLabel, setCountsByLabel] = useState(null)
-
-  const isDifferentialExpressionEnabled = annotHasDe(exploreInfo, exploreParams)
 
   const plotContainerClass = 'explore-plot-tab-content'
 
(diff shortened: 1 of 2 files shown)


app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -48,6 +48,34 @@ const tabList = [
...
+  let annotHasDe = false
+  const annotList = exploreInfo.annotationList
+  let selectedAnnot
+  if (exploreParams?.cluster) {
+    selectedAnnot = exploreParams.annotation
+  } else {
+    selectedAnnot = exploreInfo.annotationList.default_annotation
+  }
+
+  const matchingAnnot = annotList.annotations.find(annot => {
```
devonbush, line 67: BLOCKER: this duplicates `getMatchedAnnotation` from `cluster-utils.js`

devonbush (commented):
Nice to see this getting cleaned up. One easy blocker of making sure to use an existing utility function

devonbush (changes requested):
nice to sse this getting cleaned up -- one blocker around using an existing utility function

bistline (approved):
Looks good, modulo Devon's comment.

One note: I've noticed a side-effect that when you switch to a different annotation that may/may not have DE results, we keep the visualization pinned on what it was last (same gene query).  Do we want to reset this?  Or is this just the application doing what it's supposed to?  I could see the argument that when you click the arrow to exit the DE panel, it should reset the expression view, but I'm not super strong on that.

app/javascript/components/explore/ExploreDisplayTabs.jsx
eweitz: Good find!  Some day, a Copilot-like linter will catch these things.  336d270

devonbush:
Failure is AzulSearch service

devonbush (approved):

eweitz:
Thanks for looking into the build failure, Devon -- I confirm.

eweitz:
> I've noticed a side-effect that when you switch to a different annotation that may/may not have DE results, we keep the visualization pinned on what it was last (same gene query). Do we want to reset this? Or is this just the application doing what it's supposed to? I could see the argument that when you click the arrow to exit the DE panel, it should reset the expression view, but I'm not super strong on that.

I thought I'd addressed a similar concern that Leyla raised as part of the MVP, but upon closer reading I see I only partially resolved it.  Your concern is similar.  I've gone ahead and ticketed this for future DE work.

PR #1529, opened 2022-06-08
delaying disabling of study gene search until after explore data loads (SCP-4363)

devonbush:
CHANGES: The study gene search bar will no longer be disabled while the explore data for a study is loading.  While this might result in some odd behavior if the user manages to type something in it and hit search before the study loads (they will see a modal indicating that the gene wasn't found in the study), that's probably more desirable than seeing the box flash every time a study loads.

TO TEST:
1. Load a study with clusters and genes
2. confirm the study gene search bar appears and is enabled throughout the loading of the study
3. Load a study with clusters and no genes (e.g. by deleting expression matrices from an existing study)
4. confirm the study gene search is initially enabled, but then disabled once the study data loads


eweitz (approved):
Nice polish!

PR #1530, opened 2022-06-08
streamlining view count update query (SCP-4297)

devonbush:
BACKGROUND
We want to eliminate any N+1 queries related to clusters, as those may cause performance issues for the U19 study that has 100+ clusters
CHANGES
This changes the view count incrementer to skip study model validations.  These validations aren't needed since we're just updating an integer field.  This saves ~20 queries, including one query per study file.  The good news is that in development, given that queries take ~0.05 seconds, this saves almost a second in study loading time.

TO TEST:
1. Uncomment lines 154-155 from config/develoment.rb to turn on mongoid logging
2.  Load a study with 3+ clusters (e.g. the human_lymph_node synthetic study)
3.  Confirm that there are no batches of cluster_group queries repeated for each study file.  They will have ` {"find"=>"cluster_groups", "filter"=>{"study_file_id"=>...`

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -149,7 +149,8 @@ class SiteController < ApplicationController
 
   # load single study and view top-level clusters
   def study
-    @study.update(view_count: @study.view_count + 1)
+    # this skips all validation/callbacks for efficiency
+    @study.update_attribute(:view_count, @study.view_count + 1)
 
     # set general state of study to enable various tabs in UI
     # double check on download availability: first, check if administrator has disabled downloads


eweitz (approved):
A succinct optimization!

Here is empirical data from point measurements.  Times below are time-to-first-byte (TTFB) for the first request in Study Overview, which is the time spent on the server, measured via Chrome DevTools.  These are from my local SCP for `human_lymph_node` ("Tuberculosis in female human lymph").  Times are in seconds.

`development`
* 5 samples: 4.4, 4.6, 4.5, 4.5, 4.9
* Mean: 4.6

`db-cluster-db-calls`
* 5 samples: 3.2, 4.1, 3.1, 3.0, 3.1
* Mean: 3.3

Saving 1.3 seconds on every study page load is a solid development speed-up.

PR #1531, opened 2022-06-09
Less restrictive study gene search term check (SCP-4417)

ehanna4:
Previously was using the geneoptions which were the more restrictive genes that appeared as options in the dropdown to decide if a gene search term was valid. Instead using all the genes present in the study will be less restrictive and as such prevent the false negatives the previous change was making happen.

I don't have any local studies that have this situation so I can't 100% test but the functionality still works for my local studies and I can test better on staging.

This should fix:
2. "Searching with a "valid" but non-existent gene is causing some strange behavior including in certain sequences triggering the modal but overall causing an error"

from the ticket: https://broadworkbench.atlassian.net/browse/SCP-4417?atlOrigin=eyJpIjoiYjQxM2YxM2NmMGVmNGM4NGJiZWE5MzQyMjQwMDNkZTciLCJwIjoiaiJ9

### DIFF
--- a/app/javascript/components/explore/StudyGeneField.jsx
+++ b/app/javascript/components/explore/StudyGeneField.jsx
@@ -48,7 +48,7 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
     if (newGeneArray) {
       newGeneArray.forEach(gene => {
         // if an entered gene is not in the valid gene options for the study
-        if (geneOptions.length > 0 && !geneOptions.find(geneOpt => geneOpt.label.toLowerCase() === gene.label.toLowerCase())) {
+        if (allGenes.length > 0 && !allGenes.find(geneOpt => geneOpt.toLowerCase() === gene.label.toLowerCase())) {
           newNotPresentGenes.add(gene.label)
         }
       })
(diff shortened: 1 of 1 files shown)


app/javascript/components/explore/StudyGeneField.jsx
```diff
@@ -48,7 +48,7 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
     if (newGeneArray) {
       newGeneArray.forEach(gene => {
         // if an entered gene is not in the valid gene options for the study
-        if (geneOptions.length > 0 && !geneOptions.find(geneOpt => geneOpt.label.toLowerCase() === gene.label.toLowerCase())) {
+        if (allGenes.length > 0 && !allGenes.find(geneOpt => geneOpt.toLowerCase() === gene.label.toLowerCase())) {
```
devonbush, line 51: lowercasing the comparison string outside of the find loop will save 30K+ toLowerCase operations.

```suggestion
        const geneLowercase = gene.label.toLowerCase()
        if (allGenes.length > 0 && !allGenes.find(geneOpt => geneOpt.toLowerCase() === geneLowercase)) {
```

eweitz (approved):
I confirm this works locally!

Here's how it looked:

https://user-images.githubusercontent.com/1334561/172880922-84e9872c-bcf6-4974-ae4a-7b66b4c42a2f.mov

Out-of-scope refinements to consider for future work:
* Log when this user error occurs
* The modal has no obvious way to dismiss.   Clicking someone else works, but I can see folks being confused at least briefly.
* Inline error messaging would be slicker, e.g. text right below search input
* "Invalid Search" -> "Invalid search" in message

PR #1535, opened 2022-06-13
Hotfix: Decrease server load for global gene search (SCP-4432)

eweitz:
This mitigates outage risk by crudely decreasing server CPU load for global gene searches.

Previously, global gene search used a
* Page size of 10
* Default subsampling threshold of 100k cells

Now:
* Page size of 5
* Subsampling of 10k
 
This aims to decrease server load caused by recent scientific refinements related to gene expression filtering (#1510), while keeping that science intact.  Unmitigated server load from such filtering seems to have been a significant cause in transient outages on June 9 after releasing SCP 1.19.

To test:
* Ensure you have some large visualization studies in your local environment
* Open DevTools Network tab
* Do a global gene search
* Note 5 violin plot requests, with none having a subsampling of 100000, and ideally some having a subsampling of 10000

This satisfies SCP-4432.

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -371,7 +371,16 @@ module Api
         @matching_accessions = @studies.map { |study| self.class.get_study_attribute(study, :accession) }
 
         logger.info "Final list of matching studies: #{@matching_accessions}"
-        @results = @studies.paginate(page: params[:page], per_page: Study.per_page)
+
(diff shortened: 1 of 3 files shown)


bistline (approved):
Does what it says it is supposed to do, but based on conversations we're having now, I'm not sure this is addressing the actual issue.

eweitz:
More context is in [this Slack discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1655132860534149).

PR #1537, opened 2022-06-14
Eh handle paren

ehanna4:
We got this error in Sentry:
Where a user had an unmatched parenthesis. This caused an error that got reported to Sentry. We also noted that the users search input matched closely to the helper text so I revised that to not use extraneous parenthesis to hopefully avoid this sort of issue in the future.

To test:
- Pull this branch
- Open up the home page of SCP
-


devonbush (approved):
Excellent, thanks for updating the explainer text

bistline (approved):

bistline:
Test failure is unrelated, and in a rarely-used section of code.  May be an actual regression due to changes in orchestration (unknown) but would not impact any functionality.

PR #1540, opened 2022-06-14
Storing references for valid differential expression results (SCP-4437, SCP-4438, SCP-4404)

bistline:
#### BACKGROUND
After backfilling differential expression results on candidate studies in production, it was noted that several corner cases arose where some results were considered invalid, or simply failed to output anything in the study bucket.  These include:

* For a given source annotation, while it may have multiple possible values, only one of those values may be represented in a given clustering file
* Some values may only be represented by a single cell in a given clustering file, meaning there is not enough data to compute results
* There was no way of knowing what combinations of clustering/annotations in a given study would have valid results, as the information was only stored at the annotation level.

#### CHANGES
SCP-4437: A new model called `DifferentialExpressionResult` will now store references to a given clustering object and specific annotation, denoting that there are valid differential expression results available for visualization.  The attribute `observed_values` keeps track of the individual annotation values that successfully generated output files.  When the `DifferentialExpressionResult` object is created, validations run to compute what values would have succeeded in the differential expression job, and if enough values (2+) had enough cells represented (2+) the object will save.  If either of those validations fail, the object will not save, and therefore will not be available from the differential expression picker menu.  This same validation code is also used in the `DifferentialExpressionService` to determine what jobs to launch, meaning that moving forward, fewer invalid output files should be generated.  This data is now available in the `/explore` API endpoint for a given study under the `differentialExpression` key.  Each object represents a set of outputs for a specific cluster and annotation, and has a `select_options` array that can be used to directly populate the differential expression picker in the UI.

SCP-4438: A data migration was created to move data from the `is_differential_expression_enabled` flag on annotations into `DifferentialExpressionResult` instances.  While performing the migration, any invalid outputs files that were generated as a part of a differential expression PAPI job will be removed from the bucket.

SCP-4404: Each `DifferentialExpressionResult` will cache the current name of the `ClusterGroup` object at the time of valid outputs being written to the study bucket.  This will prevent links to files breaking if a user changes the name of a `ClusterGroup`, and alleviates the need to rename files in buckets on said updates.  These values will be sourced from the `/explore` API response, and will contain a map of specific annotation labels to outputs files with the correct names.

#### MANUAL TESTING
IMPORTANT: If you do not have any differential expression outputs available in your local instance, please follow the instructions from #1512 to backfill some example data before running the manual testing steps

1. Pull branch and run any pending migrations with `./rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Watch the `development.log` file, and look for entries like the following:
```
Checking DE results for SCP67:UMAP_crush (cell_type__ontology_label--group--study)
Results for SCP67:UMAP_crush (cell_type__ontology_label--group--study) valid, saving
Save complete, keeping valid outputs for pericyte cell, macrophage, perineuronal satellite cell, fibroblast, Schwann cell, myeloid cell, erythrocyte, primary neuron, endothelial cell, innate lymphoid cell
Found 1 invalid outputs: N/A, deleting
```
3. Boot local instance, and open the Network tab in Chrome Dev Tools
4. Load the above study, and filter for the `/explore` API response
5. Click "Preview" and confirm that there is an entry called `differentialExpression` with outputs similar to the following:
differentialExpression

### DIFF
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -285,7 +285,7 @@ class ClusterVizService
   #   - +annotation_name+  (String) => Name of requested annotation
   #   - +annotation_type+  (String) => Type of requested annotation (should be 'group')
   #   - +annotation_scope+ (String) => Scope of requested annotation ('study' or 'cluster')
-  def self.get_cells_by_label(cluster, annotation_name, annotation_type, annotation_scope)
+  def self.cells_by_annotation_label(cluster, annotation_name, annotation_type, annotation_scope)
     study = cluster.study
     cells = cluster.concatenate_data_arrays('text', 'cells')
     annotation = { name: annotation_name, scope: annotation_scope, type: annotation_type }
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -275,4 +275,26 @@ class ClusterVizService
 
     raw_matrix
   end
+
+  # find what annotation values are represented for this combination of cluster & annotation
+  # primarily important as a validation for differential expression, as DE cannot be run on single-value annotations,
+  # or on labels that have only one cell
+  #
+  # * *params*
+  #   - +cluster+    (ClusterGroup) => Clustering object being used as control cell list
+  #   - +annotation_name+  (String) => Name of requested annotation
+  #   - +annotation_type+  (String) => Type of requested annotation (should be 'group')
+  #   - +annotation_scope+ (String) => Scope of requested annotation ('study' or 'cluster')
+  def self.get_cells_by_label(cluster, annotation_name, annotation_type, annotation_scope)
+    study = cluster.study
+    cells = cluster.concatenate_data_arrays('text', 'cells')
+    annotation = { name: annotation_name, scope: annotation_scope, type: annotation_type }
+    labels = get_annotation_values_array(study, cluster, annotation, cells, nil, nil)
+    cells_by_label = {}
+    labels.each_with_index do |label, index|
+      cells_by_label[label] ||= []
+      cells_by_label[label] << cells[index]
+    end
+    cells_by_label
+  end
 end
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -200,6 +200,13 @@ class DifferentialExpressionService
     identifier = "#{annotation_name}--#{annotation_type}--#{annotation_scope}"
     raise ArgumentError, "#{identifier} is not present" if annotation.nil?
     raise ArgumentError, "#{identifier} cannot be visualized" unless can_visualize
+
+    # last, validate that the requested annotation & cluster will provide a valid intersection of annotation values
+    # specifically, discard any annotation/cluster combos that only result in one distinct label
+    cells_by_label = ClusterVizService.get_cells_by_label(cluster, annotation_name, annotation_type, annotation_scope)
+    if cells_by_label.keys < 2
+      raise ArgumentError, "#{identifier} does not have enough labels represented in #{cluster.name}"
+    end
   end
 
   # validate a given study is able to run DE job
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -203,7 +203,7 @@ class DifferentialExpressionService
 
     # last, validate that the requested annotation & cluster will provide a valid intersection of annotation values
     # specifically, discard any annotation/cluster combos that only result in one distinct label
-    cells_by_label = ClusterVizService.get_cells_by_label(cluster, annotation_name, annotation_type, annotation_scope)
+    cells_by_label = ClusterVizService.cells_by_annotation_label(cluster, annotation_name, annotation_type, annotation_scope)
     if cells_by_label.keys < 2
       raise ArgumentError, "#{identifier} does not have enough labels represented in #{cluster.name}"
     end
--- a/app/models/differential_expression_result.rb
+++ b/app/models/differential_expression_result.rb
@@ -20,6 +20,7 @@ class DifferentialExpressionResult
 
   before_validation :set_observed_values, :set_cluster_name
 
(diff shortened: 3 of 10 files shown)


devonbush:
Didn't look at the code in detail, but the high-level approach looks great -- thanks Jon

app/lib/cluster_viz_service.rb
```diff
@@ -275,4 +275,26 @@ def self.raw_matrix_for_cluster_cells(study, cluster)
...
+  #
+  # * *params*
+  #   - +cluster+    (ClusterGroup) => Clustering object being used as control cell list
+  #   - +annotation_name+  (String) => Name of requested annotation
+  #   - +annotation_type+  (String) => Type of requested annotation (should be 'group')
+  #   - +annotation_scope+ (String) => Scope of requested annotation ('study' or 'cluster')
+  def self.cells_by_annotation_label(cluster, annotation_name, annotation_type, annotation_scope)
+    study = cluster.study
+    cells = cluster.concatenate_data_arrays('text', 'cells')
+    annotation = { name: annotation_name, scope: annotation_scope, type: annotation_type }
```
eweitz, line 291: Blocker: We've discussed the extraneous `annotation_type` parameter for DE a few times, and typically left it as simply doing what Ingest Pipeline expects, but that strikes me as not a good enough reason to thread this extraneous parameter through all our DE code.

In other words, the fact that the DE pipeline expects an `annotation_type` parameter that only ever takes one value does not entail that all upstream code should also have that parameter.  Required parameters that take only 1 value shouldn't be parameters.  Wiring them through our interfaces is poor encapsulation -- noise that decreases readability and maintainability.

So let's omit `annotation_type` parameters from signatures in Rails DE code.

```suggestion
  def self.cells_by_annotation_label(cluster, annotation_name, annotation_scope)
    study = cluster.study
    cells = cluster.concatenate_data_arrays('text', 'cells')
    annotation = { name: annotation_name, scope: annotation_scope, type: 'group' }
```

app/lib/differential_expression_service.rb
```diff
@@ -200,6 +200,13 @@ def self.validate_annotation(cluster_file, study, annotation_name, annotation_ty
     identifier = "#{annotation_name}--#{annotation_type}--#{annotation_scope}"
     raise ArgumentError, "#{identifier} is not present" if annotation.nil?
     raise ArgumentError, "#{identifier} cannot be visualized" unless can_visualize
+
+    # last, validate that the requested annotation & cluster will provide a valid intersection of annotation values
+    # specifically, discard any annotation/cluster combos that only result in one distinct label
+    cells_by_label = ClusterVizService.cells_by_annotation_label(cluster, annotation_name, annotation_type, annotation_scope)
```
eweitz, line 206: Blocker: let's start addressing the extraneous `annotation_type` parameter.  I don't object to further refining it later, but I think it's important we start.

```suggestion
    cells_by_label = ClusterVizService.cells_by_annotation_label(cluster, annotation_name, 'group', annotation_scope)
```

eweitz (changes requested):
Code looks good overall!  I note two trivial blockers.

app/lib/cluster_viz_service.rb
bistline: I won't argue the point - the parameter is technically unnecessary, and was in place solely to match the interface for DE via ingest - but it is out of scope for this PR.  The changes are more substantive than just these two edits - there are quite a few more that would need to be made, especially considering tests.  I'm happy to ticket that work and do it elsewhere, but it's not part of the scope for this PR, and shouldn't hold it up.

bistline: @eweitz https://broadworkbench.atlassian.net/browse/SCP-4454

eweitz: Perfect, another ticket this sprint makes sense!

eweitz (approved):

app/models/differential_expression_result.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  # individual filename of label-specific result
+  def filename_for(label)
+    basename = [
+      cluster_name,
+      annotation_name,
+      label,
+      annotation_scope,
+      'wilcoxon'
```
ehanna4, line 57: Is there a chance we'd want to change this type of DE - is there any case for having the analysis type as a param rather than hardcoded?

ehanna4, line 57: Is there a chance we'd want to change this type of DE - is there any case for having the analysis type as a param rather than hardcoded?

ehanna4: Or maybe is there a case to store this value similar to `MIN_OBSERVED_VALUES` ?

ehanna4: Or maybe is there a case to store this value similar to `MIN_OBSERVED_VALUES` ?

bistline: That's an interesting suggestion - I asked Jean and she thinks that there's a possibility that we might do that.  Absent any concrete plans to do so, I think it's probably out of scope for this PR, but I have another one for removing the `annotation_type` parameter across differential expression-related code, and that would fold in nicely with that.  I'll open a ticket for that as well.

ehanna4: nice yeah agreed out of scope for this PR, thanks for following up on that!

jlchang (approved):
@ehanna4 and I performed a functional review - ran database migrations, saw DE results check in `development.log`. Further verified that there is an entry called `differentialExpression` in the /explore API response with outputs that matched my local study.

Note that for cluster name caching for the ClusterGroup object, we will want to perform cleanup is the cluster, metadata or expression matrix file of DE files if the associated cluster, metadata or expression matrix is deleted so we don't show DE results for deleted data. That work is tracked in follow-on ticket SCP-4462.

jlchang:
@devonbush Before this PR is merged to production, we should check staging [SCP294](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP294) when this PR is deployed to staging to confirm that invalid annotation/label combinations are no longer available in the DE dropdown menu (and/or the DE button does not show up - for annotations with a single label).

PR #1545, opened 2022-06-22
Remove `annotation_type` parameter for DE, adding `computational_method` (SCP-4454, SCP-4456)

bistline:
#### BACKGROUND
The parameter `annotation_type` has been identified as superfluous for differential expression jobs, as it can only ever have the value of `group` - `numeric` annotations do not qualify for DE.  In addition, it was proposed that we could add a new parameter - `computational_method` - in the case where we decide to support other DE methodologies beside `wilcoxon`.  Currently, our DE library in [scanpy](https://scanpy.readthedocs.io/en/stable/generated/scanpy.tl.rank_genes_groups.html#scanpy.tl.rank_genes_groups) supports 3 others besides`wilcoxon`: `logreg`, `t-test`, and `t-test_overestim_var`.

#### CHANGES
Wiring for passing `annotation_type` has largely been removed, though it is still present in `DifferentialExpressionParameters` as it is still a required input to DE jobs in `scp-ingest-pipeline`, but is now hard-coded to `group`.  In addition, `computational_method` is now being stored in `DifferentialExpressionResult`, with the default of `wilcoxon`, and support for the other 3.  Extra validation code has been added to ensure that only `group` based annotations are used.  This update also adds test coverage for `ClusterVizService.cells_by_annotation_label`, as it was previously uncovered.

#### MANUAL TESTING
_Note: you may use an existing study that has raw counts data, and both a group- and numeric-based annotation.  The instructions below are for creating a new, minimally-initialized study._

1.  Pull branch and ensure all migrations have been run with `bin/rails db:migrate`, then boot as normal and sign in
2. Create a new study, and upload `test/test_data/expression_matrix_example.txt` as a raw count file, and `test/test_data/cluster_example.txt` as a cluster file and wait for ingest to complete (should be ~2-3 min)
3. In a Rails console session, using the study above, run a differential expression job with the following parameters and validate it returns `true`:
```
study = Study.find_by(accession: '<accession of new study>')
DifferentialExpressionService.run_differential_expression_job(study.default_cluster.study_file, study, study.user, annotation_name: 'Category', annotation_scope: 'cluster' )
=> true
```
4. Attempt to run another DE job with the following parameters and ensure that an `ArgumentError` is raised:
```
DifferentialExpressionService.run_differential_expression_job(study.default_cluster.study_file, study, study.user, annotation_name: 'Intensity', annotation_scope: 'cluster' )
...
Traceback (most recent call last):
        3: from (irb):11
        2: from app/lib/differential_expression_service.rb:136:in `run_differential_expression_job'
        1: from app/lib/differential_expression_service.rb:201:in `validate_annotation'
ArgumentError (Intensity--group--cluster is not present or is numeric-based)
```

### DIFF
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -283,12 +283,11 @@ class ClusterVizService
   # * *params*
   #   - +cluster+    (ClusterGroup) => Clustering object being used as control cell list
   #   - +annotation_name+  (String) => Name of requested annotation
-  #   - +annotation_type+  (String) => Type of requested annotation (should be 'group')
   #   - +annotation_scope+ (String) => Scope of requested annotation ('study' or 'cluster')
-  def self.cells_by_annotation_label(cluster, annotation_name, annotation_type, annotation_scope)
+  def self.cells_by_annotation_label(cluster, annotation_name, annotation_scope)
     study = cluster.study
     cells = cluster.concatenate_data_arrays('text', 'cells')
-    annotation = { name: annotation_name, scope: annotation_scope, type: annotation_type }
+    annotation = { name: annotation_name, scope: annotation_scope, type: 'group' } # only valid for group-based
     labels = get_annotation_values_array(study, cluster, annotation, cells, nil, nil)
     cells_by_label = {}
     labels.each_with_index do |label, index|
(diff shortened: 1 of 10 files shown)


eweitz (approved):
Nice simplification!

PR #1552, opened 2022-06-27
Using pre-built Docker image in CI/CD (SCP-4493)

bistline:
### BACKGROUND
Currently, around 7.5 minutes is spent building a Docker image during a CI run to use for that job.  Most of this is running `bundle install` to install all required Ruby gems, and then the remainder running `yarn install` to install all required JavaScript packages.  In local Dockerized development, it can take as much as 25-30 minutes to build/spin up a local instance, making using Docker locally extremely unpalatable.

### CHANGES
This update introduces a pre-built Docker image that has all gems and packages pre-installed.  Now, deployed environments simply pull this image and call `docker run` directly via our convenience scripts.  This mounts the `node_modules` folder from the image into the project directory, allowing for JS components to be rendered directly via `rake assets:precompile` in deployed environments, or emitted via `vite` in local development.

In comparing CI runs before/after this change, there was a savings of ~4.5 minutes in startup time.  In local development, the time required to spin up a local instance fell by an order of magnitude (~25m to 2.5m).  While this hasn't been tested yet for a deployment, as there is additional work do be done in SCP-4494 to wire this into our CD pipeline, the expectation is that it will reduce the amount of actual downtime for the portal from ~5m to ~2.5m.

NOTE: Before this PR is merged, SCP-4494 will need to be addressed so that our deployment automation does not break due to missing Docker images.

### MANUAL TESTING
1. Pull this branch and make sure you have Docker running
2. Run the following command to boot the portal in Dockerized mode (this will pull the pre-built image using the `development` tag:
```
bin/load_env_secrets.sh -p secret/kdux/scp/development/$(whoami)/scp_config.json \
                        -r secret/kdux/scp/development/$(whoami)/read_only_service_account.json \
                        -s secret/kdux/scp/development/$(whoami)/scp_service_account.json
```
3. Confirm that you see the startup process complete in roughly 2-3 minutes (you can watch for the following text in `STDOUT`) and that you can bring the portal up locally at [https://localhost/single_cell](https://localhost/single_cell):
```
*** Runit started as PID 256
ok: run: /etc/service/nginx-log-forwarder: (pid 264) 0s
Jun 28 17:13:56 localhost cron[263]: (CRON) INFO (pidfile fd = 3)
Jun 28 17:13:56 localhost cron[263]: (CRON) INFO (Running @reboot jobs)
[ N 2022-06-28 17:13:56.1629 271/T1 age/Wat/WatchdogMain.cpp:1373 ]: Starting Passenger watchdog...
[ N 2022-06-28 17:13:56.1882 274/T1 age/Cor/CoreMain.cpp:1340 ]: Starting Passenger core...
[ N 2022-06-28 17:13:56.1909 274/T1 age/Cor/CoreMain.cpp:256 ]: Passenger core running in multi-application mode.
[ N 2022-06-28 17:13:56.2100 274/T1 age/Cor/CoreMain.cpp:1015 ]: Passenger core online, PID 274
[ N 2022-06-28 17:13:59.3326 274/T5 age/Cor/SecurityUpdateChecker.h:519 ]: Security update check: no update found (next check in 24 hours)
```


bin/boot_docker
```diff
@@ -51,7 +51,8 @@ GOOGLE_PRIVATE_KEY=$GOOGLE_PRIVATE_KEY
 GOOGLE_CLIENT_EMAIL=$GOOGLE_CLIENT_EMAIL
 GOOGLE_CLIENT_ID=$GOOGLE_CLIENT_ID
 GOOGLE_CLOUD_PROJECT=$GOOGLE_CLOUD_PROJECT
-DOCKER_IMAGE_VERSION='latest'
+DOCKER_IMAGE_NAME='gcr.io/broad-singlecellportal-staging/single-cell-portal'
+DOCKER_IMAGE_VERSION='development'
```
eweitz, line 55: Blocker: please add code comments that explain why we reference "staging" and "development" in this artifact that (IIUC) is used in production deployment.  I likely misunderstand something, but at a glance and without much context it's alarming.  Code comments here would help allay such concerns.

I don't think I'd appreciated that we use staging namespaces for artifacts that affect production deployment -- e.g. `gcr.io/broad-singlecellportal-staging/rails-baseimage:2.1.0` in the Dockerfile.  That alarms me, because I can imagine someone else on the team presuming "staging" or "development" artifacts don't require a ton of care before deleting, and thus inadvertently breaking our production deployment process.

Ideally, I think, we'd typically omit use the the strings "staging" or "development" in Docker image names and versions referenced in our code.  However, I haven't wrapped my head fully around when these images propagate to Google Cloud Registry / GCR, and I don't have a sense of how feasible it is to omit those strings.  So I think it's worth considering, but I don't insist it's blocking.

In other words, refining the name and version strings would be great, but my only blocking concern is that we need code commentary briefly explaining why we use these "staging" and "development" strings, wherever they're used.

bin/build_image.sh
```diff
@@ -0,0 +1,54 @@
...
+cat <<EOF
+$0 [OPTION]
+-v VALUE  specify version tag for docker image
+-h        print this text
+EOF
+)
+
+function main {
+  VERSION_TAG=$(extract_release_tag "0")
+  IMAGE_NAME='gcr.io/broad-singlecellportal-staging/single-cell-portal'
```
eweitz, line 21: Blocker: see note beginning "please add code comments".

.github/workflows/run-all-tests.yml
```diff
@@ -29,8 +29,11 @@ jobs:
           CI: true
         run: |
           export VAULT_TOKEN=$( vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID )
+          DOCKER_IMAGE_NAME="gcr.io/broad-singlecellportal-staging/single-cell-portal"
```
eweitz, line 32: Blocker: see note beginning "please add code comments".  (That said, this is a case where using "staging" is a bit more intuitive.)

bin/build_image.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# build the portal Docker image and tag accordingly, then push to GCR
+# cleans up untagged images after push when overwriting tags
```
eweitz, line 4: Your PR description here seems like it could help maintainers down the road.

```suggestion
# cleans up untagged images after push when overwriting tags; run in CI/CD
# More context: https://github.com/broadinstitute/single_cell_portal_core/pull/1552
```

bin/build_image.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# build the portal Docker image and tag accordingly, then push to GCR
```
eweitz, line 3: Could you add some code commentary around what calls this `build_image.sh` script?  Searching this file's name as a string returns no other code calling to reference it.

test/Dockerfile-test
```diff
@@ -0,0 +1,13 @@
+# use SCP base Rails image, configure only project-specific items here
+FROM gcr.io/broad-singlecellportal-staging/single-cell-portal:development
```
eweitz, line 2: FWIW, like the reference in the GitHub Actions .yml file, this case of using "staging" and "development" in the name / version for the Docker image doesn't alarm me.

eweitz (changes requested):
Speeding up CI from 30 minutes to 25 minutes is a solid development velocity improvement!

I note a blocking concern about naming.  For me, those would be sufficiently resolved via trivial code comments; see below.

bin/boot_docker
bistline: Great questions - the TL;DR is that the naming scheme for Google Container Registry is `[service]/[repository]/[image]:[tag]`, where `gcr.io` is the service, `broad-singlecellportal-staging` is the repository, `single-cell-portal` is the image name, and `development` is the tag of the image.  Therefore, we can't can't change the repository, as that's how Docker knows what GCP project to look in.  The argument for using `development` is a reference to the fact that this image will be continually rebuilt to reflect the state of gems/packages required on the `development` branch of `single_cell_portal_core`.  For production releases, the tag would correspond to the version number of the upcoming release, e.g. `1.20.0`.

"Why did we put everything in staging", you ask?  Basically, the main reason was that you cannot specify permission on a per-image basis.  Everything is either public, or private.  At the time, we decided we didn't want a lot of publicly accessible resources hosted out of our production GCP project.  Probably not a real security concern, but that was the decision we made.  After that, it was simply "well, that's where all the other Docker images are..."

I agree that it isn't ideal, but migrating all of our images over to our production project would entail changing a great deal of automation - basically our entire CI/CD pipeline for our 3 main repositories.  That means changes to Jenkins as well as CircleCI/Github Actions jobs, on top of renaming everything in each repository. All that being said, I think it's a completely valid concern - one that I share - and that we should in fact change it, but not in the scope of this PR.  We have bandwidth reserved for tech debt this quarter, and this is a nice project to lump in with that.

bin/build_image.sh
bistline: I think that the work that will be done as a part of SCP-4494 will clarify this - I haven't completely decided on the exact mechanism.  It will either be a separate Jenkins job (which I've started working on already), or this script will just get added to each existing deployment job.

The clean way is the separate job which then triggers the deployment on a successful build (Jenkins supports this, and could give us completely automated deployments on merges if we wanted).  However, adding a single line to each existing job would be much easier: `bin/build_image.sh && bin/deploy.sh [params]`.

bin/boot_docker
eweitz: Very helpful context, and I agree fully resolving this tech debt is out of scope for this PR.  

My only blocking concern here (and elsewhere I note "Blocker" in this PR) is that we need code comments.  Perhaps:

```suggestion
# The "broad-singlecellportal-staging" GCR repository is used in production.  
# The "development" tag is used in non-production deployment.  For production deployment, tag is version number of upcoming release, e.g. 1.20.0.  
# More context: https://github.com/broadinstitute/single_cell_portal_core/pull/1552#discussion_r910424433
DOCKER_IMAGE_NAME='gcr.io/broad-singlecellportal-staging/single-cell-portal'
DOCKER_IMAGE_VERSION='development'
```

.github/workflows/run-all-tests.yml
eweitz: Perhaps:

`# "broad-singlecellportal-staging" GCR repository is also used in production.`

bin/build_image.sh
bistline: [9db0be5](https://github.com/broadinstitute/single_cell_portal_core/pull/1552/commits/9db0be51a5d54f69d19138edbe4fe6c7388316fa)

bin/boot_docker
bistline: [9db0be5](https://github.com/broadinstitute/single_cell_portal_core/pull/1552/commits/9db0be51a5d54f69d19138edbe4fe6c7388316fa)

eweitz (approved):

bistline:
> Are any updates needed for the README with this change?

Great question!  Nothing comes to mind off the top of my head, but I'll take a pass through the Docker-specific stuff and see.  But it should automatically pull the `development` image and run w/o needing to run `docker build`.

bistline:
Build failure is related to ongoing Terra incident (via Google): https://support.terra.bio/hc/en-us/articles/7361836697115-Service-Incident-July-15-2022-Google-resource-access-#heading-1

PR #1553, opened 2022-06-28
r-formatted csfv correction

ehanna4:
Update CSFV for R-formatted dense matrices.

an R-formatted dense matrix will be missing the 'GENE' first value in the header but should still have the spot for it.

These are valid (checked by ingesting locally via Jean :) )
```
,CELL_0001,CELL_0002,CELL_0003,CELL_0004
APC,0,0,0,0
MYH11,0,7.092,0,4
```

```
	CELL_0001	CELL_0002	CELL_0003	CELL_0004
APC	0	0	0	3
MYH11	0	7.092	0	4
```

This is NOT valid (missing value at the end):
```
GENE,CELL_0001,CELL_0002,CELL_0003,
APC,0,0,0,0
MYH11,0,7.092,0,4
```

### DIFF
--- a/app/javascript/lib/validation/expression-matrices-validation.js
+++ b/app/javascript/lib/validation/expression-matrices-validation.js
@@ -185,10 +185,9 @@ function getDenseMatrixDelimiter(rawHeader, rawNextTwoLines) {
  *
  * The "cap" for an expression matrix file is the first row also called the "header"
  *
- * A dense matrix header must start with the value 'GENE' or if the file is R-formatted it can:
- *  - Not have GENE in the header and:
- *    - Have one less entry in the header than each successive row OR
- *    - Have "" as the last value in header.
+ * A dense matrix header must start with the value 'GENE' or if the file is R-formatted it can
+ * start with a blank "" value as the first value.
+ *
  */
 function validateDenseHeader(header, nextTwoLines) {
   const issues = []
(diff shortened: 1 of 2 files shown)


app/javascript/lib/validation/expression-matrices-validation.js
```diff
@@ -185,10 +185,9 @@ function getDenseMatrixDelimiter(rawHeader, rawNextTwoLines) {
  *
  * The "cap" for an expression matrix file is the first row also called the "header"
  *
- * A dense matrix header must start with the value 'GENE' or if the file is R-formatted it can:
- *  - Not have GENE in the header and:
- *    - Have one less entry in the header than each successive row OR
- *    - Have "" as the last value in header.
+ * A dense matrix header must start with the value 'GENE' or if the file is R-formatted it can
```
eweitz, line 188: We conventionally double-quote this kind of user-facing text:

```suggestion
 * A dense matrix header must start with the value "GENE" or if the file is R-formatted it can
```

eweitz (approved):
Elegant fix for a smoother UX!  Code looks good.  I suggest a non-blocking copyedit.

bistline:
@ehanna4 I can't get the invalid CSV file to fail - it keeps passing validation.  I know CSFV is working because I can add `test/test_data/expression_matrix_example_bad.txt` and that fails with the correct message (bad `GENE` keyword).

ehanna4:
@bistline thanks for catching that! Should be good to go now

jlchang (approved):
Functional review:
R-formatted test files pass CSFV and invalid format fails :)

PR #1555, opened 2022-06-30
Automated cleanup of differential expression results on file deletion (SCP-4462)

bistline:
#### BACKGROUND
If a study has differential expression results, and then the study owner deletes a file that was used to create those results, there was no automated way of cleaning up those database records & DE output files.  Also, there is no way to directly associate a particular result file with an associated study file.  This can mean that invalid DE results can potentially be viewed in the Explore tab.

#### CHANGES
This update integrates removing MongoDB records & GCS output files for DE results into `DeleteQueueJob`, so that when a user deletes an associated study file, any corresponding DE records/files will be automatically deleted.  This will prevent visualizing invalid outputs, and keep the database/bucket in sync regarding these files.  The rules for deletion are that for any clustering/raw counts matrix, deleting that file will remove all DE outputs that used those files.  For metadata files, it will remove all `study` scoped DE outputs.  This is leveraged by adding getters in the `DifferentialExpressionResult` model to access associated clustering/annotation/matrix files.

#### MANUAL TESTING
1. Boot as normal and start DelayedJob
2. Either create a new study w/ minimal DE outputs (such as the instructions from [here](https://github.com/broadinstitute/single_cell_portal_core/pull/1545)), or load an existing study w/ DE outputs in the upload wizard
3. In a separate tab, open the study's GCS bucket (accessible from the "study details" page) and verify there are DE output files there
4. Back in the upload wizard, delete the associated raw count file, and wait ~10s for the delete job to complete
5. In the bucket, verify that the `_scp_internal/differential_expression` folder has no more DE outputs
6. Go to the Explore tab for this study, and verify that the Differential Expression button is no longer shown


eweitz (approved):

bistline:
Test failures are tracked in https://broadworkbench.atlassian.net/browse/SCP-4506, unrelated to this PR.

ehanna4 (approved):

PR #1559, opened 2022-07-05
Fix broken tests (SCP-4506)

ehanna4:
Broke the tests with the addition of:
`header.lastIndexOf('') !== 0` in https://github.com/broadinstitute/single_cell_portal_core/pull/1553
which was checking that the only blank columns would be in position 0 for the header for an r-formatted dense matrix. However, if there are no indexes of `''` then `header.lastIndexOf('') !== 0` returns -1 so the check needed to be adjusted to account for that.

### DIFF
--- a/app/javascript/lib/validation/expression-matrices-validation.js
+++ b/app/javascript/lib/validation/expression-matrices-validation.js
@@ -204,10 +204,10 @@ function validateDenseHeader(header, nextTwoLines) {
     isValid = false
   }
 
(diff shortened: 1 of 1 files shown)


bistline (approved):
Looks good - I can confirm `yarn ui-test` passes all JS tests:
```
Test Suites: 47 passed, 47 total
Tests:       179 passed, 179 total
Snapshots:   0 total
Time:        82.092 s
```

PR #1563, opened 2022-07-07
Fix UX issues with create annotation button (SCP-)

ehanna4:
Fixes:
- Plus/Minus button is better positioned
- Added tooltip for the button to help users understand what it is
- Cancel button closes create annotation panel
- Better positioning of spinner

To test:
- pull this branch
- sign in and go to a study you can edit
- add an annotation

https://user-images.githubusercontent.com/54322292/177869250-69d1de1b-3c7b-47e0-aca9-2e953dd6eaf3.mov

### DIFF
--- a/app/javascript/components/visualization/controls/CreateAnnotation.jsx
+++ b/app/javascript/components/visualization/controls/CreateAnnotation.jsx
@@ -57,6 +57,7 @@ function CreateAnnotation({
   function handleCancel() {
     setUserLabels([])
     setAnnotationName('')
+    handlePanelToggle()
   }
 
   /** renders an appropriate modal and updates the cluster params with the response from the server */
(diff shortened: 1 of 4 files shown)


jlchang (approved):
Functional review: confirmed all noted "buggy" issues with create annotations UI have been resolved.

The general gist of the code changes seemed to make sense to me :)

app/javascript/components/visualization/controls/CreateAnnotation.jsx
```diff
@@ -352,3 +342,29 @@ function validationMessages(userLabels, annotationName, annotations) {
...
+      if (expanded) {
+        return 'Explore existing annotations'
+      } else {
+        return 'Create custom annotations'
+      }
+    }
+  }
+
+  return <button className = {buttonClass}
+    data-analytics-name = {!isUserLoggedIn() ? 'toggle-create-annotation-signedout' :'toggle-create-annotation'}
```
eweitz, line 364: I don't know if it these values were used before, but if not, then using only one seems most conventional.

```suggestion
    data-analytics-name = 'toggle-create-annotation'
```

Regarding convention: a few other elements have different displays when unauthenticated -- e.g. Download button on home page, Download tab in Explore page -- but it seems they all have the same Mixpanel identifier / name whether the user is signed in or not.  Auth state is somewhat of a separate concern, and can be queried via e.g. `authenticated: false` in Mixpanel.

If you can find pre-existing example of this pattern, or you simply think it's worth adopting in certain cases, then the current code works for me!

eweitz (approved):
Code looks good!  I suggested a non-blocking refinement.

app/javascript/components/visualization/controls/CreateAnnotation.jsx
ehanna4: This was the `data-analytics-name` prior to my changes so changing that now would mean needing to go through our existing mixpanel reports to update them and that feels like a separate effort that should be combined with combing through and generally standardizing metric names in the codebase and updating mixpanel reports accordingly. If that seems like a worthwhile effort I can make a ticket for the backlog?

PR #1565, opened 2022-07-11
Service worker cache for faster local frontend development (SCP-4379)

eweitz:
This speeds up frontend-only development. Cycle iteration times are often 5 seconds (~25%) faster.

### Background
Previously, backend SCP API responses always hit the server.  Since these responses are equivalent when changing only frontend code, such GET requests were usually basically wasted time.  Now -- using [service worker cache](https://web.dev/learn/pwa/caching/) (aka web platform `Cache` or `CacheStorage` API) -- we can store those responses on our local web browser client, and completely eliminate backend processing time for them.

### Impact
This can save around 5 seconds per local page load, [reducing e.g. time to first scatter plot from 17.5 s to 12.6 s (28%)](https://mixpanel.com/s/3CUjc6).  It mainly optimizes frontend cycle times not covered by Vite hot module replacement (HMR).  That said, it can also benefit those if e.g. the updated component triggers downstream SCP API requests.

### Configuration
Service worker (SW) caching for local development is currently opt-in.  

The following basic Vite command won't have caching:
```
bin/vite dev
```

To enable SW caching, use this new Vite command in your local Terminal:
```
VITE_FRONTEND_SERVICE_WORKER_CACHE="true" bin/vite dev
```

You'll also notice a new file `version.txt` in our root directory.  This is used to configure the SW cache key, and as part clearing old caches.  It's used only locally right now, so it won't affect staging or production.  The value is now also logged as `scpVersion` for all development Mixpanel `app-frontend` events, which might be worth refining and extending to our other environments and loggers.

### How to clear
Clear this cache like any other service worker cache:
* Open DevTools
* Go to Application -> Cache
* Click on the latest `scp-...` key (e.g. `scp-development-1.20.0`)
* Click the "x" button to delete whole cache

You can also delete specific cache entries by highlighting and deleting them in the same view, at right.

Service_worker_cache_for_faster_local_frontend_development__SCP_2022-07-11

### Analytics
Whether SW cache was enabled and hit is now recorded as part of Mixpanel events, e.g. in:
* `'perfTime:serviceWorkerCacheHit': true`
* `'perfTime:cache': 'service-worker'`

Values for `perfTime:backend` and other eliminated phases are now -1, re-using a pattern from plot data caching (#1036).

### Testing

An automated test verifies some caching here.  To manually test:

1. `yarn install` to get new JS dependencies
2. Stop Vite, then run in terminal: `VITE_FRONTEND_SERVICE_WORKER_CACHE="true" bin/vite dev`
3. Go to home page
4. Open DevTools, go to Application -> Cache
5. Confirm a cache `scp-development-1.20.0` appears and is populated with some SCP API requests
6. Go to Network panel, filter to "cell/api", confirm some SCP API requests were sent
7. Refresh home page
8. Go to Network panel, filter to "cell/api", confirm some SCP API requests were not sent
9. Go to Console panel, click "Default levels" dropdown, select "Verbose"
10. Confirm some messages in Console begin "Service worker cache hit"
11. Stop Vite, then run in terminal: `bin/vite dev`
12. Refresh home page.  If you get an error, refresh once more.
13. Go to Network panel, filter to "cell/api", confirm some SCP API requests were sent
14. Confirm no messages in Console begin "Service worker cache"
15. Deselect "Verbose" logging in Console levels dropdown (resetting to state before step 9)

This satisfies SCP-4379.


app/javascript/lib/scp-api.jsx
```diff
@@ -791,19 +795,36 @@ export default async function scpApi(
...
+    // Milliseconds taken to fetch data from API
+    // Suboptimal, but retained until at least Q4 2021 for continuity.
+    // Use `perfTime:full` for closest measure of user-perceived duration.
+    legacyBackendTime = performance.now() - perfTimeStart
+  }
+
+  perfTimes.legacyBackend = legacyBackendTime
+  perfTimes.serviceWorkerCacheHit = isServiceWorkerCacheHit
+
+  // response = { ok: true }
```
ehanna4:
```suggestion

```

app/javascript/lib/metrics-perf.js
```diff
@@ -171,19 +181,63 @@ export function calculatePerfTimes(perfTimes) {
       // so, e.g. the cluster plot will wait until after the expression data is fetched, and then render itself
       // based on a subset of that data
       'perfTime:frontend:other': now - perfTimes.requestStart - perfTimes.plot,
-      'perfTime:frontend:plot': perfTimes.plot,
-      'cache': 'client'
+      'perfTime:frontend:plot': perfTimes.plot
     }
     const perfProps = roundValues(Object.assign({}, rawPerfProps))
     perfProps['perfTime:url'] = perfTimes.url
+
```
devonbush, line 188: This calculatePerfTimes function is now too long-- I'm not sure the best places to slice out helper functions, but we definitely need some

app/javascript/lib/scp-api.jsx
```diff
@@ -11,16 +11,20 @@ import _compact from 'lodash/compact'
...
+import {
+  clearOldServiceWorkerCaches, fetchServiceWorkerCache
+} from './service-worker-cache'
+import { getSCPContext } from '~/providers/SCPContextProvider'
+import { STEP_NOT_NEEDED } from './metrics-perf'
 
-
-const env = getSCPContext().environment
+// On each page load, check for old SCP caches, delete any found
+clearOldServiceWorkerCaches()
```
devonbush, line 27: I'd lean towards having this line of code in application.js.  I think it's a generally healthier pattern to have startup code there, and have modules not do anything automatically when they are loaded.  This saves us, for example, from running this method every time scp-api is imported in test

app/javascript/lib/scp-api.jsx
```diff
@@ -791,18 +795,34 @@ export default async function scpApi(
...
+    isServiceWorkerCacheHit = fetchSWCacheResult[1]
+    legacyBackendTime = STEP_NOT_NEEDED
+  } else {
+    response = await fetch(url, init).catch(error => error)
+
+    // Milliseconds taken to fetch data from API
+    // Suboptimal, but retained until at least Q4 2021 for continuity.
+    // Use `perfTime:full` for closest measure of user-perceived duration.
+    legacyBackendTime = performance.now() - perfTimeStart
+  }
```
devonbush, line 821: Let's keep as much of the cache and perftime logic as we can in `service-worker-cache.js`. 
```suggestion
   if (isServiceWorkerCacheEnabled) {
      response = await fetchServiceWorkerCache(url, init, perfTimes)
   }
   if (!response) {
      // fetch normally
   }
```

app/javascript/lib/service-worker-cache.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Fetch, leveraging service worker cache if enabled and available
+ *
+ * TODO (SCP-4508): Account for same URL, different sign-in state in service worker cache
+ */
+export async function fetchServiceWorkerCache(url, init) {
+  const swCache = await caches.open(serviceWorkerCacheKey)
+  let response = await swCache.match(url)
+  let hitOrMiss = 'hit'
```
devonbush, line 24: BLOCKER: we shouldn't use magic strings to track boolean state
```suggestion
  let isHit = true
```

app/javascript/lib/service-worker-cache.js
```diff
@@ -0,0 +1,55 @@
...
+  if (typeof response === 'undefined') {
+    response = await fetch(url, init).catch(error => error)
+    await swCache.put(url, response.clone())
+    hitOrMiss = 'miss'
+  }
+  console.debug(`Service worker cache ${hitOrMiss} for SCP API fetch of URL: ${url}`)
```
devonbush, line 30:
```suggestion
  console.debug(`Service worker cache ${isHit ? 'hit' : 'miss'} for SCP API fetch of URL: ${url}`)
```

vite.config.ts
```diff
@@ -1,9 +1,16 @@
...
 import react from '@vitejs/plugin-react'
+import { readFileSync } from 'fs'
 
+// Match latest non-draft at https://github.com/broadinstitute/single_cell_portal_core/releases
+const version = readFileSync('version.txt', { encoding: 'utf8' })
 
 export default defineConfig({
+  'define': {
+    '__SCP_VERSION__': version,
+    '__FRONTEND_SERVICE_WORKER_CACHE__': process.env.VITE_FRONTEND_SERVICE_WORKER_CACHE
```
devonbush, line 12: the service worker cache doesn't really have anything to do with Vite specifically, right?  If we went back to webpacker it would still work.  So renaming the variable to just FRONTEND_SERVICE_WORKER_CACHE is probably clearer

devonbush (approved):
Glad to see this is giving you big speedups in development.  One blocker and a bunch of suggestions for making it cleaner.  I'm going to go ahead and approve since I trust you to fmake the fixes more than I trust myself to have time to come back to SCP for the re-review :)

app/javascript/lib/metrics-perf.js
```diff
@@ -171,19 +181,63 @@ export function calculatePerfTimes(perfTimes) {
...
     return perfProps
   }
 
   const plot = perfTimes.plot ? perfTimes.plot : 0
 
-  const perfEntry =
-    performance.getEntriesByType('resource')
-      .filter(entry => entry.name === perfTimes.url)[0]
+  let perfEntry = null
+  const resourcePerfEntries = performance.getEntriesByType('resource')
```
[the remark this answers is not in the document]
eweitz: Good question!  It caught my eye because `filter` is clearly suboptimal, then I assumed `for` would be faster than `find` based on `for` often being faster than newer equivalents.  It turns out [`.find()` is faster than `for...break`](https://stackoverflow.com/a/50843994), so I'll use that.

vite.config.ts
eweitz: It'd indeed be clearer, but Vite unfortunately requires custom environment variables passed in via terminal command to be prefixed with `VITE_`.

app/javascript/lib/metrics-perf.js
eweitz: Indeed -- I refactored in 8143beb and b00b600.

app/javascript/lib/scp-api.jsx
eweitz: I began this refactor, then realized it'd be more complex and unduly couple the SW cache to SCP API.  But your suggestion did help me realize a subtle bug in backend time accounting when this cache is enabled, which I fixed.

app/javascript/lib/service-worker-cache.js
eweitz: Good catch, this was a vestige from early work.  Refactored in 0c53771.

app/javascript/lib/service-worker-cache.js
eweitz: Done as part of work per my [other comment](https://github.com/broadinstitute/single_cell_portal_core/pull/1565#discussion_r923602407).

PR #1566, opened 2022-07-13
Enable feature flag for CSFV again (SCP-)

ehanna4:
At some point we lost the use of the CSFV feature flag so that going at setting the flag to 'false' did not actually stop the validation from happening on file uploads.

This PR seeks to reinstate this logic.

To test:
- Pull this PR and start it up locally
- load up localhost:3000 and sign in as your admin account

I found this easiest to open up 2 tabs for the following

Test scenario using user FF:
- in one tab open the uploadwizard for a study you own
- in another tab open the featureflag settings
- set the ff for `clientside_validation` for your user to be `true`
- refresh the tab with the upload wizard
- attempt to upload an invalid file in the upload wizard and observe error messages
-----------------------------
- now go back to the ff settings tab and set the `clientside_validation` flag to false
- refresh the tab with the upload wizard
- attempt to upload an invalid file in the upload wizard and observe no error messages

### DIFF
--- a/app/javascript/lib/validation/validate-file.js
+++ b/app/javascript/lib/validation/validate-file.js
@@ -6,6 +6,7 @@ import { oneGiB, oneMiB } from '~/lib/validation/io'
 import ValidateFileContent from './validate-file-content'
 import { logFileValidation } from './log-validation'
 import { fetchBucketFile } from '~/lib/scp-api'
+import { getFeatureFlagsWithDefaults } from '~/providers/UserProvider'
 
 
 /** take an array of [category, type, msg] issues, and format it */
@@ -49,6 +50,13 @@ function validateFileName(file, studyFile, allStudyFiles, allowedFileExts=['*'])
  * @param allowedFileExts { String[] } array of allowable extensions, ['*'] for all
  */
 async function validateLocalFile(file, studyFile, allStudyFiles=[], allowedFileExts=['*']) {
+
+  // if clientside file validation feature flag is false skip validation
+  const flags = getFeatureFlagsWithDefaults()
+  if (!flags?.clientside_validation) {
+    const issues = formatIssues([])
+    return issues
+  }
   const nameIssues = validateFileName(file, studyFile, allStudyFiles, allowedFileExts)
 
   let issuesObj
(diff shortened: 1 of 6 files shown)


jlchang (approved):
Functional review - check that both user- and study-level feature flags for CSFV now work as expected.

app/javascript/lib/validation/validate-file.js
```diff
@@ -49,6 +50,13 @@ function validateFileName(file, studyFile, allStudyFiles, allowedFileExts=['*'])
  * @param allowedFileExts { String[] } array of allowable extensions, ['*'] for all
  */
 async function validateLocalFile(file, studyFile, allStudyFiles=[], allowedFileExts=['*']) {
+
+  // if clientside file validation feature flag is false skip validation
+  const flags = getFeatureFlagsWithDefaults()
+  if (!flags?.clientside_validation) {
```
eweitz, line 56: Blocker: this will skip CSFV if this flag does not exist here, which seems feasible inadvertently.  We should be more robust by skipping CSFV if and only if this flag _exists_ and is false.

```suggestion
  if (clientside_validation in flags && !flags.clientside_validation) {
```

For reference:
```
foo = {'a': true, 'b': false}
if (!foo?.clientside_validation) {console.log('CSFV flag is missing, skipping CSFV')}
CSFV flag is missing, skipping CSFV
if ('clientside_validation' in foo && !foo.clientside_validation) {console.log('CSFV flag is missing, skipping CSFV')}
undefined
```

This would also let us de-noise test code.

test/js/upload-wizard/file-upload-control.test.js
```diff
@@ -5,14 +5,22 @@ import '@testing-library/jest-dom/extend-expect'
...
   })
 
+  beforeEach(
+    jest
+      .spyOn(UserProvider, 'getFeatureFlagsWithDefaults')
+      .mockReturnValue({
+        clientside_validation: true
+      })
+  )
+
```
eweitz, line 23: This is not a blocking concern in itself, but per my blocking comment, this code would be extraneous with the more robust, default-to-CSFV conditional.

```suggestion
```

The same applies to equivalent code in other automated test changes in this PR.

eweitz (changes requested):
Thanks for fixing this!  I think it broke as part of migrating the upload UI to React.

I note a functional blocker, and a broadly applicable, non-blocking maintainability suggestion.

eweitz (approved):

PR #1572, opened 2022-07-26
Allow setting custom GCE machine_type for differential expression jobs (SCP-4408)

bistline:
#### BACKGROUND
Some differential expression jobs fail after being invoked due to an out of memory (OOM) exception, which is denoted by an exit code of `137`.  This prevents these particular annotations from having differential expression results calculated for viewing in the Explore tab.

#### CHANGES
Now, when launching a differential expression job, the default machine type is increased from `n1-highmem-4` (4 cores, 26 GB RAM) to `n1-highmem-8` (8 cores, 52 GB RAM).  The reason for using the pre-built image is that the `n1-highmem` class of machine types in GCE all have the maximum amount available of RAM per core.  Therefore, to get more memory, we must add cores.  

In addition, it is now possible to override the new default of `n1-highmem-8` with any other `n1-highmem` machine type (from `n1-highmem-2` up to `n1-highmem-96`).  This will allow SCP admins to custom tailor individual jobs that may require more resources without needing to deploy new code to the server.  This can be configured with the `machine_type` parameter in `DifferentialExpressionService.run_differential_expression_job`.

This also fixes an issue where duplicate `DifferentialExpressionResult` instances could both point to the same collection of output files.  Now, a uniqueness constraint is placed so that only one result can exist for a given study/cluster/annotation combination.  Also, `DifferentialExpressionService` will check to see if there are already existing results for the requested combination and throw an error if they are present.

#### MANUAL TESTING
1. Pull this branch and start `Delayed::Job` with `bin/rails jobs:work`
2. In a separate terminal, open the Rails console
3. Select any pre-existing `DifferentialExpressionResult` and re-populate the parameters:
```
result = DifferentialExpressionResult.all.sample
study = result.study
cluster_file = result.cluster_file
user = study.user
params = { annotation_name: result.annotation_name, annotation_scope: result.annotation_scope }
```
4. Delete the existing result, and launch a new job to repopulate them:
```
result.destroy
DifferentialExpressionService.run_differential_expression_job(cluster_file, study, user, **params)
```
5. In `development.log`, note the new custom machine type in the `:resources` section of the running pipeline:
```
:resources:
  ...
  :virtual_machine:
    :boot_disk_size_gb: 300
    :machine_type: n1-highmem-8
    :preemptible: false
    ...
```

### DIFF
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -119,7 +119,7 @@ class DifferentialExpressionService
   #   - +user+             (User) => User initiating parse action (for email delivery)
   #   - +annotation_name+  (String) => Name of requested annotation
   #   - +annotation_scope+ (String) => Scope of requested annotation ('study' or 'cluster')
-  #   - +vm_ram+           (Integer) => Override to set RAM for DE VM (will default to 52 GB)
+  #   - +machine_type+     (String) => Override default VM machine type
   #
   # * *yields*
   #   - (IngestJob) => Differential expression job in PAPI
@@ -129,7 +129,7 @@ class DifferentialExpressionService
   #
   # * *raises*
   #   - (ArgumentError) => if requested parameters do not validate
-  def self.run_differential_expression_job(cluster_file, study, user, annotation_name:, annotation_scope:, vm_ram: nil)
+  def self.run_differential_expression_job(cluster_file, study, user, annotation_name:, annotation_scope:, machine_type: nil)
     validate_study(study)
     validate_annotation(cluster_file, study, annotation_name, annotation_scope)
 
@@ -158,9 +158,7 @@ class DifferentialExpressionService
       de_params[:matrix_file_type] = 'dense'
     end
     params_object = DifferentialExpressionParameters.new(de_params)
-    if vm_ram.present?
-      params_object.ram_in_mb = vm_ram
-    end
+    params_object.machine_type = machine_type if machine_type.present? # override :machine_type if specified
 
     if params_object.valid?
       # launch DE job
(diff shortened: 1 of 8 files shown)


app/models/differential_expression_parameters.rb
```diff
@@ -2,6 +2,10 @@
 class DifferentialExpressionParameters
   include ActiveModel::Model
 
+  # acceptable Google N1 machine times
```
eweitz, line 5:
```suggestion
  # acceptable Google N1 machine types
```

app/lib/differential_expression_service.rb
```diff
@@ -183,6 +185,15 @@ def self.validate_annotation(cluster_file, study, annotation_name, annotation_sc
     cluster = study.cluster_groups.by_name(cluster_file.name)
     raise ArgumentError, "cannot find cluster for #{cluster_file.name}" if cluster.nil?
 
+    if DifferentialExpressionResult.where(study: study,
+                                          cluster: cluster,
+                                          annotation_name: annotation_name,
+                                          annotation_scope: annotation_scope).exists?
+      raise ArgumentError,
+            "#{annotation_name} already exists for #{study.accession}:#{cluster_file.name}, " \
+            'please delete the existing results before retrying'
```
eweitz, line 194: Non-blocker: Do these get logged to either Sentry or Mixpanel?  It's unessential now, but seems worthwhile before integrating DE into the upload process.

test/models/differential_expression_parameters_test.rb
```diff
@@ -70,4 +71,9 @@ class DifferentialExpressionParametersTest < ActiveSupport::TestCase
     assert_equal '--annotation-name', DifferentialExpressionParameters.to_cli_opt(:annotation_name)
     assert_equal '--foo', DifferentialExpressionParameters.to_cli_opt('foo')
   end
+
+  test 'should provide default ram_in_mb for DE jobs' do
```
eweitz, line 75: Blocker: this trivial typo could mislead maintainers.

```suggestion
  test 'should provide default ram_in_gb for DE jobs' do
```

eweitz (changes requested):
Nice extension!  I note a trivial blocker on maintainability.

app/lib/differential_expression_service.rb
bistline: No - none of these get logged to Sentry because they're not indicative of something being broken.  We're just throwing an `ArgumentError` if the user specifies a parameter that would result in an invalid DE job.

eweitz: Gotcha, thanks for clarifying.  If the user error could ever be triggered by non-SCP-staff, I think it'd be worth eventually logging to at least Mixpanel.  But my understanding now is that these errors would never be encountered by an end user, and thus aren't valuable to durably log.

test/models/differential_expression_parameters_test.rb
bistline: Yes, exactly - it's totally fine to label something as a blocker, but proactively approving for more trivial changes makes the process smoother.  Thanks again!

jlchang (approved):
Functional review - manual tests work as described, confirmed new default machine_type for PAPI jobs seen in logs.

Also tested custom machine type for DE: Tried running on Breast Milk study data (the original impetus for this ticket, job with default matchine_type still fails with exit 137; manually set machine_type to `n1-highmem-16` to get successful result.

bistline:
> Also tested custom machine type for DE: Tried running on Breast Milk study data (the original impetus for this ticket, job with default matchine_type still fails with exit 137; manually set machine_type to `n1-highmem-16` to get successful result.

@jlchang That's great to hear - hopefully this will make it much easier to generate DE data moving forward.

PR #1574, opened 2022-07-27
Disable TCell integration in local development (SCP-4554)

bistline:
#### BACKGROUND
The TCell web application firewall has begun flagging requests on `localhost` as suspicious and blocking them, preventing developing locally.  This is in part due to the Chrome `sec-ch-ua` header, which has been flagged previously as an issue with requests being forbidden.

#### CHANGES
TCell has been moved to a separate Gemfile group so that the client libraries are only loaded in deployed environments (i.e. `staging` and `production`).  In addition, this update removes unneeded dependencies for the `travis` gem that was previously taken out in favor of Github Actions.

#### MANUAL TESTING
1. Pull branch and run `bundle install`, then start the server normally
2. Pull up the home page in Chrome and validate that it loads

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,6 @@ gem 'sentry-raven'
 gem 'rubyzip'
 gem 'rack-brotli'
 gem 'time_difference'
-gem 'tcell_agent'
 gem 'sys-filesystem', require: 'sys/filesystem'
 gem 'browser'
 gem 'ruby-prof'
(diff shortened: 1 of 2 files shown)


ehanna4 (approved):
looks good! I was not experiencing the same issues as others, but this branch does run and open localhost on chrome and bonus removes the noise i would get from tcell (due to lacking M1 compatability) in terminal when booting up

jlchang (approved):
I was affected by the `403` issue and this change resolves the issue.

Gemfile changes make sense to me! Thanks for the speedy PR :)

eweitz (approved):

PR #1575, opened 2022-07-28
Add file upload for Anndata (.h5ad) and Seurat data to wizard behind Feature flag (SCP-4525)

ehanna4:
Changes:
- Added a new section to the upload wizard for the non-visualizable files
- Added sections for uploading Seurat Data files and H5ad Anndata files
- Rename the Miscellaneous/other section to be just Miscellaneous 
- Added a migration and feature flag to configure the ability to add the H5ad & Seurat files
- Added test for the Seurat Data files and H5ad Anndata files utilizing the style of test for the reference image test

Technical Notes:
- Currently there is a bit of code duplication between the 3 non-visualizable file types components, however, this was done with the idea being that moving forward these will become more different than same when we add validation etc. 
- I toyed with the idea of doing a migration of current files that were uploaded as "Other" that have the h5ad or seurat data file extensions to the new categories however I would like to get more feedback from folks on if they think this is a good idea or not. Also since the feature flag currently hides the H5ad/Seurat sections would need to wait until we made those sections GA.

To test:
- Pull this branch
- Run the DB migration with: `bin/rails db:migrate`
- Start up the Rails server and navigate to localhost
- Start up the delayed jobs in a separate terminal with `./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails jobs:work`

- Go to a study that you own
- Go to the upload wizard
- You should see the new section for all the non-viz files
- You should only see the Miscellaneous file type in that section

- Go to the feature flags section in the portal
- Set the feature flag `h5ad_and_seurat_upload` to true for your user
- Go back to the upload wizard of the study you own 
- You should now see the Seurat Data and Anndata (.h5ad) sections available
- Go to the new sections and try to upload appropriate files

If you go to the HCA site: https://data.humancellatlas.org/explore/projects
you can filter by file type and download data to try uploading

For easy testing for the Anndata (.h5ad) you will want a file with extention `.h5ad `and for the Seurat data a file with extension `.rds`

Screen Shot 2022-07-27 at 5 56 52 PM
Screen Shot 2022-07-27 at 5 57 03 PM


app/javascript/components/upload/H5adFileForm.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
```
eweitz, line 1: Blocker: As part of substantive architecture refactor for h5ad -> AnnData: 

Let's rename this file to e.g. `AnnDataFileForm`.

app/javascript/components/upload/H5adStep.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { useEffect } from 'react'
```
eweitz, line 1: Blocker: As part of substantive architecture refactor for h5ad -> AnnData:

Let's rename this file to e.g. `AnnDataStep.jsx`.

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -35,6 +35,9 @@ import MiscellaneousStep from './MiscellaneousStep'
 import SequenceFileStep from './SequenceFileStep'
 import GeneListStep from './GeneListStep'
 import LoadingSpinner from '~/lib/LoadingSpinner'
+import H5adStep from './H5adStep'
```
eweitz, line 38: Blocker: As part of substantive architecture refactor for h5ad -> AnnData:

```suggestion
import AnnDataStep from './AnnDataStep'
```

eweitz (changes requested):
Nice work on a huge part of this supported file format expansion!

I note a bunch of blockers, and suggest some non-blocking UI polish.  The blockers fall into three categories:

* Substantive refactor for architectural concerns.  The use of "H5ad" throughout the proposed file type naming is inconsistent with our general file type naming.  All other file types are named as a broad category, but never as specific file formats -- e.g. "Image" instead of "PNG".  Similarly, the new file type should be named "AnnData" after the general type of file, not "H5ad" after one (of many) specific file formats of AnnData.  This pattern is a little subtle and not obvious, but seems important for easier future extensibility and UI consistency.  So I asked about it at standup, and Jean agreed.

* Trivial fixes important for maintainability.  These are a varied collection of issues that are likely to mislead maintainers, and need trivial effort to resolve.

* Trivial fixes important for user-facing text.  Mostly typos.  Correcting these avoids distracting users, and otherwise keeps UI quality high.

The "request changes" review status here is only due to those architectural concerns, not the trivial-but-important fixes.  To ease the substantive refactor for the former, all my related suggestions have a common prefix.  I hope that helps!

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -51,7 +54,8 @@ const STEPS = [
 ]
 
 const MAIN_STEPS = STEPS.slice(0, 4)
-const SUPPLEMENTAL_STEPS = STEPS.slice(4)
+const SUPPLEMENTAL_STEPS = STEPS.slice(4, 8)
```
bistline, line 57: Non-blocking: this seems like something that could easily break if we aren't careful.  I'm wondering if explicitly declaring them would be better for `SUPPLEMENTAL_STEPS` and `NON_VISUALIZABLE_STEPS`.

bistline (approved):
Thanks for figuring out the issues - works as advertised now.

eweitz (approved):
Code looks good!

jlchang (approved):
Functional review: uploaded both .h5 and .Rds files successfully.

app/javascript/components/upload/UploadWizard.jsx
ehanna4: That's a good point - I've made a follow up ticket for looking into CSFV for non-viz files and I can add this idea into that to take a look at

ehanna4:
test failures all still coming from `FireCloudClientTest`

PR #1571, opened 2022-07-25
Server-side rendering of expression scatter plots (SCP-4520, SCP-4542)

eweitz:
This is a proof-of-concept for a static image pipeline for gene expression scatter plots.

### Product
The goal is to view expression plots for > 1M cells in < 1s for all genes.  The underlying principle is decoupling "time to view" from "time to interactive".  Previously, users would have to wait [> 15 seconds](https://docs.google.com/document/d/1-mhtoWrg3RHoDjWrHGv2h-KtlnnYGPg5-lxlgRyWaK0/edit#heading=h.7ajhwt5vead9) to see such plots, because processing for interactivity was tightly coupled to processing for display.  If this proof matures as hoped, although users won't be able to interact with them right away, users will _see_ such plots _instantly_.  That visualization speed enables drastically better scientific exploration for large datasets.

### Engineering
In this nascent phase, the pipeline uses [Puppeteer](https://pptr.dev/) to crawl staging SCP, and screenshots a series of plots as static images.  These images will be loaded to a GCP bucket and instantly fetched and displayed by the SCP frontend, while the traditional Plotly.js interactive plot is progressively loaded in the background.

### More details
Further technical plans, motivations, etc. are described in the "[Batch cache visualization](https://docs.google.com/document/d/1-mhtoWrg3RHoDjWrHGv2h-KtlnnYGPg5-lxlgRyWaK0/edit)" technical design document.

### Install
Ensure you have the prerequisites:
* [nvm](https://github.com/nvm-sh/nvm/blob/master/README.md#installing-and-updating)
* Node >= 18 (`nvm install 18`)

Then:
```
cd image-pipeline
yarn install
```

### Run
Connect to Broad VPN, ensure you're in `image-pipeline`, then:

```
node expression-scatter-plots.js --accession="SCP303"
```

If that succeeds, you'll see terminal output like:
```
Accession: SCP303
(node:81747) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Number of genes: 36384
Navigating to Explore tab: https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303#study-visualize
Completed loading Explore tab
Awaiting expression plot for gene: A2M-AS1
Expression plot time: 531 ms
Wrote images/A2M-AS1.webp
Awaiting expression plot for gene: A2ML1
Expression plot time: 2195 ms
Wrote images/A2ML1.webp
[...]

PR #1577, opened 2022-08-01
Parallelize render, pre-fetch expression data in Image Pipeline (SCP-4542)

eweitz:
This speeds up how long it would take after upload to instantly view gene expression scatter plots.

[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/1571), the proof of concept rendered only four genes.  It did so serially and fetched expression data from SCP API.  

### Parallelization
Now, multiple instances of Puppeteer are launched in parallel.  For a machine with 4 CPU cores; this speeds up Image Pipeline by 3x.  One CPU core is left unused, to accommodate non-Image-Pipeline tasks.  This opens the door to reducing runtime by e.g. 30-60x by using many-CPU machines types from GCP.

### Get expression data from files
Requests for expressed data are now also intercepted.  By default, these frequent requests to SCP API from Puppeteer often fail due to opaque errors like `net::ERR_SSL_PROTOCOL_ERROR`.  So Image Pipeline does a direct `fetch` for the relevant SCP API before searching for a gene through the UI (which interestingly avoids those errors).  Then, when Puppeteer runs, it intercepts the frontend's request for expression from SCP API, and seamlessly replaces it with a file system read for pre-fetched expression data.

Importantly, this shows we can re-route SCP API expression data requests to use files.  So we could likely prepare all files in a preceding [scatter data cache job](https://docs.google.com/document/d/1-mhtoWrg3RHoDjWrHGv2h-KtlnnYGPg5-lxlgRyWaK0/edit#heading=h.5b86j42k5u0d), then use them in Image Pipeline to warm the image cache without notably burdening the Rails app server, and dramatically faster.  Image Pipeline would read expression data files from a local disk or bucket.  That's much safer and faster than flooding SCP API with requests that hit an uncached remote database undergirding production.

For a 1.3M cell study, this version of Image Pipeline can produce around 167 images in 26 minutes using 3 CPUs.  Using pre-generated scatter plot data would speed it up.  Handling timeouts and "Target closed" errors would improve robustness.   That potential follow-up work is tracked in SCP-4562 and SCP-4563.

To test:
0. This assumes you're set up per [previous PR](https://github.com/broadinstitute/single_cell_portal_core/pull/1571)
1. Start your local SCP
2. Get an accession for a public study where gene expression plots are visualizable
3. Run `export NODE_TLS_REJECT_UNAUTHORIZED=0; time node expression-scatter-plots.js --accession="<your accession>"`
4. Let it run for 2 minutes
5. Confirm output files exist in `output` directory

This satisfies SCP-4542.


image-pipeline/expression-scatter-plots.js
```diff
@@ -20,87 +21,280 @@ const options = {
...
 
-// Make `images` directory if absent
-access('images', async err => {
-  if (err) {
-    await mkdir('images')
-  }
-})
+// Candidates for CLI argument
+// CPU count on Intel i7 is 1/2 of reported, due to hyperthreading
+const numCPUs = os.cpus().length / 2 - 1
```
bistline, line 26: Non-blocking: shouldn't this also take into account newer m1/m2 chips?  Or is the calculation the same?

bistline (approved):
Looks good.  Very interested to see how much this will eventually speed up using the pre-render data artifacts.

image-pipeline/expression-scatter-plots.js
eweitz: Good point!  Running on Emily's M1 Pro reports: `Number of CPUs to be used on this client: 4`, so I think this is OK for now.  Ultimately, I imagine we'll drop the `/ 2` for the default value and add a CLI argument `--num-cpu` to override that.

ehanna4 (approved):
Unable to run this locally, paired with Eric and successfully ran against Staging to see the changes. We noted that my IP is IPv6 rather than IPv4 and I also am on an M1 so likely at least one of these factors contributed to my inability to get it working locally but don't think this should block this PR and can be investigated further in a future ticket. But this looks good and was cool to see working well on Staging 👍

jlchang (approved):
Tried to do a functional review and re-discovered that my laptop is only dual-core so:
`Number of CPUs to be used on this client: 1`
Can confirm that both image and json files were generated in the expected directory structure but can't comment on the parallelism.

PR #1580, opened 2022-08-03
Add more info to the error logging for failed fetches (SCP-4548)

ehanna4:
Increase the information sent to Sentry when a fetch request fails in the JS code. This pretty much only occurs for connecting to Bard or Google resources (though we've only seen this from Bard fetches). This adds extra context to the Sentry log which will hopefully help us diagnose these issues in the future. 

To test:
- Pull this branch
- Edit line 73 in sentry-logging.js to return false
- Edit line 86 in sentry-logging.jsto `1` to not throttle any requests
- Load home page
- Open DevTools in browser
- Go to the Network tab and find an `event` request and right click on it
- Block the request URL 
Screen Shot 2022-08-03 at 2 54 24 PM
- mouse around the site, until you see in the Network tab something like `?sentry_key=a713dcf8bbce...`
- click on that and check out the request payload where you should see something like below:
Screen Shot 2022-08-03 at 2 54 47 PM

Which is showing the contexts now includes error info

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -409,14 +409,15 @@ export function log(name, props = {}) {
   init = Object.assign(init, body)
 
   if ('SCP' in window || metricsApiMock) {
-    fetch(`${bardDomain}/api/event/`, init).then(response => {
+    const url = `${bardDomain}/api/event/`
+    fetch(url, init).then(response => {
       // log failed attempts to connect with Bard to Sentry
       if (!response.ok) {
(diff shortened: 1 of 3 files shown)


bistline (approved):

jlchang (approved):
Functional review - saw the new contexts info for the error as described.
Looking forward to having more insight into these errors we've been unable to understand fully!

eweitz (approved):
Promising observability enhancement!  Code looks good.

ehanna4:
Test failures are the known Firecloud related failures we've been seeing

PR #1585, opened 2022-08-10
Fix DE backfill convenience method bug and enable custom machine_type and dry_run (SCP-4591)

jlchang:
**BACKGROUND**

DE backfill convenience method `.run_differential_expression_on_all` allows easy backfill of DE results for all eligible annotations in a study. A subtle error in this convenience function allows annotations with the same name that occur in more than one cluster file to launch duplicative/extraneous DE jobs.

**CHANGES**

Now, when launching `.run_differential_expression_on_all`, duplicative jobs are no longer launched. Two additional errors in the convenience method in defining `annotation_identifier` caused function output to lack `annotation_name` and `annotation_scope` information, complicating job tracking and troubleshooting.

This PR also enables override of the default `machine_type` in the convenience functions. This will allow SCP admins to do test runs on the default annotation with non-default machine_type settings before launching DE jobs for all annotations using a non-default machine_type setting.

In addition, it is now possible to perform a "dry run" of either convenience method to see how many jobs will be run/skipped and/or determine the ID of `DifferentialExpressionResult` objects that might need deletion before running DE jobs on all annotations.

**AUTOMATED TEST**

To manually test:
1. Run in terminal: `bin/rails test test/integration/lib/differential_expression_service_test.rb`
2. Confirm `test_should_not_run_differential_expression_job_if_dry_run` passes

**MANUAL TESTING**

1. Pull this branch and start `Delayed::Job` with `bin/rails jobs:work`
3. In a separate terminal, open the Rails console
4. Use a pre-existing `DifferentialExpressionResult` to find an appropriate DE study for testing:

```
result = DifferentialExpressionResult.all.sample
study = result.study
accession = study.accession
user = study.user
Perform a dry run:
DifferentialExpressionService.run_differential_expression_on_all(accession, user: user, machine_type: 'n1-highmem-4', dry_run: true)
```

Confirm that:
1. output of `.run_differential_expression_on_all` shows (`annotation_name`--group--`annotation_scope`) instead of (--group--). 
2. output of `.run_differential_expression_on_all` indicates [`machine_type`] when machine_type is set.
3. output of jobs skipped due to existing results indicate the  `DifferentialExpressionResult` ID that should be deleted to allow reanalysis.
5. when `dry_run: true`, no DE jobs are launched

[optional] To confirm that the `machine_type` parameter is successfully passed to the individual DE jobs, delete the existing result, and run the convenience method to repopulate it. Caution - while only one job will be launched if `.run_differential_expression_on_all` was previously run on the study, multiple jobs will be launched if the study isn't previously fully populated:
```
result.destroy
DifferentialExpressionService.run_differential_expression_on_all(accession, user: user, machine_type: 'n1-highmem-4')
```
In development.log, note the new custom machine type in the `:resources section` of the launched job:
```
:resources:
  ...
  :virtual_machine:
    :boot_disk_size_gb: 300
    :machine_type: n1-highmem-4
    :preemptible: false
    ...
```
    
[optional] To test elimination of duplicative DE jobs, set up a study using production SCP1539 and confirm that `.run_differential_expression_on_all` now launches 23 jobs instead of 25.


app/lib/differential_expression_service.rb
```diff
@@ -184,14 +198,20 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
...
     raise ArgumentError, "cannot find cluster for #{cluster_file.name}" if cluster.nil?
-
     if DifferentialExpressionResult.where(study: study,
                                           cluster_group: cluster,
                                           annotation_name: annotation_name,
                                           annotation_scope: annotation_scope).exists?
+      # Question: Does each study/cluster_group/annotation_name/annotation_scope combo
+      # map to a single DifferentialExpressionResult? If yes, find_by should be fine
+      # If no, please raise a PR comment to use .where instead
+      result = DifferentialExpressionResult.find_by(study: study,
```
bistline, line 208: Yes - this is correct.  That quaternion (had to look that up!) is used as the uniqueness constraint here: https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/differential_expression_result.rb#L29

This means that `find_by` will always return the correct answer as there can only be one.  That being said, I think this whole block starting at the `if` clause can be simplified now to the following:
```suggestion
      result = DifferentialExpressionResult.find_by(study: study,
                                              cluster_group: cluster,
                                              annotation_name: annotation_name,
                                              annotation_scope: annotation_scope)
      if result.present?
        raise ArgumentError,
              "#{annotation_name} already exists for #{study.accession}:#{cluster_file.name}, " \
              "please delete result #{result.id} before retrying"
      end
```

app/lib/differential_expression_service.rb
```diff
@@ -160,7 +171,10 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
     params_object = DifferentialExpressionParameters.new(de_params)
     params_object.machine_type = machine_type if machine_type.present? # override :machine_type if specified
 
-    if params_object.valid?
+    if dry_run
```
bistline, line 174: Really like this enhancement - I can see why it would be useful.  The existing validation for uniqueness would have caught this, but the error message is certainly unhelpful, and would not point you at the existing record as easily as this does.

bistline (approved):
Great work!  Some non-blocking refinements.

eweitz (approved):
Code looks good!

app/lib/differential_expression_service.rb
```diff
@@ -184,14 +198,20 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
   def self.validate_annotation(cluster_file, study, annotation_name, annotation_scope)
     cluster = study.cluster_groups.by_name(cluster_file.name)
     raise ArgumentError, "cannot find cluster for #{cluster_file.name}" if cluster.nil?
-
```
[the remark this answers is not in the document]
jlchang: `946b49e`

app/lib/differential_expression_service.rb
```diff
@@ -160,7 +171,10 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
     params_object = DifferentialExpressionParameters.new(de_params)
     params_object.machine_type = machine_type if machine_type.present? # override :machine_type if specified
 
-    if params_object.valid?
+    if dry_run
+      self.validate_annotation(cluster_file, study, annotation_name, annotation_scope)
```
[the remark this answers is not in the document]
jlchang: `946b49e`

app/lib/differential_expression_service.rb
jlchang: `946b49e` Thanks for the simplification - much cleaner!

eweitz (approved):

jlchang:
Build failure was a false positive due to [Azul issues](https://broadinstitute.slack.com/archives/CBEHTH601/p1660164430062849).

bistline (approved):
Looks good - there's a merge conflict now with `DelayedJobAccessor` because I found (yet another) class that needed to be added to the `SAFE_CLASS_LOADERS` array - `SearchFacet`.  Somewhat regretting changing to `YAML.safe_load` at this point...

PR #1591, opened 2022-08-15
Removing `skip_push` parameter for `IngestJob.push_remote_and_launch_ingest` (SCP-4576)

bistline:
#### BACKGROUND & CHANGES
The Ruby 3.1 update exposed a regression with passing parameters to methods when invoking them via `delay`.  This is only used in `IngestJob.push_remote_and_launch_ingest` where we pass the `skip_push` parameter, which will skip attempting to push a file to a bucket if it is already there.  However, inspection of the code revealed that this would never actually be exercised as we check first to see if the file has in fact been pushed to the bucket.  Therefore, this update removes the parameter from the method definition.

~~It is unclear why `delay` is not passing parameters properly - we are on the latest release of `DelayedJob`.  It is likely related to the Ruby 3.1 update and the hash/literal shorthand updates.  We may need to open a support ticket with `DelayedJob` if we ever need to use this feature.~~ UPDATE: It appears this is a [known issue](https://github.com/collectiveidea/delayed_job/issues/1134) in `DelayedJob` and methods with keyword arguments.  Thankfully, we do not require them at this time, but if we do ever need to support kwargs via `delay`, we will need this bug fixed.  There appear to be several open PRs aimed at adding better support for Ruby 3.1, so hopefully this will be fixed soon.

This also updates `sentry-raven` to a more recent release to silence the `multipart-post` deprecation warnings in the logs.

#### MANUAL TESTING
1. Pull branch, start `DelayedJob` with `./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails jobs:work`
2. In a separate terminal window, open a Rails console session
3. Re-run a DE job with the following:
```
result = DifferentialExpressionResult.last
study = result.study
cluster_file = result.cluster_group.study_file
annotation_name = result.annotation_name
annotation_scope = result.annotation_scope
result.destroy
DifferentialExpressionService.run_differential_expression_job(cluster_file, study, study.user, annotation_name:, annotation_scope:)
```
4. In `development.log`, validate the job launched by looking for the PAPI parameters:
```
Remote found for UMAP_crush.txt:6272a06f3ea6aec6b795f2ac, launching Ingest job
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 62729e0894ec8f9704b9981e
  - "--study-file-id"
  - 6272a06f3ea6aec6b795f2ac
  - "--user-metrics-uuid"
  - f775823b-94f7-46af-b1c7-789582efaa18
  - differential_expression
...
```


eweitz (approved):
Code looks good!

bistline:
Seems this is a known issue in `DelayedJob` that they are working on fixing: https://github.com/collectiveidea/delayed_job/pull/1130

ehanna4 (approved):
Works as advertised!

PR #1593, opened 2022-08-17
Prime first search with gene leads ideogram (SCP-4596)

eweitz:
This drafts a way to prime the user's initial gene search in a single-cell study.

### Impact
Previously, our default search UX lacked "information scent"; there were no are relevant hints of what genes might be interesting to explore.  This new UI component solves that problem by showing "gene leads" to prime search, using an engaging ideogram genomic visualizations and data mined from linked publications, differential expression, and elsewhere.

The basic UI got enthusiastic product endorsement [at demo](https://drive.google.com/file/d/1S5TkGbL7j223imxqXOxSm-4c8LP28p51/view?t=0m21s) yesterday.  Gene lead data is generated by (an also draft) Python module in Ingest Pipeline; more details in forthcoming PR there.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/185139256-aa9f81ba-4c4f-4e5e-8ac4-c46d7cdc64f9.mov

### To test
* Run `rake db:migrate`
* Go to the "Human milk - differential expression" study in your local SCP
* In another browser tab, enable the "gene_leads_ideogram" feature flag for that study
* Refresh study in browser
* Note gene leads ideogram

This was designed and implemented in Developer's Choice Days.  It's targeted for a feature branch because it uses an unpublished version of Ideogram.js, which enables this.

This relates to SCP-4596.


jlchang (approved):
Functional Review - works as described.

Suggestions:
1. There are two types of "Gene Leads" 
• genes from publication 
• genes from DE
It would be useful to differentiate the two (stretch: let the words "linked publication" be a link to the publication). Perhaps gene leads that are only from the publication should have black arrows? It's unclear why the arrow in the example below is blue... 

Screen Shot 2022-08-18 at 11 26 05 AM

2. Clarify what "global interest" means (maybe a link to the source), otherwise that info seems a bit arbitrary (it could have come from https://www.genecards.org/List/HotGenes or by number of publications the gene is cited in http://emptypipes.org/2014/12/08/gene-popularity/)

I see that the arrows for DE gene leads are color coded for the cell type they're found in - which is a nice touch. 

3. DE gene leads where the gene is in more than one cell type should probably be colored differently so they're not suggestive of DE specific to one annotation label. 
Screen Shot 2022-08-18 at 11 25 22 AM

4. Why "**Very** differentially expressed"? "Very" implies a quantitative judgement that it surpassed some threshold beyond just "differentially expressed".

eweitz:
Insightful feedback, Jean -- thank you.

> There are two types of "Gene Leads"
• genes from publication
• genes from DE
> 
> It would be useful to differentiate the two (stretch: let the words "linked publication" be a link to the publication). Perhaps gene leads that are only from the publication should have black arrows? It's unclear why the arrow in the example below is blue.

To clarify, all gene leads in this prototype are mentioned in the publication.  Among those, some have strong DE; those lacking it are colored "SCP chrome blue".  It's the same hue as the header, footer, etc.  It's softer than black, but maybe lacks sufficient contrast with other colors in the palette.  I've opened SCP-4601 to refine color handling.

Linking the publication makes sense!

> Clarify what "global interest" means (maybe a link to the source), otherwise that info seems a bit arbitrary

Good point, this indeed seems worth at least documenting somewhere users could reasonably find.   

Fun fact: the "global interest" rank is from [Gene Hints](https://github.com/broadinstitute/gene-hints), which in turn was inspired by Peter Kerpedjiev's (@pkerpedjiev's) work in http://emptypipes.org/2014/12/08/gene-popularity/ that you note.  This rank uses Wikipedia page views, but global interest ranks for non-human genes would likely use citation count ranks that are computed via [code](https://github.com/broadinstitute/gene-hints/blob/main/gene_hints/citations/citations.py) inspired by [Peter's original](https://github.com/pkerpedjiev/gene-citation-counts).

> DE gene leads where the gene is in more than one cell type should probably be colored differently so they're not suggestive of DE specific to one annotation label.

Whether via color or some other visual sign, I agree that indicating multi-label DE would be valuable.  Tracked: SCP-4602.

> Why "Very differentially expressed"? "Very" implies a quantitative judgement that it surpassed some threshold beyond just "differentially expressed".

Yes, correct.  The current crude threshold for "very" is being in the top 2% or so in our DE file for the given labeled group.  The threshold in definitely worth refining to account for p-value as you'd mentioned, and likely other factors.  I've opened SCP-4603 to track that and its wording.

ehanna4 (approved):
code looks good, and test works as advertised

PR #1596, opened 2022-08-17
Add titles to unique pages in SCP (SCP-4599)

ehanna4:
Page titles are an important piece of helping users quickly orient themselves and scan information but previously all pages on SCP shared the title “Single Cell Portal” this update seeks to add more context to unique pages in SCP.

This adds unique titles to a variety of SCP pages, with a default for any/all to be "Single Cell Portal"

To test:
- pull this branch
- click around and check out the title of the page (visible in the tab and also visible when viewing the back history)
- confirm that the title updates for unique pages like studies and pages available from the user drop down section

Screen Shot 2022-08-17 at 4 13 00 PM


app/views/site/_study_title_bar.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for(:html_title) { "#{@study.accession} - Single Cell Portal" } %>
```
eweitz, line 1: Conditionally adding study names to these titles would improve SEO and human glanceability.

Only showing accession for _private_ studies is reasonable.  However, for public studies, no such privacy concerns exist.  In those cases it'd enhance search engine optimization to also show the study's name.  Names are also more useful than accessions for humans, e.g. tracing one's past in browser history -- or even glancing at currently-open tab titles; the first few characters of a name are usually better hooks for recall than an accession.

So something like "if private, use accession; if public, use name and accession" would enrich this work's value.

eweitz (approved):
Code looks good!  This is valuable UX polish.

Adding names to public study titles would further improve SEO and human glanceability.

app/views/branding_groups/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for(:html_title) { 'Collections - Single Cell Portal' } %>
```
jlchang, line 1: I can see html_title for Upload but not for Collections... Also not seeing identifying titles for the individual collection pages..
Screen Shot 2022-08-18 at 10 29 49 AM
.

jlchang (changes requested):
I'm really excited to see this functionality available in production!

Functional review - works as described. Requesting changes as I'm not seeing the more granular titles for Collection pages.

app/views/branding_groups/index.html.erb
ehanna4: Oh nice catch - clicking to the Features and the Collections pages from the dropdown profile menu vs clicking the buttons from the home page bring you to different pages (which I hadn't noticed) so I've updated those. Also added titles for the collections.

Screen Shot 2022-08-18 at 11 16 01 AM

ehanna4: Screen Shot 2022-08-18 at 11 25 22 AM

I wonder if the naming convention for Collections should be "<collection title> - Collections - Single Cell Portal" or "<collection title> - Single Cell Portal" though?

jlchang: How about "Collection title - Single Cell Portal Collections"

jlchang (approved):
Functional review - confirmed that I can now see granular title names for collections pages.

Looks great!

app/views/branding_groups/index.html.erb
ehanna4: Would be a nice time to have a UX partner to check in on this haha, from what I could find online the general convention is to keep the web app name ("Single Cell Portal") and the end as the overarching title, and then chain subsequent section titles in order of most granular/specific title being the leftmost. So I think following that we would want "Collection title - Collections - Single Cell Portal"

Looking at this in the history i do think this looks okay as it basically adds an extra descriptor each level you drill down.

Screen Shot 2022-08-18 at 1 37 44 PM

PR #1597, opened 2022-08-18
Instant gene expression scatter plots: prototype (SCP-4583)

eweitz:
This adds progressive loading.  Gene expression scatter plots now display instantly, even for extremely many cells.

Previously, gene expression scatter plots had to be fully interactive in order to be visually displayed.  [Time-to-view can be decoupled from time-to-interactive](https://github.com/broadinstitute/single_cell_portal_core/pull/1571).  In theory, that massively improves render times.  This work implements that in practice.

Here's how it looks:

https://user-images.githubusercontent.com/1334561/185487425-7a7d450b-394c-4987-a711-7fa7cb9f2c01.mov

The plot displayed 0.5 seconds after searching is a static image.  It's visible, but not interactive.  The nearly identical plot with tooltips shown ~5 seconds after searching is the interactive version. 

Graphical scaling and transforming will be refined in SCP-4600.  That'll fix the slight jank / point-mass shift upon transition.

To test:
* Run `rake db:migrate`
* Find the accession for your local "Human milk -- differential expression" study
  * If you don't have that study, you can add it via Terminal like so: `echo "SyntheticStudyPopulator.populate('human_milk_de_49k_cell')" | bundle exec rails console -e development`
* Enable "progressive_loading" feature flag for the study
* In the study, search "A1BG-AS1"
* Quickly hover over points in the gene expression scatter plot, note no tooltips
* Confirm slight change in plot, and appearance of tooltips on hover

This satisfies SCP-4583.


jlchang (changes requested):
Functional review - progressive loading works as described but possibly breaks differential expression.
DE for the breast milk study works on development but not on this branch. My local instance reports a 404 trying to get the DE file.
Screen Shot 2022-08-19 at 10 18 02 AM
Checking the bucket, the file exists:
Screen Shot 2022-08-19 at 10 17 47 AM
It's possible that this branch is merely missing some commits for DE that are in development (but I see that this branch is looking for a file where + -> pos which is the last DE update I can recall)
Please confirm that the DE issue is unrelated to this PR and I will be happy to approve.

eweitz:
That was a true-positive bug, caused by an encoding error.  Fixed!

image-pipeline/expression-scatter-plots.js
```diff
@@ -23,8 +25,8 @@ const { values } = parseArgs({ args, options })
 
 // Candidates for CLI argument
 // CPU count on Intel i7 is 1/2 of reported, due to hyperthreading
-const numCPUs = os.cpus().length / 2 - 1
-// const numCPUs = 2
+// const numCPUs = os.cpus().length / 2 - 1
```
ehanna4, line 28: Should there be a comment here about why the numCPUs is just 1 rather than using this commented out logic?

image-pipeline/expression-scatter-plots.js
```diff
@@ -83,14 +85,67 @@ async function makeExpressionScatterPlotImage(gene, page, preamble) {
...
+      }
+    })
+    .toFile(imagePath)
+
+  // const metadata = await sharp(imagePath).metadata()
+  // console.log('metadata:')
+  // console.log(metadata)
 
   print(`Wrote ${imagePath}`, preamble)
+  if (imagePath === 'output/SCP138/images/A1BG-AS1.webp') {
```
ehanna4, line 146: Possibly a comment here to explain this will only work for a study with the gene A1BG-AS1 or something similar just to explain why this is hardcoded here

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -232,7 +373,7 @@ function RawScatterPlot({
     })
   }, [cluster, annotation.name, subsample, consensus, genes.join(','), isAnnotatedScatter])
 
-  // Handles custom scatter legend updates
+  // // Handles custom scatter legend updates
```
ehanna4, line 376: double commented out might want to remove one set of `//`

ehanna4 (approved):
Code review only - looks good! I left a few comments for some potential comments in the code but nothing blocking

image-pipeline/expression-scatter-plots.js
eweitz: Done in 505c492.

app/javascript/components/visualization/ScatterPlot.jsx
eweitz: Done in 505c492.

image-pipeline/expression-scatter-plots.js
eweitz: Done in 505c492.

bistline (approved):

PR #1599, opened 2022-08-23
Fixing broken dependency from react-scripts upgrade (SCP-4617)

bistline:
#### BACKGROUND & CHANGES
Due to the upgrade for `react-scripts` in #1592, a transitive dependency for `autoprefixer` was omitted due to upgrades in other libraries.  While this didn't affect deployed environments, it shows up as spurious errors in CI, and prevented some local development in Docker due to `vite` not being able to load the missing dependency.  This update upgrades some related JS packages to restore the missing dependency.  

In addition, it reorders steps in the Dockerfile to allow faster iteration when updating startup scripts in Docker.  Additionally, it removes some broken/unneeded steps from `bin/run_tests.sh` that were obsoleted due to the Docker image upgrade from #1552.

#### MANUAL TESTING
1. Pull this branch
2. From the project root, rebuild the Docker image with:
```
docker build -t gcr.io/broad-singlecellportal-staging/single-cell-portal:development --no-cache .
```
3. Once the build is complete, start the container with the new image:
```
docker run --rm -it gcr.io/broad-singlecellportal-staging/single-cell-portal:development bash
```
4. Verify that the `autoprefixer` package was installed by checking the `node_modules` directory:
```
ls -l node_modules | grep autoprefixer
```
You should see something like `drwxr-xr-x   6 root root  4096 Aug 23 15:28 autoprefixer`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,15 +14,17 @@ COPY Gemfile /home/app/webapp/Gemfile
 COPY Gemfile.lock /home/app/webapp/Gemfile.lock
 WORKDIR /home/app/webapp
 RUN bundle install
+
+# install JS dependencies
 COPY package.json /home/app/webapp/package.json
 COPY yarn.lock /home/app/webapp/yarn.lock
 RUN chmod a+w /home/app/webapp/yarn.lock
+RUN yarn install --force
+
+# copy startup scripts
 COPY set_user_permissions.bash /etc/my_init.d/01_set_user_permissions.bash
 COPY generate_dh_parameters.bash /etc/my_init.d/02_generate_dh_parameters.bash
 COPY rails_startup.bash /etc/my_init.d/03_rails_startup.bash
-
-# install JS dependencies
-RUN sudo -E -u app -H yarn install --force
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -14,16 +14,18 @@ COPY Gemfile /home/app/webapp/Gemfile
 COPY Gemfile.lock /home/app/webapp/Gemfile.lock
 WORKDIR /home/app/webapp
 RUN bundle install
+
+# install JS dependencies
```
jlchang, line 18: Thanks for the comments - helpful for current and future me :)

jlchang (approved):
Functional review - confirmed:
`drwxr-xr-x   6 root root  4096 Aug 24 09:22 autoprefixer`

ehanna4 (approved):
This is the result in my node_modules: `drwxr-xr-x    9 ehanna  staff    288 Aug 25 10:11 autoprefixer/`

and saw that previously changes to CSS files didn't update (due to vite not actually building/running) and with this fix they work again! 👏🤩

PR #1603, opened 2022-08-29
Temporarily revert CSP for instant plot prototype (SCP-4583)

eweitz:
The prototype in #1597 updated content security policy (CSP) without AppSec review, which was unintentional.  This reverts that until AppSec approves the change.  

That presumably will be OK, but worth a separate task given the lag, timelines, and scope of the work in that prior PR.  The main goal here is to not push that insufficiently reviewed security policy change to production.

To test:
* Go to a visualizable study
* Search a gene
* Click legend label
* Confirm above yields no obvious error

This relates to SCP-4583.  The CSP change will be handled separately in SCP-4626.


bistline (approved):
While getting approval is always the right thing to do, I'd be surprised if there was an issue.  We already have `storage.googleapis.com` listed as a `connect_src`, so adding it to `img_src` shouldn't raise any alarms.

bistline:
Build failure has been addressed in #1598

ehanna4 (approved):

PR #1611, opened 2022-09-07
Ignore common/stop words when constructing Azul queries (SCP-4531)

bistline:
#### BACKGROUND
A common error found during triage is one received from Azul when attempting to convert a keyword search into something that Azul will accept.  For certain common terms, the amount of matches found are so great that Azul throws an `HTTP Payload Too Large` error since the query string payload can be over 8K.

#### CHANGES
This leverages similar work done for [rejecting stop words](https://github.com/broadinstitute/single_cell_portal_core/pull/1357) in SCP-based searches.  In addition to ignoring those words when finding matches for possible entries in Azul, a short list of exceedingly common terms found in Azul filter values are also filtered out before finding any possible hits.  These include terms like `cell`, `rna`, `sequencing`, etc., as well as any terms less than 3 characters long.  While these terms may be relevant in some cases, the amount of matches that they return are such that no query can ever be performed against Azul if they are included.  It is worth noting that if they are included in a quoted string (e.g. "B cell"), then that will be honored as a distinct term.

It is worth noting that this does not address the related issue of large compound queries not returning results from Azul.  This is due to the fact that Azul does not natively support keyword searching, and providing a large list of terms - or commonly found ones - will always result in a query with no intersection as Azul only support `AND` logic across facets.  The above work is solely to mitigate the HTTP 413 error by constraining possible matches.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Enter the following query in the text search box: `Tumor and immune reprogramming during immunotherapy in advanced renal cell carcinoma` (no quotes)
3. Confirm the search request completes and some SCP-based result are found, such as the `Human HIV and Tuberculosis, blood study` synthetic study
4. In `development.log`, confirm that a search was performed against Azul with something similar to the below object (will depend on facets in your local instance and how recently they've been updated against Azul) and that no error was thrown:
```
Executing Azul project query with: {"organ"=>{"is"=>["adrenal gland", "renal system"]}, "sampleDisease"=>{"is"=>["breast adenocarcinoma", "breast carcinoma", "breast ductal adenocarcinoma", "breast lobular carcinoma", "carcinoma of floor of mouth", "carcinoma of supraglottis", "diffuse gastric adenocarcinoma", "endometrioid adenocarcinoma", "hepatocellular carcinoma", "invasive breast carcinoma", "invasive ductal breast carcinoma", "laryngeal carcinoma", "luminal B breast carcinoma", "non-small cell lung carcinoma", "oral cavity carcinoma", "pancreatic adenosquamous carcinoma", "squamous cell carcinoma of buccal mucosa", "tonsil carcinoma"]}, "libraryConstructionApproach"=>{"is"=>["10x immune profiling"]}, "selectedCellType"=>{"is"=>["cd4neg cd45pos immune cell"]}, "organPart"=>{"is"=>["renal medulla", "renal pelvis"]}}
...
Found 0 results in Azul
```

### DIFF
--- a/app/models/hca_azul_client.rb
+++ b/app/models/hca_azul_client.rb
@@ -29,7 +29,7 @@ class HcaAzulClient
   }.freeze
 
   # list of words that are common among Azul filter values that result in too many matches
-  # this can cause HTTP 413 Payload Too Large or HTTP 413 URI Too Long errors when executing
+  # this can cause HTTP 413 Payload Too Large or HTTP 414 URI Too Long errors when executing
   # search requests in Azul
   COMMON_AZUL_TERMS = %w[cell cells human single-cell reveals sequencing rna-seq analysis atlas rna disease].freeze
 
@@ -308,7 +308,8 @@ class HcaAzulClient
     query
   end
 
-  # find matching facets based on keyword list, to be passed to :format_query_from_facets
+  # find matching facets based on keyword list along with filters that contain items from term_list
+  # to be passed to :format_query_from_facets
   # since Azul does not support keyword searching, we must find a match for a corresponding facet and treat this as
   # the search request
   # will ignore very common/short words, including known stop words (see StudySearchService::STOP_WORDS)
--- a/app/models/hca_azul_client.rb
+++ b/app/models/hca_azul_client.rb
@@ -28,6 +28,14 @@ class HcaAzulClient
     'x-domain-id' => "#{ENV['HOSTNAME']}"
   }.freeze
 
+  # list of words that are common among Azul filter values that result in too many matches
+  # this can cause HTTP 413 Payload Too Large or HTTP 413 URI Too Long errors when executing
+  # search requests in Azul
+  COMMON_AZUL_TERMS = %w[cell cells human single-cell reveals sequencing rna-seq analysis atlas rna disease].freeze
+
+  # union of above terms and known stop words
+  IGNORED_WORDS = (StudySearchService::STOP_WORDS + COMMON_AZUL_TERMS).uniq.freeze
+
   ##
   # Constructors & token management methods
   ##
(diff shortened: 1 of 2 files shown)


app/models/hca_azul_client.rb
```diff
@@ -28,6 +28,14 @@ class HcaAzulClient
     'x-domain-id' => "#{ENV['HOSTNAME']}"
   }.freeze
 
+  # list of words that are common among Azul filter values that result in too many matches
+  # this can cause HTTP 413 Payload Too Large or HTTP 414 URI Too Long errors when executing
+  # search requests in Azul
+  COMMON_AZUL_TERMS = %w[cell cells human single-cell reveals sequencing rna-seq analysis atlas rna disease].freeze
```
eweitz, line 34: Not being able to query Azul for "human" seems like the only notable downside of this approach.

```suggestion
  COMMON_AZUL_TERMS = %w[cell cells human single-cell reveals sequencing rna-seq analysis atlas rna disease].freeze
```

eweitz (approved):
Code looks good!

app/models/hca_azul_client.rb
bistline: Agreed, now that I think about it.  Interestingly, I re-ran the analysis, and the results of common words I got was different.  It seems that these values are not static, which seems odd.  I'm going to hold off merging this until I get a better handle on what is going on here.

ehanna4 (approved):
Works as advertised.

Not a blocker but I wonder if we should indicate in some way to the users that they won't be getting HCA results for these search terms though? Just seems like it could be confusing to see a particular HCA study in some searches but not others

app/models/hca_azul_client.rb
bistline: Turns out we have an unrelated bug in creating these queries where we're clobbering earlier matches, and then there .  Which means this problem is actually _much_ worse than we thought.  I'm going to reassess this and see if there's more that needs to be done here.

bistline:
> Works as advertised.
> 
> Not a blocker but I wonder if we should indicate in some way to the users that they won't be getting HCA results for these search terms though? Just seems like it could be confusing to see a particular HCA study in some searches but not others

The more I'm looking into this, the more this feels like not the right way to go (except for stop words and the term "cell" which is an obvious special case).  Rather than ignoring words that have possible scientific context, I think we need to address the underlying intersection/query size issue more directly.  For the term "cell", using facets like `B cell` work fine, and quoted strings work correctly as well.

PR #1612, opened 2022-09-07
Update cookie comment to reflect that Appcues relies on user_id cookie as well (SCP-4620)

ehanna4:
We are using the `user_id` cookie to send to Appcues as the identifier for unique users. This cookie item is set to expire after 1 year. Therefore we don't need another cookie for our un-auth'd users to avoid Appcues overload, since if they are using the same device they will have the same cookie for 1 year (provided they do not refresh their cookies, which even if we did instrument another Appcues specific cookie that would be blown away too). This is just to update the comment above setting the `user_id` cookie to indicate its used by Appcues as well, so if we ever change that expiration we are aware it will affect Appcues as well.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -43,7 +43,7 @@
       window.SCP.featuredSpace = 'covid19'
       <% end %>
 
-      // Set the unique cookie for tracking user IDs in GA.  Also used in Mixpanel.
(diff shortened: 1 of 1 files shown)


bistline (approved):
My favorite kind of PR.

eweitz (approved):
Nice!  Could you tweak the PR title to make it easier to glean context?

PR #1613, opened 2022-09-08
Fix stuck 'return to cluster view' tooltip (SCP-4633)

ehanna4:
Switch the tooltip for the "Return to cluster view" button to use the Overlay/Tooltip style. 

As far as I can see we use two main tooltip patterns in the code, one is using Overlay/Tooltip and one is adding `data-toggle="tooltip"` to buttons. 

Previously the button was using the `toggle="tooltip"` version and it was not handling the dismissal of the tooltip once the button disappeared upon clicking the button. This likely could have been sorted out using the different `trigger` options offered by this [bootstrap tooltip](https://getbootstrap.com/docs/4.0/components/tooltips/) however, I have generally found more success working with the other tooltip option we use from [react-bootstrap](https://react-bootstrap.github.io/components/overlays/) so I switched it to use this style.

The fix is that the tooltip is dismissed when the button is removed from the page upon clicking the button.

To test:
- Pull the branch
- Fire up a local instance
- Go to any study that has visualizable plots
- Search for any gene
- Then click the left arrow to clear out that search and observe that the tooltip goes away when the button does.

Before - see how the tooltip is stuck to the page even if you switch to the overview tab
Screen Shot 2022-09-08 at 12 59 05 PM

After - the screenshot doesn't show but the mouse is hovering over the arrow button
Screen Shot 2022-09-08 at 12 59 19 PM (2)

After - the button has been dismissed and the tooltip went away as well
Screen Shot 2022-09-08 at 12 59 36 PM

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -32,6 +32,8 @@ import LoadingSpinner from '~/lib/LoadingSpinner'
 import { log } from '~/lib/metrics-api'
 import { getFeatureFlagsWithDefaults } from '~/providers/UserProvider'
 import DifferentialExpressionPanel, { DifferentialExpressionPanelHeader } from './DifferentialExpressionPanel'
+import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger'
+import Tooltip from 'react-bootstrap/lib/Tooltip'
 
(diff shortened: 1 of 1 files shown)


bistline (approved):

eweitz (commented):
Nice fix for a definite UX papercut!

eweitz (approved):

PR #1605, opened 2022-08-29
Improve Image Pipeline scalability: use prefetched expression (SCP-4563)

eweitz:
This enables Image Pipeline to retrieve expression data from local files, so it processes more robustly and faster.

### Impact
[Previously](#1577), expression data for gene expression scatter plot images was fetched via SCP REST API in Image Pipeline.  This was convenient, but scaled poorly.  It entailed immense load on the SCP application and database servers -- enough so that we'd need to do something like transiently shard MongoDB, spin up infrastructure copies via Terraform, and generally do a bunch of complicated stuff in order to not take down production.  In terms of performance, that complicated approach would at best still be rather slow for studies with extremely many cells.

Now, expression data can be read from local, prefetched files.  With the forthcoming scatter data pipeline, we'll be able to generate expression data arrays for all genes and clusters in a given study _without_ hitting SCP API and all its attendant load and speed issues.  Combined with that work, this enhancement will essentially eliminate "backend" and "transfer" time.  

That [saves > 10 seconds per gene](https://mixpanel.com/project/2120588/view/19059/app/insights/#report/31338756/expression-scatter-plot-time-0-5-m-1-m-points), which with 25K genes <ins>_saves almost 3 days per Image Pipeline run_</ins> on large studies.

### Architecture
Below is a diagram showing where this fits in the upcoming architecture.  This derisks some of the "Image Pipeline" part at right -- specifically, that it can read standalone scatter data files (much of the green arrow).  Forthcoming work in Ingest Pipeline will prove we can write such scatter data.  

[](https://docs.google.com/document/d/1-mhtoWrg3RHoDjWrHGv2h-KtlnnYGPg5-lxlgRyWaK0/edit#heading=h.cg3befaxtgyj)

Future work will mature how Image Pipeline fetches scatter data from the bucket, handles file names, etc.

To test:
* Get the accession for your local "Human milk -- differential expression" study
* In Terminal, `cd image-pipeline`
* Substituting your accession, run the following command for 2 minutes, then terminate it:
`rm -rf output/; export NODE_TLS_REJECT_UNAUTHORIZED=0; time node expression-scatter-plots.js --accession="<your accession>"; tput bel; # localhost, human milk DE study`
  * If you immediately see `TypeError: fetch failed`, edit `expression-scatter-plots.js` to use `https://127.0.0.1:3000` instead of `https://localhost:3000`, then re-run above
* Confirm the following shows JSON files: `ls -latr output/<your_accession>/json/`
[...]

PR #1616, opened 2022-09-12
Image Pipeline: read gz files, handle 1M cells, add CLI arguments (SCP-4641)

eweitz:
This enables Image Pipeline to read gzipped files produced by the draft expression scatter data pipeline.  

Files will ultimately be decompressed in transit, we presume.  Decompressing them on the client is mostly a way to pick up draft scatter data pipeline output, until that other pipeline writes out individual gzipped files (without tarballing) and sets appropriate GCS metadata on them to enable [decompressive transcoding](https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding).

This also works around an issue in #1605 with using scatter files for the 1.3M cell study.  Using scatter files failed for that study, but not the 49k DE pilot study.  This seems caused not by the scatter data pipeline, but rather with using trimmed URLs to conditionally fetch only expression arrays (and not also annotation, cell, and coordinate) data in Image Pipeline.  The workaround still uses local data, but more than necessary, and can be optimized as a later, separate task.

To test:
* Install dependencies, if you hadn't in previous PR: `cd image-pipeline; yarn install`
* Find the accession for your local "Human milk -- differential expression" study
  * If you don't have that DE pilot study, you can add it via Terminal like so: `echo "SyntheticStudyPopulator.populate('human_milk_de_49k_cell')" | bundle exec rails console -e development`
* Confirm "progressive_loading" feature flag is _disabled_ for the study
* Download [All_Cells_UMAP_data.tar.gz](https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-public/test/All_Cells_UMAP_data.tar.gz), untar it, change untarred directory name from "All_Cells_UMAP-3d16b946-86a7-4d50-bce8-b97e2ab02734" to "All_Cells_UMAP"
* Ensure Rails is running (`rails s`)
* Changing "SCP138" to your DE pilot study accession and the `--json-dir` value to the path of your "All_Cells_UMAP" directory above, run:
`export NODE_TLS_REJECT_UNAUTHORIZED=0; time node expression-scatter-plots.js --accession="SCP138" --environment="development" --cores="1" --cluster="All_Cells_UMAP" --json-dir="/Users/eweitz/Downloads/All_Cells_UMAP/" --debug; tput bel # localhost, human milk DE study`
  * If you immediately see `TypeError: fetch failed`, edit `expression-scatter-plots.js` to use `https://127.0.0.1:3000` instead of `https://localhost:3000`, then re-run above
* Confirm `ls -latr output/<your DE pilot study accession>/images/`
  * If above times out, then simply try running above `expression-scatter-plots.js` command again

This satisfies SCP-4641.

### DIFF
--- a/image-pipeline/expression-scatter-plots.js
+++ b/image-pipeline/expression-scatter-plots.js
@@ -18,14 +18,18 @@ import { exit } from 'node:process'
 const args = process.argv.slice(2)
 
 const options = {
-  accession: { type: 'string' }
+  accession: { type: 'string' },
+  cores: { type: 'string' },
+  debug: { type: 'boolean' }
 }
 const { values } = parseArgs({ args, options })
 
+const timeoutMinutes = 0.5
+
 // Candidates for CLI argument
 // CPU count on Intel i7 is 1/2 of reported, due to hyperthreading
-// const numCPUs = os.cpus().length / 2 - 1
-const numCPUs = 1 // For easier local development
+const numCPUs = values.cores ? parseInt(values.cores) : os.cpus().length / 2 - 1
+// const numCPUs = 1 // For easier local development
 console.log(`Number of CPUs to be used on this client: ${numCPUs}`)
 
 // TODO (SCP-4564): Document how to adjust network rules to use staging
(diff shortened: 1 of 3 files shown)


image-pipeline/expression-scatter-plots.js
```diff
@@ -7,30 +7,42 @@
...
 
 // TODO (SCP-4564): Document how to adjust network rules to use staging
-// const origin = 'https://singlecell-staging.broadinstitute.org'
-const origin = 'https://localhost:3000'
+const originsByEnvironment = {
+  'development': 'https://localhost:3000',
+  'staging': 'https://singlecell-staging.broadinstitute.org',
+  'production': 'https://singlecell.broadinstitute.org'
+}
+const environment = values.environment ?? 'development'
```
ehanna4, line 44: `??` only looks at explicitly `null` or `undefined` so an empty string would get through in this case. Is that a reasonable value for `environment` to be set to here, or would it better to check for any falsy and do `const environment = values.environment || 'development'` ?

image-pipeline/expression-scatter-plots.js
```diff
@@ -7,30 +7,42 @@
...
+import { gunzipSync, strFromU8 } from 'fflate'
 import puppeteer from 'puppeteer'
-import { exit } from 'node:process'
+import sharp from 'sharp'
 
 const args = process.argv.slice(2)
 
 const options = {
-  accession: { type: 'string' }
+  'accession': { type: 'string' }, // SCP accession
```
ehanna4, line 22: Appreciate the comments here!

ehanna4 (approved):
Functional review: works as advertised
code review: left one non-blocking comment/suggestion

jlchang (approved):
Functional review - watched expression-scatter-plots.js render gene expression plots for several genes in-browser but only against staging. Something about using localhost is blocked when I try in my development environment (Note that I can't get vite-dev to communicate with my local instance so that may be part of the problem)

image-pipeline/expression-scatter-plots.js
eweitz: Nice catch!  Done in 3472483.

PR #1617, opened 2022-09-13
Convert all request params into hex digest for API caching (SCP-4654)

bistline:
#### BACKGROUND
Visualization API responses are cached for better performance, and each response encodes its parameters into a unique key to allow for easy clearing of caches.  Request parameters are included as plain text in the cache path to help generate this unique key.  This can give rise to errors when clearing caches where certain parameter values are URL-encoded (e.g. `%2f` for `/`), but then those strings would be split inside the percent expression due to filename length limits in linux.  This can cause the `CacheRemovalJob` to throw errors such as `invalid %-encoding` or `invalid byte sequence in UTF-8`.

#### CHANGES
This update now collects and sorts all request parameters and then computes a hex digest of those values to append onto the end of the request path.  This retains the uniqueness of cache entries while remaining deterministic, and results in cache paths that are fairly consistent in length.  For instance, a cache path before this change would look something like this:
```
_single_cell_api_v1_studies_SCP63_clusters_cluster_tsv_annotation_name_biosample_id_annotation_scope_study_annotation_type_group_cluster_name_cluster_tsv_fields_coordinates_cells_annotation_subsample_all
```
Now, this path is condensed into the following:
```
_single_cell_api_v1_studies_SCP63_clusters_cluster_tsv_8621f4a717a3a3fa4aaa58b596c6ee950d1aaac6d1c257bbe7235ed96e7cf8fa
```
Since the study accession is still present, we can continue clearing caches as we always have as this is the only request parameter that we use when determining which entries to remove.  

#### MANUAL TESTING
1. Boot as normal and ensure that request caching is turned on (run `bin/rails dev:cache` in the terminal until you see the message `Development mode is now being cached.`)
2. Load any study with visualizations enabled, and select various clusters/annotations, making sure to toggle back and forth between several values so that some are requested more than once
3. Confirm new requests have messages in `development.log` like the following:
```
Writing to API cache: _single_cell_api_v1_studies_SCP75_clusters_Cluster_1_5f129e12e78bbb214391f9ba4dfd31bc9f538cc0b3fdb9db27d9cf389c19d9e1
```
4. Issuing the same request again should show as a hit in the cache and that execution halts and renders immediately:
```
Reading from API cache: _single_cell_api_v1_studies_SCP75_clusters_Cluster_1_5f129e12e78bbb214391f9ba4dfd31bc9f538cc0b3fdb9db27d9cf389c19d9e1
```

### DIFF
--- a/app/lib/request_utils.rb
+++ b/app/lib/request_utils.rb
@@ -17,29 +17,24 @@ class RequestUtils
   # Cache path methods
   ##
 
+  # get a unique cache path for an individual request
+  # will result in paths with fixed lengths, such as
+  # _single_cell_api_v1_clusters_SCP1234_f6c41dbf8f1760d7dc1b36a3db7c05ec56049a0bdc54644639005c0e41bd7490
   def self.get_cache_path(request_path, url_params)
     # transform / into _ to avoid encoding as %2f
     sanitized_path = sanitize_value_for_cache(request_path)
-    # remove unwanted parameters from cache_key, as well as empty values
-    # this simplifies base key into smaller value, e.g. _single_cell_api_v1_studies_SCP123_explore_
-    # parameters must also be sorted by name to ensure cache paths are idempotent
-    params_key = url_params.reject {|name, value| CACHE_PATH_EXCLUDE_LIST.include?(name) || value.empty?}.sort_by {|k,v| k}.
-      map do |parameter_name, parameter_value|
-      if parameter_name == 'genes'
-        "#{parameter_name}_#{construct_gene_list_hash(parameter_value)}"
-      else
-        "#{parameter_name}_#{sanitize_value_for_cache(parameter_value).split.join('_')}"
-      end
-    end
-    [sanitized_path, params_key].join('_')
+    digest_key = construct_params_digest(url_params)
+    [sanitized_path, digest_key].join('_')
   end
 
-  # create a unique hex digest of a list of genes for use in get_cache_key
-  # this prevents long gene list queries from being split in the middle due to maximum filename length limits
+  # create a unique hex digest for a hash of request parameters
+  # parameters are sorted to ensure idempotency of resulting digest, which is critical for cache management
+  # this prevents long parameter lists from being split in the middle due to maximum filename length limits
   # and resulting in invalid % encoding issue when trying to clear selected cache entries
-  def self.construct_gene_list_hash(query_list)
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good!

Have we been able to reliably reproduce this error?  My understanding is "[not quite](https://broadinstitute.slack.com/archives/CBEHTH601/p1662567349466439)".

I ask because this change reduces observability, compressing long meaningful strings into shorter non-meaningful strings.   _If we see a future case_ of [the error Sentry reported](https://sentry.io/organizations/broad-institute/issues/3565574180/?end=2022-09-13T18%3A51%3A16&environment=production&project=1424198&query=is%3Aunresolved+cache_removal_job.rb&start=2022-06-15T18%3A51%3A16), given the observability cost of this change, I'd lean towards reverting this until we can reliably show it fixes an error observed in the wild.

Even if this fix is speculative, your hypothesis is plausible, and the severity of this rare bug is high.  So I approve.

bistline:
> Have we been able to reliably reproduce this error? My understanding is "[not quite](https://broadinstitute.slack.com/archives/CBEHTH601/p1662567349466439)".

That's an excellent question - I have been able to reliably reproduce the `invalid %-encoding` error, but not `invalid byte sequence in UTF-8`.  That being said, my investigation into [the latter error](https://sentry.io/organizations/broad-institute/issues/3565574180/?project=1424198&query=is%3Aunresolved+byte&statsPeriod=7d) revealed that the issue arises when traversing the directory tree and comparing values against the provided regular expression.  You can see this if you expand the error to show the full stack trace.  By this logic, if we remove potentially offending parameter values completely from the encoded path, it _should_ resolve the issue as these will be converted into hexadecimal values.  There are still some path-based parameters that might cause this problem, but I think that they still relied on the string breaking at ~255 characters to create an invalid sequence.

As to your comment about observability, while I agree on its face - it replaces human-readable values with encoded ones - in practice we have never gone through the cache directory tree looking for individual entries based on request parameters.  The fact that they are all nested 2 directories deep makes doing so very cumbersome, and the first two directory names are randomly generated values.  Furthermore, those parameters are never used anywhere in finding/removing cache paths, and are as such irrelevant in terms of existing functionality for caching.

ehanna4 (approved):
Functional review only - works as advertised, can see the reads and write to API cache and the halting as described

jlchang (approved):
Functional review: works as described. Reads and writes have hex digest instead of full URL string, filter chain halts on cache hits.

PR #1614, opened 2022-09-08
Increase search relevance, address HTTP errors by splitting keyword searches in Azul (SCP-4531)

bistline:
#### BACKGROUND
Due to how the logic of simulating keyword searches in Azul is implemented in SCP, we often run into an issue with some keywords resulting in very large compound query objects due to the amount of matches given for possible hits.  This will then result in an `HTTP 413 Payload Too Large` or `HTTP 414 URI Too Long` error from Azul, and no results are returned to the user.  Furthermore, even if the request does go through, usually no results are returned at all due to the fact that Azul does not natively support keyword based search only supports `AND` logic across facets, as opposed to `OR` logic which is expected of any keyword-based search.

#### CHANGES
This update addresses both issues by taking the compound query object created from the keyword search and splitting out each facet into a separate search, and then joining the results after.  The search requests are parallelized in separate threads to reduce overall latency.  Furthermore, all common stop words (stored in `StudySearchService::STOP_WORDS`) and the terms `cell` and `cells` are now ignored when attempting to find possible matches, as are any terms shorter than 3 characters.  This mirrors functionality in SCP-based searches, and removes many spurious matches due to the preponderance of the term `cell`, as well as partial string matches on single/double characters.  Note: quoted phrases such as `"B cell"` are still honored and treated as a single term.  This update also fixes a subtle bug with search term highlighting were actual matches were not being properly passed from the API to the UI.

It should be noted that a very large keyword search could result in slow load times for search results, especially if pagination is required.  This should be monitored in MixPanel to ascertain whether or not more performance tuning is required.  Limited testing has shown that the performance is roughly on par with current facet-based searches in Azul.

This update also addresses a recently found issue with `dependabot` PRs not being able to run CI due to a problem in formatting the Docker image tag.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Enter the following query in the text search box: `Tumor and immune reprogramming during immunotherapy in advanced renal cell carcinoma` (no quotes)
3. Confirm the search executes and returns HCA results, such as [The immune cell atlas of human neuroblastoma](https://data.humancellatlas.org/explore/projects/8f630e0f-6bf9-4a04-9754-02533152a954)
4. In `development.log`, note the following logging statements regarding query splitting:
```
[...]

PR #1620, opened 2022-09-19
Ignore case when filtering terms for Azul (SCP-4673)

bistline:
#### BACKGROUND & CHANGES
Changes introduced in #1614 resulted in a corner case where terms that should have been ignored for queries in Azul were not filtered due to case sensitivity, resulting in `HTTP 413` errors again.  This update now converts all search terms to lower case before filtering.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Run a search for `Cell atlas of aqueous humor outflow pathways in eyes of humans and four model species provides insight into glaucoma pathogenesis`
3. Confirm the request succeeds and [Single-cell RNA-Seq Investigation of Foveal and Peripheral Expression in the Human Retina](https://data.humancellatlas.org/explore/projects/4bec484d-ca7a-47b4-8d48-8830e06ad6db) is returned from Azul
4. In `development.log`, note the query for Azul does not contain `Cell`:
```
Executing Azul project query with: {"sampleDisease"=>{"is"=>["glaucoma", "glaucoma (disease)"]}}
```

### DIFF
--- a/app/models/hca_azul_client.rb
+++ b/app/models/hca_azul_client.rb
@@ -350,7 +350,7 @@ class HcaAzulClient
   #   - (Array<Hash>) => Array of facet objects to be fed to :format_query_from_facets
   def format_facet_query_from_keyword(term_list = [])
     matching_facets = []
-    sanitized_terms = (term_list - IGNORED_WORDS).reject { |t| t.size < 3 }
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good!

PR #1621, opened 2022-09-19
MVP Legend search (SCP-4622)

ehanna4:
Add a feature flagged legend search for plot legends that have more than 30 entries.

To test:
- Pull this branch
- Run the DB migration with: `bin/rails db:migrate`
- Boot up a localhost instance as usual
- log in and go to the feature flag section
- set the feature flag for legend search to true for your user
- go to a study that is visualizable that has an annotation with more than 30 entries (I looked at `Study: Aging mouse` using `cell_type_by_age` annotation)
- Confirm that legend search appears and you can filter down the legend results and the `hide all` and `show all` buttons continue to work as expected.

NOTE:
- Customizable colors UI is currently broken so I have disabled customizing colors when the legend search is enabled for now to not exacerbate the problem and not scope creep this ticket. There is a different ticket in sprint to fix the buggy custom colors.

https://user-images.githubusercontent.com/54322292/191302380-8ba4e4db-abee-4701-a205-0f5971f19dc3.mov


bistline (approved):
Looks good!  Works as advertised.

eweitz (approved):
Code looks good!

I note a trivial blocker on maintainability.  I also suggest several non-blocking changes for the same, and a few optimizations.

A few other non-blocking requests:
* Could you post a screenshot or brief video showing the enhanced UI?
* Given we intend to make this available to some end-users on production this sprint, could you add an automated test?

Finally, as a general post-MVP note, we might want to consider using `for` loops instead of ES6 iteration constructs as part of our performance strategy for search over many labels (SCP-4604).

ehanna4:
@eweitz Do you think it best to add the testing to this PR or add that to the AC of the ticket to make legend search GA?

eweitz:
I defer to you, but ideally some automated testing is in place before releasing code to production -- even if feature-flagged.  Last quarter's differential expression work is an example.  Another factor is whether you think legend search will be GA soon. If so, tests now seem less urgent; if not, tests now seem a bit more urgent -- e.g. so we know if we broke something that only an important stakeholder is using.

eweitz:
Also, thanks for the video -- it's a crisp, durable demo.

jlchang (approved):
Functional review - works as described!

PR #1622, opened 2022-09-19
Move exact study name matches to top of search results (SCP-4674)

bistline:
#### BACKGROUND
Text-based searches us `OR` logic when entering multiple terms, but quoted strings are treated as a single entity.  However, most users when searching by the title of a paper/study do not think to quote the search, expecting the highest-relevance match would be the entry with the same name.  This is not always the case, and can leave the study the user was looking for buried in pages or results.

#### CHANGES
This update now attempts to match the entire search string against the names of studies in the results array, and promotes that match to the first entry, if found.  The logic for matching will be case-insensitive and ignore non-word characters to make it more permissive.  Using [`Enumerable#partition`](https://ruby-doc.org/core-3.1.2/Enumerable.html#method-i-partition) chained to [`Array#flatten`](https://ruby-doc.org/core-3.1.2/Array.html#method-i-flatten) will exit silently if no match is found.

#### MANUAL TESTING
1. Boot as normal
2. Once studies load on the home page, copy the exact title of one of the entries that is not the first result (e.g. `Male Mouse Brain`)
3. Paste the name into the search box without quoting
4. Confirm that study returns first

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -343,7 +343,9 @@ module Api
         end
 
         # attempt to promote an exact text-string match, if possible
-        @studies = self.class.promote_exact_match(params[:terms], @studies) if params[:terms].present?
+        if params[:terms].present?
+          @studies, @match_by_data = self.class.promote_exact_match(params[:terms], @studies, @match_by_data)
+        end
 
         # save list of study accessions for bulk_download/bulk_download_size calls, in order of results
         @matching_accessions = @studies.map { |study| self.class.get_study_attribute(study, :accession) }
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good, as did functionality at demo!

ehanna4 (approved):
Demo was great and the code looks clean to me 👍

jlchang (approved):
Functional review - works as described!

Not specific to this PR, I see these suggestions the logs:
> Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
> 
> (!) Some chunks are larger than 500 KiB after minification. Consider:
> - Using dynamic import() to code-split the application
> - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/guide/en/#outputmanualchunks
> - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.

bistline:
@eweitz @ehanna4 See [aa10e83](https://github.com/broadinstitute/single_cell_portal_core/pull/1622/commits/aa10e830e011b0f4a2ab05a8674a7e89790617cf) - does this look good for metrics to report on when we get an exact title match that was "promoted" to the front?

ehanna4:
The metrics look good, I like the structure you went with, fits well with existing analytics 

Screen Shot 2022-09-21 at 2 53 55 PM

eweitz:
Looks good to me, too!  

I wish I could remember if any existing Mixpanel reports use this "search numResults" instrumentation.  It seems clearly valuable and is instrumented in #1429 and #1481, but I can't find any actual reports etc. using it in Mixpanel.

PR #1623, opened 2022-09-20
Remove unused docker artifacts after successful deployment (SCP-4551)

bistline:
#### BACKGROUND
There is currently no automation to clean up unused Docker images/layers/volumes on deployed hosts.  As such, over time disks fill up and eventually lead to [broken deployments](https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-development-to-staging/823/console).

#### CHANGES
This update adds automation to remove any unused Docker images, as well as local volumes no longer associated with a running container.  A recent manual cleanup removed ~21 GiB of unused images & volumes on our staging server.  The two most recent images will be kept - the first is the current image in use, and the second is a rollback candidate, if needed.

This PR also contains a documentation fix regarding OAuth client IDs as defined in SCP-4686 (incorrect JavaScript origins/redirects).

#### MANUAL TESTING
The easiest way to test this is to manually pull several copies of our main Docker image and then run the cleanup command directly:
1. Start Docker and run the following commands to pull 3 copies of our Docker image:
    * `docker pull gcr.io/broad-singlecellportal-staging/single-cell-portal:1.25.0`
    * `docker pull gcr.io/broad-singlecellportal-staging/single-cell-portal:1.24.0`    
    * `docker pull gcr.io/broad-singlecellportal-staging/single-cell-portal:1.23.0`
2. Navigate to the `bin` directory inside the project source directory
3. Load the `docker_utils.sh` script with `source docker_utils.sh`
4. Run `prune_docker_artifacts 'gcr.io/broad-singlecellportal-staging/single-cell-portal'` and confirm that it removes the `1.23.0` image, keeping both `1.25.0` and `1.24.0`
5. Run `docker volume ls` and confirm there are no volumes left
6. Rerun `prune_docker_artifacts 'gcr.io/broad-singlecellportal-staging/single-cell-portal'` and confirm that no images are deleted

### DIFF
--- a/bin/docker_utils.sh
+++ b/bin/docker_utils.sh
@@ -80,3 +80,37 @@ function ensure_container_running {
     done
     echo "$RUNNING"
 }
+
+# get all matching images, will return image IDs in the order of most recent first
+function get_matching_image_ids {
+    IMAGE_NAME="$1"
+    IMAGES=$(docker images | grep -F $IMAGE_NAME | awk '{ print $3 }')
+    echo $IMAGES
+}
+
+# remove all but the most recent release image
+function prune_docker_artifacts {
+    IMAGE_NAME="$1"
+    # get all matching images, then pop the first entry off
+    ALL_IMAGES=($(get_matching_image_ids $IMAGE_NAME))
+    RELEASE_IMAGE_ID="${ALL_IMAGES[1]}"
+    ALL_IMAGES=("${ALL_IMAGES[@]:1}")
+    RELEASE_IMAGE_NAME=$(get_image_tag_from_id $RELEASE_IMAGE_ID)
+    echo "Keeping $RELEASE_IMAGE_NAME image as most recent"
+    for IMAGE in $ALL_IMAGES; do
+        IMAGE_TAG=$(get_image_tag_from_id $IMAGE)
+        echo "Removing obsolete image $IMAGE_TAG"
+        docker rmi $IMAGE
+    done
+    # prune unused image layers and volumes
+    echo "pruning orphaned image layers"
+    docker image prune --force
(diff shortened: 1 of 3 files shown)


bin/docker_utils.sh
```diff
@@ -80,3 +80,37 @@ function ensure_container_running {
...
 }
+
+# get all matching images, will return image IDs in the order of most recent first
+function get_matching_image_ids {
+    IMAGE_NAME="$1"
+    IMAGES=$(docker images | grep -F $IMAGE_NAME | awk '{ print $3 }')
+    echo $IMAGES
+}
+
+# remove all but the most recent release image
```
eweitz, line 91: I can't ascertain at a glance: does this mean the image needed for a faster potential rollback would be retained, in addition to the image for the current / upcoming deployment?

bistline: Good thought - no, it wouldn't, but the rollback deployment will pull that image before stopping the existing container.  However, it would then try to clean it up by this logic (which would fail) and then mark the deployment as "failed" even though it succeeded.  We could keep the last 2 images to account for a potential rollback without altering any other deployment logic.

eweitz: Cool, keeping the last 2 images indeed seems robust, efficient, and simple for the case of rollback.

bistline: [7b1a75e](https://github.com/broadinstitute/single_cell_portal_core/pull/1623/commits/7b1a75e887a306db105ae91524897e3f05ce8aaf)

eweitz (approved):
Code looks good!

ehanna4 (approved):
Functional review - works as advertized

PR #1624, opened 2022-09-20
Enable linking clusters to external resources (SCP-4668)

eweitz:
This lets study editors show a link to an external resource for a cluster, e.g. to a Nissl stain image for a cluster scatter plot.

It's part of our U19 Mouse Atlas dataset work.  Here's how it looks:

https://user-images.githubusercontent.com/1334561/191530408-abba34e8-fa30-42f4-85cd-d40f685883cc.mp4

(The unusual transitions in above video are me cutting out sluggish local wait times.  I used [Canva](https://www.canva.com/features/video-trimmer/).)

To test:
* Go to a study with visualizations
* Go to its upload UI, then "Clustering" tab
* Add values for the new external link fields, e.g.:
  * URL: http://brainarchitecture.org/viewer4/mouse/map/13652N/3483091
  * Title: Brain architecture section 1
  * Description: Corresponding registered brain section (MD699) (section 368) 
* Click "Save"
* In your `rails console -e development`, run `Rails.cache.clear`
* Go to Study page, Explore tab
* Note link, click it, confirm it opens external web resource in new tab

This satisfies SCP-4668.


app/controllers/api/v1/study_files_controller.rb
```diff
@@ -687,6 +687,7 @@ def study_file_params
                                            :description, :is_spatial, :file_type, :status, :human_fastq_url, :human_data, :use_metadata_convention,
                                            :cluster_type, :generation, :x_axis_label, :y_axis_label, :z_axis_label, :x_axis_min,
                                            :x_axis_max, :y_axis_min, :y_axis_max, :z_axis_min, :z_axis_max, :species, :assembly,
+                                           :external_link_url, :external_link_title, :external_link_description,
```
bistline, line 690: BLOCKER: We're defining these parameters at the `StudyFile` level, but then they are also present in `ClusterFileInfo` as well (and apparently unused there).  We should settle on one over the other to avoid confusion.

My preference would be to put them in `ClusterFileInfo`, as that data is specific to cluster files, and we already have way too many misplaced attributes on `StudyFile` documents (like axis ranges, labels, etc).  I won't consider moving the data there to be a blocker - just the decision of one storage location over the other.

Passing the parameters through would be more challenging, though - you'd need to add a block like this to `study_file_params`:
```
cluster_file_info_attributes: { :external_link_url, :external_link_title, :external_link_description }
```
You would then need to add them to the correct form with the parameters namespaced accordingly so that they assign to the nested document (this may require a new component), and (possibly) add `study_file.build_cluster_file_info if study_file.file_type == 'Cluster` to `StudiesController#new_study_file`.

app/models/cluster_file_info.rb
```diff
@@ -4,10 +4,17 @@ class ClusterFileInfo
...
 
   field :custom_colors, type: String
+
   # hash of annotation names to booleans of whether the annotation should, by default, be shown split
   # on pipes (assuming it is array-based).  Unspecified is treated as false by the front end
   field :annotation_split_defaults, type: String
 
+  # Hyperlink to an external web resource for this cluster
+  # Set in upload / sync UI.  Shown in Study Overview page.
+  field :external_link_url, type: String
```
bistline, line 14: See above comment re: BLOCKER

app/models/study_file.rb
```diff
@@ -100,6 +100,12 @@ class StudyFile
   field :z_axis_max, type: Integer
   field :is_spatial, type: Boolean, default: false
 
+  # Hyperlink to an external web resource for a cluster
+  # Set in upload / sync UI.  Shown in Study Overview page.
+  field :external_link_url, type: String
```
bistline, line 105: See above comment re: BLOCKER

app/controllers/studies_controller.rb
```diff
@@ -976,6 +976,7 @@ def study_file_params
                                        :remote_location, :description, :file_type, :status, :human_fastq_url, :human_data, :use_metadata_convention,
                                        :cluster_type, :generation, :x_axis_label, :y_axis_label, :z_axis_label, :x_axis_min, :x_axis_max,
                                        :y_axis_min, :y_axis_max, :z_axis_min, :z_axis_max, :is_spatial, :taxon_id, :genome_assembly_id, :study_file_bundle_id,
+                                       :external_link_url, :external_link_title, :external_link_description,
```
bistline, line 979: See above comment re: BLOCKER

bistline (changes requested):
Functionally this works as expected, but I have concerns about where the data is being stored - it's defined in two places, but only used in one.

app/models/cluster_file_info.rb
eweitz: See [reply](https://github.com/broadinstitute/single_cell_portal_core/pull/1624#discussion_r976748196).

app/models/study_file.rb
eweitz: See [reply](https://github.com/broadinstitute/single_cell_portal_core/pull/1624#discussion_r976748196).

bistline (approved):

app/controllers/studies_controller.rb
eweitz: See [reply](https://github.com/broadinstitute/single_cell_portal_core/pull/1624#discussion_r976748196).

eweitz:
The [build failure](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/3099381539/jobs/5018452075#step:5:4255) seems to be a true positive -- I'm fixing it now.

> `ClustersControllerTest#test_show_should_return_visualization_information [/home/app/webapp/test/api/visualization/clusters_controller_test.rb:72]:`

ehanna4:
If the external link title is super long it just keeps going behind the other elements, it might make sense to truncate long titles or set a limit for the length of a title here, especially because the tooltip then can show in weird spots.

Screen Shot 2022-09-21 at 2 15 06 PM

ehanna4:
Another TOL I worry the amount of white space between the legend and the external link takes up precious real-estate from the legend, I wonder if reducing that space and maybe adding a line between the sections would do similar to show they are separate but take up less space?

eweitz:
Good finds on robustness, @ehanna4!  Your styling note is also on point.

**Overflow**
Unsightly overflow only seems to occur when the link title is both especially long _and_ lacks spaces.  This could happen if the user sets the title to a URL as in your example, but that's borderline user error and I suspect it'll be very rare.  I also think a robust fix would be worth a separate task, if we pursue it.

**External link lacking title**
Resolving this properly also seems worth a separate task.  I've opened SCP-4693 to track that, with accompanying details there.  Like the overflow issue, I suspect this user error will be rare in the wild, and often quickly noticed and trivially fixed by study editors when it occurs.

**Vertical space**
I hear you on the preciousness of vertical real estate!  I went ahead and reduced the negative spacing beneath the cluster external link and added a line to demarcate the sections, as you suggested.

Screen Shot 2022-09-21 at 4 33 26 PM

jlchang (approved):

ehanna4:
@eweitz Awesome thanks for taking a look, I think with your way of addressing those issues makes sense!

JZL:
Hooray @jlchang @eweitz!
That looks great, my only thought was that it could get hidden but I think the line + link icon does a good deal to help. I also think having it change when the coordinate dropdown is changed will help draw the eye too

This feature will be very useful for seeing the nissl side by side

(I had a fun time reading through all the comments, it really shows how much care is put into every feature :-) )

Thanks
Jonah

PR #1625, opened 2022-09-21
Fixing bug with array indexing in zsh vs bash (SCP-4551)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with the difference in array indexing between `bash` and `zsh` as seen on [this recent deployment](https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-development-to-staging/826/console) (scroll to the bottom):
```
### CLEANING UP SECRETS/OLD IMAGES ###
"docker image inspect" requires at least 1 argument.
See 'docker image inspect --help'.

Usage:  docker image inspect [OPTIONS] IMAGE [IMAGE...]

Display detailed information on one or more images
Keeping null as current,  as rollback
```
#### MANUAL TESTING
1. Pull branch and navigate to the `bin` directory in project source
2. Run `echo $SHELL` to get the value of your current shell
3. If you have `zsh`, then type `bash` to enter that shell
4. Follow steps from #1623 to confirm the cleanup script still works (note: you may have to load the script with `. docker_utils.sh` instead of `source docker_utils.sh`)

### DIFF
--- a/bin/docker_utils.sh
+++ b/bin/docker_utils.sh
@@ -90,13 +90,18 @@ function get_matching_image_ids {
 
 # remove all but the most recent release image
 function prune_docker_artifacts {
+    INDEX=0
+    if [[ "$SHELL" = '/bin/zsh' ]]; then
+        echo "zsh detected, setting array index to 1"
+        INDEX=1
+    fi
     IMAGE_NAME="$1"
     # get all matching images, then pop off the first two as "current" and "rollback" images
     ALL_IMAGES=($(get_matching_image_ids $IMAGE_NAME))
-    RELEASE_IMAGE_ID="${ALL_IMAGES[1]}"
-    ALL_IMAGES=("${ALL_IMAGES[@]:1}") # remove first element and reindex
-    ROLLBACK_IMAGE_ID="${ALL_IMAGES[1]}"
-    ALL_IMAGES=("${ALL_IMAGES[@]:1}")
+    RELEASE_IMAGE_ID="${ALL_IMAGES[$INDEX]}"
+    ALL_IMAGES=("${ALL_IMAGES[@]:$INDEX}") # remove first element and reindex
+    ROLLBACK_IMAGE_ID="${ALL_IMAGES[$INDEX]}"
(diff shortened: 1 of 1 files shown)


bin/docker_utils.sh
```diff
@@ -90,13 +90,18 @@ function get_matching_image_ids {
 
 # remove all but the most recent release image
 function prune_docker_artifacts {
+    INDEX=0
```
ehanna4, line 93: Are there any other options other than zsh or bash we need to be worried about?

I was curious about why zsh would be indexed at 1 and came across this stack [overflow](https://stackoverflow.com/questions/50427449/behavior-of-arrays-in-bash-scripting-and-zsh-shell-start-index-0-or-1) which suggests using `${array[@]:offset:length}` which, if we are worried about any other options outside `zsh` or `bash` might be more robust? Just thought I'd toss it out there, didn't look deeply into it so this might not work for us, and the current solution here seems good too!

bistline: To answer your question - not that I'm aware of.  Our GCP VMs all have `bash` as the default shell, and newer Macs use `zsh`.  I don't foresee an instance where we would need to support any other shells (like `csh` or `ksh`).

The issue with the above syntax is that it redefines and truncates the array in-place, which we don't necessarily want - we won't know how many copies of a particular image may or may not be on the VM at any given time.  I had started on that exact stack overflow thread and found that the answer wasn't what we're looking for.

ehanna4: okay cool thanks for the extra context! 👍

ehanna4 (approved):

bin/docker_utils.sh
eweitz: > newer Macs use `zsh`

Just for context, I'm on a newer macOS (Monterey, 12.5.1) and while Terminal prompts me at startup that:

```
The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
```

I still use `bash`.  Devon did similarly.  No action needed, just thought it worth noting.

PR #1631, opened 2022-09-29
Fix custom color setting buggyness (SCP-4684)

ehanna4:
TLDR: All buttons (cancel, OK, save, reset) work (both client and server side) and you do not have to hover to trigger the color update

We noticed a few bugs with setting custom colors in scatter plots, including the Cancel and Save buttons not working to set the colors in the UI and needing to hover over the legend labels to trigger the updated-but-not-saved colors to update.

This PR is fixing all the weird buggy-ness for typical custom coloring. 

To test:
- Pull this branch
- Boot up local dev as usual and log in
- Go to a study that has visualizations 
- Click Customize Colors 
- Try all variations of changing colors on labels, clicking OK or Cancel in the modal and clicking Save, Cancel, and Reset to Defaults in the customize color sections. Confirm that the colors update as you would expect. And confirm that refreshing persists Saved colors but does not persist unsaved colors.

BEFORE BUGS:

https://user-images.githubusercontent.com/54322292/193067370-75e802ab-3f18-44ef-b405-bcde08eb322a.mov

https://user-images.githubusercontent.com/54322292/193067467-f1844064-284d-49dc-a2b1-30a86c5730f1.mov

AFTER:

https://user-images.githubusercontent.com/54322292/193068726-c5470d67-2e4a-491f-8ae2-8da08f028830.mov

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -126,7 +126,7 @@ function RawScatterPlot({
   }, [editedCustomColors])
 
   /** Save any changes to the legend colors */
-  const saveCustomColors = async newColors => {
+  async function saveCustomColors(newColors) {
     const colorObj = {}
     // read the annotation name off of scatterData to ensure it's the real name, and not '' or '_default'
     colorObj[scatterData?.annotParams?.name] = newColors
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -57,6 +57,7 @@ function RawScatterPlot({
   const [activeTraceLabel, setActiveTraceLabel] = useState(null)
   // map of label name to color hex codes, for any labels the user has picked a color for
   const [editedCustomColors, setEditedCustomColors] = useState({})
+  const [customColors, setCustomColors] = useState({})
 
   const isRefGroup = getIsRefGroup(scatterData?.annotParams?.type, genes, isCorrelatedScatter)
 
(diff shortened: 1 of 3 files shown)


app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -116,13 +117,21 @@ function RawScatterPlot({
...
+  useEffect(() => {
+    if (editedCustomColors && Object.keys(editedCustomColors).length > 0) {
+      const plotlyTraces = document.getElementById(graphElementId).data
+      Plotly.react(graphElementId, plotlyTraces, scatterData.layout)
+    }
+  }, [editedCustomColors])
+
   /** Save any changes to the legend colors */
-  async function saveCustomColors(newColors) {
+  const saveCustomColors = async newColors => {
```
eweitz, line 129: Blocker: SCP JS uses classic function signatures for these things, which helps look up functions when one's IDE is having issues.  If we change that convention, I'd advocate doing it as a holistic sweep, for consistency.

```suggestion
  async function saveCustomColors(newColors) {
```

eweitz (approved):
Code looks good overall!  I appreciate the demo videos.

I note one trivial blocker on an important element of code style consistency.  

Also, given the complex manual tests needed to verify these fixes in production code, could you add an automated test, or ticket it as tech debt in the upcoming sprint?   The test doesn't need to exhaustively verify all paths -- having even just some additional coverage here would give us more confidence in future changes.

app/javascript/components/visualization/ScatterPlot.jsx
ehanna4: oops yes will flip that back, force of habit haha

ehanna4:
test failure seems it might be real - will take a look into it

bistline (approved):

PR #1633, opened 2022-10-05
Allow study editors to specify override_viz_limit_annotations (SCP-4739)

bistline:
#### BACKGROUND
The admin-only feature `Study#add_override_viz_limit_annotation` allows admins to specify annotations that do not meet our visualization criteria (2..200 values) as "visualizable" for a given study.  These values are stored in the `study.default_options` hash.  However, updating any visualization defaults through the study settings forms will not persist this value as it is not present in the form.

#### CHANGES
This update adds those values to the forms as a multi-select with checkboxes, allowing the study editor to specify annotations without admin intervention.  This also fixes the persistence bug when saving other values as it will always be passed with form data.

Screenshot of new form control:
Screen Shot 2022-10-05 at 1 50 15 PM

#### MANUAL TESTING
1. Boot as normal and sign in
2. If you don't already have it in your local instance, populate the synthetic study `Mouse stomach cells and GERD, with over 200 cell types` by running the following in the Rails console (you'll need DelayedJob running to launch the parse job): 
```
SyntheticStudyPopulator.populate('mouse_multilabel')
```
3. Load the `Mouse stomach cells and GERD` study and open the Settings -> View options tab
4. For `Annotation display override`, select the `manylabel_field` field and one other, then save
5. Open the Explore tab and confirm you can select `manylabel_field` and the other option you specified
6. Go back to the View options tab, and change the default annotation to `manylabel_field`
7. Go back to the Explore tab and confirm this is now the default annotation, and that `manylabel_field` and the other annotation are still selectable

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -140,6 +140,9 @@ class ApplicationController < ActionController::Base
         @default_cluster_annotations['Cannot Display'] << @default_cluster&.formatted_cell_annotation(cell_annotation)
       end
     end
+    # set array of viz override values, always showing current selections first
+    invalid_annotation_names = @default_cluster_annotations['Cannot Display'].map(&:first)
+    @viz_override_annotations = @study.override_viz_limit_annotations + invalid_annotation_names
     # initialize reviewer access object
     @reviewer_access = @study.reviewer_access.present? ? @study.reviewer_access : @study.build_reviewer_access
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -141,7 +141,7 @@ class ApplicationController < ActionController::Base
       end
     end
     # set array of viz override values, always showing current selections first
-    invalid_annotation_names = @default_cluster_annotations['Cannot Display'].map(&:first)
+    invalid_annotation_names = @default_cluster_annotations['Cannot Display'].map(&:first).sort
     @viz_override_annotations = @study.override_viz_limit_annotations + invalid_annotation_names
     # initialize reviewer access object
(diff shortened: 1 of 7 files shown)


eweitz (changes requested):
Code looks good!

A functional blocker: whenever many options are available and multiple can be selected in a plain form, [we should use checkboxes](https://ux.stackexchange.com/questions/129938/multi-selection-dropdown-with-or-without-checkbox).  These are the standard affordance that multiple things can be selected -- and we use this pattern in e.g. our search filter selection UI.

If Rails doesn't have support for that, I think classic jQuery form parsing on submit / click would accommodate it.  (I certainly don't think we need to extend the React components for search filter UI to apply here.)

eweitz (approved):
Slick, quick work.  Kudos!

jlchang (approved):
Functional review: Works as described

Non-blocking suggestion - give Annotation display override field wider default width
On my laptop's smaller screen, the annotation label "library_preparation_protocol__ontology_label" wraps awkwardly unless my browser window is fully expanded.
Screen Shot 2022-10-11 at 6 59 22 AM
If it's not difficult, it may be worth making the field wider because "library_preparation_protocol__ontology_label" is a required field so we know that we'll regularly encounter annotation labels of that length.

Also, before this PR, if the study owner sets an invalid label as the default without setting the override, the label was stored as `<label>-<type>-invalid`. With this PR, if the study owner then sets the override, does the `scope` part of the annotation get updated from `invalid` to `<label>-<type>-<scope>`? Currently, it doesn't seem like the `invalid` scope prevents visualization so this is a non-blocking question.

PR #1637, opened 2022-10-12
Adding labels to PAPI virtual machines (SCP-4700)

bistline:
#### BACKGROUND
The Google Cloud Platform billing dashboard allows SCP team members to get detailed information about the cost of various assets that are a part of SCP core infrastructure (VMs, disks, etc).  However, these costs are rolled up by resource type (compute CPU/RAM, cloud storage, egress, etc.) and do not provide adequate granularity to determine how much is being spent on specific actions in SCP, such as file parsing, differential expression computation, or Image Pipeline activities.

#### CHANGES
This update adds labels to virtual machines created in PAPI, which will allow us to filter billing reports based on these key/value pairs.  The following labels will be added to every VM and pipeline request:
- `study_accession`: accession of study
- `user_id`: MongDB ID of user requesting action
- `filename`: name of primary file being processed
- `action`: summary label for action: `ingest_pipeline`, `differential_expression`, or `image_pipeline`
- `docker_image`: version tag of the `scp-ingest-pipeline` Docker image
- `environment`: the Rails environment this job was launched in: `development`, `test`, `staging`, or `production`
-  `file_type`: SCP file type of primary file
- `machine_type`: GCE machine type, e.g. `n1-highmem-4`
- `boot_disk_size_gb`: size of the boot disk provisioned for VM (usually 300GB)

Note: due to restrictions on label formatting, the values provided for these keys can only contain lowercase alphanumeric characters, plus dashes: `-` and underscores: `_`.  This means that values like study accessions will appear slightly different - `scp1234` instead of `SCP1234`, and `expression_matrix` instead of `Expression Matrix`.

#### MANUAL TESTING
1. Locally, boot as normal and sign in
2. In a separate terminal, start DelayedJob: 
```
./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails jobs:work
```
3. Upload a parseable file to any study - cluster, metadata, or expression matrix
4. In `development.log`, note that labels have been applied to the virtual machine:
```
---
...

:resources:
  :project_id: broad-singlecellportal-bistlin
  :regions:
  - us-central1
  :virtual_machine:
    :boot_disk_size_gb: 300
    :labels:
      :study_accession: scp72
      :user_id: 6033f4a8e2413917a1d0ddcd
      :filename: expression_matrix_example_txt
      :action: ingest_pipeline
      :docker_image: '1_22_0'
      :environment: development
      :file_type: expression_matrix
      :machine_type: n1-highmem-4
      :boot_disk_size_gb: '300'
    :machine_type: n1-highmem-4
    :preemptible: false
```
5. In a separate browser tab, go to your developer specific GCP console (e.g. `broad-singlecellportal-{username}`), and open the Compute Engine panel
6. Look for the `google-pipelines-worker` VM and click on that entry
7. Note under the `Labels` section, the same labels have been applied to the VM:
Screen Shot 2022-10-12 at 2 29 26 PM

Note: you will not be able to see these labels yet in the billing reports section, as charges do not roll up until the following day.  However, you should be able to filter on any of these labels the day after running these tests.

### DIFF
--- a/app/models/papi_client.rb
+++ b/app/models/papi_client.rb
@@ -33,6 +33,10 @@ class PapiClient
   # default GCE machine_type
   DEFAULT_MACHINE_TYPE = 'n1-highmem-4'.freeze
 
+  # regex to sanitize label values for VMs/pipelines
+  # alphanumeric plus - and _
+  LABEL_SANITIZER = /[^a-zA-Z\d\-_]/
+
   # Default constructor for PapiClient
   #
   # * *params*
@@ -377,16 +381,17 @@ class PapiClient
   # * *returns*
   #   - (Hash)
   def job_labels(action:, study:, study_file:, user:, machine_type: DEFAULT_MACHINE_TYPE, boot_disk_size_gb: 300)
+    ingest_version = AdminConfiguration.get_ingest_docker_image_attributes[:tag]
     {
-      study_accession: study.accession,
+      study_accession: sanitize_label(study.accession),
       user_id: user.id.to_s,
-      filename: study_file.upload_file_name,
+      filename: sanitize_label(study_file.upload_file_name),
       action: label_for_action(action),
-      docker_image: AdminConfiguration.get_ingest_docker_image,
-      environment: Rails.env,
-      file_type: study_file.file_type,
+      docker_image: sanitize_label(ingest_version),
+      environment: Rails.env.to_s,
+      file_type: sanitize_label(study_file.file_type),
       machine_type: machine_type,
-      boot_disk_size_gb: boot_disk_size_gb
+      boot_disk_size_gb: sanitize_label(boot_disk_size_gb)
     }
   end
 
@@ -408,6 +413,17 @@ class PapiClient
     else
       action
     end
+  end
+
+  # sanitizer for GCE label value (lowercase, alphanumeric with dash & underscore only)
+  #
+  # * *params*
+  #   - +label+ (String, Symbol, Integer) => label value
+  #
+  # * *returns*
+  #   - (String) => lowercase label with invalid characters removed
+  def sanitize_label(label)
+    label.to_s.gsub(LABEL_SANITIZER, '_').downcase
   end
 
   private
--- a/app/models/papi_client.rb
+++ b/app/models/papi_client.rb
@@ -154,10 +154,7 @@ class PapiClient
   # * *return*
   #   - (Google::Apis::GenomicsV2alpha1::RunPipelineRequest)
   def create_run_pipeline_request_object(pipeline:, labels: {})
-    Google::Apis::GenomicsV2alpha1::RunPipelineRequest.new(
-      pipeline: pipeline,
-      labels: labels
-    )
+    Google::Apis::GenomicsV2alpha1::RunPipelineRequest.new(pipeline:, labels:)
   end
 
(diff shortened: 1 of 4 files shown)


app/models/papi_client.rb
```diff
@@ -358,6 +360,64 @@ def get_command_line_options(study_file, action)
...
+  # * *returns*
+  #   - (String) => label for action, condensing all ingest actions to 'ingest'
+  def label_for_action(action)
+    case action.to_s
+    when /ingest/
+      'ingest_pipeline'
+    when /differential/
+      'differential_expression'
+    when 'render_expression_arrays'
+      'image_pipeline'
```
eweitz, line 404: I think we'll want to reserve `image_pipeline` for Image Pipeline proper, i.e. PAPI runs that output images.  If `batch_data_pipeline` doesn't seem fitting, maybe something like `data_cache_pipeline` would work -- anything other than exactly "image_pipeline" seems best.  

```suggestion
      'batch_data_pipeline'
```

app/models/papi_client.rb
```diff
@@ -358,6 +360,64 @@ def get_command_line_options(study_file, action)
...
 
+  # set labels for pipeline request/virtual machine
+  #
+  # * *params*
+  #   - +action+ (String, Symbol) => action being executed
+  #   - +study+ (Study) => parent study of file
+  #   - +study_file+ (StudyFile) => File to be ingested/processed
+  #   - +user+ (User) => user requesting action
+  #   - +machine_type+ (String) => GCE machine type
+  #   - +boot_disk_size_gb+ (Integer) => size of boot disk, in GB
```
eweitz, line 371: Note to self: this is, indeed, documented as GB and not GiB [upstream](https://cloud.google.com/sdk/gcloud/reference/alpha/genomics/pipelines/run#--boot-disk-size).

eweitz (approved):
Code looks good!

I suggest one change regarding labeling, for clearer interpretability.

Out of scope, but: would it be easy to log `machineType` and `bootDiskSizeGb` to Mixpanel?  I imagine these could help diagnose issues, perhaps as a separate task.

bistline:
> Code looks good!
> 
> I suggest one change regarding labeling, for clearer interpretability.
> 
> Out of scope, but: would it be easy to log `machineType` and `bootDiskSizeGb` to Mixpanel? I imagine these could help diagnose issues, perhaps as a separate task.

Yes - it will mostly be `n1-highmem-4` and `300 GB`, respectively, and for the "special" jobs, we have a `machine_type` parameter available to log.

app/models/papi_client.rb
bistline: Yes - Google is not consistent in its use of nomenclature.  Billing uses GiB, but resource allocation is GB.  Totally not confusing at all.

app/models/papi_client.rb
bistline: I think I like `data_cache_pipeline` for this - it matches the job better.

bistline: [b5c6c07](https://github.com/broadinstitute/single_cell_portal_core/pull/1637/commits/b5c6c07b79446b14091c6ecd277d759362e618fc)

ehanna4 (approved):
Functional review only - worked as advertized!
Screen Shot 2022-10-13 at 3 24 03 PM
Screen Shot 2022-10-13 at 3 51 42 PM

PR #1638, opened 2022-10-12
Add delay for setting active trace label (SCP-4604)

ehanna4:
This is to address the laggy scrolling of the scatter plot legend for studies with very long lists of labels.

Write up here: https://docs.google.com/document/d/1kA5E2_OnS9yQcIADUeOSkT5g05Yt5cEhQYJxsfzO1gY/edit of approach to looking into performance of such studies and proposed options.

TLDR: The lagginess of scrolling the legend list appeared to be caused by the hover functionality that would be triggered as a user scrolled down the list since their mouse would be hovering as it went triggering re-draws of the plot. This fix adds delay so that the hover functionality of bringing labels to the front of the plot won't be triggered on every single hover during a scroll.

To test:
- You need a really big study. I copied the one from Staging here: https://singlecell-staging.broadinstitute.org/single_cell/study/SCP149/5k-unqiue-u19-multi-cell-type-merged-labels#study-visualize
- Try scrolling that study in staging ^ to see the before behavior
- Pull this branch and start up your local server and navigate to your really big study
- Try scrolling the legend and observe the lagginess is fixed.

Before:

https://user-images.githubusercontent.com/54322292/195439794-c804fb16-583c-491a-a339-0cf25bccca5e.mov

After:

https://user-images.githubusercontent.com/54322292/195439577-f45f6545-39f5-4dfc-a86e-8cbfa01b22f9.mov

### DIFF
--- a/app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
+++ b/app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
@@ -60,6 +60,19 @@ function LegendEntry({
 
   // clicking the label will either hide the trace, or pop up a color picker
   const entryClickFunction = showColorControls ? () => setShowColorPicker(true) : toggleSelection
+  // clicking the label will either hide the trace, or pop up a color picker
+
+  let scatterHoverHandle
+  let scatterHoverId
+  // do not trigger `setActiveTraceLabel()` immediately to prevent triggers on scroll
+  const scatterHoverHandler = (label, currentSelectCheck) => {
+    clearTimeout(scatterHoverHandle)
+    scatterHoverHandle = setTimeout(() => {
+      if (scatterHoverHandle && currentSelectCheck()) {
+        setActiveTraceLabel(label)
+      }
+    }, 10000)
+  }
   return (
     <>
       <div
@@ -67,9 +80,19 @@ function LegendEntry({
         role="button"
         onClick={entryClickFunction}
         onMouseEnter={() => {
-          setActiveTraceLabel(label)
+          scatterHoverId = label
+          scatterHoverHandler(
+            label,
+            () => label === scatterHoverId
+          )
         }}
-        onMouseLeave={() => setActiveTraceLabel(null)}
+        onMouseLeave={() => {
+          scatterHoverId=''
+          scatterHoverHandler(
+            null,
+            () => '' === scatterHoverId
+          )
+        }}
       >
         <div className="scatter-legend-icon" style={iconStyle}>
           { showColorControls && <FontAwesomeIcon icon={faPalette} title="Change the color for this label"/> }
--- a/app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
+++ b/app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
@@ -60,7 +60,6 @@ function LegendEntry({
 
   // clicking the label will either hide the trace, or pop up a color picker
   const entryClickFunction = showColorControls ? () => setShowColorPicker(true) : toggleSelection
-  // clicking the label will either hide the trace, or pop up a color picker
 
   let scatterHoverHandle
   let scatterHoverId


app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
```diff
@@ -60,16 +60,38 @@ function LegendEntry({
...
         onMouseEnter={() => {
-          setActiveTraceLabel(label)
+          scatterHoverId = label
+          scatterHoverHandler(
+            label,
+            () => label === scatterHoverId
+          )
+        }}
+        onMouseLeave={() => {
+          scatterHoverId=''
```
eweitz, line 89:
```suggestion
          scatterHoverId = ''
```

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
```diff
@@ -60,16 +60,38 @@ function LegendEntry({
 
   // clicking the label will either hide the trace, or pop up a color picker
   const entryClickFunction = showColorControls ? () => setShowColorPicker(true) : toggleSelection
+
+  let scatterHoverHandle
+  let scatterHoverId
+  // do not trigger `setActiveTraceLabel()` immediately to prevent triggers on scroll
+  const scatterHoverHandler = (label, currentSelectCheck) => {
```
eweitz, line 67: Let's use classic function signatures, per previous review.  

It's consistent with other SCP JS, and enables easier "Lookup definition" when IDEs misbevave.

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
```diff
@@ -60,16 +60,38 @@ function LegendEntry({
...
+    clearTimeout(scatterHoverHandle)
+    scatterHoverHandle = setTimeout(() => {
+      if (scatterHoverHandle && currentSelectCheck()) {
+        setActiveTraceLabel(label)
+      }
+    }, 10000)
```
eweitz, line 73: 10 seconds seems too long, on first impression.  To my understanding, this means it'd take 10 s after scrolling for hovered traces to possibly activate (i.e. raise to top layer in scatter plot).  

Is that right?  If so, and if we keep this approach, could we decrease wait to 1 s?

```suggestion
    }, 1000)
```

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
```diff
@@ -60,16 +60,38 @@ function LegendEntry({
...
   // clicking the label will either hide the trace, or pop up a color picker
   const entryClickFunction = showColorControls ? () => setShowColorPicker(true) : toggleSelection
+
+  let scatterHoverHandle
+  let scatterHoverId
+  // do not trigger `setActiveTraceLabel()` immediately to prevent triggers on scroll
+  const scatterHoverHandler = (label, currentSelectCheck) => {
+    clearTimeout(scatterHoverHandle)
+    scatterHoverHandle = setTimeout(() => {
+      if (scatterHoverHandle && currentSelectCheck()) {
```
eweitz, line 70: Could you add a code comment here briefly explaining currentSelectCheck?  

I vaguely grok it after looking at the code for a while, but a brief comment that touches on e.g. why this needs to be a function parameter (i.e. `(label, currentSelectCheck)`) instead of a boolean parameter (e.g. `(label, isTraceCurrentlySelected)`) would make this more maintainable.

eweitz (commented):
Great detective work here!  As usual, the demo videos are super helpful.

I suggest some low-level refinements, most notably waiting 1 second instead of 10.

At a higher level, though, I wonder: would a solution based on [whether the user is scrolling](https://stackoverflow.com/a/52414518) would be more robust and simpler?

ehanna4:
@eweitz would you have some time after standup to chat? I'm not sure I understand the proposal with:

> At a higher level, though, I wonder: would a solution based on [whether the user is scrolling](https://stackoverflow.com/a/52414518) would be more robust and simpler?

to me it looks like this would essentially do the same thing as a time out, the only differences i see are that it would add more noise (i.e. resetting the 'lastScrollTime' variable repeatedly) and adds another listener - which admittedly is more intuitively related to scroll since it's on "scroll" however, there is already the 'OnMouseEnter' listener which can be treated like the start of a scroll here

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
ehanna4: Oh dear, I completely missed this extra `0` and putting the delay down to 1 sec mostly negates the scroll fix (but of course 10 sec makes no sense for a user trying to hover). 

Perhaps an approach of increasing delay as list size increases? I need to take another pass at this as a solution :/

app/javascript/components/visualization/controls/ScatterPlotLegend.jsx
ehanna4: I don't _think_ there's actually much reason to use the function over boolean, that was what I had seen so I followed the examples, but I switched to boolean param instead and saw no difference. I will double check but likely could switch this to `isTraceCurrentlySelected`

ehanna4:
I am going to close this PR as after pairing with Eric and reworking the solution this is no longer the method used and I'd like to keep the thumbs clean.

PR #1641, opened 2022-10-17
Adding local dev setup via docker-compose (SCP-4777)

bistline:
#### BACKGROUND
Although SCP is built and deployed via Docker, SCP team members historically develop outside of Docker as using it locally has significant costs both in setup/boot time and overall performance.  Additionally, features such as hot module replacement (HMR) via the `vite` dev server do not work in our current Docker setup.  The drawback is that this requires developers to install packages/gems directly on their machines, and updates to machine hardware (e.g. Macintosh M1 chips) can cause serious issues that require extensive workarounds.  Any gains in performance are then lost as developers spend hours - or days - just getting their environments working.  These kinds of issues are exactly what Docker was designed to address.

#### CHANGES
This update adds a local Dockerized setup for SCP development via `docker-compose` and some helper shell scripts.  This setup retains all of the performance and ease-of-use advantages of native Rails development (like HMR) without the need to install anything locally other than Docker.  It leverages existing work for [pre-built Docker images](#1552) to create a small local network that runs the portal Rails application and Vite dev server in side-by-side containers.  Additionally, this contains two helper shell scripts to both setup/launch the services (bin/docker-compose-setup.sh), as well as tear down and revert configurations back to the native development setup (bin/docker-compose-cleanup.sh).  The `docker-compose-dev.yaml` file also automates installing any package/gem updates, so that any local changes are automatically incorporated on boot.   Furthermore, no manual changes to configurations are required to use this.  The helper scripts automate any config file updates, and `docker-compose` handles all networking/port mapping.  This makes it possible to switch back and forth between Dockerized and non-Dockerized with a few shell commands.

This also updates an apparent typo in a regular expression in the latest migration that prevented it from running locally (though it appeared to run fine on staging, so it is unclear why it failed locally).

#### MANUAL TESTING
1. Pull this branch and ensure Docker is running
2. For improved boot time, pull the latest copy of the development image: 
```
docker pull gcr.io/broad-singlecellportal-staging/single-cell-portal:development
```
3. Start the services with `bin/docker-compose-setup.sh`
4. The helper script will initialize the environment and then start both services.  You will know that it is done when you see both of the following log messages in the terminal:

Server running:
```
single_cell       | => Booting Puma
single_cell       | => Rails 6.1.6.1 application starting in development 
single_cell       | => Run `bin/rails server --help` for more startup options
single_cell       | Puma starting in single mode...
single_cell       | * Puma version: 5.6.4 (ruby 3.1.2-p20) ("Birdie's Version")
single_cell       | *  Min threads: 5
single_cell       | *  Max threads: 5
single_cell       | *  Environment: development
single_cell       | *          PID: 1
single_cell       | * Listening on http://127.0.0.1:3001
single_cell       | * Listening on ssl://0.0.0.0:3000?cert=/home/app/webapp/config/certs/localhost.crt&key=/home/app/webapp/config/certs/localhost.key&verify_mode=none
single_cell       | Use Ctrl-C to stop
```
Vite dev server running:
```
single_cell_vite  | [5/5] Building fresh packages...
single_cell_vite  | success Saved lockfile.
single_cell_vite  | Done in 12.97s.
single_cell_vite  | 
single_cell_vite  |   vite v2.9.13 dev server running at:
single_cell_vite  | 
single_cell_vite  |   > Local:    http://localhost:3036/vite-dev/
single_cell_vite  |   > Network:  http://172.23.0.2:3036/vite-dev/
single_cell_vite  | 
single_cell_vite  |   ready in 2733ms.
single_cell_vite  | 
```
5. Open the home page in a browser tab and confirm that it loads successfully
6. To confirm HMR is working, open `app/javascript/styles/_global.scss` and on line 10 change `background-color` to `#000`
7. Back in the UI, confirm the background changes to black without needing to reload the page (it make take a few seconds to pick up the change and recompile)
8. In the terminal, press `Ctrl-C` to stop the server, then run `bin/docker-compose-cleanup.sh` to remove the containers and volumes and revert `config/secrets/.source_env.bash` back to its original state

### DIFF
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -1,4 +1,8 @@
 #! /bin/sh
+
+# docker-compose-cleanup.sh
+# stop/remove all containers/volumes created by docker-compose and revert environment config files
+
 echo "### REMOVING CONTAINERS/VOLUMES ###"
 docker-compose -f docker-compose-dev.yaml down
 docker volume prune --force
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -5,6 +5,8 @@
 # More context: https://github.com/broadinstitute/single_cell_portal_core#hybrid-docker-local-development
 
 echo "### REMOVING CONTAINERS/VOLUMES ###"
+# set VITE_FRONTEND_SERVICE_WORKER_CACHE to silence warnings from docker-compose
+VITE_FRONTEND_SERVICE_WORKER_CACHE="$VITE_FRONTEND_SERVICE_WORKER_CACHE" \
 docker-compose -f docker-compose-dev.yaml down
 docker volume prune --force
 rm tmp/pids/*.pid
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -2,6 +2,7 @@
 
 # docker-compose-cleanup.sh
 # stop/remove all containers/volumes created by docker-compose and revert environment config files
+# More context: https://github.com/broadinstitute/single_cell_portal_core#hybrid-docker-local-development
 
 echo "### REMOVING CONTAINERS/VOLUMES ###"
 docker-compose -f docker-compose-dev.yaml down
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -2,9 +2,6 @@
 echo "### REMOVING CONTAINERS/VOLUMES ###"
 docker-compose -f docker-compose-dev.yaml down
 docker volume prune --force
-PID_PATH=tmp/pids/server.pid
-if [[ -f "$PID_PATH" ]]; then
-  rm $PID_PATH
-fi
+rm tmp/pids/*.pid
 echo "### REVERTING ENVIRONMENT ###"
 ./rails_local_setup.rb
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -1,0 +1,10 @@
+#! /bin/sh
+echo "### REMOVING CONTAINERS/VOLUMES ###"
+docker-compose -f docker-compose-dev.yaml down
+docker volume prune --force
+PID_PATH=tmp/pids/server.pid
+if [[ -f "$PID_PATH" ]]; then
+  rm $PID_PATH
+fi
+echo "### REVERTING ENVIRONMENT ###"
+./rails_local_setup.rb
--- a/bin/docker-compose-setup.sh
+++ b/bin/docker-compose-setup.sh
@@ -2,6 +2,7 @@
 
 # docker-compose-setup.sh
 # bring up local development environment via docker-compose
+# More context: https://github.com/broadinstitute/single_cell_portal_core#hybrid-docker-local-development
 
 usage=$(
 cat <<EOF
--- a/bin/docker-compose-setup.sh
+++ b/bin/docker-compose-setup.sh
@@ -8,15 +8,23 @@ usage=$(
 cat <<EOF
 $0 [OPTION]
 -d   run docker-compose in detached mode (default is attatched to terminal STDOUT)
+-c   enable VITE_FRONTEND_SERVICE_WORKER_CACHE (default is disabled)
 -h   print this text
 EOF
 )
 
-DETACHED="no"
-while getopts "dh" OPTION; do
+DETACHED=""
+VITE_FRONTEND_SERVICE_WORKER_CACHE="false"
+while getopts "dch" OPTION; do
 case $OPTION in
   d)
-    DETACHED="yes"
+    echo "### SETTING DETACHED ###"
+    DETACHED="--detach"
+    echo "### PLEASE ALLOW 30s ONCE SERVICES START BEFORE ISSUING REQUESTS ###"
+    ;;
+  c)
+    echo "### ENABLING VITE_FRONTEND_SERVICE_WORKER_CACHE ###"
+    VITE_FRONTEND_SERVICE_WORKER_CACHE="true"
     ;;
   h)
   	echo "$usage"
@@ -33,11 +41,6 @@ done
 echo "### SETTING UP ENVIRONMENT ###"
 ./rails_local_setup.rb --docker-paths
 source config/secrets/.source_env.bash
-if [[ "$DETACHED" = "yes" ]]; then
-  echo "### SERVICES STARTING IN DETACHED MODE ###"
-  docker-compose -f docker-compose-dev.yaml up --detach
-  echo "### please wait 60s before issuing requests to localhost:3000 ###"
-else
-  echo "### STARTING SERVICES ###"
-  docker-compose -f docker-compose-dev.yaml up
-fi
+echo "### STARTING SERVICES ###"
+VITE_FRONTEND_SERVICE_WORKER_CACHE="$VITE_FRONTEND_SERVICE_WORKER_CACHE" \
+docker-compose -f docker-compose-dev.yaml up $DETACHED
--- a/bin/docker-compose-setup.sh
+++ b/bin/docker-compose-setup.sh
@@ -1,6 +1,6 @@
 #! /bin/sh
 echo "### SETTING UP ENVIRONMENT ###"
-./rails_local_setup.rb -d
+./rails_local_setup.rb --docker-paths
 source config/secrets/.source_env.bash
 echo "### STARTING SERVICES ###"
 docker-compose -f docker-compose-dev.yaml up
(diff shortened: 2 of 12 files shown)


ehanna4 (approved):
Functional review - ran through the test outlined and it worked as advertised.

I then switched back to running the rails server locally and that started up fine (proving the clean up worked well)

Are there any changes needed for the dockerized local set up README with these changes?

Really excited about the prospect of using this for local development to bypass any local issues!

bistline:
> Are there any changes needed for the dockerized local set up README with these changes?

Ah great call - I'll add a section to the README about this.  I'm doing some minor refactoring to clean up the compose files so I'll add instructions on how to use everything as well.

ehanna4:
@bistline nice! [cb5ce85](https://github.com/broadinstitute/single_cell_portal_core/pull/1641/commits/cb5ce858cdce3dd1405c65cf72aaf6e2ba5cd07c) <- this looks great

bin/docker-compose-cleanup.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+echo "### REMOVING CONTAINERS/VOLUMES ###"
```
eweitz, line 6: Improve internal discoverability:

```suggestion
# More context: https://github.com/broadinstitute/single_cell_portal_core#hybrid-docker-local-development

echo "### REMOVING CONTAINERS/VOLUMES ###"
```

bin/docker-compose-setup.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/sh
+echo "### SETTING UP ENVIRONMENT ###"
```
eweitz, line 32: Improve internal discoverability:

```suggestion
# More context: https://github.com/broadinstitute/single_cell_portal_core#hybrid-docker-local-development

echo "### SETTING UP ENVIRONMENT ###"
```

docker-compose-dev.yaml
```diff
@@ -0,0 +1,33 @@
...
+    expose:
+      - 27017
+    volumes:
+      - .:/home/app/webapp
+      - type: volume
+        target: /home/app/webapp/node_modules
+  vite:
+    container_name: single_cell_vite
+    image: gcr.io/broad-singlecellportal-staging/single-cell-portal:development
+    entrypoint: ./vite-dev-entrypoint.sh
```
eweitz, line 24: For internal observability:

```suggestion
    entrypoint: ./vite-dev-entrypoint.sh --vite-env-var VITE_DEV_SETUP="docker-compose"
```

vite-dev-entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+
+yarn install
+bin/vite dev
```
eweitz, line 4: For internal observability, something like:

```suggestion
# (Add Bash to parse `--vite-env-var` argument values)

$VITE_ENV_VARS bin/vite dev bin/vite dev
```

With `$VITE_ENV_VARS` evaluated, the command could ultimately execute something like: 
> `VITE_FRONTEND_SERVICE_WORKER_CACHE="true" bin/vite dev` 

or

> `VITE_DEV_SETUP="docker-compose" VITE_FRONTEND_SERVICE_WORKER_CACHE="true" bin/vite dev`

eweitz (approved):
Code looks good!  

I suggest some copyedits, and refinements for internal discoverability and observability.  No blockers.

The observability suggestions would let us systematically answer developer experience (DX) questions like "How does local performance differ between Docker compose and not?" and "Are folks using Docker compose?".  Such insights from Mixpanel analysis that enables could inform future DX optimization.

In addition to the inline suggestions, we'd need small changes in two additional files to enable such observability.

* In `vite.config.ts`, add `'__DEV_MODE__': process.env.VITE_DEV_MODE` after line 12
* In `SCPContextProvider.jsx`, add:

  * Line 7: add `const devMode = (typeof __DEV_MODE__ !== 'undefined') ? __DEV_MODE__ : null`
  
  * Line 20: add `devMode` (and a comma to end of `isServiceWorkerCacheEnabled`, of course)

We have this kind of logging for internal observability with service worker cache usage for local development environments, which complements our standard suite of production Mixpanel / Bard logging.  (Incidentally, the suggestions also help towards enabling local service worker caches for local development in Docker compose.)

None of these matters affect end-users (or even developers, at least not immediately or directly), so I don't consider any of it blocking.  But hopefully the specificity of the suggestions makes observability easy to wrap in, if it seems valuable!

config/vite.json
```diff
@@ -8,7 +8,8 @@
     "autoBuild": true,
     "publicOutputDir": "vite-dev",
     "port": 3036,
-    "host": "127.0.0.1"
+    "host": "127.0.0.1",
+    "devServerConnectTimeout": 1.0
```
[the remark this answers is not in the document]
bistline: A few things - the `500` you see here is a side effect of the [custom error handling](https://github.com/broadinstitute/single_cell_portal_core/pull/1339) we added a while back so we could have better logging/reporting and well-formed error responses in the API.  Really, these are `404`s - the client is making a `GET` for a module that `vite` is not rendering - but then our handling of this emits a secondary error.  This only happens in `development`, and is due to the fact that `ActionController::RoutingError` exceptions happen at the middleware level, not in our code, and Rails has very different handling of errors in `development` than in production-like environments.  But that's not the real problem - `vite` not rendering the module is the proximate cause.

But to your question re: timeout and Docker - yes, the two are independent.  I originally was just attempting to solve the above issue, but then had the thought that perhaps Docker could sidestep it altogether (Ubuntu's file handler limits are much higher by default than MacOS - `1048576` instead of `256`), and then pivoted to this avenue of work.  However, while it was _exceedingly_ rare, I did encounter the issue once while doing this work, so I left time timeout in as it does seem to address that particular problem.  It also doesn't really seem to affect performance in any meaningful way, at least that I can detect.

vite-dev-entrypoint.sh
bistline: Sadly, `getopts` does not support long-form flags, and Mac has [well-known issues](https://stackoverflow.com/questions/402377/using-getopts-to-process-long-and-short-command-line-options) with `getopt`, which is what actually supports those.  That aside, I think this is doable.

bistline: [0ed7c91](https://github.com/broadinstitute/single_cell_portal_core/pull/1641/commits/0ed7c91d3d2fbbf8d18aed7fdfbc1ab1b76322d7)

docker-compose-dev.yaml
bistline: [0ed7c91](https://github.com/broadinstitute/single_cell_portal_core/pull/1641/commits/0ed7c91d3d2fbbf8d18aed7fdfbc1ab1b76322d7)

vite-dev-entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+
+yarn install
+VITE_DEV_MODE="\"docker-compose\"" VITE_FRONTEND_SERVICE_WORKER_CACHE="true" bin/vite dev
```
eweitz, line 4: Blocker: let's omit `VITE_FRONTEND_SERVICE_WORKER_CACHE="true"` from the default settings.

That variable has a very particular use case: when you're doing frontend work that requires frequent full page reloads (because frontend code you're editing isn't handled by HMR) and you want to just reuse SCP API responses (and aren't changing them).  

Such development workflows are common in JavaScript-heavy work, and this variable can save ~4-6 seconds per page load in applicable cases.  But they're not universal, and can cause frustrating cache issues -- e.g. if you cause a change in API response data and don't realize you need to clear your browser's service worker application cache.

```suggestion
VITE_DEV_MODE="\"docker-compose\"" bin/vite dev
```

I suggested a manual workaround for this setting in `vite.config.ts`; it's just a thought.  If you think it'd be better to emphasize a way to enable this non-default setting elsewhere, please do!

eweitz (commented):
Thanks for the swift updates!  I note a trivial blocker.  Code looks good otherwise.

config/vite.json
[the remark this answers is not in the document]
eweitz: Super insightful, thank you Jon!

vite-dev-entrypoint.sh
bistline: Ah, thanks for the clarification - that makes sense.  I didn't do the CLI approach because `vite-dev-entrypoint.sh` is invoked from `docker-compose-dev.yaml`, and it's not possible to pass them in via the CLI.  However I might be able to read it out of the environment - lemme take a stab at that approach.

bistline: [f4e377a](https://github.com/broadinstitute/single_cell_portal_core/pull/1641/commits/f4e377a60cb6589bcb2c71620450e9caed3914b9)

PR #1647, opened 2022-10-20
Enforce redirects after login to correct SCP hostname (SCP-4655)

bistline:
#### BACKGROUND
A recent pentest of staging uncovered a vulnerability where a malicious actor can intercept the OAuth request and replace the `Referer` header, which will cause the redirect after login to proceed to the newly specified host, rather than SCP.

#### CHANGES
This update adds a validation method that will ensure the requested hostname is the same as the current instance.  If it is, it will allow the redirect specified in `request.env['omniauth.orgin']`, which stores the record of the last page the user was on (if present).  If the hostnames do not match, then a `SecurityError` is raised and logged, then the user is redirected back to the home page.

#### MANUAL TESTING
Testing this feature requires installing a plugin to allow you to manipulate request headers in your browser.  For Chrome, I recommend [Modify Header Value](https://mybrowseraddon.com/modify-header-value.html).  Once installed, you can add a rule to modify the `Referer` header to an external domain on requests to `localhost`:
Screen Shot 2022-10-20 at 1 46 09 PM
1. Pull branch and boot as normal
2. Using `Modify Header Value` (or similar plugin), configure your browser to modify the `Referer` header to be `https://www.google.com`
3. Once enabled, open any study page to load the Study Overview
4. Now sign in as normal, and validate that you've been redirected to the home page (instead of the previous page)
5. In `development.log`, note the message that the external hostname was not allowed, and that a `SecurityError` would have been logged to Sentry/Mixpanel:
```
Invalid origin: www.google.com requested, does not match localhost
Suppressing error reporting to Sentry: SecurityError:Invalid origin: www.google.com requested, does not match localhost
Reporting error analytics to mixpanel for (SecurityError) Invalid origin: www.google.com requested, does not match localhost
```

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -63,17 +63,22 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
   # validate omniauth.origin is redirecting to same host
   # will allow redirect if hostnames match, otherwise will return nil to force redirect to home page
+  # malicious redirects are logged as SecurityError in Sentry/Mixpanel
   def set_omniauth_redirect(request, user)
     begin
       hostname = URI.parse(request.env['omniauth.origin']).hostname
       if hostname == ENV['HOSTNAME']
         request.env['omniauth.origin']
       else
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Code looks good!  I also manually verified this in my local SCP.

PR #1635, opened 2022-10-06
Remove AnnData/Seurat file FF and migrate misc files to new type (SCP-4567, SCP-4732)

ehanna4:
TLDR: Remove AnnData/Seurat file upload FF and migrate misc files to their new appropriate type

- Misc file types like `Other` and `Documentation` will no longer accept `'.Rds', '.rds'` or `'.h5', '.h5ad'`

- `AnnData` and `Seurat` file upload sections in the wizard will not be feature flagged anymore and will be GA

- Any file uploaded in the categories` 'Other', 'Documentation',` or ` 'Analysis Output'` that end in `h5ad` or `h5` are migrated to type `AnnData` and those ending in `rds` or `Rdata` will migrate to type `'Seurat'`

To Test:
1. BEFORE pulling this branch
2. Get together an `.rds` and a `.h5ad` files to test with, I used this study in HCA which has an `.rds` and a `.h5ad` file to download https://data.humancellatlas.org/explore/projects/83f5188e-3bf7-4956-9544-cea4f8997756/get-curl-command
3. Start up your local server `./rails_local_setup.rb && source config/secrets/.source_env.bash && rails s`
4. Open up localhost in a browser and sign in
5. Navigate to a study you own and go to the upload wizard
6. Start the rails job uploads by opening another terminal and running `./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails jobs:work`
7. Upload the `.rds` file and a `.h5ad` files to the `misc` file section choosing either 'Other' or 'Documentation' for the file type. 
8. Once that has completed open up another terminal window and start up a rails console with `rails c` 
9. In the console run `StudyFile.where(upload_file_name: /\.(h5ad)$/).pluck(:file_type)` This will print out the file types of any files ending in `.h5ad` and run `StudyFile.where(upload_file_name: /\.(rds)$/).pluck(:file_type)` to get any `.rds` ending files. You should see 'Other' or 'Documentation' at this point
Screen Shot 2022-10-07 at 11 16 23 AM

1. NOW Pull this branch
2. Kill the local server you've got running
3. Open up a new terminal window  and run the DB migration `bin/rails db:migrate`
5. Start up your local server
8. Open up localhost and sign in
10. Navigate to the same study you just uploaded to and go to the wizard
11. Observe that you should see now the `AnnData` and `Seurat` sections in the wizard
[...]

PR #1654, opened 2022-10-26
Add missing migration for .RDS (SCP-4208)

ehanna4:
Missed the migration of '.RDS' files in [<PR here>](https://github.com/broadinstitute/single_cell_portal_core/pull/1635). This adds a follow on migration for this.

To test, follow the instructions in this PR https://github.com/broadinstitute/single_cell_portal_core/pull/1635 but ensure your file extension is `.RDS` (case sensitive)


eweitz (approved):
Looks good!  This might also be worth doing for .RDS -- see suggestions.

eweitz:
Would case-insensitive checks let us simplify these things?  That might be more robust, too.  

If so, maybe that's worth a separate task and a TODO note or two in the code.

ehanna4:
> Would case-insensitive checks let us simplify these things? That might be more robust, too.

Yeah good point, I think like a tech debt ticket for that would be smart, I'll make one

bistline (approved):

PR #1655, opened 2022-10-27
Parse HDF5 in JavaScript: proof of concept

eweitz:
This demonstrates HDF5 being read via JavaScript, which indicates validating AnnData in JS might be feasible.

At this early stage, the code is basically a "Hello, world!" for [jsfive](https://github.com/usnistgov/jsfive).  That library is a JS port of [pyfive](https://github.com/jjhelmus/pyfive), which is a pure-Python and read-only port of [h5py](https://github.com/h5py/h5py), which is how [scanpy](https://github.com/scverse/scanpy) parses HDF5.  I've kept the code in this repo yet unintegrated into the Rails app, which keeps iteration cycles as fast as possible.

Next steps:
* Explore parsing a realistic AnnData / H5AD file
* Assess streamed / chunked reads of HDF5 in JS

To test:
* `cd hdf5-parser`
* `yarn install`
* Run following and confirm output:
```
$ node hdf5-parser.js | tail -n 2
"entry/collection_time" dataset attributes:
{ units: 's', long_name: 'total time detectors were active' }
```


ehanna4 (approved):
Functional Review - worked as expected. Code also looks good for a POC.

Non blocking at all just a thought for the future - `console.log` can be slow, I would think condensing all the logs to one multiline log would be more performant, but I'm not sure it matters here since I figure that might not even stay in here moving forward. Just wanted to mention it

bistline (approved):
Very cool to see this moving forward.  It's nice to see that this package will allow us to see both the structure and content of hdf5-based files.

PR #1656, opened 2022-10-27
Log error with counting trace labels to Sentry and Mixpanel (SCP-4801)

ehanna4:
We uncovered a bug that can occur when switching between annotations and "collapse by" methods after a user has searched a few genes in study. This adds logging to both Sentry and Mixpanel so that we can understand the user impact better and also have visibility if this occurs again in the future after we fix the bug.

To test:
- Pull this branch
- Start up your local host
- Navigate to a study on the portal that is visualizable with at least 3 searchable genes
- Open up dev tools
- Perform a search for 3 genes in the explore tab
- Change the annotation
- Select "Mean" from the "Collapse by" drop down menu
- You should see the error appear in the UI and if you go to the console you will see the error report to Sentry has been suppressed and going over to the Network tab you should find an event to Mixpanel to log the error

Screen Shot 2022-10-27 at 1 32 33 PM
Screen Shot 2022-10-27 at 1 32 22 PM

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -8,7 +8,7 @@ import _find from 'lodash/find'
 import _remove from 'lodash/remove'
 import $ from 'jquery'
 import { getDefaultProperties } from '@databiosphere/bard-client'
-import { logJSFetchExceptionToSentry, logJSFetchErrorToSentry } from '~/lib/sentry-logging'
+import { logJSFetchExceptionToSentry, logJSFetchErrorToSentry, logToSentry } from '~/lib/sentry-logging'
 
 import { getAccessToken } from '~/providers/UserProvider'
 import { getBrandingGroup } from '~/lib/scp-api'
@@ -322,11 +322,13 @@ export function logMenuChange(event) {
 }
 
 /**
- * Log front-end error (e.g. uncaught ReferenceError)
+ * Log front-end error (e.g. uncaught ReferenceError) to both Sentry and Mixpanel
  */
-export function logError(text, details = {}) {
-  const props = { text, ...details }
+export function logError(text, error = {}) {
+  const props = { text, error }
   log('error', props)
+
+  logToSentry(error)
 }
 
 /**
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -324,8 +324,8 @@ export function logMenuChange(event) {
 /**
  * Log front-end error (e.g. uncaught ReferenceError)
  */
-export function logError(text) {
-  const props = { text }
+export function logError(text, details = {}) {
+  const props = { text, ...details }
   log('error', props)
 }
 
--- a/app/javascript/lib/plot.js
+++ b/app/javascript/lib/plot.js
@@ -1,6 +1,5 @@
 import { UNSPECIFIED_ANNOTATION_NAME } from '~/lib/cluster-utils'
 import { log, logError } from '~/lib/metrics-api'
-import {logToSentry} from '~/lib/sentry-logging'
 
 // Default plot colors, combining ColorBrewer sets 1-3 with tweaks to yellows.
 const colorBrewerList = [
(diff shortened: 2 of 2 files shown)


app/javascript/lib/plot.js
```diff
@@ -295,11 +296,16 @@ PlotUtils.getLegendSortedLabels = function(countsByLabel) {
...
 */
 function countValues(array) {
-  return array.reduce((acc, curr) => {
-    acc[curr] ||= 0
-    acc[curr] += 1
-    return acc
-  }, {})
+  try {
+    return array.reduce((acc, curr) => {
+      acc[curr] ||= 0
```
bistline, line 301: Ooh I don't think that I knew Javascript had this notation for conditional assignment!  I definitely appreciate this in Ruby.

app/javascript/lib/metrics-api.js
```diff
@@ -324,8 +324,8 @@ export function logMenuChange(event) {
 /**
  * Log front-end error (e.g. uncaught ReferenceError)
  */
-export function logError(text) {
-  const props = { text }
+export function logError(text, details = {}) {
+  const props = { text, ...details }
```
eweitz, line 328: Per other comment:

```suggestion
export function logError(text, error = {}) {
  logToSentry(error)
  const details = error?.message || error;
  const props = { text, ...error }
```

bistline (approved):
I can't get the error to trigger (which doesn't surprise me given the differences we've seen in our dev envs) but regardless this change looks good.

eweitz (approved):
Code looks good overall!

I suggest a non-blocking observability improvement, to reuse a logging pattern we recently introduced on the Rails backend.

Also, while this fixes the particular observability issue that prompted this work, the general problem remains.  It seems that a broad kind of error like gets its stack trace swallowed, and is not logged to Sentry or Mixpanel.  I just updated the related ticket in SCP-4799 to capture that.  Once that's resolved, this particular try / catch might be mostly superfluous.

app/javascript/lib/metrics-api.js
ehanna4: oh okay cool yeah if we want to log all errors to Sentry as well this is way DRY-er to just add it here. For some reason I had thought we didn't necessarily want to log all of the errors we sent to Mixpanel to Sentry as well but I can't recall exactly why I thought that was the case. 

Do you anticipate any issue with over-logging and/or running up our quota by adding this more generally here @eweitz ?

eweitz: Ah, right, so we ideally only log application errors to Sentry, whereas user errors are also logged to Mixpanel.  An application error like this would be logged to Sentry, but e.g. a user error in validation would not be logged to Sentry.

`logError` is only called [in componentDidCatch in ErrorBoundary.jsx](https://github.com/broadinstitute/single_cell_portal_core/search?q=logerror), which also calls [logJSFetchErrorToSentry](https://github.com/broadinstitute/single_cell_portal_core/blob/a885ab2bf093902d98ec47837912ab66c449214c/app/javascript/lib/sentry-logging.js).  The latter call is _not throttled_, so all errors that pass through that function get logged to Mixpanel (via `logError`) and Sentry.

So I see a few issues, from trivial to significant:
1.  `logJSFetchErrorToSentry` seems like a misnomer, as it presumably gets (or could be) called for some errors that are not `fetch` errors.
2. `componentDidCatch` should probably just call something like your new `logError` function.  Routing JS errors to Mixpanel and Sentry seems best to make React-agnostic.
3. Why wasn't this particular "counting trace labels" error handled by [componentDidCatch](https://reactjs.org/docs/react-component.html#componentdidcatch)?  Even though the error happened in `plot.js`, this `plot.js` was used in a React component.  Naively, I'd expect `componentDidCatch` to catch errors thrown in lower-level code that's imported into a React component.

The notes above are loose observations and weak opinions.  You have more context here, so I defer to you!

ehanna4: hmm okay after more deeply digging, it looks like there are two different error handling components, and I think this one: `useErrorMessage()` might be swallowing the error before `ErrorBoundary` - to point 3. And so for point 2 `componentDidCatch` is not being triggered with this error then so adding logging there won't fix this particular case (though might be a good option in the future).

For this particular case I think leaving it as this try/catch solution is sufficient, but noting that indeed  [SCP-4799](https://broadworkbench.atlassian.net/browse/SCP-4799) could make this superfluous which I think would be okay.

PR #1657, opened 2022-10-28
Restore split/merge functionality (SCP-4728)

ehanna4:
To see the code changes for functionality check out the first commit. The second commit is only renaming `splitArrayLabels` and `setSplitArrayLabels` to `isSplitArrayLabels` and `updateIsSplitArrayLabels`.

This restores the functionality of merge/split array labels. 

https://user-images.githubusercontent.com/54322292/199084080-1e620263-ff2a-4623-91cf-700305d1cde8.mov
*Edit - this clip shows persistence of hidden traces between toggles but please note this was the first iteration of this PR and that persistence between split/merge will NOT be supported at this time.

To test
- You need a study that has split labels. I used the large study from staging <study> copied to my local and because of the bug with overrides hiding legend search and showing the split/merge button this worked haha. It might be simpler to use a smaller study so you don't need to worry about that at all but I don't know of one with the splits.
- Go to the explore tab for the study
- click merge or split
- then click it again to see that the legend traces do update appropriately

### DIFF
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -230,7 +230,7 @@ module Api
           custom_colors = cluster.study_file.cluster_file_info&.custom_colors_as_hash || {}
           custom_annotation_colors = custom_colors[annotation[:name]] || {}
           annotation_split_defaults = cluster.study_file.cluster_file_info&.annotation_split_defaults_as_hash || {}
-          split_label_arrays = annotation_split_defaults[annotation[:name]] || false
+          is_split_label_arrays = annotation_split_defaults[annotation[:name]] || false
 
           {
             data: plot_data,
@@ -255,7 +255,7 @@ module Api
             consensus: consensus,
             customColors: custom_annotation_colors,
             clusterFileId: cluster.study_file_id.to_s,
-            isSplitLabelArrays: split_label_arrays,
+            isSplitLabelArrays: is_split_label_arrays,
             externalLink: {
               url: cluster.study_file[:external_link_url],
               title: cluster.study_file[:external_link_title],
--- a/app/controllers/api/v1/visualization/clusters_controller.rb
+++ b/app/controllers/api/v1/visualization/clusters_controller.rb
@@ -255,7 +255,7 @@ module Api
             consensus: consensus,
             customColors: custom_annotation_colors,
             clusterFileId: cluster.study_file_id.to_s,
-            splitLabelArrays: split_label_arrays,
+            isSplitLabelArrays: split_label_arrays,
             externalLink: {
               url: cluster.study_file[:external_link_url],
               title: cluster.study_file[:external_link_title],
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/visualization/clusters_controller.rb
```diff
@@ -255,7 +255,7 @@ def self.get_cluster_viz_data(study, cluster, url_params)
             consensus: consensus,
             customColors: custom_annotation_colors,
             clusterFileId: cluster.study_file_id.to_s,
-            splitLabelArrays: split_label_arrays,
+            isSplitLabelArrays: is_split_label_arrays,
```
eweitz, line 258: Nice name refinement!

eweitz (approved):
Code and demo functionality look good!

Non-blocking, but from the demo video, split/merge looks slow for very many labels.  It seems to take ~8 seconds between click and visual response.  (You can show clicks in macOS screen recordings by clicking ["Options" -> "Show Mouse Clicks"](https://www.google.com/search?q=macos+screen+recording+options+mouse+click) -- this clarifies exactly where and _when_ a click occurred.)

Adding a spinner when there are > 200 labels would likely help.  Ideally, of course, it'd be fast enough to not need a spinner.  If we pursue that work, it strikes me as out-of-scope here and worth a separate ticket.

Nice work enabling this use case!

ehanna4:
So a few updates:
Most importantly the merge/split toggling with correct list and colors is fixed:
https://user-images.githubusercontent.com/54322292/199563699-cc14bd65-8fba-4453-aa33-186c7b666267.mov

However, I found this PR purposely disabling the persistance of hidden traces between toggles: https://github.com/broadinstitute/single_cell_portal_core/pull/1478

Essentially in really large cases if you click hideAll then toggle it can break down. So I've put back the clearing of the hidden traces. And I'd like to propose removing the "clearing hiddenTraces" lines in this ticket: https://broadworkbench.atlassian.net/browse/SCP-4623?atlOrigin=eyJpIjoiYzIzMTRkYzI3Nzg3NDliZTg5NzUwOWU4OTFkODNmMDciLCJwIjoiaiJ9 once a more general solution is made for handling large numbers of hiddenTraces, since there are more overarching issues with this.

If folks are in agreeance with this, I can add comment todos in the code at the lines that would eventually be removed, and update the ticket accordingly.

ehanna4:
Oh also to this valid point:

> Adding a spinner when there are > 200 labels would likely help. Ideally, of course, it'd be fast enough to not need a spinner. If we pursue that work, it strikes me as out-of-scope here and worth a separate ticket.

we have a ticket here to add spinners for a few different slow button clicks for 200+ label plots: https://broadworkbench.atlassian.net/browse/SCP-4778?atlOrigin=eyJpIjoiZDI0OTMxMDViYTNkNDUwNGI1ZjA0ZjA0NDdiNjFhMDAiLCJwIjoiaiJ9

@eweitz

bistline (approved):
Works as I would expect now - great job fixing the bug.  I also agree with the plan re: trace toggle/persistence.

PR #1658, opened 2022-11-01
Refine retry logic for Google Cloud Storage API calls (SCP-4730)

bistline:
#### BACKGROUND & CHANGES
This update refines the logic on how/when API calls to Google Cloud Storage are retried.  Previously, `4xx` errors would be retried with staggered backoffs when none of these calls had a chance of ever succeeding based on parameters/state.  This leads to situations where the UI can hang for 3-4 minutes with no obvious indication of what is happening.  Now, only error codes in the `502-4` range are ever retried, and all other errors will exit immediately and continue through normal error handling/reporting.

This update also adds exponential backoff and jitter via the `exponential-backoff` gem to better handle potential load/thundering herd  issues.  Lastly, it fixes a newly broken test in `DataRepoClientTest` due to an upstream change.

#### MANUAL TESTING
The easiest way to test this is via the Rails console:
1. Pull branch and enter a Rails console session
2. Run the following snippet to simulate a `4xx` level error and confirm the error returns immediately:
```
ApplicationController.firecloud_client.execute_gcloud_method(:get_workspace_file, 0, 'foo', 'bar.txt')
```
3. In `development.log`, note the following error message only appears once, and that the message denotes `Retry count exceeded`:
```
Suppressing error reporting to Sentry: Google::Cloud::PermissionDeniedError:forbidden: default-service-account@broad-singlecellportal-bistlin.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket., context: {:method_name=>:get_workspace_file, :retry_count=>1, :params=>["foo", "bar.txt"]}
Retry count exceeded calling get_workspace_file with foo, bar.txt: forbidden: default-service-account@broad-singlecellportal-bistlin.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket.
```

### DIFF
--- a/app/models/fire_cloud_client.rb
+++ b/app/models/fire_cloud_client.rb
@@ -1297,8 +1297,7 @@ class FireCloudClient
   # * *return*
   #   - Object depends on method, can be one of the following: +Google::Cloud::Storage::Bucket+, +Google::Cloud::Storage::File+,
   #     +Google::Cloud::Storage::FileList+, +Boolean+, +File+, or +String+
-
-  def execute_gcloud_method(method_name, retry_count=0, *params)
+  def execute_gcloud_method(method_name, retry_count = 0, *params)
     begin
       self.send(method_name, *params)
     rescue => e
@@ -1316,7 +1315,7 @@ class FireCloudClient
                                                                  retry_count: current_retry, params: params})
         end
         Rails.logger.info "Retry count exceeded calling #{method_name} with #{params.join(', ')}: #{e.message}"
-        raise RuntimeError, e.message
+        raise e.message # raise implicitly creates RuntimeError
       end
     end
   end
--- a/app/models/fire_cloud_client.rb
+++ b/app/models/fire_cloud_client.rb
@@ -1302,21 +1302,12 @@ class FireCloudClient
     begin
       self.send(method_name, *params)
     rescue => e
+      status_code = extract_status_code(e)
       current_retry = retry_count + 1
-      Rails.logger.info "error calling #{method_name} with #{params.join(', ')}; #{e.message} -- attempt ##{current_retry}"
-      retry_time = retry_count * ApiHelpers::RETRY_INTERVAL
-      sleep(retry_time) unless RETRY_BACKOFF_BLACKLIST.include?(method_name)
-      # only retry if status code indicates a possible temporary error, and we are under the retry limit and
-      # not calling a method that is blocked from retries.  In case of a NoMethodError or RuntimeError, use 500 as the
-      # status code since these are unrecoverable errors
-      if e.respond_to?(:code)
-        status_code = e.code
-      elsif e.is_a?(NoMethodError) || e.is_a?(RuntimeError)
-        status_code = 500
-      else
-        status_code = nil
-      end
       if should_retry?(status_code) && retry_count < ApiHelpers::MAX_RETRY_COUNT && !ERROR_IGNORE_LIST.include?(method_name)
+        Rails.logger.info "error calling #{method_name} with #{params.join(', ')}; #{e.message} -- attempt ##{current_retry}"
+        retry_time = retry_count * ApiHelpers::RETRY_INTERVAL
+        sleep(retry_time) unless RETRY_BACKOFF_BLACKLIST.include?(method_name)
         execute_gcloud_method(method_name, current_retry, *params)
       else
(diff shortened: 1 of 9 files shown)


bistline:
Build failure seems unrelated, though persistent - I may simply comment out that test for now since all DR integration is not used.

app/models/fire_cloud_client.rb
```diff
@@ -1297,26 +1297,16 @@ def get_pet_service_account_key(project_name)
...
-        status_code = e.code
-      elsif e.is_a?(NoMethodError) || e.is_a?(RuntimeError)
-        status_code = 500
-      else
-        status_code = nil
-      end
       if should_retry?(status_code) && retry_count < ApiHelpers::MAX_RETRY_COUNT && !ERROR_IGNORE_LIST.include?(method_name)
+        Rails.logger.info "error calling #{method_name} with #{params.join(', ')}; #{e.message} -- attempt ##{current_retry}"
+        retry_time = retry_count * ApiHelpers::RETRY_INTERVAL
+        sleep(retry_time) unless RETRY_BACKOFF_BLACKLIST.include?(method_name)
```
eweitz, line 1309:
```suggestion
        sleep(retry_time) unless RETRY_BACKOFF_DENYLIST.include?(method_name)
```

app/models/fire_cloud_client.rb
```diff
@@ -1297,26 +1297,16 @@ def get_pet_service_account_key(project_name)
...
-      if e.respond_to?(:code)
-        status_code = e.code
-      elsif e.is_a?(NoMethodError) || e.is_a?(RuntimeError)
-        status_code = 500
-      else
-        status_code = nil
-      end
       if should_retry?(status_code) && retry_count < ApiHelpers::MAX_RETRY_COUNT && !ERROR_IGNORE_LIST.include?(method_name)
+        Rails.logger.info "error calling #{method_name} with #{params.join(', ')}; #{e.message} -- attempt ##{current_retry}"
+        retry_time = retry_count * ApiHelpers::RETRY_INTERVAL
```
eweitz, line 1308: Could this code notably affect load on SCP or Terra servers?

If so, a more robust `retry_time` might be worthwhile.  The canonical solution is "exponential backoff with jitter", which we've [implemented in Python](https://github.com/broadinstitute/scp-ingest-pipeline/blob/71fc6587bb3fcf19a861d7b6b81ed4693b907a55/ingest/mongo_connection.py#L57) for reference.  That better prevents [thundering herds](https://en.wikipedia.org/wiki/Thundering_herd_problem).

eweitz (approved):
Code looks good overall!  I suggest two non-blocking refinements.

app/models/fire_cloud_client.rb
bistline: Agreed - that approach is better, but the existing retry logic here was written prior to that effort.  This update was only handling the logic around invoking it.  I can take a stab at adding jitter to the retry.

bistline: And this is why I love Ruby: https://github.com/pawelpacana/exponential-backoff

ehanna4 (approved):
Functional review - works as advertised 👍

How will this appear to a user in the UI? I assume this just will fail fast for the cases where the files are no longer accessible, rather than appearing to load for forever. Do we have followup tickets to address the open questions from this ticket about letting the user know they can't download files?

bistline:
> Functional review - works as advertised 👍
> 
> How will this appear to a user in the UI? I assume this just will fail fast for the cases where the files are no longer accessible, rather than appearing to load for forever. Do we have followup tickets to address the open questions from this ticket about letting the user know they can't download files?

Yes, this will now fail fast in the correct scenarios rather than pending for minutes.  In the case that prompted this work (workspace was unavailable due to billing lapse), the request will return and show a modal with an appropriate error message (`"We were unable to download the file (filename) due to: (error message)"`), and then the boilerplate "contact us if this persists" with our support email.  Also, if the case is that the workspace is there and the file is missing, then we tell the user the file is missing and also email the study owner to let them know that we couldn't find the file.

app/models/fire_cloud_client.rb
bistline: aebcde0

app/models/fire_cloud_client.rb
bistline: aebcde0

PR #1660, opened 2022-11-02
Automate submitting differential expression jobs on successful parse jobs (SCP-4716, SCP-4593)

bistline:
#### BACKGROUND
Currently, finding candidate studies for differential expression and generating results is a completely manual process.  Admins must find potential studies that meet the criteria, then manually launch a job to generate results for eligible annotations.  This limits the reach of the differential expression feature as it requires direct intervention on our part to extend it to new studies.

#### CHANGES
This update adds automation such that on a successful clustering/metadata/expression parse job, the portal will check to see if the study is eligible for running differential expression.  The study must meet 4 criteria:
1. Public
2. Has clustering/metadata
3. Has raw count data
4. Has visualizable group-based annotations

If all 4 are true, then a job for each eligible cluster/annotation combo will be automatically submitted.  This automation will also skip any annotation/clustering combos that already have results to reduce overall load and associated costs.  The success/failure emails of these jobs also will be delivered directly to admins instead of end users, as they will not be aware that these jobs are being submitted, and could potentially be confused by those notifications.

Note: this does not automatically backfill data on studies that already exist.  These will still have to be handled in a manual fashion, though there now is the convenience method of `DifferentialExpressionService.study_eligible?` for checking if a study qualifies.

Other changes included but not part of the DE work:
* Fixed an issue when using `docker-compose` locally where emails are not delivered due to port `587` not being bound to the host, and the `DeveloperMailInterceptor` not working since `git config user.email` does not register from inside Docker.  Developers can now use the "QA Dev Email" admin configuration to ensure email delivery to their Broad email from inside Docker locally.
* Removed an extraneous tab character that caused `raw1_human_5k_cells_80_genes.raw_dense.txt` to fail validation/ingest.
* Changed `YAML.load` to `YAML.unsafe_load` in `DelayedJobAccessor` as per SCP-4593 due to persistent errors when testing this work.  Since this code cannot be called anywhere other than Rails console, there is no security issue with calling `unsafe_load`.

#### MANUAL TESTING
The simplest way to test this is to use the `SyntheticStudyPopulator` and ingest the "Human raw counts" study.  Even if you already have this study locally, the populator will remove and re-ingest all files, thus triggering the automation. 
1. Boot as normal, and start DelayedJob (using `bin/docker-compose-setup.sh` will automate all of this if you aren't already doing so)
2. In a separate terminal, open a Rails console session and run the following (you do not need to attach to the Docker instance to do so, though that will also work):
```
SyntheticStudyPopulator.populate('human_raw_counts')
```
3. In `development.log`, you should see numerous instances of PAPI jobs being launched to ingest files
4. In 2-3 minutes, as files begin to finish parsing, you should start to see messages denoting that differential expression jobs are being launched.  The exact number of jobs will vary on the order in which files finish parsing, but there can be a maximum of 4.  In the example below, the 2D cluster file failed to ingest due to a transient MongoDB error, so that cluster was skipped.  This demonstrates both the automation and error handling for differential expression jobs:
```
SCP79 is eligible for differential expression, launching available jobs
SCP79 has annotations eligible for DE; validating inputs
Checking DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_2D.txt (CLUSTER--group--study)
  Skipping DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_2D.txt (CLUSTER--group--study) due to: cannot find cluster for raw1_human_5k_cells_80_genes.cluster_2D.txt
Checking DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_2D.txt (SUBCLUSTER--group--study)
  Skipping DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_2D.txt (SUBCLUSTER--group--study) due to: cannot find cluster for raw1_human_5k_cells_80_genes.cluster_2D.txt
Checking DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_3D.txt (CLUSTER--group--study)
==> DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_3D.txt (CLUSTER--group--study) successfully launched
Checking DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_3D.txt (SUBCLUSTER--group--study)
Remote found for raw1_human_5k_cells_80_genes.cluster_3D.txt:6362d0ac94ec8f12627bf512, launching Ingest job
==> DE job for SCP79: raw1_human_5k_cells_80_genes.cluster_3D.txt (SUBCLUSTER--group--study) successfully launched
SCP79 yielded 2 differential expression jobs; 2 skipped
```
5.  Once the DE jobs finish, you should be able to navigate to the Explore tab and view them.  This is a screenshot of the 3D cluster viewing the `SUBCLUSTER` annotation:
Screen Shot 2022-11-02 at 4 31 18 PM


app/lib/differential_expression_service.rb
```diff
@@ -185,6 +153,93 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
     end
   end
 
+  # find all eligible annotations for DE for a given study
+  #
+  # * *params*
+  #   - +study+        (Study) => Associated study object
+  #   - +machine_type+ (String) => Override default VM machine type
+  #   - +dry_run+      (Boolean) => Indication of whether or not this is a pre-flight check
+  #   - +skip_existing+   (Boolean) => Skip annotations that already have DE results
```
eweitz, line 162: I wonder if there's a good way to DRY up this sort of copy-and-paste docstring.  

Various motivations and ideas to do so are discussed at length over in e.g. https://github.com/matplotlib/matplotlib/issues/9414.  Alas, it seems no solutions were found (and it's a different programming language anyways), so just consider this food for thought.

app/lib/differential_expression_service.rb
```diff
@@ -185,6 +153,93 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
...
+  #
+  # * *returns*
+  #   - (Array<Hash>) => Array of annotation objects available for DE
+  def self.find_eligible_annotations(study, machine_type: nil, dry_run: nil, skip_existing: false)
+    annotations = []
+    metadata = study.cell_metadata.where(annotation_type: 'group').select(&:can_visualize?)
+    annotations += metadata.map do |meta|
+      { annotation_name: meta.name, annotation_scope: 'study', machine_type: machine_type, dry_run: dry_run }
```
eweitz, line 170:
```suggestion
      { annotation_name: meta.name, annotation_scope: 'study', machine_type:, dry_run: }
```

app/lib/differential_expression_service.rb
```diff
@@ -185,6 +153,93 @@ def self.run_differential_expression_job(cluster_file, study, user, annotation_n
...
+      end
+      cell_annotations += cell_annots
+    end
+    annotations += cell_annotations.map do |annot|
+      {
+        annotation_name: annot[:name],
+        annotation_scope: 'cluster',
+        cluster_file_id: annot[:cluster_file_id],
+        machine_type: machine_type,
+        dry_run: dry_run
```
eweitz, line 189: It's not a big deal in terms of maintainability, but I can't help but feel we should go beyond the trivial refactor I suggest below.  

It seems we're simply tacking on the same `machine_type` and `dry_run` to all of these annotation objects.  I suspect we could add them more simply / elegantly downstream of this method call, which would reduce visual noise and improve maintainability in this method.

```suggestion
        machine_type:,
        dry_run:
```

eweitz (approved):
Code looks good!  I suggest some non-blocking maintainability refinements.

We'll also need a way to trigger DE jobs when a study goes from private to public, right?

app/lib/differential_expression_service.rb
bistline: Yeah, that would be nice, but I think there's probably no good solution that will work with IDEs universally.

app/lib/differential_expression_service.rb
bistline: See above - we do not allow mixed Hash literal styles in our linting.  I personally think this is a good rule, but I'm open to changing it if there's a salient case for it.  To me the mixed styles are hard to read and somewhat obtuse.

bistline: See above for the linting comment.

As per refactoring: you're correct in that they're being passed all the way through because of the logic flow, and we need to get them at the final `:run_differential_expression_job`.   Let me take a pass at seeing if we can omit them until the very end to reduce the recapitulation throughout the stack.

bistline: [d56bd23](https://github.com/broadinstitute/single_cell_portal_core/pull/1660/commits/d56bd23878ca4ea5107662706cfdc4991eed56c0)

app/lib/differential_expression_service.rb
eweitz: Ah, thanks for noting.  It seems that rule was [recently added](https://github.com/broadinstitute/single_cell_portal_core/commit/1b98cc62f24e10a71cc20eeb7e61c632fff735f7) and I overlooked this implication.

Could you elaborate on the rationale for that rule?  Mixing reduces visual noise and is quite successful throughout our JavaScript.  So allowing it in Ruby seems like it'd make our code easier to read, and facilitate jumping across languages.  But maybe I'm missing a big downside.

bistline:
> We'll also need a way to trigger DE jobs when a study goes from private to public, right?

Ah yes!  Good catch.  Let me figure out the best place to hook that in.

eweitz:
Another likely rarer but foreseeable edge case:

* Study has computed DE results
* User deletes a file that makes study DE-eligible
* DE results should no longer be shown, and/or be deleted

That strikes me as likely unhandled right now and probably out-of-scope here, but worth at least noting for consideration.

bistline:
> Another likely rarer but foreseeable edge case:
> 
> * Study has computed DE results
> * User deletes a file that makes study DE-eligible
> * DE results should no longer be shown, and/or be deleted
> 
> That strikes me as likely unhandled right now and probably out-of-scope here, but worth at least noting for consideration.

I believe this is covered already.  We have automation in `DeleteQueueJob` to remove any DE results on file deletion, where applicable.  See https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/delete_queue_job.rb#L185 for that code.

That handles removing invalid results.  As to the study no longer being eligible, the only thing that wouldn't be covered by that case is if the study becomes private again.  I don't think it's worth deleting results in that case, because the study could just as easily become public again, and we'd have to re-run everything at that point.

bistline:
@eweitz [a077a8c](https://github.com/broadinstitute/single_cell_portal_core/pull/1660/commits/a077a8cc08a11cd282039a4187da5d171001c3a9)

ehanna4 (approved):
Functional Review - had to run on a different study as Human Raw Counts is not convention compliant (going to chronicle that away as a tech debt thing to address in a future dev choice lol) but got it to run on two other synthetic studies and saw the log outputs as advertised. Very exciting stuff!

PR #1662, opened 2022-11-04
Update metrics_service log to handle nil user (SCP-4797)

ehanna4:
This prevents the errors we are seeing in Sentry related to an unregistered user trying to search 50+ genes on a study. Previously the function MetricsService.log() did not handle `nil` users appropriately. This update utilizes the pattern set in MetricsService.report_error() to handle both logged in and not logged in users. Also this updates the function `get_default_headers()` to conditionally handle not signed in users as well.

Note - It seemed like the way `get_default_headers()` was designed was to sometimes produce Authentication headers that were blank, seemingly for ease of testing? I'm not sure, I couldn't sort out why we'd want to do that, so I updated it to only add real Auth bearer token headers and not add Auth headers if the user is not registered rather than adding and removing as was previously done and updated the tests to handle actual Auth bearer tokens. But please let me know if I'm missing something there.

To test:
- Pull this branch
- Boot up your local server
- Navigate to [swagger](https://localhost:3000/single_cell/api/v1#/Visualization/study_visualization_expression_show)
- Add a study accession for a public study in your local that has 50+ genes you can search
- Set the `data_type` to `heatmap` 
- Add 50+ genes to the gene search input
- Execute
- You should see no 500 error in the results on the Swagger page, you can also further confirm this by opening up `development.log` 
- To see the broken behavior do these same steps using the `development` branch and you will see an error like:

> registeredForTerra: user.registered_for_firecloud,
> single_cell       |                               ^^^^^^^^^^^^^^^^^^^^^^^^^ excluded from capture: DSN not set
> single_cell       |   
> single_cell       | NoMethodError (undefined method `registered_for_firecloud' for nil:NilClass

Before:
Screen Shot 2022-11-07 at 3 22 14 PM

After:
Screen Shot 2022-11-08 at 10 02 15 AM

### DIFF
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -41,7 +41,7 @@ class MetricsService
         'Authorization' => "Bearer #{access_token.dig(:access_token)}",
       })
     end
-    return headers
+    headers
   end
 
   # Merges unauth’d and auth’d user identities in Mixpanel via Bard
@@ -93,7 +93,7 @@ class MetricsService
   # @param {String} name Name of the event
   # @param {Hash} props Properties associated with the event
   # @param {User} user User model object
-  def self.log(name, props={}, user=nil)
+  def self.log(name, props={}, user=nil, request: nil)
     Rails.logger.info "#{Time.zone.now}: Logging analytics to Mixpanel for event name: #{name}"
 
     props.merge!({
@@ -109,7 +109,8 @@ class MetricsService
     if user.present? && user.registered_for_firecloud
       props.merge!({ authenticated: true, registeredForTerra: user.registered_for_firecloud })
     else
-      props.merge!({ authenticated: false, distinct_id: user.get_metrics_uuid, registeredForTerra: false })
+      distinct_id = user&.get_metrics_uuid || request&.cookies['user_id']
+      props.merge!({ authenticated: false, distinct_id: distinct_id, registeredForTerra: false })
     end
 
     post_body = {'event' => name, 'properties' => props}.to_json
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -108,7 +108,7 @@ class MetricsService
     if user.present? && user.registered_for_firecloud
       props.merge!({ authenticated: true, registeredForTerra: user.registered_for_firecloud })
     else
-      props.merge!({ authenticated: user.present?, distinct_id: request.cookies['user_id'] })
+      props.merge!({ authenticated: user.present?, distinct_id: user.get_metrics_uuid })
     end
 
     post_body = {'event' => name, 'properties' => props}.to_json
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -34,11 +34,13 @@ class MetricsService
   end
 
   def self.get_default_headers(user)
-    access_token = user.token_for_api_call
-    {
-      'Authorization' => "Bearer #{access_token.present? ? access_token.dig('access_token') : nil }",
-      'Content-Type' => 'application/json'
-    }
+    headers = { 'Content-Type' => 'application/json'}
+    if user.present?
+      access_token = user.token_for_api_call
+      headers.merge!({
+        'Authorization' => "Bearer #{access_token.present? ? access_token.dig('access_token') : nil }",
+      })
+    end
   end
 
   # Merges unauth’d and auth’d user identities in Mixpanel via Bard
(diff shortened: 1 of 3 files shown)


app/lib/metrics_service.rb
```diff
@@ -90,28 +93,23 @@ def self.merge_identities_in_mixpanel(user, cookies={})
...
-    user_id = user.get_metrics_uuid
-
-    if access_token.nil? || !user.registered_for_firecloud
-      # User is unauthenticated / unregistered / anonymous
-      props['distinct_id'] = user_id
-      headers.delete('Authorization')
-      props['authenticated'] = false
+    # configure properties/headers depending on user presence
+    # only pass user token if user is registered for Terra to avoid 4xx/5xx errors
+    if user.present? && user.registered_for_firecloud
```
bistline, line 109: Blocker: I'm still seeing an error that is preventing reporting to Bard - if I'm not signed in, `user.get_metrics_uuid` throws a `NoMethodError` because there's no user present.

In thinking about this, there's definitely more context that was missing from the ticket.  As I understand it, the way Bard functions in the background is that it talks to SAM to set user identities, and it does so one of two ways: via the `Authorization: Bearer` token, or the `distinct_id` UUID param.  The former is only present if the user is signed in, but the latter is always accessible as a cookie in the request.  If we don't pass either value, Bard will respond `400` as it cannot determine the identity of the user.  Previously, `MetricsService.log` was only ever called in places where we knew we had an authenticated user, but the `too-many-genes` event breaks that paradigm.

If we look at `report_error`, it includes the request object as a parameter so that it can access request cookies (among other things).  I think we need to take this approach to ensure we can always report events - that will ensure we always have a useable credential.  Also - if we refactor the method signature like this:

```
def self.log(name, props={}, user=nil, request: nil)
```

Then the only place we need to update to pass in the request is this case - the `too-many-genes` event.  Everywhere else `MetricsService.log` is called, there is a user present.  Then we can rewrite this block as:

```
if user.present? && user.registered_for_firecloud
  props.merge!({ authenticated: true, registeredForTerra: user.registered_for_firecloud })
else
  distinct_id = user&.get_metrics_uuid || request&.cookies['user_id']
  props.merge!({ authenticated: false, distinct_id: distinct_id, registeredForTerra: false })
end
```

app/lib/metrics_service.rb
```diff
@@ -34,11 +34,14 @@ def self.post_to_bard(params, user)
...
-      'Content-Type' => 'application/json'
-    }
+    headers = { 'Content-Type' => 'application/json'}
+    if user.present? && user.registered_for_firecloud && user.token_for_api_call.present?
+      access_token = user.token_for_api_call
+      headers.merge!({
+        'Authorization' => "Bearer #{access_token.dig(:access_token)}",
+      })
+    end
+    return headers
```
bistline, line 44: Non-blocking: Ruby has implicit returns of the last object operated on, so unless you're halting execution you don't need to say `return headers` at the end.  You can just say `headers`.

bistline (changes requested):
See comment in `MetricsService.log` for blocker - I'm still seeing an error, and the case is more complicated than I think we originally thought.

app/lib/metrics_service.rb
ehanna4: Shoot I shouldn't have missed that, thanks for pointing this out, fixed per your suggestion: [89a0295](https://github.com/broadinstitute/single_cell_portal_core/pull/1662/commits/89a0295cc6dbc5c3b592372e872da78c353ca584)

bistline: Ah I also neglected to mention that the [`search-too-many-genes`](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/controllers/api/v1/visualization/expression_controller.rb#L172) event also needs to have the request passed in:

```
MetricsService.log('search-too-many-genes', { numGenes: num_genes }, current_api_user, request:)
```

bistline (approved):
Approving modulo the one new comment - `Api::V1::Visualization::ExpressionController#check_gene_limit` needs to add the `request` object to the parameter list for `MetricsService.log`.

app/lib/metrics_service.rb
ehanna4: geeze nope you totally did, I read too quickly sorry thanks for catching that.

app/lib/metrics_service.rb
```diff
@@ -90,28 +93,24 @@ def self.merge_identities_in_mixpanel(user, cookies={})
...
-      headers.delete('Authorization')
-      props['authenticated'] = false
+    # configure properties/headers depending on user presence
+    # only pass user token if user is registered for Terra to avoid 4xx/5xx errors
+    if user.present? && user.registered_for_firecloud
+      props.merge!({ authenticated: true, registeredForTerra: user.registered_for_firecloud })
     else
-      props['authenticated'] = true
+      distinct_id = user&.get_metrics_uuid || request&.cookies['user_id']
+      props.merge!({ authenticated: false, distinct_id: distinct_id, registeredForTerra: false })
```
eweitz, line 113: More succinctly, and easier to compare to the similar command on line 110:

```suggestion
      props.merge!({ authenticated: false, registeredForTerra: false, distinct_id: })
```

eweitz (approved):
Code looks good!  I suggest a non-blocking maintainability refinement.

Also, I really like how you _lead the pull request description with a summary of impact_: 

> This prevents the errors we are seeing in Sentry related to an unregistered user trying to search 50+ genes on a study.

In my opinion, it's best when the first line (and ideally, it's only one line) of the description conveys how the PR's changes affect the user.  Anchoring the description with "here's a single line explaining why you should care" lets maintainers recover context faster after a long time away.  It helps readers quickly assess relevance.  Further context is often important, but I personally find it worth making the brief impact statement first.  

You use that technical writing technique well here -- kudos!

PR #1670, opened 2022-11-10
Fix test idempotency for explore_controller_test.rb (SCP-4756)

bistline:
#### CHANGES
This fixes a potential regression where test order can cause a failure in `explore_controller_test.rb` due to caching of explore responses.  Now, the state is set in `before(:all)` and the `hasImageCache` is checked in the same test for other basic explore attributes.

#### MANUAL TESTING
1. Pull branch and setup environment with `./rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Run `bin/rails test test/api/visualization/explore_controller_test.rb` 3-4 times and ensure no failures, regardless of what order the individual tests run in


eweitz (approved):

ehanna4 (approved):
Ran the tests 5x in a row and never errored 👍

PR #1674, opened 2022-11-16
Wire up Rails to call to ingest-pipeline for Cluster data (SCP-4772)

ehanna4:
Wire up the Rails side to call the new "extract cluster" ingest job from https://github.com/broadinstitute/scp-ingest-pipeline/pull/276 

This also makes use of the feature flag to gate AnnData file ingestion so make sure you follow the steps in this PR: https://github.com/broadinstitute/single_cell_portal_core/pull/1669 to utilize the flag

Screen Shot 2022-11-15 at 4 07 01 PM

Open for review but will need to update the ingest-pipeline version with the small fix for the 2 dimensions check. In theory getting to the error `ValueError: Too few dimensions for visualization in obsm "X_umap", found 2, expected 2 or 3.` proves that the Rails logic is sound to get the ingest job to trigger. 

More tweaks will be needed as we iterate on this, but for the most basic wiring up of cluster data extraction wired into the rails logic this suffices. 

To test:
- pull this branch
- Make sure you've flipped your feature flag to 'true' for your user or your study for `ingest_anndata`
- go to a study you can upload to
- download this AnnData file: https://cellxgene.cziscience.com/collections/2902f08c-f83c-470e-a541-e463e25e5058
- upload that ^ file to your study 
- wait for the email, for now you will see this error `ValueError: Too few dimensions for visualization in obsm "X_umap", found 2, expected 2 or 3.`

To test that the feature flag works, flip it back to false and attempt to upload an AnnData file. Watch the rails terminal and you will see
 `Aborting parse of AnnData file <study_file.name> due to feature flag being false`
Screen Shot 2022-11-15 at 3 54 40 PM

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -13,6 +13,7 @@ class FileParseService
   def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail: false)
     logger = Rails.logger
     logger.info "#{Time.zone.now}: Parsing #{study_file.name} as #{study_file.file_type} in study #{study.name}"
+    do_anndata_file_ingest = FeatureFlaggable.feature_flags_for_instances(user, study)['ingest_anndata_file']
     if !study_file.parseable?
       return {
           status_code: 422,
@@ -99,7 +100,6 @@ class FileParseService
         end
       when 'AnnData'
         # gate ingest of AnnData using the feature flag 'ingest_anndata_file'
-        do_anndata_file_ingest = FeatureFlaggable.feature_flags_for_instances(user, study)['ingest_anndata_file']
         if do_anndata_file_ingest == true
           # Currently assuming "Happy Path" and so the AnnData file will have clustering data
           # extract and parse clustering data
@@ -116,7 +116,10 @@ class FileParseService
         end
 
       end
-      study_file.update(parse_status: 'parsing')
+      # If the AnnData ingest feature flag is false don't update the parse status since no ingest job was initiated
+      unless study_file.file_type == 'AnnData' && !do_anndata_file_ingest
+        study_file.update(parse_status: 'parsing')
+      end
       changes = ["Study file added: #{study_file.upload_file_name}"]
       if study.study_shares.any?
         SingleCellMailer.share_update_notification(study, changes, user).deliver_now
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -97,6 +97,25 @@ class FileParseService
         else
           Rails.logger.info "Aborting parse of #{@study_file.name} as #{@study_file.file_type} in study #{@study.name}; not applicable"
         end
+      when 'AnnData'
+        # gate ingest of AnnData using the feature flag 'ingest_anndata_file'
+        do_anndata_file_ingest = FeatureFlaggable.feature_flags_for_instances(user, study)['ingest_anndata_file']
+        if do_anndata_file_ingest == false
+          Rails.logger.info "Aborting parse of AnnData file 1 #{study_file}"
+          Rails.logger.info "Aborting parse of AnnData file #{study_file.name} due to feature flag being #{do_anndata_file_ingest}"
+        else
+          # Currently assuming "Happy Path" and so the AnnData file will have clustering data
+          # extract and parse clustering data
+          job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_anndata, reparse: reparse,
+          persist_on_fail: persist_on_fail)
+          job.delay.push_remote_and_launch_ingest
+
+          # Future consideration about whether to do all in one job likely in (SCP-4754)
+          # TODO extract and parse Metadata (SCP-4708)
+          # TODO extract and parse Processed Exp Data (SCP-4709)
+          # TODO extract and parse Raw Exp Data (SCP-4710)
+        end
+
       end
       study_file.update(parse_status: 'parsing')
       changes = ["Study file added: #{study_file.upload_file_name}"]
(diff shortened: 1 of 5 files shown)


app/lib/file_parse_service.rb
```diff
@@ -97,6 +97,24 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
         else
           Rails.logger.info "Aborting parse of #{@study_file.name} as #{@study_file.file_type} in study #{@study.name}; not applicable"
         end
+      when 'AnnData'
+        # gate ingest of AnnData using the feature flag 'ingest_anndata_file'
+        do_anndata_file_ingest = FeatureFlaggable.feature_flags_for_instances(user, study)['ingest_anndata_file']
+        if do_anndata_file_ingest == false
```
bistline, line 103: Non-blocking:  I think it might be more readable to invert this and have the `true` clause first.  You could write it as:
```
if do_anndata_file_ingest
  ... ingest code
else
  ... log message for aborting call
end
```

app/lib/file_parse_service.rb
```diff
@@ -97,6 +97,24 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
+          Rails.logger.info "Aborting parse of AnnData file #{study_file.name} due to feature flag being #{do_anndata_file_ingest}"
+        else
+          # Currently assuming "Happy Path" and so the AnnData file will have clustering data
+          # extract and parse clustering data
+          job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_anndata, reparse: reparse,
+          persist_on_fail: persist_on_fail)
+          job.delay.push_remote_and_launch_ingest
+
+          # Future consideration about whether to do all in one job likely in (SCP-4754)
```
bistline, line 112: Useful!

bistline (approved):
Works as advertised, and received expected errors.  Looking forward to seeing this working fully!

app/models/ingest_job.rb
```diff
@@ -14,7 +14,7 @@ class IngestJob
 
   # valid ingest actions to perform
   VALID_ACTIONS = %i[
-    ingest_expression ingest_cluster ingest_cell_metadata subsample differential_expression render_expression_arrays
+    ingest_expression ingest_cluster ingest_cell_metadata subsample differential_expression render_expression_arrays ingest_anndata
```
eweitz, line 17: Keeping more similar things closer to each other, for coherence:

```suggestion
    ingest_expression ingest_cluster ingest_cell_metadata ingest_anndata subsample differential_expression render_expression_arrays
```

app/models/ingest_job.rb
```diff
@@ -391,6 +391,13 @@ def set_study_state_after_ingest
       create_differential_expression_results
     when :image_pipeline
       set_has_image_cache
+    when :ingest_anndata
+      # currently extracting and ingesting only clustering data
+      # this will likely error until the DB inserts ingest job is done
+      set_cluster_point_count
+      set_study_default_options
+      launch_subsample_jobs
+      # TODO (SCP-4708)(SCP-4709)(SCP-4710) will duplicate a lot more from above
```
eweitz, line 400: Like we delimited TODO tickets elsewhere in code, and in PR titles:

```suggestion
      # TODO (SCP-4708, SCP-4709, SCP-4710) will duplicate a lot more from above 
```

app/models/papi_client.rb
```diff
@@ -327,6 +328,9 @@ def get_command_line(study_file:, action:, user_metrics_uuid:, params_object: ni
     when 'image_pipeline'
       # image_pipeline is node-based, so python command line to this point no longer applies
       command_line = 'node expression-scatter-plots.js'
+    when 'ingest_anndata'
+      # extract cluster data from AnnData file, currently hardcoding the obsm-keys
+      command_line +=  " --ingest-anndata --anndata-file #{study_file.gs_url} --extract-cluster --obsm-keys ['X_umap','X_tsne']"
```
eweitz, line 333: This would be a bit more coherent to put above `ingest_subsample`.  It's upstream of that in our dataflow and quite similar to the ingest jobs for our core study file types, whereas e.g. `differential_expression` and `image_pipeline` are fairly different from a dataflow / product perspective.

eweitz (approved):
Code looks good -- kudos of stretching into the Rails ingest orchestration code!

I note a few maintainability nits, nothing significant; no blockers.

jlchang (changes requested):
Currently in production, upload of h5ad does the simple-minded "can I open this file" validation (ie. --ingest-anndata without --extract-cluster). With the new extraction capability behind a feature flag, when the `upload_seurat_and_anndata` feature flag is not TRUE, we should probably continue to do the simple-minded "can I open this file" validation.

From functional testing, when I don't have the feature flag on, an uploaded h5ad stays in "parsing" state (I think its stuck in limbo because it wasn't parsed but also wasn't rejected as bad.

bistline:
> Currently in production, upload of h5ad does the simple-minded "can I open this file" validation (ie. --ingest-anndata without --extract-cluster). With the new extraction capability behind a feature flag, when the `upload_seurat_and_anndata` feature flag is not TRUE, we should probably continue to do the simple-minded "can I open this file" validation.
> 
> From functional testing, when I don't have the feature flag on, an uploaded h5ad stays in "parsing" state (I think its stuck in limbo because it wasn't parsed but also wasn't rejected as bad.

Good call - I didn't consider that!

jlchang (approved):
Thanks for fixing the parse status so the h5ad files don't stay in permanent limbo! Functional review works as expected.

Let's tackle wiring up the basic validation in a future ticket.

PR #1676, opened 2022-11-16
Restore stack traces, add Sentry logging for React errors (SCP-4799)

eweitz:
This fixes observability bugs that blocked much JavaScript error debugging.  It also improves error logging consistency.

Previously:

* JavaScript stack traces were unavailable when a JS error was encountered in a React component. 
 Only the top-level error message was displayed in the UI.  Now, the same error is shown in the UI, _and_ the full error stack trace is printed to the browser's console.  This traces the error's execution path, with files, line numbers, etc.  These traces had been available some time ago; this restores them.

* Such errors were not logged to Sentry.  Now they are.   They were only logged to Mixpanel, which also had the downside of using an idiosyncratic `error-message` event name.  Now, this Mixpanel logging for JS errors in React components uses the same event name -- `error` -- as other error logging in JS (like e.g. [#1656](https://github.com/broadinstitute/single_cell_portal_core/pull/1656)) and [in Ruby on Rails](https://mixpanel.com/s/3Ywrz).  Mixpanel Lexicon reports the conventional "error" event is used in 8 Mixpanel queries, whereas the prior React-specific "error-message" event was used in none.  So I'm not concerned we're breaking analytics.

Here's an example stack trace:
Restored_stack_trace_for_JS_error_in_React_component__SCP_2022-11-16

To test:
1. Change line 312 of ScatterPlot.jsx to `scatter = null // updateScatterLayout(scatter)`
2. Go to a Study Overview page that has visualizations
3. Open DevTools, go to Console Panel
4. Note JS error stack trace, like one in screenshot
5. Undo step 1

This satisfies SCP-4799.

### DIFF
--- a/app/javascript/components/visualization/GeneListHeatmap.jsx
+++ b/app/javascript/components/visualization/GeneListHeatmap.jsx
@@ -25,7 +25,7 @@ function RawGeneListHeatmap({
     heatmapRowCentering,
     geneList: geneListName
   })
-  const { ErrorComponent, setShowError, setErrorContent } = useErrorMessage()
+  const { ErrorComponent, setShowError, setError } = useErrorMessage()
   // we can't render until we know what the cluster is, since morpheus requires the annotation name
   // so don't try until we've received this, unless we're showing a Gene List
   const canRender = !!geneListName && geneLists.length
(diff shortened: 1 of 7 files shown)


bistline (approved):
Very happy this has been resolved.

ehanna4 (approved):

PR #1680, opened 2022-11-21
Add AnnData Ingestible and AnnData Reference file types

ehanna4:
We currently allow users to upload AnnData files to SCP. These are more of a reference type file (more akin to the Miscellaneous file types) than the ingestible file types. Once we add the ability to ingest AnnData files any AnnData files already on SCP will cause issues and could break our new requirements. To fix this issue and allow users to continue to upload reference style AnnData files while we build out and test the ingestible AnnData upload this PR introduces two new distinct AnnData file_types.

`AnnData reference` file_type will be used for all the existing AnnData files on SCP, so there is a migration script to move those. Also when the feature flag for `ingest_anndata_file` is flipped to false this will be the assigned file_type

`AnnData ingestible` file_type will be used for testing and for the future when we allow users to ingest AnnData files. This will be the assigned file_type when the feature flag is true.

This means all AnnData will still be housed under the same section in the UI wizard and in the future we can employ the tactic used in the Miscellaneous section to allow users to choose `ingestible` or `reference`

### DIFF
--- a/app/javascript/components/upload/AnnDataStep.jsx
+++ b/app/javascript/components/upload/AnnDataStep.jsx
@@ -4,11 +4,12 @@ import AnnDataFileForm from './AnnDataFileForm'
 import { AddFileButton } from './form-components'
 
 const DEFAULT_NEW_ANNDATA_FILE = {
-  file_type: 'AnnData',
+  file_type: 'AnnData Reference',
   options: {}
 }
 
-const AnnDataFileFilter = file => ['AnnData'].includes(file.file_type)
+const AnnDataFileTypes = ['AnnData Reference', 'AnnData Ingestible']
+const AnnDataFileFilter = file => AnnDataFileTypes.includes(file.file_type)
 
 export default {
   title: 'AnnData (.h5ad)',
@@ -20,6 +21,7 @@ export default {
 
 /** Renders a form for uploading one or more AnnData files */
 function AnnDataForm({
+  serverState,
   formState,
   addNewFile,
   updateFile,
@@ -27,9 +29,22 @@ function AnnDataForm({
   deleteFile
 }) {
   const AnnDataFiles = formState.files.filter(AnnDataFileFilter)
+  const featureFlagState = serverState.feature_flags
+  console.log('featureflagstate:', featureFlagState)
+  console.log('featureflagstate ingest_anndata:', featureFlagState?.ingest_anndata_file)
+
+  // if the feature flag is flipped to ingest the AnnData file update the file_type
+  let NEW_ANNDATA_FILE = DEFAULT_NEW_ANNDATA_FILE
+  if (featureFlagState?.ingest_anndata_file) {
+    NEW_ANNDATA_FILE = {
+      file_type: 'AnnData Ingestible',
+      options: {}
+    }
+  }
+
   useEffect(() => {
     if (AnnDataFiles.length === 0) {
-      addNewFile(DEFAULT_NEW_ANNDATA_FILE)
+      addNewFile(NEW_ANNDATA_FILE)
     }
   }, [AnnDataFiles.length])
 
@@ -54,10 +69,10 @@ function AnnDataForm({
         updateFile={updateFile}
         saveFile={saveFile}
         deleteFile={deleteFile}
-        annDataFileTypes={['AnnData']}
+        annDataFileTypes={AnnDataFileTypes}
         bucketName={formState.study.bucket_id}
         isInitiallyExpanded={AnnDataFiles.length === 1}/>
     })}
-    <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_ANNDATA_FILE}/>
+    <AddFileButton addNewFile={addNewFile} newFileTemplate={NEW_ANNDATA_FILE}/>
   </div>
 }
--- a/app/javascript/lib/validation/validate-file.js
+++ b/app/javascript/lib/validation/validate-file.js
@@ -61,7 +61,7 @@ async function validateLocalFile(file, studyFile, allStudyFiles=[], allowedFileE
 
   let issuesObj
 
-  const noContentValidationFileTypes = ['Seurat', 'AnnData', 'Other', 'Documentation']
+  const noContentValidationFileTypes = ['Seurat', 'AnnData Reference', 'AnnData Ingestible', 'Other', 'Documentation']
 
   const studyFileType = studyFile.file_type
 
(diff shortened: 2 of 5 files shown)


app/javascript/components/upload/AnnDataStep.jsx
```diff
@@ -4,11 +4,12 @@ import AnnDataFileForm from './AnnDataFileForm'
 import { AddFileButton } from './form-components'
 
 const DEFAULT_NEW_ANNDATA_FILE = {
-  file_type: 'AnnData',
+  file_type: 'AnnData Reference',
   options: {}
 }
 
-const AnnDataFileFilter = file => ['AnnData'].includes(file.file_type)
+const AnnDataFileTypes = ['AnnData Reference', 'AnnData Ingestible']
```
eweitz, line 11: Blocker: for succinctness and consistency, let's retain "AnnData" and interpret the lack of qualifiers as "AnnData ingestible".  It seems more maintainable in the long term to consider the default "AnnData" type to be ingestible, rather than needing to pass around the explicit qualifier.

This convention would let us simplify naming for other foreseeable future use cases, like reference and ingestible Seurat files.

```suggestion
const AnnDataFileTypes = ['AnnData', 'AnnData Reference']
```

app/javascript/components/upload/AnnDataStep.jsx
```diff
@@ -20,16 +21,30 @@ export default {
...
   formState,
   addNewFile,
   updateFile,
   saveFile,
   deleteFile
 }) {
   const AnnDataFiles = formState.files.filter(AnnDataFileFilter)
+  const featureFlagState = serverState.feature_flags
+  console.log('featureflagstate:', featureFlagState)
+  console.log('featureflagstate ingest_anndata:', featureFlagState?.ingest_anndata_file)
```
eweitz, line 34:
```suggestion
```

app/javascript/components/upload/AnnDataStep.jsx
```diff
@@ -20,16 +21,30 @@ export default {
...
+
+  // if the feature flag is flipped to ingest the AnnData file update the file_type
+  let NEW_ANNDATA_FILE = DEFAULT_NEW_ANNDATA_FILE
+  if (featureFlagState?.ingest_anndata_file) {
+    NEW_ANNDATA_FILE = {
+      file_type: 'AnnData Ingestible',
```
eweitz, line 40: Blocker: per other comment:

```suggestion
      file_type: 'AnnData',
```

app/javascript/lib/validation/validate-file.js
```diff
@@ -61,7 +61,7 @@ async function validateLocalFile(file, studyFile, allStudyFiles=[], allowedFileE
 
   let issuesObj
 
-  const noContentValidationFileTypes = ['Seurat', 'AnnData', 'Other', 'Documentation']
+  const noContentValidationFileTypes = ['Seurat', 'AnnData Reference', 'AnnData Ingestible', 'Other', 'Documentation']
```
eweitz, line 64: Blocker: per other comment:

```suggestion
  const noContentValidationFileTypes = ['Seurat', 'AnnData Reference', 'AnnData', 'Other', 'Documentation']
```

app/models/study_file.rb
```diff
@@ -24,7 +24,7 @@ class StudyFile
   # constants, used for statuses and file types
   STUDY_FILE_TYPES = ['Cluster', 'Coordinate Labels' ,'Expression Matrix', 'MM Coordinate Matrix', '10X Genes File',
                       '10X Barcodes File', 'Gene List', 'Metadata', 'Fastq', 'BAM', 'BAM Index', 'Documentation',
-                      'Other', 'Analysis Output', 'Ideogram Annotations', 'Image', 'AnnData', 'Seurat'].freeze
+                      'Other', 'Analysis Output', 'Ideogram Annotations', 'Image', 'AnnData Reference', 'AnnData Ingestible', 'Seurat'].freeze
```
eweitz, line 27: Blocker: per other comment:

```suggestion
                      'Other', 'Analysis Output', 'Ideogram Annotations', 'Image', 'AnnData', 'AnnData Reference', 'Seurat'].freeze
```

eweitz (commented):
I'd usually scrap a review where I overlooked (helpful, thank you!) draft status, but I think this early PR feedback may well save time overall.

bistline:
Thinking out loud - since we have a ticket to create a model to [track what we extract](https://broadworkbench.atlassian.net/browse/SCP-4831) (among other things) from AnnData files next sprint, would it make sense to address this issue there?  Some kind of nested `AnnDataFileInfo` object that would let us know what kind of data we pulled out of the file?  That might be a less disruptive change that adding a new file type and migrating records.

Then we could just check the feature flag state, and depending on whether it's on/off adjust the command line to pass the appropriate flags to ingest to control what kind of job to run (existing, normal "just open and validate" or full data ingest).

ehanna4:
@bistline oh yes I think that could also work. I guess we don't have an exact pattern for this sort of upload yet, I was basing this idea off of the way Miscellaneous files have two different `file_types`. But if it'd be a simpler convention to use one `file_type` with two different paths that's also fine to me. 

One other consideration would be mixpaneling, would we be able to differentiate the `ingestible` vs `reference` type AnnData files using the nested `AnnDataFileInfo` way?

bistline:
> @bistline oh yes I think that could also work. I guess we don't have an exact pattern for this sort of upload yet, I was basing this idea off of the way Miscellaneous files have two different `file_types`. But if it'd be a simpler convention to use one `file_type` with two different paths that's also fine to me.
> 
> One other consideration would be mixpaneling, would we be able to differentiate the `ingestible` vs `reference` type AnnData files using the nested `AnnDataFileInfo` way?

I believe so, yes - we could add those attributes to what we report to Mixpanel (in `IngestJob#get_job_analytics`):
```
job_props = {
  fileType: 'AnnData',
  hasClusters: true,
  hasMetadata: false,
  ...
}
```
The normal `fileType` breakdown in [reports like this](https://mixpanel.com/project/2120588/view/19059/app/dashboards#id=734187&editor-card-id=%22report-27903457%22) would just say `AnnData`, but if we had reports where we were looking at specific file types (like `Cluster`), we could check for that new property with `fileType: 'Cluster' OR (fileType: 'AnnData' AND hasClusters: true)`

jlchang (commented):
Makes sense to fold "reference" vs "ingestable" into the AnnData data model. Thanks for taking the lead on distinguishing between these two different use cases, Emily!

ehanna4:
Closing this draft and linking it to the Data Model ticket @bistline has referenced above. Thanks team for all your insight into this discussion!

PR #1681, opened 2022-11-22
Wire up the ingest for 'reference AnnData' files (SCP-4838)

ehanna4:
This adds the wiring on the Rails side to call the ingest job that does the very basic check on AnnData files uploaded as reference files. Previously these files were not being parsed at all, just uploaded.

To test:
- Pull this branch and boot up a rails server and the delayed job worker
- Open up localhost
- Make sure your feature flag for `ingest_anndata_file` is set to false
- go to a study you can upload data to
- upload an AnnData file
- observe that the file uploads and the parsing status appears then you should receive an email that says the file finished parsing
- If you want to further confirm check out the study details page to see the file and it's parsed status
Screen Shot 2022-11-29 at 5 07 20 PM

Screen Shot 2022-11-21 at 3 46 05 PM

Screen Shot 2022-11-29 at 5 07 42 PM

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -28,8 +28,7 @@ class FileParseService
       self.create_bundle_from_file_options(study_file, study)
       case study_file.file_type
       when 'Cluster'
-        job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_cluster, reparse: reparse,
-                            persist_on_fail: persist_on_fail)
+        job = IngestJob.new(study:, study_file:, user:, action: :ingest_cluster, reparse:, persist_on_fail:)
         job.delay.push_remote_and_launch_ingest
         # check if there is a coordinate label file waiting to be parsed
         # must reload study_file object as associations have possibly been updated
@@ -38,25 +37,23 @@ class FileParseService
           study_file.bundled_files.each do |coordinate_file|
             # pre-emptively set parse_status to prevent initialize_coordinate_label_data_arrays from failing due to race condition
             study_file.update(parse_status: 'parsing')
-            study.delay.initialize_coordinate_label_data_arrays(coordinate_file, user, {reparse: reparse})
+            study.delay.initialize_coordinate_label_data_arrays(coordinate_file, user, { reparse: })
           end
         end
       when 'Coordinate Labels'
         if study_file.has_completed_bundle?
-          ParseUtils.delay.initialize_coordinate_label_data_arrays(study, study_file, user, {reparse: reparse})
+          ParseUtils.delay.initialize_coordinate_label_data_arrays(study, study_file, user, { reparse: })
         else
           return self.missing_bundled_file(study_file)
         end
       when 'Expression Matrix'
-        job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_expression, reparse: reparse,
-                            persist_on_fail: persist_on_fail)
+        job = IngestJob.new(study:, study_file:, user:, action: :ingest_expression, reparse:, persist_on_fail:)
         job.delay.push_remote_and_launch_ingest
       when 'MM Coordinate Matrix'
         study_file.reload
         if study_file.has_completed_bundle?
           study_file.bundled_files.update_all(parse_status: 'parsing')
-          job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_expression, reparse: reparse,
-                              persist_on_fail: persist_on_fail)
+          job = IngestJob.new(study:, study_file:, user:, action: :ingest_expression, reparse:, persist_on_fail:)
           job.delay.push_remote_and_launch_ingest
         else
           study.delay.send_to_firecloud(study_file) if study_file.is_local?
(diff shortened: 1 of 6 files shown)


app/models/papi_client.rb
```diff
@@ -321,8 +322,11 @@ def get_command_line(study_file:, action:, user_metrics_uuid:, params_object: ni
     when 'ingest_cluster'
       command_line += " --cluster-file #{study_file.gs_url} --ingest-cluster"
     when 'ingest_anndata'
-      # extract cluster data from AnnData file, currently hardcoding the obsm-keys
-      command_line +=  " --ingest-anndata --anndata-file #{study_file.gs_url} --extract-cluster --obsm-keys ['X_umap','X_tsne']"
+        # extract cluster data from AnnData file, currently hardcoding the obsm-keys
+        command_line +=  " --ingest-anndata --anndata-file #{study_file.gs_url} --extract-cluster --obsm-keys ['X_umap','X_tsne']"
+    when 'ingest_anndata_reference'
+        command_line = "python ingest_pipeline.py --study-id #{study.id} --study-file-id #{study_file.id} " \
```
bistline, line 328: We've already initializes `command_line` with the default values so no need to re-declare them here:
```suggestion
        command_line +=  " --ingest-anndata --anndata-file #{study_file.gs_url}"
```

app/lib/file_parse_service.rb
```diff
@@ -99,27 +99,25 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
           Rails.logger.info "Aborting parse of #{@study_file.name} as #{@study_file.file_type} in study #{@study.name}; not applicable"
         end
       when 'AnnData'
-        # gate ingest of AnnData using the feature flag 'ingest_anndata_file'
+        # gate full ingest of an AnnData file using the feature flag 'ingest_anndata_file'
         if do_anndata_file_ingest == true
```
bistline, line 103: Redundant - this would evaluate to `if true == true`
```suggestion
        if do_anndata_file_ingest
```

bistline (approved):
Works as advertised - non-blocking refinements.  Nice work catching this!

app/models/papi_client.rb
ehanna4: The initialized `command_line` includes the action which in this case we don't want to be part of the command to ingest pipeline since the action`ingest_anndata_reference` isn't part of the ingest command. I made an update to duplicate less of the command_line, let me know what you think.

app/lib/file_parse_service.rb
```diff
@@ -99,27 +99,25 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
-
           # Future consideration about whether to do all in one job likely in (SCP-4754)
           # TODO extract and parse Metadata (SCP-4708)
           # TODO extract and parse Processed Exp Data (SCP-4709)
           # TODO extract and parse Raw Exp Data (SCP-4710)
         else
-          Rails.logger.info "Aborting parse of AnnData file #{study_file.name} due to feature flag being #{do_anndata_file_ingest}"
+          # launch an ingest job for parsing a reference AnnData file
+          job = IngestJob.new(study: study, study_file: study_file, user: user, action: :ingest_anndata_reference, reparse: reparse,
+          persist_on_fail: persist_on_fail)
```
eweitz, line 115: This is a good example of how specifying values that have names matching keys degrades readability.  

I suggest:

```suggestion
          job = IngestJob.new(study:, study_file:, user:, action: :ingest_anndata_reference, reparse:, persist_on_fail:)
```

Benefits:
* Faster to discern most salient part of assignment -- the part that's not implied by the key
* Better visual chunking due to assignment fitting on one line, rather than filling two
* Truly leverage one of the biggest features in Ruby 3.1, rather than our quite constrained recent usage
* Easier jump between languages, e.g. SCP JS is replete with its equivalent of "mixed" Hash literal style

I'm unaware of any costs not already incurred in our JS, where such mixing has seemed successful.

If the above persuades folks, I'd propose we [remove this line in our Rubocop lint rules](https://github.com/broadinstitute/single_cell_portal_core/blob/development/.rubocop.yml#L58), and gradually apply Ruby 3.1 shorthand syntax wherever it applies.

eweitz (approved):
Code looks good overall!

I note a nit, and resume a small unfinished [code style discussion](https://github.com/broadinstitute/single_cell_portal_core/pull/1660#discussion_r1013059291) that relates here.

app/models/papi_client.rb
bistline: Ah yes excellent point!  I missed that.

app/lib/file_parse_service.rb
bistline: I'm ambivalent about the rubocop change - personally I think consistency matters, but it does allow for better concision so I'm fine if the rest of the team wants to make that change.

bistline: 4ecb737

app/lib/file_parse_service.rb
```diff
@@ -99,27 +99,25 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
           Rails.logger.info "Aborting parse of #{@study_file.name} as #{@study_file.file_type} in study #{@study.name}; not applicable"
         end
       when 'AnnData'
-        # gate ingest of AnnData using the feature flag 'ingest_anndata_file'
-        if do_anndata_file_ingest == true
+        # gate full ingest of an AnnData file using the feature flag 'ingest_anndata_file'
```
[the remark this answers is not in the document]
bistline: 4ecb737

PR #1634, opened 2022-10-06
Creating `ImagePipelineService` for running Image Pipeline jobs in PAPI (SCP-4649)

bistline:
#### BACKGROUND
As part of ongoing Image Pipeline work, we will need a way to easily launch jobs in PAPI without having to manually construct parameter sets and poll for completion.  This will be a requirement for future work to automate the end-to-end run of generating expression arrays and then rendering static images for expression plots.

#### CHANGES
This update adds `ImagePipelineService` and `RenderExpressionArraysParameters`, both of which are modeled after the analogous `DifferentialExpression` classes.  These will allow launching of PAPI jobs through a simple convenience method of `ImagePipelineService.run_render_expression_arrays_job(study, cluster_file, matrix_file)` which will handle constructing/validating parameters, job submission/polling, and success/error reporting.  `RenderExpressionArraysParameters` will also dynamically scale the `machine_type` parameter based on the size of the expression matrix.

In addition, the `Parameterizable` concern was added to refactor out shared logic for constructing command line arrays for Docker/PAPI.  Lastly, this updates the version of `rubocop` and some of our linting rules to better support changes in Ruby 3.1

#### MANUAL TESTING
1. Pull branch, and enter a Rails console session:
```
./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails c
```
2. In a separate terminal, start DelayedJob: 
```
./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails jobs:work
```
3. Back in the rails console, load the `Murine vision and HIV with arrays` synthetic study (or any study with a small dense matrix & cluster file):
```
name = "Murine vision and HIV with arrays"
study = Study.find_by(name:)
cluster_file = study.cluster_ordinations_files.first
matrix_file = study.expression_matrices.first
```
4. Launch the job and confirm you get `true` back
```
ImagePipelineService.run_render_expression_arrays_job(study, cluster_file, matrix_file)
```
5. In `development.log`, note that the job has launched, and used the "small" machine type of `n1-highcpu-4`:
```
Remote found for cluster.tsv:6261a3fb94ec8fe535ac00a4, launching Ingest job
Request object sent to Google Pipelines API (PAPI), excluding 'environment' parameters:
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 6261a3e394ec8fe535ac009a
  - "--study-file-id"
  - 6261a3fb94ec8fe535ac00a4
  - "--user-metrics-uuid"
[...]

PR #1682, opened 2022-11-28
Speed up large expression scatter plots 3-4x via bucket cache data (SCP-4642, SCP-4792)

eweitz:
This greatly speeds up gene expression scatter plots that are >= 0.5M points, by using data cached in buckets.

Previously, this data was retrieved from MongoDB via Rails.  Now, building atop infrastructure described in #1634, this data is pulled as a pre-computed array stored in a flat JSON file in a Google bucket.  So it completely bypasses our traditional backend, by expanding our pre-existing use of buckets as a storage medium.  This will theoretically speed up plots that are 1-1.5M points from [46 s to 15 s](https://mixpanel.com/s/36K58R) (3.1x), and plots that are 0.5-1M points from [19 s to 4.5 s](https://mixpanel.com/s/2B2qli) (4.2x).

To test:
1. Enter development Rails console: 
`bin/rails console --environment development`
2. Find your local DE pilot study: 
`study = Study.find_by(name: 'Human milk - differential expression')`
3. Manually adjust default clustering: 
`study.cluster_groups.find_by(name: 'All Cells UMAP').update(has_image_cache: true)`
4. Get study's bucket ID (prints "fc-65379b91-5ded-4d28-8e51-ada209542117" for me): 
`study.bucket_id`
5. Copy test files from staging DE pilot to your development DE pilot:
```
DEST_BUCKET="<your bucket ID from step 5>"; SRC_BUCKET="fc-37c32921-4c79-4767-819d-66f157f871a9"; gcloud storage cp "gs://${SRC_BUCKET}/_scp_internal/cache/expression_scatter/images/_debug_2022-10-25T164028/A1BG-AS1.webp" "gs://${DEST_BUCKET}/_scp_internal/cache/expression_scatter/images/All_Cells_UMAP/A1BG-AS1.webp"; gcloud storage cp "gs://${SRC_BUCKET}/_scp_internal/cache/expression_scatter/images/_debug_2022-10-25T164028/A1BG.webp" "gs://${DEST_BUCKET}/_scp_internal/cache/expression_scatter/images/All_Cells_UMAP/A1BG.webp"; gcloud storage cp "gs://${SRC_BUCKET}/_scp_internal/cache/expression_scatter/data/All_Cells_UMAP/A1BG.json" "gs://${DEST_BUCKET}/_scp_internal/cache/expression_scatter/data/All_Cells_UMAP/"; gcloud storage cp "gs://${SRC_BUCKET}/_scp_internal/cache/expression_scatter/data/All_Cells_UMAP/A1BG-AS1.json" "gs://${DEST_BUCKET}/_scp_internal/cache/expression_scatter/data/All_Cells_UMAP/"
```
6.  In. your Rails console, get your DE pilot's accession:
`study.accession`
8.  In your web browser, sign in to your local SCP
9.  If you haven't already, enable the "progressive_loading" feature flag on your local DE pilot study (Upper right menu, Feature flags, search by your DE pilot's accession, set "progressive loading" to true)
10.  Search for your local DE pilot study ("Human milk - differential expression"), go to its Explore tab
11. Open Network panel
12. Search for gene "A1BG"
13. Confirm the gene expression scatter plot renders
14. In Network panel, filter requests to "storage.googleapis.com method:GET"
15. Confirm there are 2 matching successful GET requests

This satisfies SCP-4642 and SCP-4792.


app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -355,34 +372,69 @@ function RawScatterPlot({
...
+      fetchBucketFile(bucketId, imagePath).then(async response => {
+        const imageCacheHit = response.ok
+
+        // Draw plot as static image first, if it's cached
+        if (imageCacheHit) {
+          await drawPlotImage(response)
+        }
+
+        // Then make it interactive, using gene expression scatter plot data array from GCS bucket
+        renderBucketData(fetchMethod, expressionParams, dataPath)
```
bistline, line 415: Non-blocking: since this is still WIP and prototype-level, I don't consider this blocking because it's an error state that I doubt we'd encounter in normal usage.

I think this method should have error handling such that if we make a `GET` for the gene-level JSON data and can't find it, it will then fall back to the API.  While testing this PR, the image request returned `200` but the data request was `404` because it wasn't in the bucket.  I think the testing setup missed steps for copying that data or relied on the state from a previous PR review, and I had done cleanup while testing other work.

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -355,34 +372,69 @@ function RawScatterPlot({
...
       flags?.progressive_loading && isGeneExpression(genes, isCorrelatedScatter) && !isAnnotatedScatter &&
-      !scatterData &&
-      genes[0] === 'A1BG-AS1' // Placeholder; likely replace with setting like DE
+      !scatterData
     ) {
-      const bucketName = 'broad-singlecellportal-public'
-      const filePath = `test/scatter_image/${genes[0]}-v2.webp`
-      fetchBucketFile(bucketName, filePath).then(async response => {
-        renderImage(response)
+      const urlSafeCluster = cluster.replaceAll(' ', '_')
```
bistline, line 384: Blocker: we will need to handle the corner case where a cluster name has `+` in it (like `CD4+`) and had that converted to `pos` when caching images/data.  Also, the logic for the replaceAll should cover non-word characters, not just spaces.  The `ExpressionWriter` class in `scp-ingest-pipeline` has the following logic:

```
plus_converted_string = name.replace('+', 'pos')
return re.sub(r'\W', '_', plus_converted_string)
```

We should also ensure that Image Pipeline is doing the same thing.  I checked and it doesn't appear that there is any cluster name sanitization happening that I can see.  None of our testing data covers this case but there are clusters in production that contain a plus character in their name.

bistline (changes requested):
Overall looks good!  Only one trivial blocker re: cluster name sanitization, and a non-blocking error handling suggestion.

app/javascript/components/visualization/ScatterPlot.jsx
eweitz: Good point, agreed.  I've got another PR in the works that fixes some bugs in Image Pipeline and related integration.  I'll fix this code here as you request, and make the related fix in Image Pipeline in that forthcoming PR.

app/javascript/components/visualization/ScatterPlot.jsx
eweitz: Agreed, this seems very unlikely on production.  I've updated the step 5 above to account for that test data case, and opened SCP-4845 for the related robustness task.

app/javascript/components/visualization/ScatterPlot.jsx
eweitz: Done in 163be78.  I also just opened SCP-4846 to consolidate our 3-4 implementations of this, per chat.

bistline (approved):

PR #1685, opened 2022-11-29
Updating to scp-ingest-pipeline:1.23.2, adding cell_type__custom facet

bistline:
#### BACKGROUND & CHANGES
This update adds support for using the `cell_type__custom` column in compliant metadata files while no longer needing a corresponding entry in `cell_type` or `cell_type__ontology_label`.  This will unblock users who cannot find corresponding analogs for special/custom cell types not listed in the [Cell](https://www.ebi.ac.uk/ols/ontologies/cl) ontology.

#### MANUAL TESTING
1. Download [valid_cell_type__custom_v2.2.1.txt](https://github.com/broadinstitute/scp-ingest-pipeline/blob/jlc_ctcustom_nodep/tests/data/annotation/metadata/convention/valid_cell_type__custom_v2.2.1.txt) locally
2. Pull this branch, run any pending migrations and start all services:
```
.bin/docker-compose-setup.sh
```
3. On the homepage, click the `cell type` facet and confirm that `cell type (custom)` also shows (you may not have data present there depending on the state of your database)
Screen Shot 2022-11-29 at 4 20 05 PM

4. Create a new study, and upload valid_cell_type__custom_v2.2.1.txt as a metadata file and ensure that it passes ingest


eweitz (approved):
Code looks good!  

It's neat how adding a facet requires only a small change in Rails code.  The complexity is encapsulated in [recent Ingest Pipeline changes](https://github.com/broadinstitute/scp-ingest-pipeline/pull/281), I suppose.

I allude to a maintainability improvement in a detailed comment below.  It's non-blocking.

I think it's more important, but definitely out of scope and non-blocking, to refine the UX and scientific overlap for this "cell type (custom)" facet.  The example "custom" cell types in your helpful screenshot have labels that seem very likely to map to CLO-official cell type labels.  Having exactly the same label match to two different-but-similar things might not be a technical problem for us, but would facilitate terminology confusion that ontologies exist to clarify.

That's food for thought.  As you note, this [unblocks a user](https://broadinstitute.zendesk.com/agent/tickets/298615), which surely takes precedence.

bistline:
> The example "custom" cell types in your helpful screenshot have labels that seem very likely to map to CLO-official cell type labels. Having exactly the same label match to two different-but-similar things might not be a technical problem for us, but would facilitate terminology confusion that ontologies exist to clarify.

FWIW that screenshot had data from a synthetic study (`mouse_brain`) which is why it looks duplicative.  In production, hopefully we would be getting values that do not map to entries in the CL ontology (like "immature neuron", which is one of the use cases that prompted this work).

jlchang (approved):

I see `cell type (custom)` as a facet and after ingest of the test file, the new labels are visible as additional filters for the new facet. The code updates made sense to me in a general way but I was not confident I could spot any issue based on my code review. Since this is a bug fix we're releasing for a blocked user, I felt obligated to have fully performed the functional test.

I regret it took me a while to get my local dockerized instance running to complete the review. Thanks for your patience.

PR #1686, opened 2022-11-30
Ensure cache dir permissions before compiling static assets (SCP-4844)

bistline:
#### BACKGROUND & CHANGES
This fixes an issue discovered on a [recent deployment](https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-master-to-production/118/console) where directories owned by `root` could not be removed when attempting to clear asset caches and recompile.  Now, the `tmp/cache` directory will have `chown` recursively applied to ensure everything is owned by the `app` user.  That being said, this particular error only occurred once on each host (staging & production), and repeated deployments to staging have succeeded without this update.  This step is more preventative than actually addressing a known bug.  Also, the operation only takes 2-3 seconds to complete and will not substantively impact deployment times.

#### MANUAL TESTING
Since this only happens on deployments, and Docker Desktop for Mac uses different file system drivers than Ubuntu, it is not possible to test this locally.

### DIFF
--- a/rails_startup.bash
+++ b/rails_startup.bash
@@ -24,7 +24,9 @@ if [[ $PASSENGER_APP_ENV = "production" ]] || [[ $PASSENGER_APP_ENV = "staging" 
 then
     echo "*** PRECOMPILING ASSETS ***"
     export NODE_OPTIONS="--max-old-space-size=4096"
+    # ensure file permissions don't interfere with compiling assets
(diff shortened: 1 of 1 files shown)


eweitz (approved):

jlchang (approved):
Per discussion post-meeting, there's no way to test this change.

The change itself looks straightforward enough. Sorry for the delay on this review!

PR #1691, opened 2022-12-07
Fix sparse gene sets in related genes ideogram (SCP-4851)

eweitz:
This fixes poor handling by related genes ideogram when a study has relatively few genes.

Before, searching e.g. "CST3" in the new 3k PBMC demo study showed a paralog neighborhood of genes that exist, but were not assayed in the study.  So clicking the gene symbols in the paralog neighborhood tooltip unexpectedly did nothing.  The label for the searched gene (CST3, red) was also unexpectedly not shown.  

Here's how it looked with the bug:
Screen Shot 2022-12-07 at 10 26 11 AM

Now, paralog neighborhoods (pink rectangles overlaid atop the chromosome) only show genes in the expression matrix, and labels are more robust.

Here's how it looks now with the fix:
Screen Shot 2022-12-07 at 10 25 50 AM

I've updated synthetic test data to enable easy manual verification.

To test:
* `yarn install`
* (Re-)populate synthetic study:
```
echo "SyntheticStudyPopulator.populate('human_lymph_node')" | bundle exec rails console -e development
```
* Wait a few minutes, then go to that study, "Tuberculosis in female human lymph"
* Search CST3
* Confirm related genes ideogram looks like 2nd screenshot above

This satisfies SCP-4851.


ehanna4 (approved):
Code review - code change is straightforward and the additional testing data will be useful, thanks for adding that!

bistline (approved):

PR #1695, opened 2022-12-15
Track DE feature globally in Mixpanel (SCP-4717)

bistline:
#### BACKGROUND & CHANGES
This update adds the global `isDEEnabled` property to all front-end Mixpanel events.  This will allow us to create a variety of reports/funnels related to differential expression usage by users.  For instance, rather than having to manually curate a list of study accessions to know which studies have DE results, we can simply add a filter for `isDEEnabled: true`.  Similarly, we can chain this event with another for the `click:button` event where the text is `Differential expression`, and add a conversion criterion of `studyAccession` remaining the same.  This will then give us users who visited a study that has DE results available and then interacted with the DE feature without leaving that study.

Some things worth noting:
1. We are tracking this at the study level since of the 106 studies that have DE results available, 96 have results for the default annotation (90%).  Furthermore, 48 (45%) have `cell_type` as the default annotation (or some analog of that).  Given this prevalence, it does not seem necessary to track DE enablement at the plot or annotation level, as this may grossly overcount users.
2. Even though the above studies have DE results available, this does _not_ necessarily mean that users can see them.  Given the current feature flag state for `differential_expression_frontend`, we are opting in DE at the study level.  Right now, only 39 studies have DE enabled.  In order for this to have better signal, we may want to revisit this decision.
3. This feature _does_ take into account whether or not a user is able to see DE results.  For instance, if a study has not opted into DE, `isDEEnabled` is `false`.

#### MANUAL TESTING
1. Boot as normal and visit any study with DE results that you can view
2. Open DevTools > Network and filter for `event` requests
3. Click on the "Explore" tab, and find the corresponding Mixpanel event, noting that `isDEEnabled` is `true`
Screen Shot 2022-12-15 at 10 10 48 AM
4. Click the "Differential expression" button, and find the corresponding Mixpanel event, noting that `isDEEnabled` is `true`
Screen Shot 2022-12-15 at 10 11 14 AM

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -56,7 +56,9 @@ function annotHasDe(exploreInfo, exploreParams) {
   const flags = getFeatureFlagsWithDefaults()
   if (!flags?.differential_expression_frontend || !exploreInfo) {
     // set isDEEnabled to false since user cannot see DE results, so we don't want to skew usage data
-    window.SCP.isDEEnabled = false
+    if (window.SCP) {
+      window.SCP.isDEEnabled = false
+    }
     return false
   }
 
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -55,6 +55,8 @@ const tabList = [
 function annotHasDe(exploreInfo, exploreParams) {
   const flags = getFeatureFlagsWithDefaults()
   if (!flags?.differential_expression_frontend || !exploreInfo) {
+    // set isDEEnabled to false since user cannot see DE results, so we don't want to skew usage data
+    window.SCP.isDEEnabled = false
     return false
   }
 
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -55,7 +55,7 @@ const tabList = [
 function annotHasDe(exploreInfo, exploreParams) {
   const flags = getFeatureFlagsWithDefaults()
   if (!flags?.differential_expression_frontend || !exploreInfo) {
-    // set isDEEnabled to false since user cannot see DE results, so we don't want to skew usage data
+    // set isDifferentialExpressionEnabled to false as user cannot see DE results, even if present for annotation
     if (window.SCP) {
       window.SCP.isDEEnabled = false
     }
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -57,7 +57,7 @@ function annotHasDe(exploreInfo, exploreParams) {
   if (!flags?.differential_expression_frontend || !exploreInfo) {
     // set isDifferentialExpressionEnabled to false as user cannot see DE results, even if present for annotation
     if (window.SCP) {
-      window.SCP.isDEEnabled = false
+      window.SCP.isDifferentialExpressionEnabled = false
     }
     return false
   }
--- a/app/javascript/components/explore/ExploreView.jsx
+++ b/app/javascript/components/explore/ExploreView.jsx
@@ -29,7 +29,9 @@ function RoutableExploreTab({ studyAccession }) {
     const exploreResponse = await fetchExplore(studyAccession)
     setExploreInfo(exploreResponse)
     // set window.SCP.isDEEnabled so that we can track DE visibility globally
-    window.SCP.isDEEnabled = exploreResponse.differentialExpression.length > 0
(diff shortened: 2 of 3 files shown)


app/javascript/components/explore/ExploreView.jsx
```diff
@@ -28,6 +28,10 @@ function RoutableExploreTab({ studyAccession }) {
   async function loadStudyData() {
     const exploreResponse = await fetchExplore(studyAccession)
     setExploreInfo(exploreResponse)
+    // set window.SCP.isDEEnabled so that we can track DE visibility globally
```
ehanna4, line 31: For clarity could the comment say "so that we track Differential Expression (DE).." just so new folks can be sure what DE stands for

app/javascript/lib/metrics-api.js
```diff
@@ -392,13 +392,21 @@ function getTabProperty() {
  * @param {Object} props
  */
 export function log(name, props = {}) {
+  let isDEEnabled // track DE globally
```
ehanna4, line 395: Maybe it's overkill but maybe here too "track Differential Expression (DE) globally)

ehanna4 (approved):
Code looks good, the global approach seems robust and likely to allow us to capture lots of metrics for DE 
functional review works as advertised as well!

Just two small suggestions on the comments for clarity but overall looks good!

eweitz (changes requested):
This analytics instrumentation approach makes sense for now!  Thanks for the decisive background data, and the Slack discussion.  

I note one blocker with instrumentation naming.  Also, below, I outline our 1-on-1 discussion's summary on an important-but-not-yet-blocking concern with analytics validity around this subject.

**Instrumentation design**
Non-blocker: This instrumentation will soon likely be basically misleading. A [main requirement](https://broadworkbench.atlassian.net/browse/SCP-4717) of this work is to "track the % of users who saw the DE button who also clicked on it".  Only 45% of DE-enabled studies having "cell type" as default annotation is a big factor in that instrumentation design.  

In other words, given we [plan](https://broadinstitute.slack.com/archives/CBEHTH601/p1669676825309839?thread_ts=1669645954.274569&cid=CBEHTH601) to only show the DE button for cell type-equivalent annotations,  it seems _most_ studies with DE results soon will not show the DE button by default.  I haven't dug into it, but I hypothesize that most users who visit the Explore tab do not change annotations.  Thus, I'm concerned that A) we soon won't be able to use this work to answer the question "How many users who see the DE button also click it?", and -- moreso -- B) product managers will not realize that limitation, and think the opposite.

I don't consider this design concern blocking because, as is, it will only affect ~10% of cases per your data.  We agree it's essential to address before we restrict DE to "cell type", i.e. when it'll affect closer to 55% of cases.  To resolve that lurking concern, we'll likely need to make this property annotation-specific rather than study-specific.

**Instrumentation naming**
Blocker: see code suggestions below.  The code change would be trivial, but I think it's important and non-obvious / might warrant discussion.

app/javascript/lib/metrics-api.js
```diff
@@ -392,13 +392,21 @@ function getTabProperty() {
...
+  }
+
   props = Object.assign(props, {
     appId: 'single-cell-portal',
     appPath: getAnalyticsPageName(),
     appFullPath: getAppFullPath(),
     env,
     logger: 'app-frontend',
     scpVersion: version,
+    isDEEnabled,
```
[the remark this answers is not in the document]
bistline: [b836c36](https://github.com/broadinstitute/single_cell_portal_core/pull/1695/commits/b836c366a08ff706a35b608f5f12cccff02bc505)

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -55,6 +55,10 @@ const tabList = [
 function annotHasDe(exploreInfo, exploreParams) {
   const flags = getFeatureFlagsWithDefaults()
   if (!flags?.differential_expression_frontend || !exploreInfo) {
+    // set isDifferentialExpressionEnabled to false as user cannot see DE results, even if present for annotation
+    if (window.SCP) {
+      window.SCP.isDEEnabled = false
```
eweitz, line 60: Blocker: straggler

```suggestion
      window.SCP.isDifferentialExpressionEnabled = false
```

eweitz (approved):
(Note trivial blocker.)

app/javascript/components/explore/ExploreDisplayTabs.jsx
bistline: Ack I had changed that and apparently did a command-Z before committing... thanks for finding that.

bistline: [c8b8588](https://github.com/broadinstitute/single_cell_portal_core/pull/1695/commits/c8b858818738b41515d2995e2a5d6fcb4ab2591f)

PR #1698, opened 2023-01-03
Add Metadata extraction from AnnData functionality (SCP-4824)

ehanna4:
Changes: 
- Alters the way we pass flags for file extraction from AnnData files in line with the updates done in ingest pipeline. This affects cluster and metadata extraction
- Update the ingest pipeline to be the most recent version

Additions:
- Logic for metadata extraction from AnnData files

This adds the Rails side for the ingest pipeline update you can find here: https://github.com/broadinstitute/scp-ingest-pipeline/pull/287

To test:
Pre-reqs
- You will need a clean study with no files in it
- You will need to make sure your local ingest-pipeline configuration is not set or is set to the latest release of 1.24.0.
- You need to ensure you've got the feature flag for ingesting anndata set to true for your user and study
- download this h5ad file (and unzip it, just couldn't put here without compressing it): 
[compliant_pbmc3K.h5ad.zip](https://github.com/broadinstitute/single_cell_portal_core/files/10362369/compliant_pbmc3K.h5ad.zip)

Steps:
- Pull this branch
- Start up a local instance and also start up delayed jobs
- Go to your clean study that has no files in it
- Go to file upload and navigate to the AnnData upload
- Upload this file:
- Once the file completes go to the study details page
- You should see:
Cluster info:
Screen Shot 2023-01-06 at 10 59 53 AM
Metadata info:
Screen Shot 2023-01-06 at 10 59 47 AM
And the AnnData file:
Screen Shot 2023-01-06 at 11 54 17 AM

You can also check out your studies google bucket to see the individual fragments:
Screen Shot 2023-01-06 at 11 29 25 AM

https://user-images.githubusercontent.com/54322292/211059786-f5fbb124-3235-4062-aa3c-82f3d88c99eb.mov

PR #1701, opened 2023-01-09
Add parsed data cleanup for AnnData files (SCP-4870)

bistline:
#### BACKGROUND & CHANGES
This update applies the normal data cleanup policy to AnnData files.  This will remove any parsed data from MongoDB when a user deletes the parent AnnData file via `DeleteQueueJob`, including unsetting any derived values (like study defaults & counts).  This also fleshes out the `FactoryBot` entry for AnnData files to allow creating a more fully initialized study without having to manually creating database records.

#### MANUAL TESTING
1. Boot as normal (including DelayedJob) and sign in
2. Go to "My Studies" and open the upload wizard for any study that has a parsed AnnData file
3. Delete the file, then go back to "My Studies" and click "Details" for that study
4. Confirm there are no parsed clusters or annotation data for your study
5. (Optional) re-upload the AnnData file and confirm you can ingest data (note: you will only get clustering data until [metadata](https://github.com/broadinstitute/single_cell_portal_core/pull/1698) parsing is merged)

### DIFF
--- a/app/models/delete_queue_job.rb
+++ b/app/models/delete_queue_job.rb
@@ -26,8 +26,13 @@ class DeleteQueueJob < Struct.new(:object)
       file_type = object.file_type
       study = object.study
 
-      # remove expression_file_info, if present to avoid validation issues later
-      object.expression_file_info.destroy if object.expression_file_info.present?
+      # remove all nested documents if present to avoid validation issues later
+      # not all of them have validations but this guards against future errors in case they are added later
+      # :nested_attributes returns all :accepts_nested_attributes_for documents in StudyFile as a Hash
+      # e.g. { "expression_file_info_attributes"=>"expression_file_info_attributes=", ... }
+      StudyFile.nested_attributes.keys.map { |key| key.chomp('_attributes') }.each do |nested_association|
+        object.send(nested_association)&.destroy
+      end
 
       # now remove all child objects first to free them up to be re-used.
       case file_type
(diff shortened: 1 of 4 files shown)


test/factories/study_file.rb
```diff
@@ -139,6 +139,49 @@
     factory :ann_data_file do
       file_type { 'AnnData' }
       parse_status { 'parsed' }
+      transient do
+        # cell_input is an array of all cell names
+        # e.g.  ['cellA', 'cellB', 'cellC']
+        cell_input { [] }
+        # coordinate_input is a array of hashes of axes and values where the key is the name of the cluster
```
eweitz, line 146:
```suggestion
        # coordinate_input is an array of hashes of axes and values where the key is the name of the cluster
```

app/models/delete_queue_job.rb
```diff
@@ -92,6 +88,18 @@ def perform
...
+        delete_convention_data(study: study, metadata_file: object)
+        # delete user annotations first as we lose associations later
+        delete_user_annotations(study:, study_file: object)
+        delete_parsed_data(object.id, study.id, ClusterGroup, CellMetadatum, Gene, DataArray)
+        # reset default options/counts
+        study.reload
+        study.cell_count = study.all_cells_array.size
+        study.gene_count = study.unique_genes.size
+        study.default_options[:cluster] = nil
+        study.default_options[:annotation] = nil
```
eweitz, line 101: At a glance, this seems like it might only handle when the study has only 1 study file, an AnnData file?  E.g. if the user has a traditional set of metadata, clustering, and expression files _and_ an AnnData file, then presumably they could still have a non-nil default `cluster` and `annotation`.

eweitz (approved):
Code looks good.  I note a possible user-facing issue.  It's non-blocking as my concern is a bit speculative, but it seems worth considering.

app/models/delete_queue_job.rb
bistline: That is true - it does.  This was a conscious decision since we're currently only working on "happy path" for AnnData.  Currently, we would not be able to support a user uploading a traditional metadata file alongside an AnnData file.  We could support other clustering files though without any other updates, so I'll take a quick pass and making sure we wouldn't unset that if there was a default that didn't come from the AnnData file.

bistline: [1a65dcf](https://github.com/broadinstitute/single_cell_portal_core/pull/1701/commits/1a65dcf0241545adeed54578f9645491fef3c3fc)

ehanna4 (approved):
functional review, works as advertised deleting the data and AnnData file  👍

PR #1703, opened 2023-01-11
Fixing Explore tab rendering for AnnData files (SCP-4894)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where the Explore tab did not render the Scatter pane when data was parsed from AnnData files.  The issue was that our queries for clustering names only took into account files of type `Cluster`.

#### MANUAL TESTING
1. Boot as normal 
2. If you do not have a study with a parsed AnnData file, download and ingest [`compliant_pbmc3K.h5ad`](https://github.com/broadinstitute/single_cell_portal_core/files/10362369/compliant_pbmc3K.h5ad.zip) (from #1698)
3. Navigate to that study and validate that you can see data (shown below is the `X_umap` cluster w/ `louvain` annotation)
Screen Shot 2023-01-11 at 10 13 24 AM

### DIFF
--- a/app/lib/cluster_viz_service.rb
+++ b/app/lib/cluster_viz_service.rb
@@ -13,7 +13,7 @@ class ClusterVizService
 
   # helper method to load all possible cluster groups for a study
   def self.load_cluster_group_options(study)
-    non_spatial_file_ids = StudyFile.where(study: study, :is_spatial.ne => true, file_type: 'Cluster').pluck(:id)
+    non_spatial_file_ids = study.clustering_files.where(is_spatial: false).pluck(:id)
     ClusterGroup.where(study: study, :study_file_id.in => non_spatial_file_ids).pluck(:name)
   end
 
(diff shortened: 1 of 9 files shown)


app/lib/cluster_viz_service.rb
```diff
@@ -13,7 +13,7 @@ def self.get_cluster_group(study, params)
 
   # helper method to load all possible cluster groups for a study
   def self.load_cluster_group_options(study)
-    non_spatial_file_ids = StudyFile.where(study: study, :is_spatial.ne => true, file_type: 'Cluster').pluck(:id)
+    non_spatial_file_ids = study.clustering_files.where(is_spatial: false).pluck(:id)
```
eweitz, line 16: Nice catch!  Are queries involving `file_type: <Foo>` prevalent?  If so, I can imagine we'd need similar updates in a fair number of other places to account for AnnData files.

eweitz (approved):

app/lib/cluster_viz_service.rb
bistline: Great call - according to my IDE (assuming I got the regex right) there are ~20 occurrences of `StudyFile.where` that include `file_type` as a parameter.  I'll go ahead and review each one to determine if we need to amend the queries.

bistline: [0bd7122](https://github.com/broadinstitute/single_cell_portal_core/pull/1703/commits/0bd71220f6c0254dfe785e0ea7bba31baa91c539)

app/lib/summary_stats_utils.rb
```diff
@@ -43,7 +46,7 @@ def self.weekly_returning_users
     one_week_ago = today - 1.weeks
     user_count = User.where(:last_sign_in_at.gte => one_week_ago, :last_sign_in_at.lt => today)
                      .or(:current_sign_in_at.gte => one_week_ago, :current_sign_in_at.lt => today).count
-    {count: user_count, description: "Count of bleturning users from #{one_week_ago} to #{today}"}
+    {count: user_count, description: "Count of returning users from #{one_week_ago} to #{today}"}
```
ehanna4, line 49: nice catch

ehanna4 (approved):
functional review - works as advertised

PR #1704, opened 2023-01-11
Download DE results (SCP-4375)

ehanna4:
Adds the ability for a user to download the full DE results table.

https://user-images.githubusercontent.com/54322292/211916523-f614f653-9086-465d-965a-13e2c6579acb.mov

To test:

Pre-steps:
- You need a study that has DE results there are steps to follow to set this up from previous PRs: <link> <link>
- You need to make sure the study you are testing with locally has the DE feature flag flipped to true.

1. Pull this branch and start up your local environment
2. Go to the study that has DE results
3. Click the DE button and choose your gene group
4. You should see a download icon next to the info icon
5. Hovering over the download icon should produce a tooltip
6. Clicking on the download icon should result in a file download that is named like `<clustering name>--<annotation>--<group>--study--wilcoxon.tsv`

### DIFF
--- a/app/javascript/components/explore/DifferentialExpressionPanel.jsx
+++ b/app/javascript/components/explore/DifferentialExpressionPanel.jsx
@@ -71,9 +71,10 @@ export default function DifferentialExpressionPanel({
       <>
         15 most DE genes
         <a className="de-download-button"
-          onClick={async () => { await downloadBucketFile(bucketId, deFilePath)}}
-          title="Download all DE genes data for this group"
+          onClick={async () => {await downloadBucketFile(bucketId, deFilePath)}}
           data-analytics-name="differential-expression-download"
+          data-toggle="tooltip"
+          data-original-title="Download all DE genes data for this group"
         >
           <FontAwesomeIcon icon={faDownload}/>
         </a>
--- a/app/javascript/components/explore/DifferentialExpressionPanel.jsx
+++ b/app/javascript/components/explore/DifferentialExpressionPanel.jsx
@@ -6,7 +6,7 @@ import DifferentialExpressionModal from '~/components/explore/DifferentialExpres
 
 import DifferentialExpressionGroupPicker from '~/components/visualization/controls/DifferentialExpressionGroupPicker'
 import { logSearchFromDifferentialExpression } from '~/lib/search-metrics'
-import { fetchBucketFile } from '~/lib/scp-api'
+import { downloadBucketFile } from '~/lib/scp-api'
 
 
 /** Return selected annotation object, including its `values` a.k.a. groups */
@@ -71,9 +71,7 @@ export default function DifferentialExpressionPanel({
       <>
         15 most DE genes
         <a className="de-download-button"
-          onClick={async () => {
-            await fetchBucketFile(bucketId, deFilePath)
-          }}
+          onClick={async () => { await downloadBucketFile(bucketId, deFilePath)}}
           title="Download all DE genes data for this group"
           data-analytics-name="differential-expression-download"
         >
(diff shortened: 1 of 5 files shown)


app/javascript/styles/_explore.scss
```diff
@@ -209,3 +209,12 @@
...
 
+.de-download-button{
+	margin-right: 0.5em;
+	margin-left: 1em;
+	display: inline-block;
+	vertical-align: middle;
+	padding: 2px 5px;
+	border-radius: 4px;
+  }
+
```
eweitz, line 220: Could you refine this and the `.de-modal-raw-wrapper` CSS rule to use SCP-standard 2-space indentation?

bistline (approved):
Looks good - works as advertised with both public & private studies.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
```diff
@@ -60,24 +62,25 @@ export default function DifferentialExpressionPanel({
...
-            target="_blank"
-            data-analytics-name="differential-expression-download"
-            data-toggle="tooltip"
-            data-original-title="Download all DE genes data for this group"
-          >
-            <FontAwesomeIcon icon={faDownload}/></a> */}
-          <DifferentialExpressionModal />
-        </span>
+        <a className="de-download-button"
+          onClick={async () => {await downloadBucketFile(bucketId, deFilePath)}}
```
eweitz, line 74: Have you tried this without the async / await wrapper?  I.e.:

```suggestion
          onClick={downloadBucketFile(bucketId, deFilePath)}
```

Even though `downloadBucketFile` is an async function, I don't see off-hand why the onClick handler would need to be async too.  If it works without that, this would could slightly reduce visual noise / improve maintainability.

app/javascript/lib/scp-api.jsx
```diff
@@ -361,6 +361,37 @@ export async function fetchBucketFile(bucketName, filePath, maxBytes=null, mock=
...
+  // Convert data to a blob (standard for non-same-origin downloads)
+  const dataBlob = await data.blob()
+
+  // Create an element with an anchor link and connect this to the blob
+  const element = document.createElement('a')
+  element.href = URL.createObjectURL(dataBlob)
+
+  // name the file and indicate it should download
+  const filePathAsArr = filePath.split('/')
+  element.download = `${filePathAsArr.pop()}`
```
eweitz, line 385: `pop` ought to return a string, and we're not interpolating anything else, so it seems we can simplify to:

```suggestion
  element.download = filePathAsArr.pop()
```

For reference:
```
foo = 'abc/efg/123'
'abc/efg/123'
bar = foo.split('/')
(3) ['abc', 'efg', '123']
bar.pop()
'123'
```

eweitz (approved):
Code and demo look good!  Crisp video.  I note code maintainability nits, no blockers.

app/javascript/styles/_explore.scss
ehanna4: I can change those. I based the spacing for`.de-download-button` off of `.de-modal-row-wrapper`, do we have a linter or documented somewhere for the SCP-standard for indentation?

ehanna4: Actually I'm seeing a bit of differing indentation in the css file, I think if we want to standardize on 2-spaces that should be it's own work.
Screen Shot 2023-01-12 at 1 34 10 PM

eweitz: Makes sense!  So for future individual cases, let's use 2 spaces.  I agree that implementing that for all our SCSS is worth a separate task.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
ehanna4: It's most proper/standard to `await` (and therefore wrap that function with `async` or else you will get an `Uncaught SyntaxError: await is only valid in async function`) when calling an `async` function. I would argue that then makes this more maintainable since it makes it more obvious that `downloadBucketFile()` is an `async` function and is the most standard practice.

app/javascript/lib/scp-api.jsx
ehanna4: Yeah that seems safe, the reference code is nice thanks

PR #1705, opened 2023-01-12
Constrain possible test user email values in CI (SCP-4910)

bistline:
#### BACKGROUND & CHANGES
A recent review of a user cleanup process in SAM (Terra's workspace identity & access management service) found that SCP has been creating test user accounts with the email pattern of  `test.user.#{random_seed}@test.edu`, where `random_seed` is a 4-character alphanumeric string.  This means there could be over 1.6M different possible values for the email address (36^4 = 1,679,616).  Over the last several years, this has led to ~34K fake email accounts being invited to their "all users" Google group, despite the accounts not existing, and removing these accounts from the group resulted in upstream Google quota rate limits for SAM.  In an effort to avoid this moving forward, we've been asked to constrain the number of possible email values.  

This update will change the pattern to `test.user.#{n}@test.edu`, where `n` is an incrementing integer starting at `1`.  There are currently `109` calls from `FactoryBot` to create a user (though not all of them will make it to Terra as some are local mocks only).

### DIFF
--- a/test/factories/user.rb
+++ b/test/factories/user.rb
@@ -2,6 +2,7 @@
 # Rough performance timing in local (non-dockerized) development suggests that crating a user
 # using this factory takes ~0.1 seconds
 FactoryBot.define do
+  random_num = rand(1..100)
   factory :user do
     transient do
       random_seed { SecureRandom.alphanumeric(4).upcase }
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Elegant fix!  

I recall you noting this test had re-used a modest number (~5?) of test accounts like this before we expanded to the larger pool of 1.6M random seed email addresses.  I assume re-using addresses from this new ~100-count pool across runs is valid, even for tests that use production Terra as an external service.  

If we have any tests that use these random `test.edu` email addresses where we're testing something around _new users_, those tests might be worth another look to assess if reusing emails would make them invalid, e.g. a subtle false negative.  I doubt it warrants deep scrutiny, maybe just a quick second read if this hadn't crossed your mind.

N.B.: For our reference, here's a starter link for our [internal technical discussion](https://broadinstitute.slack.com/archives/C9YSEMSFR/p1673470120419579) on this.

ehanna4 (approved):

PR #1706, opened 2023-01-17
Show RNA and protein diagrams in related genes ideogram (SCP-4887)

eweitz:
This adds interactive RNA and protein diagrams to related genes ideogram.  It gives granular functional insight for genes.

### Overview 
Previously, tooltips shown upon hovering over a gene had only text: the gene's full name, genomic coordinates, and its relation to the searched gene.  Now, for protein-coding genes in human, diagrams are shown too.  Mature mRNA is at top, and aligned features in the corresponding protein are on bottom.  Hovering over parts of each diagram shows details, e.g. its length, and exon number or name.

#### Splice toggle
Clicking the scissors icon toggles the pre-mRNA into view.  While the _mature_ mRNA shown by default is basically equivalent to the cDNA molecule that is assayed in RNA-seq experiments, the pre-mRNA (i.e. preliminary mRNA) includes introns.

#### Alternative transcripts and isoforms
These diagrams exist for both canonical and alternative transcripts.  The [canonical transcript](https://useast.ensembl.org/info/genome/genebuild/canonical.html) is the default RNA transcript shown in traditional genome browsers.  Genes can also have other transcripts, e.g. due to alternative splicing or RNA editing.  These mRNA molecules and their corresponding proteins are often called "isoforms".  Different transcripts of the same gene are sometimes expressed in a cell type-specific manner.

### Use cases
These diagrams add value by giving granular functional insight and molecular references for genes.  For example:

* ***Easily learn more about what the gene does***.  Say you hover over a protein feature and see "GTP-binding domain".   Now you know the protein binds GTP!  Or maybe the protein feature notes it has kinase activity, and that's not deducible from the protein's full name.  Perhaps this helps you -- a biologist -- make other useful connections based on your scientific knowledge.  The new protein feature names convey key functional information about the gene you're looking at.

* ***Gauge known transcriptional complexity***.  Does your gene have 1 known transcript, or 10+?  Seeing many alternative transcripts in the "Transcripts" dropdown might warrant more care before assuming that your expression was due to the canonical transcript.

* ***Assess paralog similarity***.  Knowing that two genes are paralogs means knowing they are similar -- and now you can assess at a glance _how similar_ those genes are.  Does one gene have a much larger 3'-UTR?  Are their protein features basically the same, quite different, or somewhere in the middle?

There are potential future opportunities with sample data, e.g. integrating BAMs.  As is, though, these diagrams give unique and convenient biochemical context for genes.

Data on exons and transcripts comes from Ensembl, data on protein features comes from InterPro via Ensembl.

### <span id="video">Video</span>
Here's how it looks:

https://user-images.githubusercontent.com/1334561/212961840-1781d43f-9c1e-4bb3-95ec-2eb8c488aae7.mov

### Test
To manually verify:

1. **If local**, pull this branch, boot SCP via `bin/docker-compose-setup.sh`, navigate to a study that has human genes.  **If on staging**, Go to ["Human milk - differential expression"](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/human-milk-differential-expression#study-visualize) (SCP303).
2. Search for a gene
3. In related genes ideogram, hover over a gene
4. Confirm the tooltip shows RNA and protein diagrams
5. Play around with the new functionality, note anything that seems defective

Integrating this upstream Ideogram.js enhancement is a Developer's Choice Days project, tracked in SCP-4887.

### DIFF
--- a/app/javascript/components/visualization/RelatedGenesIdeogram.jsx
+++ b/app/javascript/components/visualization/RelatedGenesIdeogram.jsx
@@ -148,6 +148,9 @@ export default function RelatedGenesIdeogram({
 
   const verticalPad = 40 // Total top and bottom padding
 
+  // For Ideogram functionality only available for human
+  const showAdvanced = taxon === 'Homo sapiens'
+
   useEffect(() => {
     const ideoConfig = {
       container: '#related-genes-ideogram-container',
@@ -159,7 +162,9 @@ export default function RelatedGenesIdeogram({
       onClickAnnot,
       onPlotRelatedGenes,
       onWillShowAnnotTooltip,
-      showParalogNeighborhoods: taxon === 'Homo sapiens', // Works around bug in Ideogram 1.37.0, remove upon upgrade
+      showGeneStructureInTooltip: showAdvanced,
+      showProteinInTooltip: showAdvanced,
+      showParalogNeighborhoods: showAdvanced,
       onLoad() {
         // Handles edge case: when organism lacks chromosome-level assembly
         if (!genomeHasChromosomes()) {return}
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "core-js": "^3.6.4",
     "exifreader": "4.6.0",
     "fflate": "^0.7.3",
-    "ideogram": "1.40.0",
+    "ideogram": "1.41.0",
     "jquery": "3.5.1",
     "jquery-ui": "1.13.2",
     "morpheus-app": "1.0.18",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6885,10 +6885,10 @@ identity-obj-proxy@^3.0.0:
   dependencies:
     harmony-reflect "^1.4.6"
 
-ideogram@1.40.0:
-  version "1.40.0"
-  resolved "https://registry.yarnpkg.com/ideogram/-/ideogram-1.40.0.tgz#23768f49d6b3351dceca5e5acc3ac1983c0f4ed4"
-  integrity sha512-mnLjBQryo/OuW0YHa1ipBw85mkscBuqvMBhTqfiWUtcDKSEcibM+11KmrAXWKeSFp/WLtMH1Hw11bRPLDlrexw==
+ideogram@1.41.0:
+  version "1.41.0"
+  resolved "https://registry.yarnpkg.com/ideogram/-/ideogram-1.41.0.tgz#6bac8155c62dddee6df00a4dccb75329fd498522"
+  integrity sha512-+4ZTleo7aIsBBA/x5KRFJst3v5OX8arsaB1/1n2U/63mvyfEIdgdUC5jL5RycBx+DAlx+Io8cC2oVNKebFIIYA==
   dependencies:
     crossfilter2 "1.5.2"
     d3-array "^2.8.0"


bistline (approved):
Very cool - works as advertised from demo.  Adds a lot of rich information in a very compact UI.

My only nit is that tooltip on the button for "Unsplice exon (s)" takes a long time to load.  I know this is default behavior in all browsers for title tags on DOM elements, but it might be nicer if it could load faster somehow.  The reason I bring it up is that it is not immediately obvious what the button does - seeing it I would think that I was cutting content from somewhere, which didn't make sense.  Apologies if someone suggested this yesterday - I couldn't remember if it was brought up or not.

eweitz:
> tooltip on the button for "Unsplice exon (s)" takes a long time to load. I know this is default behavior in all browsers for title tags on DOM elements, but it might be nicer if it could load faster somehow.

Good point.  I use `title` in a few other places in the broader gene tooltip, but I agree that bare icons warrant faster text display on hover.  In the next Ideogram.js version I'll plan to add a tiny dependency to do that, or simply write the likely trivial code for it.

michelle-vel (commented):
Wow, the feature looks great! Looks great in the video

ehanna4 (approved):
Overall looks good and the hovering is nice and clean, fades nicely, the things are responsive, looks good! Only question I have is that Gene name at the top of the box appears to be a link but clicking it doesn't do anything. Is it supposed to do something when clicked? (see video attached)

https://user-images.githubusercontent.com/54322292/214084921-0057862a-f1b8-4874-afd8-f3d6342ea23f.mov

jlchang (commented):
Very cool! (I haven't watched the video of last week's demo yet so apologies if any of my comments are redundant)

Can the default state of the RNA diagram be the pre-RNA state? I'm not sure whether the scissor icon would need to change... The pre-RNA diagram (that has introns, exons and UTRs) is a very familiar viz in biology and somewhat specific to mRNA visualizations. While the mRNA diagram (FYI, not technically "protein" - for a protein "diagram" I'd expect to see a protein structure, which is quite a different viz), wasn't immediately recognizable because colored rectangles are used for many other UI functionalities. 

It seems if I toggle the scissors in one gene window, it stays using the pre mRNA view for all the rest. Is that intentional?

I'm not able to see the protein features visualization in my local instance - is this user error on my part?

Also, in my functional testing, the MTOR gene popup didn't show a dropdown for alternative transcripts that is show in the video - is that expected?

Regarding alternative transcripts, for MTX files that use both columns in the genes file AND are 10x v3 5' experiments, the data may have distinct alternative transcript data so, technically, the portal user could request a particular RNA isoform... I don't know how many of our studies have such info... We should probably look to see how we report that data when multiple isoforms are associated with the same gene name.

eweitz:
Thanks for the insightful scientific review, Jean.  

In our debugging session, we saw the local discrepancies were likely due to your instance running Ideogram.js 1.38.0 instead of 1.41.0.  Iterating with Docker on Intel Mac was very slow, and opted to do post-merge testing, especially given videos from Emily and me indicate the RNA and protein diagrams work as expected locally.

We also discussed pre-mRNA vs. mature mRNA as the default diagram.  The former is indeed more conventional.  We opted to begin by trying the latter as default, as it seems more useful since it gives a more coherent coordinate space to show aligned protein features.  We agreed the mature mRNA default would thus better help generate hypotheses, and so favored it.

Your notes on how some library preparation protocols capture preliminary mRNA were fantastic.  I hadn't appreciated that at all.  We concurred that something like library-aware display of pre- vs. mature-mRNA might be neat in the future, but isn't a blocker.

> Regarding alternative transcripts, for MTX files that use both columns in the genes file AND are 10x v3 5' experiments, the data may have distinct alternative transcript data so, technically, the portal user could request a particular RNA isoform... I don't know how many of our studies have such info... We should probably look to see how we report that data when multiple isoforms are associated with the same gene name.

This is a great point.  Your June 2022 analysis indicated the [prevalence of multiple transcripts per gene](https://docs.google.com/presentation/d/1SnOf3IS1lzdUgg-oKa8gp-Fc6QU_tV0M9U1aqj4RI2c/edit#slide=id.g13073d67193_0_118) is very low.  That said, perhaps there are a substantial number of studies that include transcript IDs, and only record 1 transcript, _and it is an alternative transcript_.  Analyzing questions like this indeed seems valuable.

In summary, per our chat, I'll go ahead and merge this now given the approvals, and we can easily revert if blockers are found when doing further scientific functional reviews on staging.

PR #1707, opened 2023-01-18
Changing default session timeout to 24h with configurable 15m opt-in (SCP-4627)

bistline:
#### BACKGROUND
The current timeout for session inactivity in SCP is 30 minutes, which was originally set as a compliance requirement.  We have since learned that this is not strictly necessary - Terra has a session timeout of 24 hours, for instance.  Increasing the timeout will allow for fewer user disruptions as they switch tasks and then come back to an active SCP window later.

#### CHANGES
This update increases the default timeout to 24 hours for all users.  In addition, there is a new "opt-in" timeout length of 15 minutes for users who do not want long sessions.  This can be useful for users with shared workstations.  This update also refines the inline form submission logic on the user profile page to be more consistent with shared functions, and reduces latency on updating the UI when toggling values (much longer in local development, but still a noticeable lag on production).  Lastly, it fixes broken code with rendering errors, and now only shows a modal if the profile update fails - successful saves have their state shown in the UI and as such do not require a "success" modal.

#### MANUAL TESTING (if desired)
1. Boot as normal and sign in
2. From the top-righthand menu, select "My profile"
3. Note that the default value for your session is now 24 hours
Screen Shot 2023-01-18 at 1 14 08 PM
4. Open Chrome DevTools and go to the Javascript console
5. Back in the UI, toggle the Session length switch and note that the console reads `Account update successfully recorded.` after a few seconds, confirming that the save was successful.
6. Toggle any other switch and confirm you see similar messages
7. (Optional) to test error rendering, open `app/controllers/profiles_controller.rb` in your IDE and update lines 43-44 to read:
```
update_params = user_params.to_unsafe_hash.merge(email: 'foo')
if @user.update(update_params)
```
8. (Optional) toggle the Session length switch again, and confirm you see the following modal:
Screen Shot 2023-01-18 at 1 14 25 PM
9. (Optional) confirm that the Session length switch reverted back to its previous state

### DIFF
--- a/app/views/profiles/show.html.erb
+++ b/app/views/profiles/show.html.erb
@@ -4,7 +4,7 @@
 
 <div id="profile-tab-root">
   <ul class="nav nav-tabs" role="tablist" id="profile-tabs">
-    <li role="presentation" class="profile-nav active" id="profile-emails-nav"><a href="#profile-emails" data-toggle="tab">Email Delivery</a></li>
+    <li role="presentation" class="profile-nav active" id="profile-emails-nav"><a href="#profile-emails" data-toggle="tab">Preferences</a></li>
     <% if @profiles_available %>
       <li role="presentation" class="profile-nav" id="profile-firecloud-nav"><a href="#profile-firecloud" data-toggle="tab">Terra Profile</a></li>
     <% else %>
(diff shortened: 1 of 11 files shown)


app/views/profiles/show.html.erb
```diff
@@ -4,29 +4,65 @@
...
-      <li role="presentation" class="profile-nav" id="profile-firecloud-nav"><a href="#profile-firecloud" data-toggle="tab">Terra Profile</a></li>
+      <li role="presentation" class="profile-nav" id="profile-firecloud-nav">
+        <a href="#profile-firecloud" data-toggle="tab">Terra Profile</a>
+      </li>
     <% else %>
-      <li role="presentation" class="profile-nav disabled" id="profile-firecloud-nav"><a href="#profile-firecloud" data-toggle="tooltip" title="FireCloud profiles are currently unavailable, please check back later.">Terra Profile</a></li>
+      <li role="presentation" class="profile-nav disabled" id="profile-firecloud-nav">
+        <a href="#profile-firecloud" data-toggle="tooltip"
+           title="FireCloud profiles are currently unavailable, please check back later.">Terra Profile</a>
+      </li>
```
eweitz, line 18: With the rebrand having happened a few years ago, I think it's worth updating "FireCloud" to "Terra" when it's in or around code being changed -- especially HTML, CSS, and JS.

```suggestion
      <li role="presentation" class="profile-nav disabled" id="profile-terra-nav">
        <a href="#profile-terra" data-toggle="tooltip"
           title="Terra profiles are currently unavailable, please check back later.">Terra Profile</a>
      </li>
```

app/views/profiles/show.html.erb
```diff
@@ -4,29 +4,65 @@
...
 <div id="profile-tab-root">
   <ul class="nav nav-tabs" role="tablist" id="profile-tabs">
-    <li role="presentation" class="profile-nav active" id="profile-emails-nav"><a href="#profile-emails" data-toggle="tab">Email Delivery</a></li>
+    <li role="presentation" class="profile-nav active" id="profile-emails-nav">
+      <a href="#profile-emails" data-toggle="tab">Preferences</a>
+    </li>
     <% if @profiles_available %>
-      <li role="presentation" class="profile-nav" id="profile-firecloud-nav"><a href="#profile-firecloud" data-toggle="tab">Terra Profile</a></li>
+      <li role="presentation" class="profile-nav" id="profile-firecloud-nav">
+        <a href="#profile-firecloud" data-toggle="tab">Terra Profile</a>
```
eweitz, line 12:
```suggestion
      <li role="presentation" class="profile-nav" id="profile-terra-nav">
        <a href="#profile-terra" data-toggle="tab">Terra profile</a>
```

app/views/profiles/update.js.erb
```diff
@@ -1,3 +1,15 @@
-var button = $("#toggle-admin-emails");
-var newText = "<%= @user.admin_email_delivery ? "On <i class='fas fa-fw fa-toggle-on'></i>".html_safe : "Off <i class='fas fa-fw fa-toggle-off'></i>".html_safe %>";
-button.html(newText);
\ No newline at end of file
+if (<%= @notice.present? %>) {
+  // modal unnecessary as toggle has already changed
+  console.log('<%= @notice %>')
+} else {
```
eweitz, line 4: Simpler, with less entangled JS and Ruby, and no unnecessary debug logging:

```suggestion
const isNoticePresent = <%= @notice.present? %>
if (!isNoticePresent) {
```

eweitz (approved):
This is a big win for usability!  It also really improves developer experience -- thank you.

Code looks good overall.  I suggest a bunch of trivial maintainability changes, none blocking.

app/views/profiles/show.html.erb
bistline: Ah great catch - totally glossed over all that.

app/views/profiles/update.js.erb
bistline: This is actually inverted from what we want - it should be `if (ifNoticePresent)`

eweitz: It looks non-inverted to me -- note how `if (!isNoticePresent) {` in the suggestion is where the `else` is in the original.  Omitting the original `if` clause also leaves out the unnecessary `console.log`, per SCP convention.

PR #1708, opened 2023-01-18
Truncate all GCP resource labels to 63 characters (SCP-4931)

bistline:
#### BACKGROUND & CHANGES
This update fixes a bug where filenames longer than 63 characters result in an invalid GCP resource label. This prevents the file from being ingested as the job cannot be queued due to `Google::Apis::ClientError:badRequest: Error: validating pipeline: invalid VM label value` being thrown.  This is not recoverable unless the user shortens the filename, which is completely opaque to the end user.  All resource labels are now truncated at 63 characters.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to the upload wizard for any study
3. Open a new terminal window, navigate to project source, and run the following command to create a cluster file with a filename longer than 63 characters:
```
cp test/test_data/cluster_example.txt ~/Desktop/$(openssl rand -hex 40).txt
```
4. Upload this new file as a clustering file, confirm the job queues in PAPI, and that you can see the truncated filename as a label on the virtual machine in `development.log`:
```
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
...

  :virtual_machine:
    :boot_disk_size_gb: 300
    :labels:
      :study_accession: scp52
      :user_id: 6033f4a8e2413917a1d0ddcd
      :filename: d8bfeb07f0299f75ed260e8adcf14279a24b87d224f93ed50ade9bddad56f6a
      :action: ingest_pipeline
...

Ingest run initiated: projects/broad-singlecellportal-bistlin/operations/10279437626310948347, queueing Ingest poller

```

### DIFF
--- a/app/models/papi_client.rb
+++ b/app/models/papi_client.rb
@@ -442,7 +442,7 @@ class PapiClient
     end
   end
 
-  # sanitizer for GCE label value (lowercase, alphanumeric with dash & underscore only, 64 characters)
+  # sanitizer for GCE label value (lowercase, alphanumeric with dash & underscore only, 63 characters)
   # see https://cloud.google.com/compute/docs/labeling-resources#requirements for more info
   #
   # * *params*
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good.

What happens when different GCP resources get the same label?  I suspect that'd merely degrade the granularity of cloud spend reporting, and have no operational effect.  Is that right?

ehanna4 (approved):
Code change seems very straightforward, thanks for adding the google documentation to the comment for easy reference

PR #1709, opened 2023-01-18
Reducing false positives for Codecov integration (SCP-4502)

bistline:
#### BACKGROUND & CHANGES
This update alters metrics for Codecov in an effort to reduce false positives on coverage calculations that fail PR checks.  The `patch` check has been disabled in favor of `project`, since global coverage is the published metric for this repository.  Additionally, an analysis of recently closed PRs with a failed Codecov check has shown that these failures were due to a drop in global coverage of ~0.05%.  The threshold for failing has now been increased to 0.1% to reduce these uninformative failed checks.

#### MANUAL TESTING
The checks section below should only display the `codecov/project` check, and it should pass since coverage is not affected by this update.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,3 +9,4 @@ coverage:
   project:
     default:
       threshold: 0.1%
+
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,8 +5,7 @@ comment:
   behavior: default
 coverage:
   status:
+    project:
+      default:
+        threshold: 0.1%
     patch: off
-  project:
-    default:
-      threshold: 0.1%
-
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,3 +3,9 @@ codecov:
 comment:
   layout: "reach, diff, files"
   behavior: default
+coverage:
+  status:
+    patch: off
+  project:
+    default:
+      threshold: 0.1%


eweitz (changes requested):
Nice signal-to-noise improvement!  Please update "false negative" to "false positive" in the title and description of this PR, though.

For some time, we've used "[false positive](https://github.com/broadinstitute/single_cell_portal_core/search?q=false+positive&type=issues)" to refer to tests that fail in the absence of a real bug.  This logic for software tests is borrowed from [medical tests](https://en.wikipedia.org/wiki/False_positives_and_false_negatives).  For example, if a medical test for a virus says you have the virus and you are not really infected, then that's a false positive.  Similarly, if a software test for a defect says you have a defect but it's not really a bug, then that's a false positive.

The principle here is that a "positive" result _reports_ a problem.  Whether there is _actually_ a problem is the "true" and "false" part.  Check out [others' thoughts on this](https://testfully.io/blog/false-positive-false-negative), if it's of interest.

Here's how the logic applies in the context of code coverage tests:

* True positive: Code coverage _did_ drop below threshold we care about, and was reported as such
* True negative: Code coverage _did not_ drop below threshold we care about, and was reported as such
* False positive: Code coverage _did not_ drop below threshold we do not care about, but was reported as dropping below it
* False negative: Code coverage _did_ drop below threshold we care about, but was reported as _not_ dropping below it

bistline:
Looks like I broke `codecov.yaml`:

```
Commit YAML is invalid
Coverage data is unable to be displayed, as the commit YAML appears to be invalid.
```

eweitz (approved):

michelle-vel (commented):
Interesting change! I see that this will change the threshold for failures to make the failed checks more reliable

ehanna4 (approved):
Seems like a smart change, I have indeed almost always ignored the Patch percentage when it was so small, this seems like it will improve the cognitive load of opening and approving a PR

PR #1713, opened 2023-01-24
Updating base Docker image, Rails to 6.1.7.1, automating local dev tasks (SCP-4938)

bistline:
#### BACKGROUND & CHANGES
This updates to the latest release of the base Docker image `gcr.io/broad-singlecellportal-staging/rails-baseimage:2.2.0`.  This contains normal security patches to the underlying Ubuntu 20.04 OS, as well as upgrading the Ruby version to 3.1.3.  Also, this updates the Rails version to `6.1.7.1` to address CVE-2023-22794.

Additionally, it was discovered while working on this update that the current `docker-compose` automation does not work terribly well when changes have been made to the upstream Docker image or locally to the Gemfile.  It necessitated rebuilding the image locally and tagging it correctly with `gcr.io/broad-singlecellportal-staging/single-cell-portal:development` before calling `bin/docker-compose-setup.sh`.  This borrows from some recent work in [scp-ingest-pipeline](https://github.com/broadinstitute/scp-ingest-pipeline/pull/290) to automatically pull in updates to the Docker image, as well as allowing the user to specify alternate image tags, or completely rebuild the image locally and use that instead if changes have been detected.  The latter option allows testing of this PR without any additional steps.  

Lastly, this also automates rebuilding the development image locally in CI for the same reasons as any upstream changes will cause builds to fail since the image is out of sync.  This will only happen if a change is detected to the base `Dockerfile` to keep build times as short as possible.

#### MANUAL TESTING
1. Pull this branch
2. From project source, run `bin/docker-compose-setup.sh` (this should automatically detect that the Dockerfile has changed and rebuild the image locally)
3. After the image builds and the Rails server & vite start, confirm you can bring up the home page
4. (Optional) Attach to the Rails console and confirm that you see the version of Rails has updated to `6.1.7.1` by running `docker exec -it single_cell bin/rails c` in a separate terminal:
```
Loading development environment (Rails 6.1.7.1)
irb(main):001:0> 
```


.github/workflows/run-all-tests.yml
```diff
@@ -19,6 +21,19 @@ jobs:
...
+        with:
+          files: Dockerfile
+      - name: Rebuild local image if Dockerfile changed
+        if: steps.changed-dockerfile.outputs.any_changed == 'true'
+        run: |
+          # the previous step uses the changed-files action to detect if the Dockerfile has any updates
+          # this will require rebuilding the image locally as the downstream "test" image will fail otherwise
+          # see https://github.com/marketplace/actions/changed-files for more info on this action
+          echo "#### CHANGES DETECTED TO DOCKERFILE, REBUILDING IMAGE LOCALLY ####"
+          docker build -t $DOCKER_IMAGE_NAME:development -f Dockerfile .
```
eweitz, line 36: Out-of-scope, but how much effort would it be to automatically `docker build` in local SCP via `docker-compose-setup.sh`?  It'd be convenient to not need to remember to manually `docker build` in cases of upstream change like this.

eweitz (approved):
Code looks good!  Manual tests pass:

```
wmce2-f16:single_cell_portal_core eweitz$ docker exec -it single_cell bin/rails c
Loading development environment (Rails 6.1.7.1)
irb(main):001:0> 
```

I suggest a possible future refinement.

.github/workflows/run-all-tests.yml
bistline: That's a very interesting idea!  Let me play around with this for a minute... I actually think this is very doable.

bistline: This does appear to be very straightforward:
```
LOCAL_BRANCH=$(git rev-parse --abbrev-ref HEAD)
CHANGED=$(git diff "$LOCAL_BRANCH" development --name-only -- Dockerfile)
```

This will compare your local branch with what the state of the Dockerfile is on `development`.  If there are any changes (staged or unstaged), then `$CHANGED` will be set to `Dockerfile`.  This can then be passed to the `if/else` to determine whether to build or pull the image.

Great suggestion!  I'm going to incorporate this and update the instructions.

bistline: @eweitz [f2d272d](https://github.com/broadinstitute/single_cell_portal_core/pull/1713/commits/f2d272de450394c3cacbd511dca1a6f00bfb2648)

eweitz:
Oh, also, for auditability, please add a Jira ticket ID to the PR title.

ehanna4 (approved):
Functional review - running this branch through docker compose locally I can indeed load up the portal

PR #1715, opened 2023-01-30
Restrict automated differential expression jobs to cell type annotations (SCP-4953)

bistline:
#### BACKGROUND & CHANGES
This update restricts the cases in which an automated differential expression job is launched to only annotations that are a cell type analog of some kind.  Previously, the restrictions for annotations were that they had to be group-based with between 2 and 250 unique values.  Now, we check the name of the annotation against the regular expression `/cell.*type/i` as well.  It should be noted that this is only for automated jobs - it is still possible to run differential expression jobs manually via the `DifferentialExpressionService.run_differential_expression_job` method.  This update is in service of two issues: 1) reducing the amount of "churn" in the form of needless differential expression jobs that end up getting deleted when the parent file is removed, and 2) restrict automated jobs to annotations that we believe are of higher relevance for differential expression analysis.  User interviews have indicated that cell type is a good candidate to begin with.  As we move forward, we may find other annotations (like Seurat clustering, for example) that may also be good proxies.

A check of the production database showed a count of 6941 for annotations matching our current rules.  Adding in the cell type restriction drops that to 661, a reduction of over 90%.

#### MANUAL TESTING
Note: if you do not have the `Human milk - differential expression` study loaded, you can download the [metadata file](https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-public/test/studies/SCP1671/MIT_milk_study_metadata.csv.gz;tab=live_object?project=broad-singlecellportal) and ingest it locally in a new study (we don't need clustering/expression for this test)
1. Pull branch and enter a Rails console session
2. Load the `Human milk - differential expression` study
```
```
3. Find all eligible annotations for the above study and confirm you only get back `General_Celltype` and `cell_type__ontology_label`
```
 DifferentialExpressionService.find_eligible_annotations(study)
=> 
[{:annotation_name=>"General_Celltype", :annotation_scope=>"study"},                                
 {:annotation_name=>"cell_type__ontology_label", :annotation_scope=>"study"}]
```

### DIFF
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -1,5 +1,9 @@
 # handle launching differential expression ingest jobs
 class DifferentialExpressionService
+
+  # regex for matching annotation names for possible cell type analogs
+  CELL_TYPE_MATCHER = /cell.*type/i
+
   # run a differential expression job for a given study on the default cluster/annotation
   #
   # * *params*
@@ -155,6 +159,7 @@ class DifferentialExpressionService
   end
 
   # find all eligible annotations for DE for a given study
+  # will restrict to cell type analog annotations
   #
   # * *params*
   #   - +study+        (Study) => Associated study object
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good.

> A check of the production database showed a count of 6941 for annotations matching our current rules. Adding in the cell type restriction drops that to 661, a reduction of over 90%.

Thanks for quantifying this impact -- 90% is a big change.

"Why restrict so much?", a reviewer might ask.  The reason is scientific accuracy -- to mitigate false positives.  A [2021 publication](https://www.nature.com/articles/s41467-021-25960-2) goes into detail.  [Internal survey data](https://docs.google.com/presentation/d/1ttcK5RFtM0iYMuFEzfb1j-0oZU9KAvA8khWAxh7I5-A/edit#slide=id.g189a6b3e7ce_0_26) is another factor.  [Recent discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1674833851434869?thread_ts=1674749942.977789&cid=CBEHTH601) gives additional context.

Hopefully the upcoming UI refinements will help preserve overall DE engagement in this much smaller but more scientifically accurate set of DE views.

bistline:
> "Why restrict so much?", a reviewer might ask. Expanding on the rationale you noted, the reason is scientific accuracy -- to mitigate false positives. A [2021 publication](https://www.nature.com/articles/s41467-021-25960-2) goes into detail. [Internal survey data](https://docs.google.com/presentation/d/1ttcK5RFtM0iYMuFEzfb1j-0oZU9KAvA8khWAxh7I5-A/edit#slide=id.g189a6b3e7ce_0_26) is another factor, as well as the user interviews you noted. [Recent discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1674833851434869?thread_ts=1674749942.977789&cid=CBEHTH601) gives more context.

Thanks for this!  Very relevant information.

ehanna4 (approved):
Functional review - thanks for including the metadata file for this as i do not have the Human milk study locally. I uploaded the metadata file and can confirm i saw:

> DifferentialExpressionService.find_eligible_annotations(study)
=> [{:annotation_name=>"General_Celltype", :annotation_scope=>"study"}, {:annotation_name=>"cell_type__ontology_label", :annotation_scope=>"study"}]

as expected and can also confirm that for a local study i have that i ran DE on manually i get different/more results as you noted in the description

PR #1716, opened 2023-02-02
Find any gene by name in differential expression table (SCP-4814)

eweitz:
This adds a search box to the DE table, which lets users find any gene in the table by its name.

Beyond seeing the 15 most differentially expressed genes, users can now quickly and easily learn if their gene of interest is significantly differentially expressed, and if so by how much.  Or they can see if genes with similar names to those listed by default -- i.e., often their paralogs -- are also very differentially expressed.

### Video <span id="video"></span>
Here's how it looks:

https://user-images.githubusercontent.com/1334561/216365576-b0ddce42-951e-47ba-98cb-b3f7ee2be99e.mov

### Analytics <span id="analytics"></span>
Instrumentation on this new search feature aligns with our other search usage metrics.  It lets us answer many questions, e.g.:
* How much usage does this feature get?
* What gene was searched in the DE table?
* Does usage differ by species?
* How many times did the user do this search on this page view?

Notably, while a search is done after each key press, a search is only logged as of 1 second after the last key press.  So typing e.g. "PIK" in relatively quick succession would log only 1 search, not 3, even though the UI technically updated instantly with 3 sets of search results.  We're really only interested in analyzing interactions where the user could reasonably be inferred to visually process the results -- not ultra-short inadvertent or in-transit interactions where the user does not look at what the UI responds with.  This improves the signal-to-noise ratio for our analytics.  The "delayed logging" technique introduced here might help denoise similar analytics, like scatter plot legend hover logging.

Check out the [Mixpanel report preview](https://mixpanel.com/s/2NikvP).  The same query will populate the [production Mixpanel report](https://mixpanel.com/s/Sq4YS) after release.  You can also find it from the [differential expression analytics board](https://mixpanel.com/s/2ZxI7z).

### Test <span id="test"></span>
An automated test confirms behavior.  Optionally, if you want to manually verify:
1.  **If local**, after pulling, since we're only testing frontend changes, try booting with the frontend service worker cache enabled for faster page loads: `bin/docker-compose-setup.sh -c`.  Then go to a study with DE results.  **If on staging**, go to ["Human milk - differential expression"](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/human-milk-differential-expression#study-visualize) (SCP303).
2. Click the blue "Differential expression" button
3. Select an annotation group from the drop-down
4. Type in the "Find gene" search box
5. Confirm results have a gene name that includes your queried text
6. Click the "x" to clear results
7. Open DevTools, go to Network panel
8. Add "method:POST bard" to the filter at upper left
9. In "Find gene", type 3 characters quickly, like a user typically would
10. In DevTools Network panel, confirm only 1 new post to Bard for "search" was sent

If you see an error in DevTools Console like the following, please disregard:
```
Warning: A component is changing an uncontrolled input of type radio to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (at DifferentialExpressionPanel.jsx:107)
    ...
```

This seems to be a pre-existing issue, and has likely existed since the DE table was created.  I spent much of a day trying to fix it without success.  It's reproducible for Emily and me on local SCP on `development`, but not staging with the same code.  It also has no user impact, so it seems reasonable to not invest more time here to resolve it.

This satisfies SCP-4814.


bistline (approved):
Looks good, and works as advertised.

As a follow-on, do we care about the number of matches found for reporting?  We're logging what the user entered as the search value, and I'm just curious if it would be worthwhile to know if/how many results were found.  Either way, it will be nice to see a report seeing how many people used search => click as opposed to just using the top 15 default genes.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
```diff
@@ -31,26 +36,146 @@ function initChecked(deGenes, checkedGene) {
...
+                      changeRadio(event)
+                    }}/>
+                  {deGene.name}</label></td>
+              <td>{deGene.log2FoldChange}</td>
+              <td>{deGene.pvalAdj}</td>
+            </tr>)
+        })}
+      </tbody>
+    </table>
+    <a href="https://forms.gle/qPGH5J9oFkurpbD76" target="_blank" title="Take a 1 minute survey">
```
ehanna4, line 104: I think this would benefit from an icon indicating that clicking it will open a new tab, the one that is the square with an arrow

app/javascript/components/explore/DifferentialExpressionPanel.jsx
```diff
@@ -67,71 +192,45 @@ export default function DifferentialExpressionPanel({
...
+        <div className="de-search-box">
+          <span className="de-search-icon">
+            <FontAwesomeIcon icon={faSearch} />
+          </span>
+          <input
+            className="de-search-input no-border"
+            name="de-search-input"
+            type="text"
+            autoComplete="off"
+            placeholder="Find gene"
```
ehanna4, line 206: Would "Find a gene" make more sense here to help suggest that it's only for searching for a single gene?

ehanna4 (approved):
Functional review works as advertised, and code looks good overall. I added two small suggestions though.

eweitz:
> do we care about the number of matches found for reporting? We're logging what the user entered as the search value, and I'm just curious if it would be worthwhile to know if/how many results were found.

It might be slightly valuable, but I couldn't get in working in 15-20 minutes, so I'm not sure it's worth the effort.  For example, a naive solution of just passing e.g. `numResults: genesToShow.length` reported the _previous_ search's number of results.  Maybe moving the logging into a `useEffect` would work, but I'd need to futz to preserve e.g. `trigger` with that approach.

Also, while we log number of results for global searches, I unfortunately can't think of when we've used those particular properties in interpreting usage or product planning.  So while existing `numResults` data may have some future value, it strikes me as OK to defer implementing here since it's not trivial.  I'd be happy to open a ticket if you'd like!

app/javascript/components/explore/DifferentialExpressionPanel.jsx
eweitz: This might be worth a design discussion.  Looking into this, I found:

* UK.GOV recommends not using the icon, for [reasons they explain](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/).
* Other single-cell products don't use the icon for such links, see e.g. [CELLxGene](https://cellxgene.cziscience.com/) and [10x Genomics](https://www.10xgenomics.com/).
* Major technology companies like [Google](https://blog.google/outreach-initiatives/diversity/lunar-new-year-2023/?_ga=2.25771709.1095051699.1675451359-452518495.1675451359), [YouTube](https://blog.youtube/creator-and-artist-stories/5-creator-ceos-entrepreneur-brand-building-on-youtube/), [Apple](https://www.apple.com/newsroom/2023/01/apple-music-launches-rihannas-road-to-halftime-ahead-of-super-bowl-lvii/) omit the icon in such links.
  * [Microsoft](https://blogs.microsoft.com/blog/2023/01/04/the-future-of-mobility-is-now-five-themes-to-watch-at-ces-2023/), [Amazon](https://aws.amazon.com/blogs/opensource/aws-teams-with-ostif-on-open-source-security-audits/), and [Facebook](https://about.fb.com/news/2023/01/instagram-quiet-mode-manage-your-time-and-focus/) don't open external links in new tabs, and their external links omit the icon
* Terra often omits it, as almost always does SCP.
* The link on production appears the same as this link; this PR doesn't change the link.

If we do add icons for links that open in new tabs, I suggest a ticket to implement it SCP-wide.  I'm not sure it'd be valuable though.  In my opinion, the icon is confusing.  I interpret it as "external link", not necessarily "new tab".  Research linked above finds users get similarly confused by the icon.  And although some web UIs use the icon to indicate links open in a new tab, it seems they don't do so consistently, and most web UIs omit the icon.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
eweitz: Nice refinement!  Done in 4452fb5.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
ehanna4: > The link on production appears the same as this link; this PR doesn't change the link.
Ah that's my bad, I thought this was updated in this PR, indeed a more holistic ticket/review of the use/omission of this icon for SCP seems more appropriate.

PR #1717, opened 2023-02-02
Patch for bug in Explore Tab after multi gene search and annotation & consensus toggling (SCP-4800)

ehanna4:
There is an issue with the data available in the ExploreTab when a user performs a multi-gene search and then attempts to toggle between Annotations and Consensus mechanisms. This results in a few errors mostly like "Cannot read properties of undefined (reading 'reduce')" from the plots attempting to update but not having good data to complete. I believe there are a few issues involved here around stale state and data however, this would require substantial refactor work of how these multiple plots and inputs interact. 

Instead here I am suggesting two updates that I believe patches this bug. 

The main section that is causing this issue comes from the `UseEffect` that is called when any plot data parameter updates.  Included in this `UseEffect` is a fetch to retrieve data. 

1. This fetch was previously called utilizing the javascript Promise construct of `.then()` but I am changing to use the construct of `await` instead. `.then()` and `await` are similar in that they are used to handle the result of a Promise however, a key difference is that `.then()` does not block the current execution (is more truly asynchronous) while `await` does block. In this case the fetch is being used to retrieve data and the next action is using this data, so it would be best to block and wait for this data to finish resolving before continuing. This is achieved using `await`. 

2. Retry on failure of good data. The fetch of the data in this `UseEffect` attempts to use the cached data if available, however in case this does not return good data I have added a "retry" to call the API for the data. Future work could certainly include digging into the cache code to fix any possibility of "bad" data from this, however, I believe this will be much more involved and I think it is best to have a fallback to the API data which, presumably, is the source of truth, in general.

To test: 
- Pull this branch and boot up a local server
- You do not need to sign in, navigate to a study with multiple genes and go to the explore tab
- Search for two genes
- Use the "Collapse by" menu and collapse to Mean or Median
- Change the annotation
- Now use the "Collapse by" menu again and choose whichever (mean or median) you hadn't chose before
- This should not result in an error
- Feel free to continue playing around with toggling annotations and consensus mechanisms to ensure no error comes up.
- Compare to Staging or Prod following these same steps to see the bug

Example error
Screen Shot 2023-02-03 at 3 32 03 PM

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -181,13 +181,9 @@ function RawScatterPlot({
       isSplitLabelArrays: isSplitLabelArrays ?? scatter.isSplitLabelArrays,
       isRefGroup: isRG
     })
-    // debugger
     if (isRG) {
       setCountsByLabel(labelCounts)
     }
-
-    //EMiLY the traces here are legit and coming thru not undefined in teh debug
-
     return traces
   }
 
@@ -212,10 +208,9 @@ function RawScatterPlot({
       const response = await fetchMethod(expressionParams)
       processScatterPlot(response)
     } catch (error) {
-      debugger
       setIsLoading(false)
-      // setShowError(true)
-      // setError(error)
+      setShowError(true)
+      setError(error)
     }
   }
 
@@ -329,7 +324,7 @@ function RawScatterPlot({
   function processScatterPlot(clusterResponse=null) {
     let [scatter, perfTimes] =
       (clusterResponse ? clusterResponse : [scatterData, null])
-      debugger
+    // debugger
     scatter = updateScatterLayout(scatter)
     const layout = scatter.layout
 
@@ -384,7 +379,7 @@ function RawScatterPlot({
 
       let expressionArray
 
-      // const fetchMethod = dataCache ? dataCache.fetchCluster : fetchCluster
+      const fetchMethod = dataCache ? dataCache.fetchCluster : fetchCluster
 
       // use an image and/or data cache if one has been provided, otherwise query scp-api directly
       if (
@@ -431,7 +426,8 @@ function RawScatterPlot({
         })
       } else {
         try {
-          const respData = await fetchCluster({
+          // attempt to fetch the data using the cache if that exists
+          const respData1 = await fetchMethod({
             studyAccession,
             cluster,
             annotation: annotation ? annotation : '',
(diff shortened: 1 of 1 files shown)


bistline (approved):
Interestingly, I can't trigger the error locally but can reproduce it on staging.  There might be some subtleties to this we haven't accounted for - all supporting your hypothesis that there are many different things contributing to this.

Regardless, it works - thanks for spending the time trying to tease this apart.

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -370,72 +368,97 @@ function RawScatterPlot({
...
-        expressionArray
-      }).then(processScatterPlot).catch(error => {
-        setIsLoading(false)
-        setShowError(true)
-        setError(error)
-      })
+      }
     }
-  }, [cluster, annotation.name, subsample, consensus, genes.join(','), isAnnotatedScatter])
+    getData()
```
eweitz, line 460: SCP JS conventionally includes "fetch" in a function name when it involves sending network requests for something (and when it might be cached).  This helps distinguish at a glance that the function doesn't simply look up a value from memory.

```suggestion
    fetchData()
```

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -370,72 +368,97 @@ function RawScatterPlot({
...
-    if (
-      flags?.progressive_loading && isGeneExpression(genes, isCorrelatedScatter) && !isAnnotatedScatter &&
-      !scatterData
-    ) {
-      const urlSafeCluster = cluster.replaceAll('+', 'pos').replace(/\W/g, '_')
-      const gene = genes[0]
-      const stem = '_scp_internal/cache/expression_scatter/'
-      const leaf = `${urlSafeCluster}/${gene}`
+    /** retrieve and process data */
+    async function getData() {
```
eweitz, line 372: Per other comment.

```suggestion
    async function fetchData() {
```

eweitz (approved):
Code looks good overall!  Caching issues are hard.

It seems this is hard to verify manually or automatically.  In that case, it might be worth keeping an eye on related analytics.  Could you add links in the description to related errors in Sentry and/or Mixpanel?

I'm curious if we'll see any effect on how often the "client plot data cache" gets leveraged. 
 I just created [Scatter plot: client plot cached vs. uncached](https://mixpanel.com/s/1ywnvI) to track that.  That measures how often plot data is retrieved from e.g. `dataCache.fetchCluster` vs. `fetchCluster` (excluding the very first scatter plot in the page load, which never is).

ehanna4:
We should hope to see this sort of error reduce: https://broad-institute.sentry.io/issues/3776283354/?project=1424198&query=is%3Aunresolved+reduce&referrer=issue-stream&statsPeriod=14d

Also here in Mixpanel: https://mixpanel.com/s/1kzw5x

PR #1723, opened 2023-02-10
Adjust GitHub Actions for branch deletion (SCP-4389)

ehanna4:
This fixes a papercut found where a hot-fix release branch would be automatically deleted after merging it to Master which is a problem since it was needed for merging to Development after. 

This PR splits our development and master merge actions and also suggests a new requirement for hot-fix branches
1. Merge to master will not delete a branch that starts with 'hot-fix' or 'hotfix' (also does not delete development or master)
2. Merge to development will delete any branch that is not master (or development)

The new requirement that I will add to the playbook if folks agree is that for a hot-fix you must name the branch starting with 'hot-fix' or 'hotfix' for the Actions to act appropriately. This practice appears to already be something many folks are doing informally, so this is more so just formalizing that practice. 

https://user-images.githubusercontent.com/54322292/218143345-edbdd445-b2a3-452f-b62b-b83a8db46276.mov

To test:
- Use this Repo: https://github.com/ehanna4/testing-actions
- I have made a variety of branches available there that include hotfix branches and non-hotfix branches
- Try out merging a hotfix branch to master and see that it doesn't delete
- Try out merging a hotfix branch to development and see that it does delete
- Try out merging development to master and see it does not delete
- etc

Alternatively you can take a look at the outputs from the already successful runs of the github actions in the Actions tab

### DIFF
--- a/.github/workflows/delete-branch-on-merge-to-master.yml
+++ b/.github/workflows/delete-branch-on-merge-to-master.yml
@@ -1,4 +1,4 @@
-name: Selective delete feature branch after merge to master
+name: Selectively delete feature branch after merge to master
 
 # Only run the workflow when a PR that targets master is closed
 on:
--- a/.github/workflows/delete-branch-on-merge-to-master.yml
+++ b/.github/workflows/delete-branch-on-merge-to-master.yml
@@ -1,9 +1,12 @@
-name: Delete feature branch after merge
+name: Selective delete feature branch after merge to master
 
-# Controls when the workflow will run
+# Only run the workflow when a PR that targets master is closed
 on:
   pull_request:
-    types: [ closed ]
+    types:
+      - closed
+    branches:
+      - 'master'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
@@ -19,7 +22,6 @@ jobs:
       uses: koj-co/delete-merged-action@master
       with:
       # do not delete master, development, or hot-fix-release branches
-      # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
         branches: "!master, !development, !hot-fix*, !hotfix*, *"
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/delete-branch-on-merge-to-master.yml
+++ b/.github/workflows/delete-branch-on-merge-to-master.yml
@@ -21,7 +21,7 @@ jobs:
     # from the github actions marketplace repo here: https://github.com/pabio/delete-merged-action
       uses: koj-co/delete-merged-action@master
       with:
-      # do not delete master, development, or hot-fix-release branches
-        branches: "!master, !development, !hot-fix*, !hotfix*, *"
+      # do not delete master, development, or hotfix-release branches
+        branches: "!master, !development, !*hotfix*, *"
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/delete-branch-after-pr-merge.yml
+++ b/.github/workflows/delete-branch-after-pr-merge.yml
@@ -19,7 +19,8 @@ jobs:
     # from the github actions marketplace repo here: https://github.com/pabio/delete-merged-action
       uses: koj-co/delete-merged-action@master
       with:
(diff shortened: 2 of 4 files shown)


.github/workflows/delete-branch-on-merge-to-master.yml
```diff
@@ -0,0 +1,27 @@
...
+  merge_job:
+    # this job will only run if the PR has been merged
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+    - name: Delete merged branch
+    # from the github actions marketplace repo here: https://github.com/pabio/delete-merged-action
+      uses: koj-co/delete-merged-action@master
+      with:
+      # do not delete master, development, or hot-fix-release branches
```
eweitz, line 24: Can we standardize on "hotfix" and not write it as "hot-fix"?  Hyphenating it as "hot-fix" seems quite rare -- e.g. Broad Slack has 3118 message hit for "hotfix", and 43 for "hot-fix".  

Accepting both in the code on the line below this one seems reasonable enough, but having differing conventions for how we write it would needlessly complicate e.g. naive string search.  For example, it'd be best for folks to only need to search "hotfix" and not also "hot-fix" if trying to find references to the subject via in-page find in a document, or via GitHub search in our PRs, etc.

```suggestion
      # do not delete master, development, or hotfix-release branches
```

.github/workflows/delete-branch-on-merge-to-master.yml
```diff
@@ -0,0 +1,27 @@
...
+        branches: "!master, !development, !hot-fix*, !hotfix*, *"
```
eweitz, line 25: Blocker: SCP conventionally starts all feature branch names with the author's initials.  Of the 64 search [hits for "hotfix" in this repo](https://github.com/broadinstitute/single_cell_portal_core/search?q=hotfix&type=issues), I haven't seen any examples of branch names beginning with "hotfix" in the handful of cases I glanced over.  So this new requirement strikes me as:

* quite feasible overlook,
* cause delays in urgent changes when that happens, and
* not of notable cost to omit

Regarding that last point, in other words, _not_ deleting a branch named e.g. `eh-hotfix-violin-plots` or `jb-registration-hotfix` seems basically fine.  The cost of not deleting it is very low -- some extra cruft among open branches that can be cleaned up in a few clicks.  The cost of deleting the branch unintentionally is notably higher -- it can (IIUC) substantially delay the release of an urgent fix.  

So relaxing the rules to not delete a branch -- e.g. don't delete any branches that have "hotfix" _anywhere_ in the name -- would reduce those expensive false positives (unintended deletes) for a negligible cost in false negatives (unintended keeps).

```suggestion
        branches: "!master, !development, !*hotfix*, !*hot-fix*, *"
```

eweitz (changes requested):
Thanks for this refinement!

I note one blocker for what seems like a new lurking release-process bug, as well as a few non-blocking maintainability suggestions.

.github/workflows/delete-branch-on-merge-to-master.yml
ehanna4: I didn't want to be overly prescriptive on the wording but if everyone agrees on "hotfix" as the single best term I am totally onboard with that. @bistline @jlchang @michelle-vel curious your opinions on "hotfix"

.github/workflows/delete-branch-on-merge-to-master.yml
ehanna4: Ah I am realizing I checked PR titles not branch titles when checking for how folks had been doing hotfix branches. Making the check for hotfix in the title rather than beginning with it makes sense to me I will make that update. 

As an aside I am not convinced this is a blocker and would push back on that/using request-changes for this update.

bistline (approved):
Looks good, modulo the discussion.  I agree with establishing the prerequisite of having `hotifx` in the branch name for this scenario.

.github/workflows/delete-branch-on-merge-to-master.yml
eweitz: Thanks for raising that concern on the "blocker" prefix and "request changes" status.  

My previous comment describes a functional problem that directly relates, which seems to warrant a "Blocker".  The fix itself did seem seem trivial, so I agree "Request changes" wasn't ideal!

eweitz (approved):

PR #1479, opened 2022-04-29
adding MIXPANEL SECRET to nginx and docker env (SCP-4327)

devonbush:
This correctly loads the new MIXPANEL_SECRET into the server environment

TO TEST:
<<probably much easier and more accurate to test by just waiting until this is merged and then testing on staging, but you can test this locally by>>
1. load SCP in dockerized mode (via `load_env_secrets.sh`)
2. go to a study, and click 'Usage stats' in 'Settings'
3. confirm the stats load, and no errors

PR #1724, opened 2023-02-13
Bring back study usage page! (SCP-4616)

ehanna4:
This adds back the links to the study usage page and restores the functionality

The fix ended up being super simple - apparently a small ruby syntax error? After barking up the wrong tree for a few days I did confirm Prod has the appropriate mixpanel secret and the mixpanel API works as expected and has had no updates since Devon added this feature.

To test
See this [PR](https://github.com/broadinstitute/single_cell_portal_core/pull/1479) about ensuring you've got the Env Variables loaded in your local
- pull this branch
- boot up a local server and sign into the portal
- navigate to a study and go to the study settings page
- click the usage link and wait for the page to actually load!

https://user-images.githubusercontent.com/54322292/218538304-46f94537-fc0f-452b-9305-c3f26e57df48.mov

### DIFF
--- a/app/controllers/api/v1/studies_controller.rb
+++ b/app/controllers/api/v1/studies_controller.rb
@@ -112,9 +112,8 @@ module Api
             where_string: "\"#{@study.accession}\" in properties[\"studyAccessions\"]"
           }
         }
-
         responses = Parallel.map(usage_stats_params.keys) do |key|
-          response = MixpanelClient.fetch_segmentation_query(usage_stats_params[key])
+          response = MixpanelClient.fetch_segmentation_query(usage_stats_params[key][:event],usage_stats_params[key][:type], usage_stats_params[key][:where_string] )
         end
         usage_stats = {}
         usage_stats_params.keys.each_with_index do |key, index|
--- a/app/javascript/components/my-studies/MyStudiesPage.jsx
+++ b/app/javascript/components/my-studies/MyStudiesPage.jsx
@@ -199,7 +199,7 @@ function StudyActionLinks({ study }) {
 
   const actionList = <ul className="list-style-none-menu">
     <li><a href={`/single_cell/studies/${studyId}`}>Details</a></li>
-    <li><a href={`/single_cell/studies/${studyId}/usage_stats`}>Usage stats <sup className="alert-warning">NEW</sup></a></li>
+    <li><a href={`/single_cell/studies/${studyId}/usage_stats`}>Usage stats</a></li>
     <li><a href={`/single_cell/studies/${studyId}/edit`}>Study settings</a></li>
     <li><a href={`/single_cell/studies/${studyId}/edit`}>Edit name &amp; description</a></li>
     <li><a href={`/single_cell/studies/${studyId}/upload`}>Upload &amp; edit files</a></li>
--- a/app/javascript/components/my-studies/MyStudiesPage.jsx
+++ b/app/javascript/components/my-studies/MyStudiesPage.jsx
@@ -199,6 +199,7 @@ function StudyActionLinks({ study }) {
 
   const actionList = <ul className="list-style-none-menu">
     <li><a href={`/single_cell/studies/${studyId}`}>Details</a></li>
+    <li><a href={`/single_cell/studies/${studyId}/usage_stats`}>Usage stats <sup className="alert-warning">NEW</sup></a></li>
     <li><a href={`/single_cell/studies/${studyId}/edit`}>Study settings</a></li>
     <li><a href={`/single_cell/studies/${studyId}/edit`}>Edit name &amp; description</a></li>
     <li><a href={`/single_cell/studies/${studyId}/upload`}>Upload &amp; edit files</a></li>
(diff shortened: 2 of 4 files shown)


app/controllers/api/v1/studies_controller.rb
```diff
@@ -112,9 +112,8 @@ def usage_stats
             where_string: "\"#{@study.accession}\" in properties[\"studyAccessions\"]"
           }
         }
-
         responses = Parallel.map(usage_stats_params.keys) do |key|
-          response = MixpanelClient.fetch_segmentation_query(usage_stats_params[key])
+          response = MixpanelClient.fetch_segmentation_query(usage_stats_params[key][:event],usage_stats_params[key][:type], usage_stats_params[key][:where_string] )
```
eweitz, line 116: You note:

> fix ended up being super simple - apparently a small ruby syntax error? 

Was this the line?  If so, intriguing!  [This line hasn't changed](https://github.com/broadinstitute/single_cell_portal_core/blame/a0943b993b8ccfdd13d087691b00c91e78566ddd/app/controllers/api/v1/studies_controller.rb#L117) since it was first written.  So I see two possibilities:

1.  This never worked.
2. It worked, but then something beyond this line changed.

[Analytics for "Usage stats"](https://mixpanel.com/s/3PO78c) make (2) seem likelier.  The linked chart shows [errors of type `ArgumentError` and text "wrong number of arguments (given 1, expected 0)"](https://mixpanel.com/s/9qCTR) starting on 2022-08-23, several months after this feature was [released](https://github.com/broadinstitute/single_cell_portal_core/releases/tag/1.15.0) on 2022-04-28.

In the [closest release preceding 2022-08-23](https://github.com/broadinstitute/single_cell_portal_core/releases/tag/1.24.0), we upgraded a major Ruby version, which strikes me as a plausible cause but not a smoking gun.  If it wasn't Ruby, then perhaps the API response from Mixpanel changed.

Knowing what caused the error would indicate the most fruitful test to prevent regression.  I don't consider that blocking, but it's at least food for thought.

eweitz (approved):
Code looks good!  Thanks for the demo video.

The fix evidently works, but got me pondering; see below.  I also suggest removing some outdated UI signage.

app/controllers/api/v1/studies_controller.rb
ehanna4: I think you are correct that it is option 2 that you have outlined. As noted in the description the Mixpanel API has not changed at all. I also remember testing this feature when Devon first introduced it and it worked so it can't be option 1. 
I believe the issue was something with the definition of `fetch_segmentation_query` since the error claimed to expect 0 arguments but the function header has like 7. I changed the way the default values are set to use `=` rather than `:` and changed the call to explicitly pass 3 of those arguments rather than passing a single object. I suspect you are correct that something occurred in the Ruby update that changed how object param de-structuring works here.

app/javascript/components/my-studies/MyStudiesPage.jsx
```diff
@@ -199,6 +199,7 @@ function StudyActionLinks({ study }) {
 
   const actionList = <ul className="list-style-none-menu">
     <li><a href={`/single_cell/studies/${studyId}`}>Details</a></li>
+    <li><a href={`/single_cell/studies/${studyId}/usage_stats`}>Usage stats <sup className="alert-warning">NEW</sup></a></li>
```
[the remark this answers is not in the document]
ehanna4: good call👍

app/controllers/api/v1/studies_controller.rb
ehanna4: Screen Shot 2023-02-13 at 3 38 29 PM
I tested out using that `:` syntax on an object arg like the old code in the console (aka no API hits or data) and received the same error so it does indeed seem to suggest something in Ruby changed that made this syntax style no longer valid

eweitz: Good find!

bistline: I believe that the culprit in this case is in fact the Ruby 3.1 update.  The main impetus for that was getting the new hash key literal expansion feature that was introduced in that version.  This allowed the `{ key_1:, key_2: }` shorthand.  I did notice while doing this update that many other methods in the codebase that used keyword-based arguments began to throw odd errors and required some refactoring to get them to work again.  It had to do with how the arguments were getting merged in (some really deep weirdness with the `**` operator and calling `merge!`).  Since `MixpanelClient.fetch_segmentation_query` took keyword arguments and we are passing them in as values from a nested hash, I think that's where the regression crept in.

Specifically, things like this worked pre-Ruby 3.1:
```
MyClass.some_method({ kwarg_a: 'foo', kwarg_b: 'bar' })
```

And now after the update only work like this:
```
MyClass.some_method(kwarg_a: 'foo', kwarg_b: 'bar')
```

I'm not absolutely positive on this but it seems pretty likely.  More context: https://rubyreferences.github.io/rubychanges/3.1.html#values-in-hash-literals-and-keyword-arguments-can-be-omitted

bistline (approved):

PR #1727, opened 2023-02-16
A/B test for showing Explore tab by default (SCP-4671)

bistline:
#### BACKGROUND & CHANGES
Now that we have a framework for running A/B tests, we can perform our first test by making the Explore tab the default active tab for some users and run the comparison in Mixpanel.  This update adds the convenience method `AbTest.override_for_feature?` to handle the business logic of determining when a particular user should see an updated UX.

The following checks are performed sequentially with the given outcomes ("current" is the Summary tab, "updated" is Explore tab):
1. If the `FeatureFlag` doesn't exist or the A/B test is not enabled, the user sees the current UX
2. The merged state of the `FeatureFlag` and any user overrides is checked - if either are `true` then the user will see the updated UX
3. The group assignment is checked for the `metrics_uuid` for this current user, and if they are assigned to the `intervention` group, they see the updated UX

This logic will apply to all subsequent A/B tests as well. Since we will always need to add code to handle the actual switching of the UX, this is a good example of how this framework makes that simpler.  Outside of the A/B test infrastructure (which ideally does not need to be updated moving forward), there is only the server-side logic of checking which tab to render by default:

`site_controller.rb`:
```
@explore_tab_default = @study.can_visualize? &&
                       AbTest.override_for_feature?('explore_tab_default', request.cookies['user_id'])
```

And then the front-end code to switch the active tabs:

`_study_tabs_nav.html.erb:`
```
<li role="presentation" class="study-nav <%= @explore_tab_default ? nil : ' active' %>" id="study-summary-nav">
```

`study.html.erb:`
```
<div class="tab-pane <%= @explore_tab_default ? nil : ' active in' %>" id="study-summary" role="tabpanel">
```

#### MANUAL TESTING
1. Boot normally with `bin/docker-compose-setup.sh` to start all services and run the migration to create the necessary feature flag
2. Sign in and go to the "[Feature flags](https://localhost:3000/single_cell/feature_flags)" page
3. Next to `explore_tab_default`, click the gear icon to configure the A/B test
4. Click "Create A/B Test"
5. Click the checkbox to enable the test and assign yourself to a particular group and save
6. In another tab, navigate to any study, and observe the following:
    - if you are in the `control` group, you see the "Summary" tab 
    - if you are in the `intervention` group, you see the "Explore" tab
7. Open the DevTools panel in Chrome, go to the network tab and refresh the page.  You should see an event for `page:view:site-study` and your `abTests` value should be `explore-tab-default-group-{name}`, where `name` is either `control` or `intervention`
Screen Shot 2023-02-16 at 12 09 45 PM

7. Back in the A/B test config page, click the `Group testing` tab and assign yourself to the other group
9. Back in the study page, refresh and confirm that you see the other tab by default, and in the corresponding Mixpanel event, your `abTests` assignment has changed as well

### DIFF
--- a/app/models/ab_test.rb
+++ b/app/models/ab_test.rb
@@ -50,7 +50,7 @@ class AbTest
   # will consider flag default state, possible user overrides, and A/B test state
   #
   # * *params*
-  #   - +flag_name+ (String)      => name of feature_flag
+  #   - +flag_name+ (String)     => name of feature_flag
   #   - +metrics_uuid+ (UUID)    => value of cookies['user_id'] for current user
   #   - +groups+ (Array<String>) => array of groups that qualify to override UX (default: 'intervention')
   #
--- a/app/models/ab_test.rb
+++ b/app/models/ab_test.rb
@@ -39,6 +39,35 @@ class AbTest
     where(enabled: true).map { |ab_test| ab_test.assignment(metrics_uuid) }.reject(&:flag_override?).map(&:tag)
   end
 
+  # determine if assignment for metrics_uuid qualifies for showing an updated feature
+  def override_feature?(metrics_uuid, groups: [])
+    return false unless enabled
+
+    groups.include? assignment(metrics_uuid).group_name
+  end
+
+  # determine what state a potential feature_flag and metrics_uuid combine to show
+  # will consider flag default state, possible user overrides, and A/B test state
+  #
+  # * *params*
+  #   - +flag_name+ (String)      => name of feature_flag
+  #   - +metrics_uuid+ (UUID)    => value of cookies['user_id'] for current user
+  #   - +groups+ (Array<String>) => array of groups that qualify to override UX (default: 'intervention')
+  #
+  # * *returns*
+  #   - (Boolean) => T/F if current merged state constitutes an updated UX
+  def self.override_for_feature?(flag_name, metrics_uuid, groups: %w[intervention])
+    user = User.find_by(metrics_uuid:)
+    feature_flag = FeatureFlag.find_by(name: flag_name)
+    # return false if flag isn't found or A/B test is not enabled
+    return false if feature_flag.blank? || !feature_flag.ab_test_enabled?
+
+    # FeatureFlaggable#merged_value_for will check the state of feature_flag default and any user overrides
+    # if that is false, then fall back to checking if user group assignment constitutes an override to the normal UX
+    ab_test = feature_flag.ab_test
+    FeatureFlaggable.merged_value_for(feature_flag.name, user) || ab_test.override_feature?(metrics_uuid, groups:)
+  end
+
   private
 
   # ensure consistent group name formatting - all lowercase word characters (plus dashes)
--- a/app/views/site/_study_tabs_nav.html.erb
+++ b/app/views/site/_study_tabs_nav.html.erb
@@ -1,27 +1,61 @@
 <ul class="nav nav-tabs"  data-analytics-name="study-overview" role="tablist" id="study-tabs">
-  <li role="presentation" class="study-nav active" id="study-summary-nav"><a href="#study-summary" data-toggle="tab">Summary <i class="far fa-fw fa-file-alt"></i></a></li>
+  <li role="presentation" class="study-nav <%= @explore_tab_default ? nil : ' active' %>" id="study-summary-nav">
+    <a href="#study-summary" data-toggle="tab">Summary <i class="far fa-fw fa-file-alt"></i></a>
+  </li>
   <% if @study.can_visualize? %>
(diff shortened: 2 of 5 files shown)


app/models/ab_test.rb
```diff
@@ -39,6 +39,35 @@ def self.load_assignments(metrics_uuid)
...
 
+  # determine if assignment for metrics_uuid qualifies for showing an updated feature
+  def override_feature?(metrics_uuid, groups: [])
+    return false unless enabled
+
+    groups.include? assignment(metrics_uuid).group_name
+  end
+
+  # determine what state a potential feature_flag and metrics_uuid combine to show
+  # will consider flag default state, possible user overrides, and A/B test state
```
eweitz, line 50: I had trouble understanding this docstring, as I initially read it all as a single sentence.  For clarity:

```suggestion
  # determine what state a potential feature_flag and metrics_uuid combine to show.
  # will consider flag default state, possible user overrides, and A/B test state
```

Prose-conventional uppercase ("Determine", "Will") would also help distinguish what's a new sentence vs. what's merely a line break, though maybe not totally necessary.

app/models/ab_test.rb
```diff
@@ -39,6 +39,35 @@ def self.load_assignments(metrics_uuid)
...
+  # * *returns*
+  #   - (Boolean) => T/F if current merged state constitutes an updated UX
+  def self.override_for_feature?(flag_name, metrics_uuid, groups: %w[intervention])
+    user = User.find_by(metrics_uuid:)
+    feature_flag = FeatureFlag.find_by(name: flag_name)
+    # return false if flag isn't found or A/B test is not enabled
+    return false if feature_flag.blank? || !feature_flag.ab_test_enabled?
+
+    # FeatureFlaggable#merged_value_for will check the state of feature_flag default and any user overrides
+    # if that is false, then fall back to checking if user group assignment constitutes an override to the normal UX
```
eweitz, line 66: Helpful, nice detail.

app/views/site/_study_tabs_nav.html.erb
```diff
@@ -1,27 +1,61 @@
 <ul class="nav nav-tabs"  data-analytics-name="study-overview" role="tablist" id="study-tabs">
-  <li role="presentation" class="study-nav active" id="study-summary-nav"><a href="#study-summary" data-toggle="tab">Summary <i class="far fa-fw fa-file-alt"></i></a></li>
+  <li role="presentation" class="study-nav <%= @explore_tab_default ? nil : ' active' %>" id="study-summary-nav">
+    <a href="#study-summary" data-toggle="tab">Summary <i class="far fa-fw fa-file-alt"></i></a>
+  </li>
   <% if @study.can_visualize? %>
-    <li role="presentation" class="study-nav" id="study-visualize-nav"><a href="#study-visualize" data-toggle="tab">Explore <i class="fas fa-fw fa-eye"></i></a></li>
+    <li role="presentation" class="study-nav <%= @explore_tab_default ? ' active' : nil %>" id="study-visualize-nav">
```
eweitz, line 6: This is quite simple, at its core!

app/views/site/study.html.erb
```diff
@@ -4,10 +4,10 @@
 <div id="tab-root">
   <%= render :partial => 'study_tabs_nav' %>
   <div class="tab-content top-pad">
-    <div class="tab-pane active in" id="study-summary" role="tabpanel">
+    <div class="tab-pane <%= @explore_tab_default ? nil : ' active in' %>" id="study-summary" role="tabpanel">
       <%= render partial: 'study_description_view' %>
     </div>
-    <div class="tab-pane" id="study-visualize" role="tabpanel">
+    <div class="tab-pane <%= @explore_tab_default ? 'active in' : nil %>" id="study-visualize" role="tabpanel">
```
eweitz, line 10: Can you remind me how these tabs from the tabs in `_study_tabs_nav.html.erb`?  Are both related to the Explore tab in Study Overview, or is there an Explore tab outside Study Overview that I'm forgetting?

eweitz (approved):
Code looks good!

app/models/ab_test.rb
bistline: Ah great point - I can do that moving forward.

app/views/site/study.html.erb
bistline: Yes, it is a little confusing - I had to remind myself while doing this.  The `_study_tabs_nav.html.erb` partial is just the navigation portion of the Study Overview page.  These `div` elements here are the content panes that get shown/hidden when the nav items are clicked.

ehanna4 (approved):
Functional review - works as advertised, was able to toggle between groups and saw the expected Overview or Explore tab show when I clicked the study

PR #1728, opened 2023-02-16
Create exitStatus field for Mixpanel (SCP-4948)

michelle-vel:
Adds a field to Mixpanel reports that will have an exitStatus, meaning a 0 if all works as expected and a 1 or specific error code if there was a failure

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -154,7 +154,7 @@ class IngestJob
       # expression matrices currently cannot be ingested in parallel due to constraints around validating cell names
       # this block ensures that all other matrices have all cell names ingested and at least one gene entry, which
       # ensures the matrix has validated
-      other_matrix_files = StudyFile.where(study_id: study.id, file_type: /Matrix/, :id.ne => study_file.id)
+      other_matrix_files = study.expression_matrices.where(:id.ne => study_file.id)
       # only check other matrix files of the same type, as this is what will be checked when validating
       similar_matrix_files = other_matrix_files.select {|matrix| matrix.is_raw_counts_file? == study_file.is_raw_counts_file?}
       similar_matrix_files.each do |matrix_file|
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -154,7 +154,7 @@ class IngestJob
       # expression matrices currently cannot be ingested in parallel due to constraints around validating cell names
       # this block ensures that all other matrices have all cell names ingested and at least one gene entry, which
       # ensures the matrix has validated
-      other_matrix_files = study.expression_matrices.where(:id.ne => study_file.id)
+      other_matrix_files = StudyFile.where(study_id: study.id, file_type: /Matrix/, :id.ne => study_file.id)
       # only check other matrix files of the same type, as this is what will be checked when validating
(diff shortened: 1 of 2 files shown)


app/models/ingest_job.rb
```diff
@@ -716,7 +732,8 @@ def get_job_analytics
       trigger: trigger,
       jobStatus: failed? ? 'failed' : 'success',
       machineType: vm_info['machineType'],
-      bootDiskSizeGb: vm_info['bootDiskSizeGb']
+      bootDiskSizeGb: vm_info['bootDiskSizeGb'],
+      exitStatus: exit_code #takes integer value from exit_code method
```
eweitz, line 736: This example would clarify for maintainers how `exitStatus` is useful beyond `jobStatus`:

```suggestion
      exitStatus: exit_code # integer exit code from PAPI, e.g. `137` for out of memory (OOM)
```

eweitz (approved):
Code looks good!

Could you create a Mixpanel report that uses this new property in our production analytics, link it from this PR, and add it to the "[Ingest - Single Cell Portal](https://mixpanel.com/project/2120588/view/19059/app/boards)" board?  Like report "[Ingest job status](https://mixpanel.com/project/2120588/view/19059/app/boards#id=734187&editor-card-id=%22report-11688551%22)", but break down by `exitStatus` rather than `jobStatus`.  Maybe even add a filter like "`exitStatus` does not equal `0`", as the motivating marginal value of this property is to report _why_ ingest (e.g. differential expression) jobs fail.

I was concerned to see no indication we'd verified this handles "137" errors -- e.g., it seems plausible that exit status code 137 could be thrown upstream in PAPI, but then get swallowed and presented as a `1`.  I searched around and found no concrete examples of how PAPI presents 137's.  However, the "BACKGROUND" section of #1572 talks in some detail about it, mostly allaying my concern.

So it'd be nice to at least manually test this handles "137" / out of memory (OOM) errors as we expect, as that seems like the chief value of this work.  Jean might have a suitable example.  But I suspect this handles that properly, so that's not a blocking concern for me.

ehanna4 (commented):
It looks like the test failure is related to these changes, I think there need to be updates to the tests for `ingest_job.rb` 

Looking at the test failure I see:

> *** Rails test failures ***
> 
>   IngestJobTest#test_should_gather_job_properties_to_report_to_mixpanel:
>   NoMethodError: unmocked method :done?, expected one of [:error, :metadata]
> Exiting with code: 1
> Error: Process completed with exit code 1.

bistline:
> I was concerned this may well not log "137" / out of memory (OOM) errors, which seems like the chief value of this work. E.g., it seems plausible that exit status code 137 could be thrown upstream in PAPI, but then get swallowed and presented as a `1`. 

@eweitz when I was helping out with this PR, I did some investigation regarding the `exit_code` for pipelines.  From what I could tell there is only one status reported, and it is done so at the metadata event level, _not_ the pipeline level.  This _should_ correctly report `137` when the ingest pipeline process runs out of memory.  We did correctly see `65` for non-convention metadata files when we tried to validate them against the Alexandria convention, so it holds that OOM will be reported correctly as well.

eweitz:
Thanks Jon, that updates my position on a manual test for 137 from "nice but non-blocking" to "likely unworthwhile".

bistline (approved):
Worked with Michelle to fix the mock errors in the tests - looks good otherwise!

PR #1737, opened 2023-03-02
Adding configurable RAM limits for docker-compose in local development (SCP-5011)

bistline:
#### BACKGROUND & CHANGES
Previously, local Docker instances would occasionally die with the error message `Killed: single_cell exited with code 137`.  This is due to Docker not being able to manage two instances that are consuming more RAM than is currently allocated to the Docker engine.  It can manage a single instance without setting limits, but multiple instances running concurrently will result in Docker killing one of the instances when the limit is exceeded.

This update adds configurable memory limits for both Docker containers when running locally via `docker-compose`.  They default to 2 & 6GB for the `single_cell_vite` and `single_cell` containers, respectively.  These limits can also be changed when calling `bin/docker-compose-setup.sh` via the `-p` and `-v` flags:
```
-p  {PORTAL_RAM_GB}  specify as integer the amount of RAM in GB for the single_cell container
-v  {VITE_RAM_GB}  specify as integer the amount of RAM in GB for the single_cell_vite container
```

Note: if you do not have at least 8GB of RAM allocated to your Docker engine, you should either increase this from the "Settings > Resources" menu in the Docker app, or set your own limits via the shell script.  1 & 4GB for the vite and portal instances is sufficient, though more is obviously better.

#### MANUAL TESTING
1. Boot via `bin/docker-compose-setup.sh`
2. Once both containers have started, in a separate terminal window run the following commands to confirm that memory limits have been enforced:
```
$ docker inspect single_cell | jq '.[0].HostConfig.Memory'
6442450944

$ docker inspect single_cell_vite | jq '.[0].HostConfig.Memory'
2147483648
```

### DIFF
--- a/bin/docker-compose-cleanup.sh
+++ b/bin/docker-compose-cleanup.sh
@@ -6,8 +6,8 @@
 
 # suppress 'variable is not set' warning, actual values are irrelevant
 export GCR_IMAGE="gcr.io/broad-singlecellportal-staging/single-cell-portal:development"
-export PORTAL_RAM_GB="6gb"
-export VITE_RAM_GB="2gb"
+export PORTAL_RAM_GB="6"
+export VITE_RAM_GB="2"
 echo "### REMOVING CONTAINERS/VOLUMES ###"
 # set VITE_FRONTEND_SERVICE_WORKER_CACHE to silence warnings from docker-compose
 VITE_FRONTEND_SERVICE_WORKER_CACHE="$VITE_FRONTEND_SERVICE_WORKER_CACHE" \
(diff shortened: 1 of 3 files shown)


bistline:
> When I run `bin/docker-compose-setup.sh -p PORTAL_RAM_G` the portal does not start up and says "docker.errors.DockerException: Failed converting the string value for memory (PORTAL_RAM_GB) to an integer." Am I running that wrong? If I do `bin/docker-compose-setup.sh -p 11` for instance though it does work and I can see the memory limit has increased

Apologies - those instructions may have been confusing.  You do need to pass a value as an integer for the number of gigabytes of RAM for each container if you want to override the defaults.  Docker was complaining because `PORTAL_RAM_GB` is the name of the variable we're interpolating, not its value.  I'll update the instructions to be clearer.

ehanna4 (approved):
Thanks for the clarification, looks good!

eweitz (approved):

PR #1738, opened 2023-03-07
Restore fast native local development environment (SCP-4997)

eweitz:
This describes how to run SCP natively, i.e. without Docker, which makes local page load times drastically faster.

### Background
Previously, local SCP did not work natively.  It needed Docker, which abstracted away problems with essential dependencies that emerged while shifting from old Intel- to new M1-CPU Macs and from Ruby 2 to 3.  Docker was key to enabling work on new development machines.  

But Docker was quite slow.  As most of the team shifted to new machines by January 2023, early reports of slowness became reproducible more broadly and clear in [development analytics](https://docs.google.com/presentation/d/1UoA9uZZjv6tDobx2VPDVAv-_PxCq9QJUf9rm-I-RGVg/edit#slide=id.g20f5ae807d5_0_10).  Local page load times were 2-3x slower in early Q1 '23 than in Q4 '22, coinciding with the shift to M1 and Docker.

### Impact
Now, local SCP can run natively on M1.  It's snappy.  Analytics show [native](https://mixpanel.com/s/1kbL6R) local M1 SCP is 5-10x faster than in [Docker](https://mixpanel.com/s/1edR57).  

Even more than saving 5-20 seconds per page load, the ergonomic impact is in preserving mental context.  In other words, local speeds now much better preserve _human working memory_ across page loads.  That improves productivity.

In summary, this restores our development environment to how it was before [September 2022](https://docs.google.com/presentation/d/1UoA9uZZjv6tDobx2VPDVAv-_PxCq9QJUf9rm-I-RGVg/edit#slide=id.g20f5ae807d5_0_30).  Now we can also take advantage of fast new M1 machines.

### Video
Here's a video showing local native SCP.  Note the speed of booting the three main services -- `rails s`, `bin/vite dev`, and `rails jobs:work`.  Each is nearly instant.  The equivalent in Docker took me > 20 seconds (presumably because it does more, e.g. installs dependencies).

Initial full load of the home page takes 9 seconds.  Subsequent full page load takes 5 seconds.  The equivalent in Docker took me > 70 seconds and > 24 seconds, respectively.  So empirical spot measurements show full page load speedup of > 4-7x.

https://user-images.githubusercontent.com/1334561/223493169-13702c5a-c91d-4389-9b52-5925a43822c9.mov

### Test

Follow instructions in NON_CONTAINERIZED_DEV_README.md.

### Further details
The [demo presentation](https://drive.google.com/file/d/1Q6Ya2xLWA_mwlqbSrGxBn9SieQpiHFZR/view?t=34m25s) has more context.  [Slides](https://docs.google.com/presentation/d/1UoA9uZZjv6tDobx2VPDVAv-_PxCq9QJUf9rm-I-RGVg/edit) from a related topic two weeks ago give deeper background, including local performance [milestones](https://docs.google.com/presentation/d/1UoA9uZZjv6tDobx2VPDVAv-_PxCq9QJUf9rm-I-RGVg/edit#slide=id.g20f5ae807d5_0_5), and [graphs](https://docs.google.com/presentation/d/1UoA9uZZjv6tDobx2VPDVAv-_PxCq9QJUf9rm-I-RGVg/edit#slide=id.g20f5ae807d5_0_10) showing systematic local slowdown that motivates this work

Jon gave key technical insight.  Emily, Jean, and Michelle have all successfully installed native local SCP with these instructions.  Their review has helped correct and refine this work.

This satisfies SCP-4997.

### DIFF
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -4,7 +4,11 @@ Developing on SCP natively, i.e. outside a Docker container, is less robust but 
 
 ## SETUP
 Commands below assume your CPU is Apple Silicon / M1, not Intel.
-1. Append `CFLAGS="-Wno-error=implicit-function-declaration"` to your `~/.zshrc` (if using Z shell macOS default in Terminal, or `~/.bash_profile` if Bash), then run `source ~/.zschrc`
+1. Append the following to your `~/.zshrc` (if using Z shell macOS default in Terminal, or `~/.bash_profile` if Bash), then run `source ~/.zschrc`:
+```
+# Avoid compilation errors with occasional Ruby gems, e.g. bson_ext, puma
+CFLAGS="-Wno-error=implicit-function-declaration"
+```
 2. Run `rbenv -v`.  If it is not found, `brew install rbenv`, then `rbenv init` to set up rbenv in your shell, then close terminal and reopen it.
 3. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
   - If it outputs contains "OpenSSL 1.1.1", go to Step 4.  
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -78,6 +78,15 @@ new shell.
 NOTE: If you ever use the `bin/run_tests.sh` script locally, this will write out and delete any shell env files 
 after completion.  You will need to run `./ruby_local_setup.rb` again to repopulate them.
 
+Once you're all set up, you'll want to run each of the following commands in a separate terminal (or tab of your terminal):
+* `rails s # Run Rails app server`
+* `bin/vite dev # Run Vite -- enables HMR for reloading JS / CSS changes instantly`
+* `rails jobs:work # Run Delayed Job worker, to enable Ingest Pipeline jobs that parse uploaded files`
+
+Here's what starting up a regular development session in native local SCP looks like:
+
+https://user-images.githubusercontent.com/1334561/223493169-13702c5a-c91d-4389-9b52-5925a43822c9.mov
+
 ## KNOWN ISSUES
 1. Developing outside the docker container inherently runs more risk that your code will not work in the docker environment in staging/production.  Be careful!  If your changes are non-trivial, confirm your changes work in the containerized deploy before committing (especially changes involving package.json and/or the Gemfile)
 2. You may experience difficulty toggling back and forth between containerized and non-containerized deployment, as node-sass bindings are OS-specific.  If you see an error like 'No matching version of node-sass found'
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -7,7 +7,7 @@ live css/js reloading, faster build times, and byebug debugging in rails.
 Commands below assume your CPU is Apple Silicon / M1, not Intel.
 1. Run `rbenv -v`.  If it is not found, `brew install rbenv`, then `rbenv init` to set up rbenv in your shell, then close terminal and reopen it.
 2. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
-  - If it outputs contains "OpenSSL 1.1.1", go to Step 4.  
+  - If it outputs contains "OpenSSL 1.1.1", go to Step 3.  
   - Else, if output contains something like "OpenSSL 3.0.7", install the needed version of OpenSSL, and install Ruby so that it compiles with the needed version of OpenSSL:
      - Run `brew install openssl@1.1`
      - Run `rbenv uninstall 3.1.3`
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -6,14 +6,14 @@ live css/js reloading, faster build times, and byebug debugging in rails.
 ## SETUP
 Commands below assume your CPU is Apple Silicon / M1, not Intel.
 1. Run `rbenv -v`.  If it is not found, `brew install rbenv`, then `rbenv init` to set up rbenv in your shell, then close terminal and reopen it.
-2. Run `ruby -v` to ensure Ruby 3.1.3 is installed on your local machine.  If not, run `rbenv install 3.1.3` (current [as of 2023-01-26](https://github.com/broadinstitute/single_cell_portal_core/pull/1713)) then `rbenv global 3.1.3`.
-3. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
+2. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
   - If it outputs contains "OpenSSL 1.1.1", go to Step 4.  
   - Else, if output contains something like "OpenSSL 3.0.7", install the needed version of OpenSSL, and install Ruby so that it compiles with the needed version of OpenSSL:
      - Run `brew install openssl@1.1`
      - Run `rbenv uninstall 3.1.3`
      - Run `LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib" CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include" CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" rbenv install 3.1.3`
      - Run `ruby -v` to ensure Ruby 3.1.3 is installed
+3. Run `ruby -v` to ensure Ruby 3.1.3 is installed on your local machine.  If not, run `rbenv install 3.1.3` (current [as of 2023-01-26](https://github.com/broadinstitute/single_cell_portal_core/pull/1713)) then `rbenv global 3.1.3`.
 4. Run `bundler -v` to ensure Bundler is installed.  If not, `gem install bundler`.
 5. Run `node -v` to ensure Node is installed. If not, install via https://nodejs.org/en/download/
 6. Run `yarn -v` to ensure Yarn is installed. If not, install via `brew install yarn`
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -20,7 +20,8 @@ Commands below assume your CPU is Apple Silicon / M1, not Intel.
 7. `cd` to where you have the `single_cell_portal_core` Git repo checked out.
 8. Run `bundle install`
   - That might fail with a message that contains "An error occurred while installing bson_ext"
-  - If so, follow steps from https://stackoverflow.com/a/64248633, i.e.:
+  - If so, run: ``gem install bson_ext -v '1.5.1' --source 'https://rubygems.org/' -- --with-cflags="-Wno-error=implicit-function-declaration"``
+  - If that fails, follow steps from https://stackoverflow.com/a/64248633, i.e.:
     - Run `cd /Users/$(whoami)/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/bson_ext-1.5.1/ext/cbson`
     - Insert one line, near to the top of the file, `bson_buffer.h`:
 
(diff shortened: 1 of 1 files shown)


NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,96 @@
...
 
 ## SETUP
+Commands below assume your CPU is Apple Silicon / M1, not Intel.
+1. Run `rbenv -v`.  If it is not found, `brew install rbenv`, then `rbenv init` to set up rbenv in your shell, then close terminal and reopen it.
+2. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
+  - If it outputs contains "OpenSSL 1.1.1", go to Step 3.  
+  - Else, if output contains something like "OpenSSL 3.0.7", install the needed version of OpenSSL, and install Ruby so that it compiles with the needed version of OpenSSL:
+     - Run `brew install openssl@1.1`
+     - Run `rbenv uninstall 3.1.3`
+     - Run `LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib" CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include" CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" rbenv install 3.1.3`
```
bistline, line 13: FWIW I think it's safe to have the values for `LDFLAGS`, `CONFIGURE_OPTS`, and `RUBY_CONFIGURE_OPTS` pre-loaded in your shell.  I've been doing this for years without any issue.  That being said, I am still on an Intel chip.  Curious what your thoughts are about having to declare these during install.

My main reason for wanting them in the shell is that I suspect whenever we have to install newer versions of some gems, it will fail with similar errors on M1 machines unless these configuration options are set.  It specifically will address the `bson_ext` error you outlined below.

This is what I have exported in my shell (paths will be different for other people but you get the idea).
```
export CFLAGS="-Wno-error=implicit-function-declaration"
export MAKE_OPTS=-j1 RUBY_CONFIGURE_OPTS="--with-openssl-dir=/usr/local/opt/openssl@1.1 --with-readline-dir=$(brew --prefix readline) --with-libyaml-dir=$(brew --prefix libyaml)"
```

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,96 @@
...
-10. Add `config/local_ssl/localhost.crt` to your system's trusted certificates. On macOS, you can drag this file into the 
-keychain access app, use the "System" keychain, and the "Certificates" category. Then you will likely need to open the 
-newly added certificate in the keychain access app and update the "Trust" setting to be "Always Trust" rather than "Use 
-System Defaults".
-11. Run `rails s`
-12. If you're developing JS, for hot module replacement and live reload, in a separate terminal, run `bin/vite dev`
-13. If you are working on functionality that involves delayed jobs, like uploading data:
+11. Run the source command the script outputs -- this will export those needed variables into the current shell
+12. Add `config/certs/localhost.crt` to your system's trusted certificates. 
+  -  Automatic route (preferred), run `sudo security add-trusted-cert -d -r trustAsRoot -k /Library/Keychains/System.keychain config/certs/localhost.crt`
```
bistline, line 58: Nice!

bistline (approved):
Great work!  Questions about configuration options being pre-loaded.

michelle-vel (approved):
Works great for me! Thanks for creating this, it's a huge help with using SCP locally

NON_CONTAINERIZED_DEV_README.md
jlchang: So we could suggest the shell modification here in this README (maybe where we're setting CFLAGS="-Wno-error=implicit-function-declaration")? Do those env changes apply to any other build processes (ie. are there other places in our onboarding docs where it should also be noted? or is there a "higher level" process that this shell mod belongs?)

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,97 @@
-# Developing on SCP without a Docker container
+# Developing on SCP natively, without a Docker container
 
-Developing on SCP without a Docker container, while less robust, opens up some faster development paradigms, including 
-live css/js reloading, faster build times, and byebug debugging in rails.
+Developing on SCP natively, i.e. outside a Docker container, is less robust but faster.  In native mode, full pages load are 4-7x faster, tests run faster, and rich [byebug debugging](https://github.com/deivid-rodriguez/byebug/blob/master/GUIDE.md) in Rails is possible.
 
 ## SETUP
+Commands below assume your CPU is Apple Silicon / M1, not Intel.
+1. Append `CFLAGS="-Wno-error=implicit-function-declaration"` to your `~/.zshrc` (if using Z shell macOS default in Terminal, or `~/.bash_profile` if Bash), then run `source ~/.zschrc`
```
jlchang: Let's note that this CFLAGS declaration is specific to the bson_ext issue (if we see other build issues, it might be good to know that this one is not a general use environment setting when troubleshooting)

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,97 @@
...
+2. Run `rbenv -v`.  If it is not found, `brew install rbenv`, then `rbenv init` to set up rbenv in your shell, then close terminal and reopen it.
+3. Run `ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'`.  
+  - If it outputs contains "OpenSSL 1.1.1", go to Step 4.  
+  - Else, if output contains something like "OpenSSL 3.0.7", install the needed version of OpenSSL, and install Ruby so that it compiles with the needed version of OpenSSL:
+     - Run `brew install openssl@1.1`
+     - Run `rbenv uninstall 3.1.3`
```
jlchang, line 17: Just checking that this step is a no-op if ruby 3.1.3 has never been installed?

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,97 @@
...
+     - Run `LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib" CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include" CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)" rbenv install 3.1.3`
+     - Run `ruby -v` to ensure Ruby 3.1.3 is installed
```
jlchang, line 19: This line is duplicative with the first step of the next line and can be removed.

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,97 @@
...
+4. Run `ruby -v` to ensure Ruby 3.1.3 is installed on your local machine.  If not, run `rbenv install 3.1.3` (current [as of 2023-01-26](https://github.com/broadinstitute/single_cell_portal_core/pull/1713)) then `rbenv global 3.1.3`.
```
jlchang, line 20: Perhaps "check current Ruby version used in production docker image" in
https://github.com/broadinstitute/single_cell_portal_core/blob/development/Dockerfile
Look for `<version>`
```
 # Set ruby version
RUN bash -lc 'rvm --default use ruby-<version>'
```

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,97 @@
...
+5. Run `bundler -v` to ensure Bundler is installed.  If not, `gem install bundler`.
+6. Run `node -v` to ensure Node is installed. If not, install via https://nodejs.org/en/download/
+7. Run `yarn -v` to ensure Yarn is installed. If not, install via `brew install yarn`
+8. `cd` to where you have the `single_cell_portal_core` Git repo checked out.
+9. Run `bundle install`
+  - That might fail with a message that contains "An error occurred while installing bson_ext"
```
jlchang, line 26: Suggestion: move the two troubleshooting options into the Troubleshooting section at the bottom of this doc

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,101 @@
...
+    - Run `gem list bson_ext`, and confirm it outputs:
+
+```
+*** LOCAL GEMS ***
+bson_ext (1.5.1)
+```
+  -
+    - Change directories back to `single_cell_portal_core`
+10. Run `yarn install`
+11. Run `./rails_local_setup.rb` to will write out required variables into an shell env file (using your Broad username
```
jlchang, line 59: For my new M1, I had to go searching for info on how to set up Vault. Here's the link to the DevOps doc:
https://docs.google.com/document/d/11pZE-GqeZFeSOG0UpGg_xyTDQpgBRfr0MLxpxvvQgEw/edit#heading=h.trehvi2yr632
For onboarding, do steps 5-8. For migrating to a new laptop, skip step 7 and copy over your GitHub Personal Access Token.

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,101 @@
...
-newly added certificate in the keychain access app and update the "Trust" setting to be "Always Trust" rather than "Use 
-System Defaults".
-11. Run `rails s`
-12. If you're developing JS, for hot module replacement and live reload, in a separate terminal, run `bin/vite dev`
-13. If you are working on functionality that involves delayed jobs, like uploading data:
+12. Run the source command the script outputs -- this will export those needed variables into the current shell
+13. Add `config/certs/localhost.crt` to your system's trusted certificates. 
+  -  Automatic route (preferred), run `sudo security add-trusted-cert -d -r trustAsRoot -k /Library/Keychains/System.keychain config/certs/localhost.crt`
+  - Manual route (alternative): On macOS, drag the certificate file into the Keychain Access app, use the "System" keychain, and the "Certificates" category. Then left click on the newly added certificate in the Keychain Access app, click "Get Info", then toggle open the "Trust" section, then set "When using this certificate" to "Always Trust" rather than "Use System Defaults"; and ensure "Always Trust" also gets set on all other drop-down menus in the "Trust" section.
```
jlchang, line 64: Perhaps this can be moved to Troubleshooting and used to check on the cert's status and settings if the automated route fails?

NON_CONTAINERIZED_DEV_README.md
```diff
@@ -1,49 +1,101 @@
...
-10. Add `config/local_ssl/localhost.crt` to your system's trusted certificates. On macOS, you can drag this file into the 
-keychain access app, use the "System" keychain, and the "Certificates" category. Then you will likely need to open the 
-newly added certificate in the keychain access app and update the "Trust" setting to be "Always Trust" rather than "Use 
-System Defaults".
-11. Run `rails s`
-12. If you're developing JS, for hot module replacement and live reload, in a separate terminal, run `bin/vite dev`
-13. If you are working on functionality that involves delayed jobs, like uploading data:
+12. Run the source command the script outputs -- this will export those needed variables into the current shell
+13. Add `config/certs/localhost.crt` to your system's trusted certificates. 
+  -  Automatic route (preferred), run `sudo security add-trusted-cert -d -r trustAsRoot -k /Library/Keychains/System.keychain config/certs/localhost.crt`
```
jlchang, line 63: Blocker: Please update -r trustAsRoot to -r trustRoot
Both options are mentioned at https://apple.stackexchange.com/questions/80623/import-certificates-into-the-system-keychain-via-the-command-line but only trustRoot worked

jlchang (changes requested):
Thanks for working this out, Eric! Very happy to be able to run locally again :)

Mostly suggestions, one blocker (trivial update of command line param).

NON_CONTAINERIZED_DEV_README.md
jlchang: Ah okay - I wasn't aware of the previous instances where `export CFLAGS="-Wno-error=implicit-function-declaration"` had resolved other gem install issues. Thanks for the additional context!

NON_CONTAINERIZED_DEV_README.md
eweitz: Good catch.  I [copied that from a Slack comment](https://broadinstitute.slack.com/archives/CBEHTH601/p1678203008450679?thread_ts=1677782368.756229&cid=CBEHTH601).  Could you edit that comment so folks who stumble onto it use the working argument, and I'll plan to edit this doc as part of the PR that's opened in this repo?

PR #1747, opened 2023-03-23
Reliably load differential expression table, and other bucket data (SCP-4964)

eweitz:
This fixes a bug where the differential expression table would often break.  It improves reliability for some other features, too.

## Bug
Previously, the DE table broke for almost [1 in 5 users who tried it](https://docs.google.com/presentation/d/1JIxltvjxyCLIo_zN1PaGs2KdF9RQEINMUUvPdT5ubmw/edit#slide=id.g1dfdc26ed34_0_64).  Although frequent in real usage, the bug is elusive.  The [problem](#problem) only occurs if the user loads a new DE table an hour or more after having loaded the page.  (This made the cause of the bug hard to diagnose with certainty, and the fix lengthy to fully test.)  The workaround is to refresh the page.  However, the bug was completely unhandled, so the user just saw a weirdly empty area where the DE table should be.  A message like "Refresh page to see this data" could convey the workaround, but would still be a rather poor user experience.

## Fix
Now, the DE table is much more reliable.  No page refresh needed!  The [solution](#solution) is to renew the access token that the frontend JavaScript client uses to fetch data from the study's GCS bucket.  The 1-hour token is renewed 5 minutes before it expires.  This access token renewal is secure because it happens not only while the access token is still valid, but also within the context of a valid, longer-lived refresh token and (for signed-in users) a 24-hour authentication session.

The fix [applies to 2 of 3 scenarios](https://github.com/broadinstitute/single_cell_portal_core/pull/1747#authentication-timeout-scenarios) for authentication timeout, which represent the vast majority of DE table usage.  The unresolved 3rd scenario is still fairly common, but different in cause and presentation.  This work has [clarified](https://broadworkbench.atlassian.net/browse/SCP-5064) that prevalent and elusive problem.

Diagrams below give more detail on the problem, the solution, and [how it integrates with existing authentication mechanics](#detail).

[AppSec approved](https://broadinstitute.slack.com/archives/CADU7L0SZ/p1679688996621989?thread_ts=1679685000.179099&cid=CADU7L0SZ) this as secure.

## Problem <span id="problem"></span>
Unreliable_but_secure_client-side_bucket_access__SCP_2023-03-24

## Solution <span id="solution"></span>
Reliable_and_secure_client-side_bucket_access__SCP_2023-03-24

## Detail <span id="detail"></span>
Access_token_grants_for_reliable_and_secure_client-side_bucket_access__SCP_2023-03-24

<span id="authentication-timeout-scenarios"></span>
## Authentication timeout scenarios
Below are details on the 3 broad cases that can manifest auth timeouts, where the user loads a DE table an hour or more after having loaded the page.

| Public? | Signed in? | DE table fixed? | Staging example | Production example |
| --- | --- | --- | --- | --- |
| Yes | No | Yes | [SCP344](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP344/#study-visualize) | [SCP1671](https://singlecell.broadinstitute.org/single_cell/study/SCP1671#study-visualize) |
| Yes | Yes | [No](https://broadworkbench.atlassian.net/browse/SCP-5064) | [SCP344](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP344#study-visualize) | [SCP1671](https://singlecell.broadinstitute.org/single_cell/study/SCP1671#study-visualize) |
| No | Yes | Yes | [SCP340](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP340#study-visualize) | [SCP2165](https://singlecell.broadinstitute.org/single_cell/study/SCP2165#study-visualize)|

Per above, the unfixed scenario is when the user is signed in, on a public study, and hasn't refreshed the page in over an hour.  The bug seems related but orthogonal to the GCS bucket auth bug that broke the DE table.  More details are in SCP-5064.

The private studies required special setup to be representative of those created by real users; they're created with a non-admin account so as to be unaffected by a related bug that only affects admins, SCP-5033.

## Test
A truly valid manual test requires > 1 hour and intricate setup.  I've done the test a few times now.  In my opinion, that effort is not worth the value of others reporting confirmation.  [Monitoring this error in Sentry](https://broad-institute.sentry.io/issues/3782541638/?project=1424198) after deployment would confirm the fix works in production.

Nonetheless, in case others consider it worthwhile, manual test instructions are documented below.  

* Create a public study using `mouse_brain` synthetic data.
* Upload raw counts, processed matrix, metadata, and cluster file
* Using a non-admin account like "[registeredinscpandterra](https://broadinstitute.slack.com/archives/CBEHTH601/p1600365349124700?thread_ts=1600365317.124600&cid=CBEHTH601)", create a private study using `mouse_brain` synthetic data.
* Upload raw counts, processed matrix, metadata, and cluster file
* Load 3 pages:
  * A) In your normal Chrome window, ensure you're signed in and go to the private study
  * B) In your normal Chrome window, ensure you're signed in and go to the public study
  * C) Open an incognito Chrome window and go to the public study
* Refresh each of the 3 pages in series, quickly, to ensure they finish loading at roughly the same wall-clock time
* In each of the 3 pages: 
  * Click "Differential expression", select an annotation with DE, select a group
  * Ensure DE table loads
  * Open Console panel in Chrome DevTools, paste in `window.SCP.readOnlyTokenObject`, hit enter
  * Ensure you see an object with properties `accessToken`, `expiresIn`, and `expiresAt`
  * Set your phone's alarm to 7 minutes before the time in `expiresAt`.  (Alarm should be < 55 minutes from now.)
  * Go to the Network panel in Chrome DevTools.
  * Do something else until your alarm sounds.  Don't refresh the page.
  * When your alarm sounds, look at the Network panel, confirm no `renew_token` request has been sent
  * Wait 1-3 minutes
  * Confirm `renew_token` request has been sent since your alarm sounded
  * Select a different group in the DE dropdown menu
  * Confirm DE table does not break

An automated test verifies an important part of handling the above.

Page (B) is the unfixed authentication scenario, i.e. "Public: Yes, Signed in: Yes", so it will fail, though it might take 2 hours instead of 1.  As noted above, SCP-5064 tracks that issue.

This satisfies SCP-4964.

### DIFF
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -160,8 +160,12 @@ module Api
       end
 
       def renew_token
-        Rails.logger.info "Renewing token via SCP API for user #{user.id} in study #{@study.accession}"
-        render json: DifferentialExpressionService.get_read_access_token(@study, current_api_user, renew=true)
+        renewing_user = "an unauthenticated user (via read-only service account)"
+        if current_api_user
+          renewing_user = "user #{current_api_user.id}"
+        end
+        Rails.logger.info "Renewing token via SCP API for #{renewing_user} in study #{@study.accession}"
+        render json: RequestUtils.get_read_access_token(@study, current_api_user, renew: true)
       end
 
       swagger_path '/site/studies/{accession}/download' do
--- a/app/controllers/api/v1/site_controller.rb
+++ b/app/controllers/api/v1/site_controller.rb
@@ -9,7 +9,8 @@ module Api
       before_action :set_analysis_configuration, only: [:get_analysis, :get_study_analysis_config]
       before_action :check_study_detached, only: [:download_data, :stream_data, :get_study_analysis_config,
                                                   :submit_study_analysis, :get_study_submissions,
-                                                  :get_study_submission, :sync_submission_outputs]
+                                                  :get_study_submission, :sync_submission_outputs,
+                                                  :renew_token]
       before_action :check_study_view_permission, except: [:studies, :analyses, :get_analysis]
       before_action :check_study_compute_permission,
                     only: [:get_study_analysis_config, :submit_study_analysis, :get_study_submissions,
@@ -125,8 +126,42 @@ module Api
         end
       end
 
-      def view_study
+      swagger_path '/site/studies/{accession}/renew_token' do
+        operation :get do
+          key :tags, [
+              'Site'
+          ]
+          key :summary, 'Renew a soon-expiring GCS access token for a study'
+          key :description, 'Get a new 1-hour access token, within the authentication session duration'
+          key :operationId, 'site_study_renew_token_path'
+          parameter do
+            key :name, :accession
+            key :in, :path
+            key :description, 'Accession of Study to renew access for'
+            key :required, true
+            key :type, :string
+          end
+          response 200 do
+            key :description, 'Access token for Google Cloud Storage'
+          end
+          response 401 do
+            key :description, ApiBaseController.unauthorized
+          end
+          response 403 do
+            key :description, ApiBaseController.forbidden('view study')
+          end
+          response 404 do
+            key :description, ApiBaseController.not_found(Study)
+          end
+          response 406 do
+            key :description, ApiBaseController.not_acceptable
+          end
+        end
+      end
 
+      def renew_token
+        Rails.logger.info "Renewing token via SCP API for study #{@study.accession}"
+        render json: DifferentialExpressionService.get_read_access_token(@study, current_api_user, renew=true)
       end
 
       swagger_path '/site/studies/{accession}/download' do
@@ -196,8 +231,8 @@ module Api
         rescue RuntimeError => e
           ErrorTracker.report_exception(e, current_api_user, @study, params.to_unsafe_hash)
           MetricsService.report_error(e, request, current_api_user, @study)
-          logger.error "Error generating signed url for #{params[:filename]}; #{e.message}"
-          render json: {error: "Error generating signed url for #{params[:filename]}; #{e.message}"}, status: 500
+          logger.error "Error generating signed URL for #{params[:filename]}; #{e.message}"
+          render json: {error: "Error generating signed URL for #{params[:filename]}; #{e.message}"}, status: 500
         end
       end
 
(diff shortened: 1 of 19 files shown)


app/lib/differential_expression_service.rb
```diff
@@ -308,4 +308,27 @@ def self.log_message(message)
     puts message
     Rails.logger.info message
   end
+
+  # Get token for client-side access to GCS bucket, depending on study privacy
+  # Context: https://github.com/broadinstitute/single_cell_portal_core/pull/1747
+  def self.get_read_access_token(study, user, renew=false)
```
bistline, line 314: Blocker: while this particular issue is scoped to the differential expression results table, the _actual_ underlying issue is much broader than that.  It can affect any client-side fetch of a remote object, not just DE files.  As such, this method doesn't really belong in the `DifferentialExpressionService` class.

That being said, we don't really have an obvious place to put this code outside of a controller.  I could see it being placed in `app/lib/request_utils.rb`, or possibly a new "service" class (though that may be overkill but possibly not).

Also, see the previous comment re: kwargs and boolean arguments.

app/views/studies/_shared_sync_functions.js.erb
```diff
@@ -1,7 +1,8 @@
 // shared JS functions for sync forms
 
 // Used for sync CSFV
-window.SCP.readOnlyToken = '<%= get_read_access_token(@study, current_user) %>'
+window.SCP.readOnlyTokenObject = <%= raw get_read_access_token(@study, current_user).to_json %>
```
bistline, line 4: Good call to catch this!  Wouldn't have known to update this here but I guess that's what global find/replace is for...

bistline (approved):
Approving preemptively though I do think the blocker needs to be addressed (though I can see it be handled in a downstream ticket if we think that's a better plan).  Other than that, very nice work, and one of the most detailed/exhaustive PR descriptions I've ever seen!

app/controllers/api/v1/site_controller.rb
```diff
@@ -125,8 +126,42 @@ def studies
...
+          end
+          response 406 do
+            key :description, ApiBaseController.not_acceptable
+          end
+        end
+      end
 
+      def renew_token
+        Rails.logger.info "Renewing token via SCP API for study #{@study.accession}"
+        render json: DifferentialExpressionService.get_read_access_token(@study, current_api_user, renew=true)
```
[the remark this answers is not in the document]
eweitz: I just opened SCP-5065 to implement handling for bucket auth errors unhandled in the UI, which I think covers this and suggestions from demo yesterday.

jlchang (approved):
Functional review - set up three browser windows per description and confirmed that `renew_token` requests were sent at the appropriate interval.

PR #1751, opened 2023-03-28
Mark legacy AnnData files as reference, enforce metadata convention (SCP-4928, SCP-5042)

bistline:
#### BACKGROUND & CHANGES
This update marks all previously uploaded AnnData files as "reference" files, meaning they will not be parsed, even if the user has their `ingest_anndata_file` feature flag set to `true`.  In addition, all AnnData files that are parsed will now have the metadata convention enforced.  Lastly, this fixes a new bug that caused all uploaded AnnData files in "classic" mode to be parsed, when they should not have.

#### MANUAL TESTING
1. Boot as normal, start Delayed::Job and sign in
2. Go to any study that is not in the "AnnData" mode and upload any `.h5ad` file as a reference file
3. Confirm it pushes to the bucket and does not attempt to parse.  Note: you will still see an ingest job, but the `--extract` parameter will not be present, and legacy behavior of simply opening the file and validating that it is readable is applied

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -301,6 +301,8 @@ module Api
           study_file.build_ann_data_file_info if study_file.ann_data_file_info.blank?
 
           safe_file_params = study_file.ann_data_file_info.merge_form_data(study_file_params.to_unsafe_hash)
+          # override to enforce metadata convention for all AnnData parsing
+          safe_file_params[:use_metadata_convention] = true
         else
           safe_file_params = study_file_params
         end
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -101,7 +101,8 @@ class FileParseService
         end
 
         # enable / disable full ingest of AnnData files using the feature flag 'ingest_anndata_file'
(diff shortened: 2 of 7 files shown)


bistline:
Same `IDX_WS_UNIQUE_NAMESPACE_NAME` error.

ehanna4 (approved):
Works as advertised - was able to upload a "reference anndata file" in classic mode and it saved without extraction

jlchang (approved):
Functional review: confirmed that when uploading `.h5ad` file to a "classic mode" study, the `--extract` parameter is not sent to `ingest_pipeline.py` and the file is checked for open-ability and available for download.

eweitz:
Details on the [build failure](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/4547366636/jobs/8017157124#step:7:4584) that caused this false positive:

```
### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  StudyCleanupToolsTest#before_all:
  Mongoid::Errors::Validations: 
Exiting with code: 1
```

Jean and I are looking into the cause of other build failures.  Log search isn't available for GitHub Actions, so making the causative error searchable by putting it in a PR comment seems like a decent workaround for assessing build failures across PRs.

PR #1753, opened 2023-04-03
Adding Differential Expression study file (SCP-5004)

bistline:
#### BACKGROUND & CHANGES
This update adds a new `StudyFile` type of "Differential Expression" to allow users to upload their own custom differential expression results.  In addition, an embedded `DifferentialExpressionFileInfo` document has been added to keep track of high-level information about the file, such as the source annotation, computational method type, as well as required associations/validations.  

The update additionally adds the `pairwise_comparisons` field to `DifferentialExpressionResult` entries.  This is a hash of possible combinations of labels detected when the (yet-to-be-implemented) ingest process completes for the user-uploaded file.  This will be used to generate UX elements & file URLs similar to how `observed_values` is used (which denotes "one-vs-rest" comparisons).

Lastly, it also consolidates some common annotation-specific methods into the `Annotatable` concern to reduce code duplication, and fixes a logic bug in validating that source annotations are present.

#### MANUAL TESTING
**Note: if you tested this previously and ran the associated migration, please run this first before continuing:**
```
migration = DataMigration.find_by(version: '20230329181342')
migration.destroy if migration.present?
```
Since there is no upload UI or ingest process for handling these files yet, we can only show that they are valid as `StudyFile` objects:
1. Pull branch, initialize environment, and enter a Rails console session
2. Load the `Human milk - differential expression` study:
```
```
3. Simulate a new file upload with the following block (using a dummy temp file for the upload):
```
FileUtils.cp('test/test_data/table_1.xlsx', 'tmp/table_1.xlsx')
upload = File.open('tmp/table_1.xlsx')
study_file = study.study_files.build(file_type: 'Differential Expression', upload:)
cluster_group = study.default_cluster
annotation_name = 'cell_type__ontology_label'
annotation_scope = 'study'
diff_info = study_file.build_differential_expression_file_info(annotation_name:, annotation_scope:, cluster_group:)
```
4. Save and show the file is valid:
```
study_file.valid?
=> true
study_file.save
=> true
```
5. Inspect the attributes to confirm they are the same (IDs will be different) and then destroy the file to clean up:
```
study_file.attributes
=> 
{"_id"=>BSON::ObjectId('642c4ac894ec8f78cb5401cd'),    
 "parse_status"=>"unparsed",                           
 "x_axis_label"=>"",
 "y_axis_label"=>"",
 "z_axis_label"=>"",
 "spatial_cluster_associations"=>[],
 "remote_location"=>"",
 "options"=>{},
 "file_type"=>"Differential Expression",
 "upload_content_type"=>"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
 "upload_file_size"=>41692,
 "human_data"=>false,
 "use_metadata_convention"=>false,
 "is_spatial"=>false,
 "queued_for_deletion"=>false,
 "study_id"=>BSON::ObjectId('62718fbb94ec8f51542049d6'),
 "name"=>"table_1.xlsx",
 "data_dir"=>"22e840a57b04cc28a9c0a2b451ea4e6ce14c5d11846c2f2e95fd7033df1169b9",
 "upload_file_name"=>"table_1.xlsx",
 "updated_at"=>2023-04-04 16:06:14.932949 UTC,
 "created_at"=>2023-04-04 16:06:14.932949 UTC,
 "version"=>1,
 "differential_expression_file_info"=>
  {"_id"=>BSON::ObjectId('642c4aca94ec8f78cb5401ce'),
   "computational_method"=>"wilcoxon",
   "annotation_name"=>"cell_type__ontology_label",
   "annotation_scope"=>"study",
   "cluster_group_id"=>BSON::ObjectId('627190374fe67c088c82a84a')}}

study_file.destroy
=> true

```

### DIFF
--- a/app/models/differential_expression_file_info.rb
+++ b/app/models/differential_expression_file_info.rb
@@ -1,51 +1,19 @@
 # top-level info about DE results contained in a user-uploaded file
 class DifferentialExpressionFileInfo
   include Mongoid::Document
-
-  ANNOTATION_PARAMS = %i[annotation_name annotation_scope].freeze
+  include Annotatable # handles getting/setting annotation objects
 
   embedded_in :study_file
   belongs_to :cluster_group
 
-  # user-uploaded files could contain results for multiple annotations
-  field :annotations, type: Array, default: []
+  field :annotation_name, type: String
+  field :annotation_scope, type: String
   field :computational_method, type: String, default: DifferentialExpressionResult::DEFAULT_COMP_METHOD
-  field :analysis_type, type: String, default: DifferentialExpressionResult::DEFAULT_ANALYSIS
 
   validates :computational_method, inclusion: { in: DifferentialExpressionResult::SUPPORTED_COMP_METHODS }
-  validates :analysis_type, inclusion: { in: DifferentialExpressionResult::ANALYSIS_TYPES }
-  validate :validate_annotations
+  validates :annotation_name, presence: true, uniqueness: { scope: %i[annotation_scope cluster_group] }
+  validates :annotation_scope, presence: true
+  validate :annotation_exists?
 
   delegate :study, to: :study_file
-
-  # retrieve source annotation object
-  def annotation_object(annotation)
-    safe_annot = annotation.with_indifferent_access
-    annotation_name = safe_annot[:annotation_name]
-    annotation_scope = safe_annot[:annotation_scope]
-    case annotation_scope
-    when 'study'
-      study.cell_metadata.by_name_and_type(annotation_name, 'group')
-    when 'cluster'
-      cluster_group.cell_annotations.detect do |annot|
-        annot[:name] == annotation_name && annot[:type] == 'group'
-      end
-    end
-  end
-
-  private
-
-  # ensure annotation info has required keys and there are no duplicate entries
-  def validate_annotations
-    annotations.each do |annotation|
-      if annotation.keys & ANNOTATION_PARAMS != ANNOTATION_PARAMS
-        errors.add(:annotations,
-                   "has an invalid entry; must have values for #{ANNOTATION_PARAMS.join(', ')}: #{annotation}")
-      end
-    end
-    duplicates = annotations.map(&:with_indifferent_access).tally.select { |_, v| v > 1 }.keys
-    if duplicates.any?
-      errors.add(:annotations, "contains duplicate entries: #{duplicates.join(', ')}")
-    end
-  end
 end
--- a/app/models/differential_expression_file_info.rb
+++ b/app/models/differential_expression_file_info.rb
@@ -1,0 +1,49 @@
+# top-level info about DE results contained in a user-uploaded file
+class DifferentialExpressionFileInfo
+  include Mongoid::Document
+
+  ANNOTATION_PARAMS = %i[annotation_name annotation_scope].freeze
+
+  embedded_in :study_file
+  belongs_to :cluster_group
+
+  # user-uploaded files could contain results for multiple annotations
+  field :annotations, type: Array, default: []
+  field :computational_method, type: String, default: DifferentialExpressionResult::DEFAULT_COMP_METHOD
+  field :analysis_type, type: String, default: DifferentialExpressionResult::DEFAULT_ANALYSIS
+
+  validates :computational_method, inclusion: { in: DifferentialExpressionResult::SUPPORTED_COMP_METHODS }
+  validates :analysis_type, inclusion: { in: DifferentialExpressionResult::ANALYSIS_TYPES }
+  validate :validate_annotations
+
+  # retrieve source annotation object
+  def annotation_object(annotation)
+    safe_annot = annotation.with_indifferent_access
+    annotation_name = safe_annot[:annotation_name]
+    annotation_scope = safe_annot[:annotation_scope]
+    case annotation_scope
+    when 'study'
+      study.cell_metadata.by_name_and_type(annotation_name, 'group')
+    when 'cluster'
+      cluster_group.cell_annotations.detect do |annot|
+        annot[:name] == annotation_name && annot[:type] == 'group'
+      end
+    end
+  end
+
+  private
+
+  # ensure annotation info has required keys and there are no duplicate entries
+  def validate_annotations
+    annotations.each do |annotation|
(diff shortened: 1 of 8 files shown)


app/models/differential_expression_result.rb
```diff
@@ -13,22 +13,31 @@ class DifferentialExpressionResult
     DEFAULT_COMP_METHOD, 'logreg', 't-test', 't-test_overestim_var'
   ].freeze
 
+  # analysis types, e.g. A vs. B and A vs. All
+  DEFAULT_ANALYSIS = 'a-vs-all'.freeze
+  ANALYSIS_TYPES = [DEFAULT_ANALYSIS, 'a-vs-b'].freeze
```
eweitz, line 18: (If we don't end up requiring separate uploads for pairwise and one-vs-all, then this comment might be moot.)

A few weeks ago, I recall the team came to rough consensus on referring to "A vs. B" as "pairwise", to avoid confusion with A/B testing.  This would also align SCP vocabulary with [diffxpy](https://diffxpy.readthedocs.io/en/latest/api/index.html#multiple-tests-per-gene), and some other scientific works.

I'd also suggest we use the most community-aligned terms for other kinds of DE analysis.  Scanpy, diffxpy, and at least some [publications](https://www.biorxiv.org/content/10.1101/2022.05.09.490241v1.full.pdf) basically replace "all" with "rest" (which actually seems more accurate), e.g. "one-vs-rest" or "versus-rest".

```suggestion
  # analysis types, e.g. pairwise (a.k.a. A vs. B) and one-vs-rest (a.k.a. A vs. all)
  DEFAULT_ANALYSIS = 'one-vs-rest'.freeze
  ANALYSIS_TYPES = [DEFAULT_ANALYSIS, 'pairwise'].freeze
```

bistline: Ah yes you are correct!  Thanks.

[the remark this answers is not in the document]
bistline: > Our [product docs](https://broadworkbench.atlassian.net/browse/SCP-4860) note the PM suggested just "asking for a single .csv file that contains the DE results for all comparisons", and our SME thought this would make sense.

Oh I totally misunderstood that - I had somehow conflated "comparisons" to "annotations".  Yes, in re-reading this it seems that the model would be for a single annotation per file.  I can make that change.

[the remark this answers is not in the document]
bistline: > I suspect it'd be simpler implementationally to have users upload a single CSV per annotation that contains DE results for _any_ comparisons in that annotation -- whether it's one-vs-rest, pairwise, just one type, both, covering every group, some incomplete subset of groups, etc. Then, the process of [splitting the user's custom DE file](https://broadinstitute.slack.com/archives/CBEHTH601/p1680281337118449?thread_ts=1680279219.304729&cid=CBEHTH601) into per-comparison files for use in the DE UI could be more delegated to Ingest Pipeline.

I think a quick chat about this might be helpful - there's some detail that I'm failing to glean from re-reading Slack conversations.

bistline: > (If we don't end up requiring separate uploads for one-vs-rest and pairwise, then this comment might be moot.)

My feeling is that even if we allow a unified upload that has both pairwise and one-vs-rest in the same file, there may be a benefit to keeping track of this at the result level as there potentially will be a slightly different UX when showing the results.  If we decide later that this isn't needed, it's trivial to remove the data with a migration.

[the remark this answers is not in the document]
bistline: [922ab25](https://github.com/broadinstitute/single_cell_portal_core/pull/1753/commits/922ab25a49483d5e2c0aee0c5667fedf8a79a4d7)

bistline: [922ab25](https://github.com/broadinstitute/single_cell_portal_core/pull/1753/commits/922ab25a49483d5e2c0aee0c5667fedf8a79a4d7)

eweitz (approved):
Code looks good!

Beyond my resolved questions on [scope](https://github.com/broadinstitute/single_cell_portal_core/pull/1753#discussion_r1157285690) and  [nomenclature](https://github.com/broadinstitute/single_cell_portal_core/pull/1753#discussion_r1157303957), I ask below about how much flexibility we want in custom DE files uploaded by users.

app/models/differential_expression_file_info.rb
```diff
@@ -0,0 +1,19 @@
...
+  include Annotatable # handles getting/setting annotation objects
+
+  embedded_in :study_file
+  belongs_to :cluster_group
+
+  field :annotation_name, type: String
+  field :annotation_scope, type: String
+  field :computational_method, type: String, default: DifferentialExpressionResult::DEFAULT_COMP_METHOD
+
+  validates :computational_method, inclusion: { in: DifferentialExpressionResult::SUPPORTED_COMP_METHODS }
```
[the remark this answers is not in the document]
jlchang: There are two ways to think about "computational method" that apply to our method of exploratory analyses:

1. the "tool" level (scanpy vs diffxpy)
2. the "statistical tool" level (wilcoxon vs t-test etc)

Thinking pragmatically, we don't expect much traffic with this feature, perhaps the best thing to do is to add "custom" to the list?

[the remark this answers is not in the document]
bistline: I don't think this is really being leveraged anywhere, it's more just record keeping on our end.

bistline:
@eweitz @jlchang I'm going to incorporate our discussion from earlier today re: pairwise comparisons into this branch, and do away with `analysis_type` altogether.

app/models/differential_expression_result.rb
```diff
@@ -132,14 +119,24 @@ def set_cluster_name
     self.cluster_name = cluster_group.name
   end
 
-  def has_observed_values?
-    if observed_values.count < MIN_OBSERVED_VALUES
-      errors.add(:observed_values, "must have at least #{MIN_OBSERVED_VALUES} values")
+  def comparisons_available?
```
eweitz, line 122: Would this add an error when the user uploads a DE file that contains only 1 one-vs-rest comparison?

bistline: No - this would throw an error if there was only one label (meaning there's nothing to compare against).

PR #1759, opened 2023-04-10
Fixing AbTest dependencies/validations and associated tests (SCP-5025)

bistline:
#### BACKGROUND & CHANGES
This update addresses an issue where duplicate `AbTest` entries could be created for the same feature flag, and would therefore cause some test assertions to fail when a delete was made but only one copy removed.  The root cause appeared to be that there was no uniqueness constraint placed for the `FeatureFlag` association.  Also, since the association is `has_one` from `FeatureFlag`, the dependent clause should have been `:destroy` and not `:delete_all` as there will only ever be one.  Both of these issues have been addressed, and a new test added to check for the uniqueness validation.

#### MANUAL TESTING
Since the whole point of this PR was to fix a test that only failed in CI (for the most part), and was not uniformly reproducible, manual testing will simply attempt to prove the absence of a negative as a positive condition.  However, you can run both updated test suites to confirm that they work locally for you:
1. Pull branch and load secrets into your environment with `./rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Run both test suites and confirm there are no errors:
```
# run controller tests
$ bin/rails test test/controllers/ab_tests_controller_test.rb  

SUITE AbTestsControllerTest starting
... # test output
SUITE AbTestsControllerTest completed (12.595 sec)
Finished in 12.598215s, 0.3175 runs/s, 0.8731 assertions/s.

4 runs, 11 assertions, 0 failures, 0 errors, 0 skips

# run model tests 
$ bin/rails test test/models/ab_test_test.rb 

SUITE AbTestTest starting
... # test output
SUITE AbTestTest completed (4.83 sec)


Finished in 4.833363s, 1.2414 runs/s, 2.8965 assertions/s.

6 runs, 14 assertions, 0 failures, 0 errors, 0 skips


```

### DIFF
--- a/app/models/ab_test.rb
+++ b/app/models/ab_test.rb
@@ -16,6 +16,7 @@ class AbTest
   field :enabled, type: Boolean, default: false
 
   before_validation :sanitize_group_names
+  validates :feature_flag, uniqueness: true
   validate :two_groups?
   after_validation :migrate_assignments, on: :update
 
--- a/app/models/feature_flag.rb
+++ b/app/models/feature_flag.rb
@@ -11,7 +11,7 @@ class FeatureFlag
   field :name, type: String
   field :default_value, type: Boolean, default: false
   field :description, type: String
-  has_one :ab_test, dependent: :delete_all
+  has_one :ab_test, dependent: :destroy
 
   # pointer to all per-model feature_flag_options, will delete all if this flag is removed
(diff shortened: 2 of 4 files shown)


jlchang (approved):
manual review - confirmed that tests pass.

ehanna4 (approved):
Code looks good

PR #1718, opened 2023-02-02
Update instructions for Terra Terms of Service (SCP-4929)

eweitz:
This refines instructions for accepting updated Terra Terms of Service.  It clarifies steps in a non-obvious critical user story.

### Impact
Some users might not be able to use Single Cell Portal, if they do not closely read a Terms of Service UI and ascertain a few non-obvious steps <a href="#details">detailed below</a>.  Errors began flaring up yesterday (2023-02-01), and [today (2023-02-02) it has affected 24 users](https://broad-institute.sentry.io/issues/3869811457/?project=1424198&query=is%3Aunresolved&referrer=issue-stream) -- around 3% of the day's daily visitors.  There is a workaround, noted [in Zendesk](https://broadinstitute.zendesk.com/agent/tickets/303184).  

So likely < 3% of daily users experiece a severe form of this problem.

### Problem details <span id="details"/>
Previously, if a user had registered for SCP and Terra and then Terra updated the Terms of Service, we would show the user an error modal that generically directed them to Terra.  However, to accept the Terra Terms of Service, the user must be signed in on Terra -- not just signed in on SCP.  The common case seems to be that users are signed in on SCP but not Terra.  

So the user must ascertain that they must sign in, then they must ascertain how to get to the Terms of Service UI, then click "ACCEPT" for the Terms -- a button that is physically close to an unrelated "AGREE" button for cookies.  A few SCP team members independently had trouble with this UX flow, and a user reported trouble on 2023-02-02.

### Fix
Now, that scenario will bring such users to an error page with clearer, more explicitly instructions that account for a frequent case -- where the user is signed in on SCP _but not Terra_.  The instructions have the user sign in _on Terra_, click "ACCEPT" (not "AGREE"!), etc.

Here's how it looks:

Updated_Terra_Terms_of_Service_handling__SCP_2023-02-03

Jon and I paired on this; he navigated, I drove.

To test:
At most one person should test this pre-merge, one person post-merge, one person post-production-deployment.
* This likely isn't feasible to manually verify unless you previous have a "special account", i.e. an account registered for SCP and Terra, but have not yet accepted the Terra Terms of Service as updated in January 2023.
* If you have such an account and want to verify pre-merge, then pull this branch locally
* Go to your local SCP in your web browser
* Sign in with your special account
* Confirm you see https://localhost:3000/single_cell/terra_tos
* Go through steps there
* Confirm your local SCP loads any page as expected

[...]

PR #1760, opened 2023-04-11
Speed up page load for signed-in users registered for Terra (SCP-4971)

eweitz:
This makes page views 10-30% faster for authenticated users who are registered for Terra.

Previously, such users had to wait for SCP to check they had accepted the current Terra terms of service (ToS) on every page load.  That pre-flight check blocked page load.  However, the Terra orchestration API does the same check internally, and throws a specific error on failure.  So while an additional blocking check was useful to patch a problem (#1718), it's not needed for security.

Now, SCP page loads are faster for such users.  That's because the extra Terra ToS check has been eliminated or made non-blocking.  On local SCP, this speeds up essentially all SCP page loads for signed-in Terra users by 0.6-2.3 seconds, or 10-30%.

Like previous Terra ToS handling in SCP, these failure modes have a simplistic UX.  If _any_ part of the page requires Terra and you haven't accepted current Terra ToS, then you see _none_ of the page (beyond perhaps a brief glimpse of non-Terra data).  We could theoretically refine this "complete degradation" UX to an experience that's only partially degraded in the rare case that a signed-in registered Terra user hasn't accepted the current Terra ToS.   But achieving that smoother edge-case UX while retaining these more-common-case speedups is likely complex, and represents a separate task.  

In other words, this work focuses on relatively easy changes that significantly speed up the common case, while retaining a somewhat crude UX for the rare case.

### Scenarios
Which new Terra ToS handling flow is invoked depends on whether data from Terra (and thus the Terra ToS check) is needed on the server or client.
* _On server_ (e.g. home page, "My studies"): SCP now catches a specific ToS-related error thrown by Terra orchestration API, and redirects the user to SCP's page about Terra ToS acceptance.
* _On client_ (e.g. study page):, SCP now does a non-blocking request, i.e. _after the page loads_, to detect if user has accepted the current Terra ToS, and on failure redirects the user to SCP's page about Terra ToS acceptance.

### Performance data
Mean time for initial page HTTP response, before (on local `development` branch) and after (on local feature branch):
<table>
  <tr><th></th><th colspan="2">Home</th><th colspan="2">Private study</th><th colspan="2">Public study</th><th colspan="2">My studies</th></tr>
  <tr><th></th><th>Before</th><th>After</th><th>Before</th><th>After</th><th>Before</th><th>After</th><th>Before</th><th>After</th></tr>
  <tr><td>Signed in, accepted</td><td>2.75</td><td>1.92</td><td>8.59</td><td>7.44</td><td>13.77</td><td>13.14</td><td>2.65</td><td>1.92</td></tr>
  <tr><td>Not signed in</td><td>1.17</td><td>1.14</td><td>N/A</td><td>N/A</td><td>9.67</td><td>7.36</td><td>N/A</td><td>N/A</td></tr>
</table>

### Test <span id="test"></span>
To manually test, pull this branch, then:
1. [Sign in as single.cell.ad1@gmail.com](https://docs.google.com/document/d/1QslqdIN0o7OCQ_bNY9YFvVgU6v_JtdQqnV4i33BwakQ/edit), an account which is registered for Terra but has not accepted current Terra ToS
2. Go to home page
3. Confirm you get redirected to SCP page about Terra ToS
4. Go to a "My studies" page
5. Confirm you get redirected to SCP page about Terra ToS
6. Sign out
7. Go to a study page
8. Confirm page loads as expected
9. Copy URL
10. Sign back in as single.cell.ad1@gmail.com
11. Paste URL to study page in web browser address bar, go to it
12. Confirm you get redirected to SCP page about Terra ToS after briefly seeing page (which does not contain Terra data)

This satisfies SCP-4971.  More context is in [slides](https://docs.google.com/presentation/d/1jXMwHRSKjXwj1uwilHmrbm4UCNMDWu4bstQAcPZVC4A/edit#slide=id.p).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -76,10 +76,14 @@ class ApplicationController < ActionController::Base
     end
     begin
       yield
-    rescue RuntimeError => e
-      # Efficiently handle when user needs to accept updated Terra ToS
-      if e.message == "User must accept the latest terms of service."
-        redirect_to exceptions_terra_tos_path and return
+    rescue RestClient::Exception => e
+      if e.http_code === 401
+        # Efficiently and robustly check for error that corresponds to
+        # "User must accept the latest terms of service." for Terra,
+        # but without relying on that particular error message.
+        if user_signed_in? && current_user.must_accept_terra_tos?
+          redirect_to exceptions_terra_tos_path and return
+        end
       end
     end
   ensure
--- a/app/models/fire_cloud_client.rb
+++ b/app/models/fire_cloud_client.rb
@@ -232,6 +232,11 @@ class FireCloudClient
       context = " encountered when requesting '#{path}', attempt ##{current_retry}"
       log_message = "#{e.message}: #{e.http_body}; #{context}"
       Rails.logger.error log_message
+
+      if e.http_code == 401
+        raise e
+      end
+
       # only retry if status code indicates a possible temporary error, and we are under the retry limit and
       # not calling a method that is blocked from retries
       if should_retry?(e.http_code) && retry_count < ApiHelpers::MAX_RETRY_COUNT && !ERROR_IGNORE_LIST.include?(path)


bistline (approved):
Tidy implementation!  I have some small qualms about just doing string matching on the error message, but on the whole I like this approach while we work on the more holistic remediation plan.

app/controllers/application_controller.rb
```diff
@@ -75,7 +74,18 @@ def set_current_user
     if current_user.present?
       current_user.update_last_access_at!
     end
-    yield
+    begin
+      yield
+    rescue RestClient::Exception => e
+      if e.http_code === 401
```
bistline, line 80: I like this a _lot_ better.  Just to be on the safe side, we might want to say:
```suggestion
      if [401, 403].include?(e.http_code)
```

I'm _fairly_ certain that orchestration replies `401` in this case, but it also might differ by endpoint/service.  I've definitely never seen `5xx`, there have been [incidences](https://broad-institute.sentry.io/issues/3910539286/events/0024a467f66c4822a66db85abb765762/?project=1424198) of `403` that could potentially be related.  I'm a little iffy on this but it seems relatively safe to do so.

eweitz: Yeah, when the user needs to accept updated Terra ToS, I see 401 is the only HTTP status code used for the response from the Orchestration API.  That derives from [this line in Sam](https://github.com/broadinstitute/sam/blob/f8cb6929291d798f27185977f593b270a0a8cca1/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala#L118); and 401 is mapped to that "Unauthorized" code in [this line from the akka-http Scala library](https://github.com/akka/akka-http/blob/main/akka-http-core/src/main/scala/akka/http/scaladsl/model/StatusCode.scala#L132) (and, of course, the [HTTP RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-401-unauthorized)).

eweitz:
> would it make sense to add a test for this?

I just tried for 15-30 minutes, but couldn't get one to work.  As you note, maybe the future enhancements can address that!

PR #1765, opened 2023-04-24
Avoid "Too many open files" error in Vite HMR (SCP-5023)

eweitz:
This fixes "Too many open files" errors due to Vite, which have slowed native local SCP development.

Previously, with Vite running, 5-20% of full page reloads yielded a broken page.  The cause is that Vite loads frontend files (mostly JS) in an unbundled, per-file manner rather than as one bundled file.  This enables HMR, which is key for fast frontend development.  

But it also means ~283 HTTP requests are done for each local page load.  Each of these requests is an "open file", i.e. a TCP network socket that is established or awaiting connection on your Mac.  (You can see the number of sockets Vite uses at a given moment via `netstat -a -n | grep 127.0.0.1.3036 | wc -l`.)  The default limit on macOS is 256 sockets.  So local pages would break, with diagnostic reports like below.

Chrome DevTools console:
> GET https://localhost:3000/vite-dev/components/search/controls/FiltersBox.jsx net::ERR_ABORTED 500 (Internal Server Error)

rails s stdout:
> 2023-04-05 15:13:13 -0400 Rack app ("GET /vite-dev/components/search/controls/FiltersBox.jsx" - (127.0.0.1)): #<Errno::EMFILE: Failed to open TCP connection to 127.0.0.1:3036 (Too many open files - socket(2) for "127.0.0.1" port 3036)>

Now, with the changes described in these updated onboarding docs, full reloads never return "Too many open files" errors.  Increasing that default limit has fixed the problem for me -- I haven't observed it in a few weeks.

This relates to SCP-5023.

### DIFF
--- a/NON_CONTAINERIZED_DEV_README.md
+++ b/NON_CONTAINERIZED_DEV_README.md
@@ -68,6 +68,7 @@ to determine which `vault` paths to read from).
 ```
 # Avoid errors caused by Vite in native local SCP using many sockets, e.g.:
 #   "Errno::EMFILE: Failed to open TCP connection to 127.0.0.1:3036 (Too many open files - socket(2) for "127.0.0.1" port 3036"
+# More context: https://github.com/broadinstitute/single_cell_portal_core/pull/1765
 ulimit -Sn 10240
 ```
   - Run `source ~/.zshrc` (or `source ~/.bash_profile`), to update your running shell with those changes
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Increasing ulimit makes sense to me! Thanks for finding and documenting this solution.

bistline (approved):
First off, thanks for doing this - I hope that it works.  My only caveat was that I had actually tried this earlier when I was looking at Vite in a [previous PR](https://github.com/broadinstitute/single_cell_portal_core/pull/1641).  It seemed to work on my machine for a while, and then came back.  That being said, I agree that this should be the solution to the issue - it's even on Vite's [troubleshooting page](https://vitejs.dev/guide/troubleshooting.html#dev-server).  Fingers crossed!

PR #1766, opened 2023-05-01
DRAFT: Addressing CI errors with creating Terra workspaces

bistline:
Draft PR to see resilience of CI runs.

### DIFF
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -274,7 +274,7 @@ class StudiesController < ApplicationController
       # delete firecloud workspace so it can be reused (unless specified by user), and raise error if unsuccessful
       # if successful, we're clear to queue the study for deletion
       # if a study is detached, then force the 'persist' option as it will fail otherwise
-      if params[:workspace] == 'persist' || @study.detached
+      if params[:workspace] == 'persist' || @study.detached?
         @study.update(firecloud_workspace: SecureRandom.uuid)
       else
         begin
@@ -1112,7 +1112,7 @@ class StudiesController < ApplicationController
 
   # check if a study is 'detached' and handle accordingly
   def check_study_detached
-    if @study.detached
+    if @study.detached?
       redirect_to merge_default_redirect_params(request.referrer, scpbr: params[:scpbr]),
                   alert: "We were unable to complete your request as the study is question is detached from the workspace " \
                          "(maybe the workspace was deleted?).  #{SCP_SUPPORT_EMAIL}" and return
--- a/app/views/api/v1/site/_study.json.jbuilder
+++ b/app/views/api/v1/site/_study.json.jbuilder
@@ -6,7 +6,7 @@ json.set! :public, study.public
 json.set! :detached, study.detached
 json.set! :cell_count, study.cell_count
 json.set! :gene_count, study.gene_count
-if study.detached
+if study.detached?
   json.set! :study_files, 'Unavailable (cannot load study workspace or bucket)'
 else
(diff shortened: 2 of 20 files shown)


eweitz (approved):
Speeding up CI by ~9 minutes (35%) is a solid guaranteed benefit.  The greater test isolation also seems like a promising and pragmatic mitigation for these `IDX_WS_UNIQUE_NAMESPACE_NAME` CI false positives, and perhaps others.  And it makes further test parallelization (i.e., drastic speedups) more feasible.  Kudos!

Is there a way to easily assess which tests verify that SCP Rails successfully interacts with external services?  IIUC all tests under `test/integration` previously did that, but now many don't seem to exercise upstream APIs over the network.  Calling these "integration" tests still seems appropriate, FWIW.  I'm just looking for a quick way to find non-isolated integration tests as they're still a useful (and now better-proportioned) part of our test corpus.

Relatedly, with this increased reliance on mocks, we might want to consider introducing [contract tests](https://martinfowler.com/bliki/ContractTest.html).  Those help mocks / stubs / doubles remain accurate.  Definitely slow and unreliable tests seem worse here than potential false-negative tests.  But this excellent work nevertheless does increase the risk of false negatives, so contract tests are worthwhile food for thought.  If this piques your interest, there's a new-ish community of practice in [#dsp-contract-testing](https://broadinstitute.slack.com/archives/C043YJ40719), and [pact-ruby](https://github.com/pact-foundation/pact-ruby) looks intriguing.

bistline:
> Is there a way to easily assess which tests verify that SCP Rails successfully interacts with external services? IIUC all tests under `test/integration` previously did that, but now many don't seem to exercise upstream APIs over the network. Calling these "integration" tests still seems appropriate, FWIW. I'm just looking for a quick way to find non-isolated integration tests as they're still a useful (and now better-proportioned) part of our test corpus.

Great question.  I took a look at the organization we have, and we do have some mixed metaphors.  Most of what is in `test/integration` are `ActionDispatch::IntegrationTest` suites, though almost everything in `test/integration/lib` are `ActiveSupport::TestCase` suites, which are unit-based.  These are mostly local service-level tests.  But many of the `ActionDispatch::IntegrationTest` suites are focused on local API or controller calls, not necessarily external services.  These are still "integration" tests, but I agree it would be nice to have internal separated from external.  Let me see if I can take a pass at reorganizing things in a way that makes this more explicit.
 
> Relatedly, with this increased reliance on mocks, we might want to consider introducing [contract tests](https://martinfowler.com/bliki/ContractTest.html). Those help keep mocks / stubs / doubles accurate. Definitely slow and unreliable tests seem worse here than potential false-negative tests. But this excellent work does increase risk around false negatives, so contract tests are worthwhile food for thought. If this piques your interest, there's a new-ish community of practice in [#dsp-contract-testing](https://broadinstitute.slack.com/archives/C043YJ40719), and [pact-ruby](https://github.com/pact-foundation/pact-ruby) looks intriguing.

That's a great point!  The test suites `fire_cloud_client_test.rb` and `hca_azul_client_test.rb` are two good examples of test libs that might benefit from this.  I can look at `pact-ruby` and see if theres anything in there that looks promising.

bistline:
@eweitz [c60cea7](https://github.com/broadinstitute/single_cell_portal_core/pull/1766/commits/c60cea721945d442b01a9318081e1299abb3f24e) - I added `external` and `local` folders to the `test/integration` directory.  I also moved the service tests to `test/service` since none of these seem to make external calls and rely heavily on mocks.  There's a couple other moves but I think on the whole this makes more sense organizationally.  Thanks for the suggestion!

PR #1771, opened 2023-05-02
Allow clustering updates in AnnData file upload (SCP-5039)

ehanna4:
Allow multiple clusterings, updating and deleting clusterings for an AnnData file upload

To test:
- Pull this branch and start up delayed jobs and a rails server
- Ensure you have the flag for AnnData ingest flipped to true
- Start from a fresh study (necessary as there was a bug in the way clusterings were handled previously)
- Fill in expression and clustering info
- For clustering start with one and make the name "first" and the obsm key "X_umap"
- upload the AnnData file 
- allow this to complete uploading and parsing

To test updating a name
- Now go to the clustering page and change the name of the clustering and click save
- Observe in `development.log` that an ingest job will be kicked off to re-parse this
- Observe in the Study details page the clustering will be deleted and then replaced with the new name

To test adding another clustering
- go to the clustering page and click to add a clustering
- name it "second" and give it an obsm key name of "X_tsne" 
- click save
- now wait, locally this takes a bit of time for me, but eventually you will see the job in `development.log` and refreshing the study details page a few times you will see this new clustering option available.

You can also delete a clustering (it won't present a delete button if there is only one though) and you will see this reflected in the study details page as well.

Here is a screenrecording of me going through all this - warning it's 16min haha this is not quick on local https://drive.google.com/file/d/1egNdyYy081KW5xHPdlsBAsFn3XWOOzUk/view?usp=sharing


app/models/ann_data_file_info.rb
```diff
@@ -106,9 +106,22 @@ def merge_form_fragments(form_data, fragments)
 
   # find a data_fragment of a given type based on arbitrary key/value pairs
   # any key/value pairs that don't match return false and fail the check for :detect
+  # also supports finding values as both strings and symbols (for data_type values)
   def find_fragment(**attrs)
     data_fragments.detect do |fragment|
-      !{ **attrs }.map { |k, v| fragment[k] == v }.include?(false)
+      !{ **attrs }.map { |k, v| fragment[k] == v || fragment[k] == v.send(transform_for(v)) }.include?(false)
```
bistline, line 112: I had forgotten I sent this code to you... I included it on my draft PR that covers expression parsing (due to the same bug we found).  Depending on who merges first, we'll need to remove or deal with the conflict then.

app/models/study_file.rb
```diff
@@ -1331,6 +1333,86 @@ def set_cluster_group_ranges
...
+        end
+      end
+    end
+
+  # Copied from ann_data_ingest_parameters.rb
+  # generate a GS URL to a derived fragment that was extracted from the parent AnnData file
+  # File name structure is: <input_filetype>_frag.<file_type>.<file_type_detail>.tsv
+  #   file_type = cluster|metadata|matrix
+  #   file_type_detail [optional] = cluster name (for cluster files), raw|processed (for matrix files)
+  def fragment_file_gs_url(bucket_id, fragment_type, h5ad_file_id, file_type_detail = "")
```
bistline, line 1406: This looks like a copy of `AnnDataIngestParameters#fragment_file_gs_url` - we should reference that method rather than copying it here.  That being said, if we do end up going the above route, then we wouldn't need this here.

bistline (changes requested):
See the note in `study_file.rb` about renaming database objects rather than the full delete/re-parse job.

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -206,12 +221,50 @@ function DeleteButton({ file, deleteFile, setShowConfirmDeleteModal }) {
  * @returns {boolean} Whether to show upload control buttons or not
  */
 function getIsUploadEnabled(isAnnDataExperience, allFiles, file) {
+  const alreadyUploaded = allFiles.forEach(fileObj => {
+    if (fileObj?.ann_data_file_info?.has_clusters) {
+      return true
+    }
+  })
```
eweitz, line 228: Blocker: the current code makes `alreadyUploaded` always `undefined`, and thus always falsy, which I believe makes `getIsUploadEnabled` always return true, which would be a bug.

For reference:
```
allFiles = [  
  {name: 'foo', file_type: 'Expression Matrix'},
  {name: 'bar', file_type: 'AnnData', ann_data_file_info: {has_clusters: false}},
  {name: 'baz', file_type: 'Expression Matrix'},
  {name: 'moo', file_type: 'Metadata'},
  {name: 'quix', file_type: 'AnnData', ann_data_file_info: {has_clusters: true}}
]

alreadyUploaded = allFiles.forEach(fileObj => {
  if (fileObj?.ann_data_file_info?.has_clusters) {
    return true
  }
})

alreadyUploaded
undefined
```

I think you intend `some` or `all` instead of `forEach`, and a different inner function.  For reference:

```
// Some AnnData files are uploaded (assuming `has_clusters: true` reliably indicates file is uploaded)
allFiles = [  
  {name: 'foo', file_type: 'Expression Matrix'},
  {name: 'bar', file_type: 'AnnData', ann_data_file_info: {has_clusters: false}},
  {name: 'baz', file_type: 'Expression Matrix'},
  {name: 'moo', file_type: 'Metadata'},
  {name: 'quix', file_type: 'AnnData', ann_data_file_info: {has_clusters: true}}
]

alreadyUploaded = allFiles.some(fileObj => fileObj?.ann_data_file_info?.has_clusters)

alreadyUploaded
true
```

I don't have a complete enough understanding to confidently assert which of `some` or `all` is appropriate, but I do think there's a bug (or at least unintentionally misleading code) here.

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -206,12 +221,50 @@ function DeleteButton({ file, deleteFile, setShowConfirmDeleteModal }) {
...
+  // if it's not AnnDataExperience don't gate the save/delete buttons
+  if (!isAnnDataExperience) {return true}
+
+  // only allow save/delete for updates after an AnnData file has been uploaded
+  let alreadyUploaded = false
+  allFiles.forEach(fileObj => {
+    if (fileObj?.ann_data_file_info?.data_fragments.length > 0) {
+      alreadyUploaded = true
+    }
+  })
```
eweitz, line 252: More concisely, and idiomatically:

```suggestion
  // only allow save/delete for updates after an AnnData file has been uploaded
  const alreadyUploaded = 
    allFiles.some(fileObj => fileObj?.ann_data_file_info?.data_fragments.length > 0)
```

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -206,12 +221,50 @@ function DeleteButton({ file, deleteFile, setShowConfirmDeleteModal }) {
...
+  const isExpressionMatrix = file.data_type === 'expression'
+
+  if ((isClustering || isExpressionMatrix || isAnnData) && alreadyUploaded && isAnnDataExperience) {
+    return true
+  } else if (isAnnData && isAnnDataExperience) {
+    return true
+  } {
+    return false
+  }
+}
```
eweitz, line 266: Less cyclomatic complexity, more concision:

```suggestion
  if ((isClustering || isExpressionMatrix || isAnnData) && alreadyUploaded && isAnnDataExperience) {
    return true
  } else {
    return isAnnData && isAnnDataExperience
  }
```

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -89,13 +89,10 @@ export function RawUploadWizard({ studyAccession, name }) {
 
 
   // used for toggling between the split view for the upload experiences
-  const [choiceMade, setChoiceMade] = useState(false)
+  const [overrideExperienceMode, setOverrideExperienceMode] = useState(false)
```
eweitz, line 92: For maintainability, let's either use:

```suggestion
  const [isOverrideExperienceMode, setIsOverrideExperienceMode] = useState(false)
```

like `const [isAnnDataExperience, setIsAnnDataExperience] = useState(false)`, or let's make the latter `const [annDataExperience, setAnnDataExperience] = useState(false)`.  

I personally think `isFoo` for ideal for booleans, but I don't feel strongly.  My only qualm here is with the locally inconsistent naming style.

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -269,49 +270,72 @@ export function RawUploadWizard({ studyAccession, name }) {
...
-          // mulitple clustering forms are allowed so add each as a fragment to the AnnData file
-          const annDataFile = formState.files.find(f => f.file_type === 'AnnData')
-          annDataFile?.ann_data_file_info ? '': annDataFile['ann_data_file_info'] = {}
-          const fragments = annDataFile.ann_data_file_info?.data_fragments || []
-          fragments.push(fileFormData)
-
-          annDataFile.ann_data_file_info.data_fragments = fragments
+      // necessary check for when updating a Clustering or Expression matrix form the fileToBeSaved needs to be updated to the AnnData file
+      // if the file update is on an existing fragment it will have a data_type rather than file_type
+      // or if it's a new clustering it won't have a data_type but will have an obsm key name field
```
eweitz, line 280: Could you break this up with some punctuation?  

The comment strikes me as important, but I've tried reading it a few times, and I unfortunately can't really understand it.  (I don't intend to wordsmith code comments; I'm just seeking to ensure maintainers can clearly understand what's written here.)

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -269,49 +270,72 @@ export function RawUploadWizard({ studyAccession, name }) {
...
+      } else {
+        // enable ingest of data by setting reference_anndata_file = false
+        if (fileToSave.file_type === 'AnnData') {fileToSave['reference_anndata_file'] = false}
+        formState.files.forEach(fileFormData => {
+          if (fileFormData.file_type === 'Cluster') {
+            fileToSave = formState.files.find(f => f.file_type === 'AnnData')
+            studyFileId = fileToSave._id
+            fileFormData.data_type = 'cluster'
+
+              // mulitple clustering forms are allowed so add each as a fragment to the AnnData file
```
eweitz, line 301:
```suggestion
              // multiple clustering forms are allowed so add each as a fragment to the AnnData file
```

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -269,49 +270,72 @@ export function RawUploadWizard({ studyAccession, name }) {
 
   /** save the given file and perform an upload if a selected file is present */
   async function saveFile(file) {
+    let fileToSave = file
     let studyFileId = file._id
 
+    // in AnnData case of saving file when you're doing an update
     if (isAnnDataExperience) {
```
eweitz, line 277: For maintainability, could you extract lines 278 - 314 into a separate function?

Those 36 lines add substantially complexity to this `saveFile` function that's already likely a bit too complex.  Having the `isAnnDataExperience` logic in its own function would enable the containing `saveFile` function to be understand more easily at a high level, while also making it more testable (which would help robustness).

eweitz (commented):
Good progress!

I got about halfway through the review before you conscientiously noted it's worth ceasing review, as a chat with Jon entails substantial changes here.  I suspect my findings so far are mostly orthogonal to that, so I've noted them here.  

I found two trivial blockers that seem like they'd cause user-facing bugs.  I also suggest a handful of non-blocking refinements for maintainability.  All specific code suggestions are just for convenience, and worth a quick check if accepted.

app/javascript/components/upload/UploadWizard.jsx
ehanna4: Oh in my mind I was reading 'override' as the action verb like override the experience mode? Yes or no. But I see how that could be confusing since 'override' could be read as a noun like is the override-experience-mode on or off? I can change it.

app/javascript/components/upload/ExpandableFileForm.jsx
ehanna4: ah i think that was actually a leftover from a previous attempt that I missed removing, good catch!

app/models/study_file.rb
```diff
@@ -1331,6 +1333,45 @@ def set_cluster_group_ranges
     end
   end
 
+  # update clustering fragments to be called on updates to study_file
+  def call_ingest_on_updated_clustering_fragments
+    if self.ann_data_file_info.present? && self.ann_data_file_info.data_fragments_changed? && self.file_type == "AnnData" && (!is_reference_anndata? || is_reference_anndata? != nil)
```
bistline, line 1338: There are some convenience methods that will make this simpler, I think.  `StudyFile#is_reference_anndata?` also calls `is_anndata?`, and all `AnnData` files will already have an `AnnDataFileInfo` object initialized:
```suggestion
    if !!is_reference_anndata? && ann_data_file_info.data_fragments_changed?
```
That will cover the 4 checks you already had, and `!!` does the same thing in Ruby it does in Javascript.

Also (I know I used to be bad about this, and you probably copied stuff of mine from elsewhere), you don't need to call `self` in Ruby unless you're doing a direct assignment in an initializer or need to pass a reference.  Any method/attribute can be called directly  - `file_type` and `is_reference_anndata?` are the same as `self.file_type` and `self.is_reference_anndata?` - and the `self` is implicit, depending on the method scope.

app/models/study_file.rb
```diff
@@ -1331,6 +1333,45 @@ def set_cluster_group_ranges
...
+          matcher = { data_type: :cluster, obsm_key_name: fragment }
+          cluster_data_fragment = self.ann_data_file_info.find_fragment(**matcher)
+          curr_id = cluster_data_fragment&.[]("_id")
+
+          # if it's an existing clustering just update don't parse
+          if prev_ids.include? curr_id 
+            name = cluster_data_fragment&.[](:name) || fragment # fallback if we can't find data_fragment
+
+            if prev_ids[curr_id]["name"] != name
+              @cluster = ClusterGroup.find_by(study_id: study.id)
```
bistline, line 1358: Blocker: this query needs to include the previous name of the `ClusterGroup` as well as the `study_file` (and here we do need `self`).  Also, we can just pass the study association to the query for the sake of brevity.  And I find `dig` easier to work with than multiple `[]` calls:
```suggestion
              @cluster = ClusterGroup.find_by(study:, study_file: self, name: prev_ids.dig(curr_id, 'name'))
```

app/lib/file_parse_service.rb
```diff
@@ -11,6 +11,9 @@ class FileParseService
   # * *returns*
   #   - (Hash) => Status object with http status_code and optional error message
   def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail: false)
```
bistline, line 13: Blocker: this method declaration needs `obsm_key: nil` added to it so we can call it from `StudyFile#call_ingest_on_updated_clustering_fragments`.

app/lib/file_parse_service.rb
```diff
@@ -100,15 +103,19 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
...
 
-        # enable / disable full ingest of AnnData files using the feature flag 'ingest_anndata_file'
-        # will ignore reference AnnData files (includes previously uploaded files) as the default for
-        # ann_data_file_info.reference_file is true legacy files were covered in data migration
         if do_anndata_file_ingest && !study_file.is_reference_anndata?
-          params_object = AnnDataIngestParameters.new(
-            anndata_file: study_file.gs_url, obsm_keys: study_file.ann_data_file_info.obsm_key_names
-          )
+          # obsm_key is only set for parsing a new singular clustering
+          if obsm_key != nil
```
bistline, line 108:
```suggestion
          if obsm_key.present?
```

app/models/study_file.rb
```diff
@@ -1331,6 +1333,45 @@ def set_cluster_group_ranges
...
+              # before updating, check if the defaults also need to change
+              if study.default_cluster == @cluster
+                study.default_options[:cluster] = name
+                study.save
+              end
+              @cluster.update(name:)
+            end
+          else
+          FileParseService.run_parse_job(self, study, study.user, obsm_key: fragment)
```
bistline, line 1367:
```suggestion
            FileParseService.run_parse_job(self, study, study.user, obsm_key: fragment)
```

app/models/study_file.rb
```diff
@@ -1331,6 +1333,45 @@ def set_cluster_group_ranges
     end
   end
 
+  # update clustering fragments to be called on updates to study_file
+  def call_ingest_on_updated_clustering_fragments
```
bistline, line 1337: The method name seems pretty long (though descriptive).  Maybe `handle_data_fragment_updates` since some are just database calls and others new parse jobs?

bistline (approved):
Two blockers, but both pretty straightforward so I'm preemptively approving so you're not blocked tomorrow.  Thanks for working this all out!

app/models/study_file.rb
ehanna4: Since this method (currently) only does stuff on clustering fragments I'll go with `handle_clustering_fragment_updates`

eweitz (changes requested):
Getting closer!  This is a tough area of the code, kudos on keeping with it.

I see only one significant blocker: the `deleteFile` function needs rework.  The [biggest issue is uploading content as part of deleting it](https://github.com/broadinstitute/single_cell_portal_core/pull/1771#discussion_r1186407623), which strikes me as needing non-trivial effort to resolve, and also of significant functional impact for the end user.  (Ideally, I'm wrong about one or both of those concerns.)

Also, not quite blocking, but I think we need some frontend testing for these AnnData updates.  The JavaScript is complex.  With no tests for this significant new JS, I'm concerned we won't be able to change it confidently and quickly.  We ought to be able to use mocks in Jest to mimic server responses if needed; I don't think more realistic end-to-end tests are required.

I'm free to pair any time for help with these!

Finally, thanks for the demo video.  Almost 17 minutes of video-only recording is too long and raw to be very useful for product or technical review, but it _does_ seem reasonably helpful as a full-fidelity reference for maintenance.  One idea to make such resources more valuable without much effort: consider adding timestamped links to salient parts (e.g., [Update cluster in AnnData file at 5m30s](https://drive.google.com/file/d/1egNdyYy081KW5xHPdlsBAsFn3XWOOzUk/view?t=5m30s)).  This is just food for thought, not a request for refinement!

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -236,8 +236,9 @@ export function RawUploadWizard({ studyAccession, name }) {
     })
   }
 
-  /** handler for progress events from an XMLHttpRequest call.
-   *    Updates the overall save progress of the file
+  /**
+   * Handler for progress events from an XMLHttpRequest call.
```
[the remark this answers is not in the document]
ehanna4: This isn't my comment, I think the linter picked it up or maybe in clean up I just fixed the missing capitalization.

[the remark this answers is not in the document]
eweitz: Two clicks to accept this suggestion and pull strikes me as a decent effort-value ratio to address this technically-incorrect comment (which I appreciate wasn't written by you!) -- but that said, this isn't a blocker.

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -170,12 +167,15 @@ export function RawUploadWizard({ studyAccession, name }) {
   }
 
   /** handle response from server after an upload by updating the serverState with the updated file response */
-  function handleSaveResponse(response, uploadingMoreChunks, requestCanceller) {
+  function handleSaveResponse(response, uploadingMoreChunks, requestCanceller, originalFile) {
     const updatedFile = formatFileFromServer(response)
     const fileId = updatedFile._id
     if (requestCanceller.wasCancelled) {
       Store.addNotification(failureNotification(`${updatedFile.name} save cancelled`))
```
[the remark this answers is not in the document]
ehanna4: Made a ticket to more holistically address this: https://broadinstitute.slack.com/archives/CBEHTH601/p1683320088149879

Seems reasonable but would prefer to have done in a different ticket.

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -336,32 +325,135 @@ export function RawUploadWizard({ studyAccession, name }) {
...
-          isDeleting: false
-        })
+      } else {
+        updateFile(fileId, { isDeleting: true })
+        try {
+          await deleteFileFromServer(fileId)
+          deleteFileFromForm(fileId)
+          Store.addNotification(successNotification(`${file.name} deleted successfully`))
+        } catch (error) {
+          Store.addNotification(failureNotification(<span>{file.name} failed to delete<br />{error.message}</span>))
```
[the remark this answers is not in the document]
ehanna4: Mentioned in other comment - seems like a totally reasonable change but not something I would like to address in this PR so made a ticket to address: https://broadinstitute.slack.com/archives/CBEHTH601/p1683320088149879

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -336,32 +325,135 @@ export function RawUploadWizard({ studyAccession, name }) {
...
+              const chunkApiData = formatFileForApi(file, chunkStart, chunkEnd)
+              response = await sendStudyFileChunk({
+                studyAccession, studyFileId, studyFileData: chunkApiData, chunkStart, chunkEnd, fileSize, requestCanceller,
+                onProgress: e => handleSaveProgress(e, studyFileId, fileSize, chunkStart)
+              })
+            }
+            handleSaveResponse(response, false, requestCanceller, file !== fileToDelete ? file : null)
+          }
+        } catch (error) {
+          Store.addNotification(failureNotification(<span>{file.name} failed to delete<br />{error}</span>))
```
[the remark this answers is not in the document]
ehanna4: https://broadinstitute.slack.com/archives/CBEHTH601/p1683320088149879

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -137,9 +144,17 @@ export function SaveDeleteButtons({ file, saveFile, deleteFile, validationMessag
...
+  const saveDisabled = isAnnDataExperience ? false : Object.keys(validationMessages).length > 0
+  let saveButton
+
+  if (file.serverFile?.parse_status === 'parsing' && isAnnDataExperience) {
+    saveButton = <OverlayTrigger trigger={['hover', 'focus']} rootClose placement="top" overlay={parsingPopup}>
+      <span className="detail">Parsing <LoadingSpinner/></span>
+    </OverlayTrigger>
+  }
+
+  saveButton = <button
```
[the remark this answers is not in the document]
ehanna4: Oh woops got lost in the mix, addressing this now

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -205,17 +219,43 @@ function DeleteButton({ file, deleteFile, setShowConfirmDeleteModal }) {
...
+ * @returns {boolean} Whether to show upload control buttons or not
+ */
+function getIsSaveEnabled(isAnnDataExperience, allFiles, file) {
+  // if it's not AnnDataExperience don't gate the save/delete buttons
+  if (!isAnnDataExperience) {return true}
+
+  // allow save/delete for updates after an AnnData file has been uploaded
+  let alreadyUploaded = false
+
+  alreadyUploaded = allFiles.some(fileObj => fileObj?.ann_data_file_info?.data_fragments.length > 0)
```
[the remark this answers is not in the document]
ehanna4: Hmm that's a good thought. I'm not totally sure how this piece would work within the environment of multiple AnnData files. Using the logic as is with `every` and multiple AnnDatas would mean you couldn't save on any AnnData file if there is another AnnData file not yet uploaded but is being filled in. Whereas exact logic with `some` would allow saving on AnnDatas you are currently uploading as long as one is up. So both not quite right, with `some` being too permissive and `every` being too restrictive.

So I think instead of looking at `allFiles` when we go multi-AnnData we'd have to add some logic for each individual AnnData. 

So I think you are correct that `some` won't play right if keeping this exact logic when allowing multiple AnnData but I don't think `every` would work exactly right either and so either option will require some re-working to support multi.

eweitz (approved):
Thanks for investigating further!

It seems a trivial blocker might have been overlooked, see "Blocker: this saveButton assignment".

My "Request changes" blocking concern about re-uploading whole AnnData files to delete a cluster is indeed resolved.  Code comments suggested below would help maintainability, but I don't consider that blocking.

app/javascript/components/upload/ExpandableFileForm.jsx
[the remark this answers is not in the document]
ehanna4: Hey I just updated this a bit to deal better with showing parsing status for fragments and added some comments to make it a bit more clear what the conditional saveButton assignments do and re-arranged a bit so I think it's fixed. 

Unfortunately now the lines aren't matching up so I can't 100% confirm to these lines but what is now on line 158 is the initial `saveButton` after that it is only updated when saving, parsing or saveDisabled.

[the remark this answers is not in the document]
eweitz: Awesome, that looks good!

PR #1773, opened 2023-05-10
Adding daily download quota exemption for users (SCP-5066)

bistline:
#### BACKGROUND
All users in SCP have a daily quota that limits the amount of data they can download within a 24 hour period.  This is currently set to 0.5TB in production, and is enabled as a cost management feature to prevent runaway egress charges.  However, there are some files in public studies that are larger than this quota, meaning they can never be directly downloaded by users unless they have direct access to the study bucket via `gsutil` (this is not counted against the quota as it bypasses SCP services).  As such, the only way that users can download theses files is by admins setting their daily quota directly in the database to a sufficiently large negative number that will stay under the 0.5TB limit.  Furthermore, it is completely opaque to the user that this is what needs to happen, as in some scenarios (like bulk download) the error is completely hidden from the end user, and no appropriate messaging is provided.

#### CHANGES
This update adds functionality to temporarily exempt a single user from their download quota.  All download quotas are now managed by the `DownloadQuotaService`, which contains logic for checking/setting/exempting user quotas.  By setting a download quota to `nil`, no values are incremented during downloads, and the user is free to export as much data as they want until their quota resets at midnight the next day.  This allows individual users to export large datasets while still maintaining quotas for all other users.  Additionally, appropriate messaging has now been added to the single- and bulk-download UX to inform users when they are potentially exceeding their quota, and how to contact SCP support to request an exemption.  This is done by throwing a non-401/403 error code (424: Failed Dependency was chosen) that can be handled at the UI level. 

This is the new button (under Admin Config > User Actions) to grant individual download exemptions.  This can be done by any SCP admin and does not require database access, meaning it can be handled by frontline support when a user emails Zendesk inquiring about an exemption.
Screenshot 2023-05-10 at 11 22 46 AM

#### MANUAL TESTING
1. Boot as normal and sign in with an admin account
2. Go to the Admin config panel, and create/update the `Daily User Download Quota` entry:
   * Configuration Type: `Daily User Download Quota`
   * Type of Value: `Numeric`
   * Value: `100`
   * Byte Operator: `megabyte`
3. Load the `Human Milk - differential expression` study (or any study with files larger that 100 megabytes)
4. Note that larger files now display the `Exceeds Quota` button, and hovering on the button displays the help text about contacting SCP support:

Screenshot 2023-05-10 at 11 04 20 AM
5. Return to the homepage, and run a faceted search for `organ: milk`
6. Click the `Download` button, then click `Next`
7. Note that after attempting to generate a `cURL` command the following error message is shown:

Screenshot 2023-05-10 at 11 04 51 AM
8. Go back to the Admin config page, and click the `Disable Quota`  button next to your email entry in the `User Actions` tab
9. Once completed, go back and try both download actions and note that you are no longer prevented from downloading data
10. **CLEANUP**: Delete the download quota configuration entry, and in a Rails console session manually reset your quota:
```
email = "<your email>"
user = User.find_by(email:)
user.update(daily_download_quota: 0)
```


app/javascript/components/search/controls/download/DownloadCommand.jsx
```diff
@@ -44,7 +47,14 @@ export default function DownloadCommand({ fileIds=[], azulFiles }) {
       </div>
     }
     {
-      !isLoading &&
+      !isLoading && !!authCodeError &&
+      <div className="text-center text-danger">
+        <h4>There was a problem authorizing your request</h4>
+        <p dangerouslySetInnerHTML={{__html: authCodeError }} />
```
eweitz, line 53: Could you remind me why `dangerouslySetInnerHTML` is needed?

eweitz (approved):
Code, screenshots, and [live demo](https://drive.google.com/file/d/1YNrvPM1dL468eo8dHg09a2F6YhARluPA/view?t=4m36s) look good!

> This can be done by any SCP admin and does not require database access, meaning it can be handled by frontline support when a user emails Zendesk inquiring about an exemption.

This might have already been done, but if not: we should ensure Frontline Support knows about this new feature.  It also seems worth documenting in the [SCP Frontline Support playbook](https://docs.google.com/document/d/1IUHVskwVaD-gB-mZDCc0JqUu5FSyAFvs-Y65plfTE1c/edit).  

I could see a link to your live demo from yesterday or even a link to your PR sufficing for that. 
 The main thing is making your already-pretty-excellent docs easily discoverability by FS.

app/javascript/components/search/controls/download/DownloadCommand.jsx
bistline: I don't know that it is specifically, but I couldn't figure out any other way to render the error message as HTML so that the `mailto:` link works.  If there's a different/better paradigm for this I'm all ears.

bistline:
> This might have already been done, but if not: we should ensure Frontline Support knows about this new feature. It also seems worth documenting in the [SCP Frontline Support playbook](https://docs.google.com/document/d/1IUHVskwVaD-gB-mZDCc0JqUu5FSyAFvs-Y65plfTE1c/edit).

It is not!  Good call.  I'll do that now.

bistline:
@eweitz https://docs.google.com/document/d/1IUHVskwVaD-gB-mZDCc0JqUu5FSyAFvs-Y65plfTE1c/edit#heading=h.ctgz8zt8hzuu

ehanna4 (commented):
Functional review - works well as described and is overall smooth
Two suggestions:
1. Could the color for the "Exceed quota" button be a slightly darker shade of red? As it is that shade doesn't have enough contrast with the text to pass accessibility standards. (see screenshot)
2. Would it be too much effort to add some sort of visual indicator to the "Disable Quota" button in the Admin Config to indicate when it has already been enabled? Even just disabling the button if it's already been done for a user while it's active would be helpful to know which users have their quota disabled.
Screenshot 2023-05-12 at 11 06 38 AM

PR #1774, opened 2023-05-10
Add migration script for new feature flag to gate explore tab UX updates (SCP-5128)

ehanna4:
Add a feature flag defaulted to false that we can use for gating the UX updates that we are doing for the explore tab this quarter

Testing: There is not real functionality added with this PR but you can run the migration script to see the new flag.
* Pull this branch
* Open up terminal and run bin/rails db:migrate
* Start a server and open up localhost
* Navigate to the Feature Flags list page: https://localhost:3000/single_cell/feature_flags and see the new feature flag
Screenshot 2023-05-10 at 12 11 53 PM


bistline (approved):

eweitz (approved):
Tidy!

PR #1775, opened 2023-05-15
Handle keyword searching an accession that was entered lowercased (SCP-5076)

ehanna4:
Currently if you try to search for an accession in the home page keyword search you must capitalize the letters "SCP" in order for it to work. This change allows you to type 'scp<#>' and it will update the term to 'SCP' before querying.

Current search:

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/97c712e6-6d9d-4ff2-9a81-90d3aae97411

With this change:

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/b338ee6f-c29b-4763-adaa-2d951d53962f

### DIFF
--- a/app/models/study_accession.rb
+++ b/app/models/study_accession.rb
@@ -8,7 +8,7 @@ class StudyAccession
   validates_uniqueness_of :accession
 
   # exclude everything that does not start with "SCP" and end with digits
-  ACCESSION_SANITIZER = /[^SCP\d+$]/
+  ACCESSION_SANITIZER = /[^SCP\d+$]/i
   # match on accepted accession format of "SCP" and ending with digits
   ACCESSION_FORMAT = /^SCP\d+$/
 


app/javascript/providers/SearchSelectionProvider.jsx
```diff
@@ -70,10 +70,36 @@ export default function SearchSelectionProvider(props) {
     }
     setSelection(newSelection)
   }
+  /**
+   * if there are lowercased accessions in the search terms,
+   * make them uppercased before performing the server search
+   */
+  function uppercaseAccessionTerms(selection) {
```
bistline, line 77: Thinking out loud - why did you opt for doing the conversion client-side as opposed to server-side?  The approach you adopted works perfectly well, but if someone makes a direct search request through the API, then the conversion will not take place.  Granted, this is likely exceedingly rare, but it could happen.

After looking through the code, it seems we could have the same effect with very minimal changes if we updated the `StudyAccession.sanitize_accessions` method (this is what does the matching in search requests):

```
# line 11 in study_accession.rb
ACCESSION_SANITIZER = /[^SCP\d+$]/i
...

# line 37 in study_accession.rb
accession_string = term.strip.gsub(ACCESSION_SANITIZER, '').upcase
```

This generates the following output:
```
irb(main):001:0> terms = %w[foo bar scp1 SCP2 scp scpblah]
=> ["foo", "bar", "scp1", "SCP2", "scp", "scpblah"]
irb(main):002:0> StudyAccession.sanitize_accessions(terms)
=> ["SCP1", "SCP2"]

```

bistline (approved):

eweitz (approved):
Nice refinement!  Code and demo video look good.  Below is some food for thought, not necessarily in scope here.

Do you think it'd be worthwhile to _load_ the study that's returned, automatically?  In other words, I wonder how valuable and difficult it'd be to auto-load a study when it's the only item in the result set and the query was an exact (case insensitive) match for the study's accession.

ehanna4:
> Do you think it'd be worthwhile to _load_ the study that's returned, automatically? In other words, I wonder how valuable and difficult it'd be to auto-load a study when it's the only item in the result set and the query was an exact (case insensitive) match for the study's accession.

That's an interesting idea, and seems it would have good benefit since it would save the user from having to wait for the load of the results list and clicking the link. I've got a list of ideas for dev choice projects and i'll add that there to consider for next time! Thinking out loud actually I wonder if we ought to make a "sprint" in the backlog of ideas for dev choice things in case someone needs a project in the future..

PR #1782, opened 2023-05-23
Show all Explore tabs by default (SCP-5050)

eweitz:
This shows more tabs by default in the Explore view, which aims to boost how well users can discover existing features.

Previously, only enabled tabs were shown in a given Explore view.  Now, essentially all tabs are shown, with some tabs in a new "disabled" state.  The overall visual styling for these was implemented in #1772.  This work enhances the tab display logic to show disabled tabs.  It also slightly refines visual styling for the tabs, and related visual elements, for greater consistency.  These updates are feature flagged, so the combined effect of previously-unmeasured Explore UX can be assessed as a batch when more changes are ready for production.

## Demo
Here's how it looks.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/e0e3ced8-f6d6-4546-a032-7ad53a97a337

### No genes
Before:
Before_0_genes_more_tabs_Explore_UX__SCP_2023-05-22

After:
After_0_genes_more_tabs_Explore_UX__SCP_2023-05-22

### 1 gene
Before:
Before_1_gene_more_tabs_Explore_UX__SCP_2023-05-22

After:
After_1_gene_more_tabs_Explore_UX__SCP_2023-05-22

### 2 genes
Before:
Before_2_genes_more_tabs_Explore_UX__SCP_2023-05-22

After:
After_2_genes_more_tabs_Explore_UX__SCP_2023-05-22

### 3 genes
Before:
Before_3_genes_more_tabs_Explore_UX__SCP_2023-05-22

After:
After_3_genes_more_tabs_Explore_UX__SCP_2023-05-22

## Test
Automated tests verify the updated tab logic.  To manually verify:

* Go to study with visualizations initialized
* In another browser tab, set "show_explore_tab_ux_updates" feature flag to true
* Refresh study page
* Confirm you see disabled tabs in no-genes study UI: Distribution, Correlation, Dot plot, Heatmap

This satisfies SCP-5050.

PR #1783, opened 2023-05-25
Ease gene search: execute query in one step, not two (SCP-5053)

eweitz:
This decreases the [interaction cost](https://en.wikipedia.org/wiki/Interaction_cost) to execute a study gene search, mitigating occasional user confusion.

## Overview
Previously, users had to do two interactions to execute a search: 1) select the gene from autocomplete, then 2) submit the search by clicking the magnifying glass search icon or pressing "Enter" / "Return".  Users occasionally reported confusion with how to search a gene, and this two-step select-then-submit UX seemed to be a factor.  

Now, selecting a gene also searches it.  This combined interaction gesture is also seen in common search UIs like [Google web search](https://www.google.com/) and [Google Maps](https://www.google.com/maps).  Like #1782, this Explore UX update is only available behind a feature flag, so its effect can be assessed as part of that batch of changes.

### Before
https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/c57cf347-7f6d-4068-84c1-2885c65d6989

### After
https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/58973f1e-d77f-4f5c-82ed-08b33fab8cb0

## Observability
This sensitive trigger will likely increase apparent [study gene search activity reported in analytics](https://mixpanel.com/project/2120588/view/19059/app/boards#id=737684) by [2-3x](https://mixpanel.com/s/faOL5).  That seems unavoidable, and also reasonable.  When searching for 3 genes by serially and manually typing in their gene symbols, one after another, displaying "interstitial" plots -- e.g. the gene expression plot upon entering the 1st gene, then the correlation plot upon entering the 2nd gene -- seems valuable as a way to serendipitously discover features at essentially no cost to the user experience.  Pasting (from system clipboard) multiple gene symbols into the search box also executes search immediately, without another interaction, and only logs a single (multi-gene) search to Mixpanel.  

This change updates the `trigger` for study gene search from either `click` or `submit` to `change:multiselect`.  Previously, `change:multiselect` was a full event, rather than a property of the `search`.  The prior `change:multiselect` event is no longer logged; it was among our most frequent events, albeit less signal than this updated `search` event.

The cost in VM resources on the [app server](https://console.cloud.google.com/compute/instancesDetail/zones/us-central1-a/instances/singlecell-01?project=broad-singlecellportal&tab=monitoring&pageState=(%22duration%22:(%22groupValue%22:%22P1M%22,%22customValue%22:null))) and [DB server](https://console.cloud.google.com/compute/instancesDetail/zones/us-central1-a/instances/singlecell-mongo-01?project=broad-singlecellportal&tab=monitoring&pageState=(%22duration%22:(%22groupValue%22:%22P1M%22,%22customValue%22:null))) can be monitored after release.  It's very likely to be negligible.

## Test
To manually verify:
* Go to study with visualizations initialized
* In another browser tab, set "show_explore_tab_ux_updates" feature flag to true
* Refresh study page
* In the search box, type in a few letters
* Click on a suggestion from autocomplete
* Confirm that executes the search, and that you see a gene expression scatter plot
* In the search box, without clearing your existing query, type in a few letters
* Press "tab" on your keyboard to select a suggestion from autocomplete
* Confirm that executes the search, and that you see a correlation scatter plot
* In the search box, without clearing your existing query, type in a few letters
* Select a suggestion from autocomplete
* Confirm that executes the search, and that you see a dot plot

You might only see 2 genes in the dot plot, in which case you've encountered SCP-5148.  That seems like it might effectively only affect local development; it's worth keeping an eye on in staging.

This satisfies SCP-5053.

### DIFF
--- a/app/javascript/components/explore/StudyGeneField.jsx
+++ b/app/javascript/components/explore/StudyGeneField.jsx
@@ -36,7 +36,6 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
     * an array of already entered genes (geneArray),
     * and the current text the user is typing (inputText) */
   const [geneArray, setGeneArray] = useState(enteredGeneArray)
-  const [showEmptySearchModal, setShowEmptySearchModal] = useState(false)
   const [showTooManyGenesModal, setShowTooManyGenesModal] = useState(false)
 
   const [notPresentGenes, setNotPresentGenes] = useState(new Set([]))
@@ -73,13 +72,6 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
           logStudyGeneSearch(genesToSearch, trigger, speciesList)
         }
         searchGenes(genesToSearch)
-      }
-    } else {
-      if (event.type !== 'change:multiselect') {
-        // Don't show empty search modal if the user manually removed gene entry
-        setShowEmptySearchModal(true)
-      } else if (isNewExploreUX) {
-        setShowEmptySearchModal(true)
       }
     }
   }
@@ -205,15 +197,6 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
           <FontAwesomeIcon className="action fa-lg" icon={faFileUpload} />
         </label>}
       </div>
-      <Modal
-        show={showEmptySearchModal}
-        onHide={() => {setShowEmptySearchModal(false)}}
-        animation={false}
-        bsSize='small'>
-        <Modal.Body className="text-center">
-          Enter at least one gene to search
-        </Modal.Body>
-      </Modal>
       <Modal
         show={showNotPresentGeneChoice}
         onHide={() => {setShowNotPresentGeneChoice(false)}}
--- a/app/javascript/components/explore/StudyGeneField.jsx
+++ b/app/javascript/components/explore/StudyGeneField.jsx
@@ -87,7 +87,6 @@ export default function StudyGeneField({ genes, searchGenes, allGenes, speciesLi
       return geneArray
     }
     const newGeneArray = geneArray.concat(getOptionsFromGenes(inputTextValues))
-    logGeneArrayChange(newGeneArray)
     setInputText(' ')
     setGeneArray(newGeneArray)
     return newGeneArray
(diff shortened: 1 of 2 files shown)


eweitz:
The [build failure](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/5081975377/jobs/9131086801?pr=1783#step:7:4515) seems like a false positive.

```
### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  StudySyncServiceTest#before_all:
  RuntimeError: Wrapped undumpable exception for: RuntimeError: workspace single-cell-portal-test/full-sync-test-gjo15 does not exist or you do not have permission to use it

  FireCloudClientTest#test_create_and_mange_user_groups:
  RuntimeError: Wrapped undumpable exception for: RuntimeError: Resource managed-group/test-group-28571f4f-fdde-4fd3-b209-384da21d4c31 not found
Exiting with code: 1
```

ehanna4 (commented):
This seems a good improvement on gene search and seems to make the code cleaner as well. Functional testing though I'm running into some console errors when I try searching multiple genes quickly that I'm not seeing when I switch to the development branch and do the same.

Some of these errors look like local dev things like the CORS error and I think we've seen the violin plot `No DOM element with id 'study-violin-10'` type errors in Sentry before but I wonder/worry if this might introduce a bunch more errors but I can't tell if any are blocking or just noise. 
 
Screenshot 2023-05-25 at 3 52 36 PM
Screenshot 2023-05-25 at 4 00 41 PM

eweitz:
Thanks for the investigation!  The console errors seem like noise, with existing tickets.  I suspect their greater frequency after these changes is largely caused by unrepresentative local times surfacing pre-existing race conditions.  If you see any user-facing problems, then those might be blockers.  

I didn't notice any user-facing problems, with the possible exception of the one noted in the "Test" section of the PR description.  Even that seems like it may well be an artifact of the local environment, or pre-existing.  It seems worth keeping an eye on in staging, but doesn't strike me as an obvious blocker.

ehanna4 (approved):
That analysis seems reasonable, I didn't have any user-facing issues and yeah local is always a bit more noisy so will be good to check out staging once merged. But yea again code looks good and definitely feels like a nicer experience not having to double click/enter

michelle-vel (approved):
This is a great change to make! This method of gene search is much more intuitive for users and much easier to use. Looks great!

PR #1784, opened 2023-05-25
Add DE upload to Upload Wizard (SCP-4999)

ehanna4:
This adds the UI to the Upload Wizard to allow users to upload their own DE files. 

Each DE file can be given an associated clustering, associated annotation, and associated computational method.

If an annotation has already been associated to a DE file (and this has been saved) that annotation option is not presented in the drop down anymore. 

The computational method dropdown uses `CreatableSelect` and a different placeholder to allow the user to type in their own method if it's not listed already. 

Actual uploading does not yet work, as the pieces are not all wired together with the Rails side (SCP-5096)

To test:
- Pull this branch and start up your local environment
- Head to a study that has clustering files already uploaded
- Go to the Upload Wizard and you will see the new Differential Expression tab
- In the Differential Expression tab you will see the form to add DE files and you can see the annotations and clusterings in the selects.

If you want to see that the used annotations are removed from the list you can go to the `DifferentialExpressionStep.jsx` file and alter the default DE file annotation_association to be some annotation that exists on your study. For example I tested on a study with `biosample_id` and so updated the default DE like so:
```
const DEFAULT_NEW_DE_FILE = {
  file_type: 'Differential Expression',
  differential_expression_file_info: {
    clustering_association: [],
    annotation_association: ['biosample_id'],
    comp_method_association: []
  }
}
```
Then when I re-loaded the page (or waited for HMR to do it for me) I checked in the annotation drop down and could see that `biosample_id` was no longer an option.

I am adding some unit tests to this PR shortly, just wanted to get the code up for review in the meantime

Screenshot 2023-05-25 at 1 28 51 PM
Screenshot 2023-05-25 at 1 29 03 PM
Screenshot 2023-05-25 at 1 29 24 PM

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -750,7 +750,10 @@ module Api
           ann_data_file_info_attributes: [
             :_id, :reference_file, :has_clusters, :has_metadata, :has_expression, :has_raw_counts,
             data_fragments: AnnDataFileInfo::DATA_FRAGMENT_PARAMS
-          ]
+          ],
+          differential_expression_file_info_attributes: [
+            :_id, clustering_association: [], annotation_association: []
+          ],
         )
       end
     end
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -752,7 +752,7 @@ module Api
             data_fragments: AnnDataFileInfo::DATA_FRAGMENT_PARAMS
           ],
           differential_expression_file_info_attributes: [
-            :_id, :clustering_association, :annotation_association, :comp_method_association
+            :_id, :clustering_association, :annotation_association, :computational_method
           ],
         )
       end
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -752,7 +752,7 @@ module Api
             data_fragments: AnnDataFileInfo::DATA_FRAGMENT_PARAMS
           ],
           differential_expression_file_info_attributes: [
-            :_id, clustering_association: [], annotation_association: []
+            :_id, :clustering_association, :annotation_association, :comp_method_association
           ],
         )
       end
--- a/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
+++ b/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
@@ -66,6 +66,7 @@ export default function DifferentialExpressionFileForm({
 
   const associatedCompMethod = compMethodOptions?.find(
     opt => opt.value === file.differential_expression_file_info.comp_method_association
+    opt => opt.value === file.differential_expression_file_info.computational_method
   )
 
   /** handle a change in the associated cluster select */
@@ -92,7 +93,7 @@ export default function DifferentialExpressionFileForm({
     if (option) {
       newVal = option.value
     }
-    updateFile(file._id, { differential_expression_file_info: { comp_method_association: newVal } })
+    updateFile(file._id, { differential_expression_file_info: { computational_method: newVal } })
   }
 
   return <ExpandableFileForm {...{
--- a/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
+++ b/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
@@ -31,27 +31,27 @@ export default function DifferentialExpressionFileForm({
   const clusterFiles = matchingFormFiles(allFiles, clusterFileFilter, isAnnDataExperience, fragmentType)
   const clusterFileOptions = clusterFiles.map(cf => ({ label: cf.name, value: cf._id }))
   const associatedCluster = clusterFileOptions?.find(
-    opt => opt.value === file.differential_expression_file_info.clustering_association?.id
+    opt => opt.value === file.differential_expression_file_info.clustering_association
   )
 
   const annotsAlreadyInUse = []
   // retrieve the annotations that are already in use on a DE file
   allFiles.filter(differentialExpressionFileFilter).filter(
-    diffExpFile => diffExpFile.differential_expression_file_info.annotation_association?.length > 0
+    diffExpFile => diffExpFile.differential_expression_file_info.annotation_association
   ).forEach(file => {
-    annotsAlreadyInUse.push(file.differential_expression_file_info.annotation_association[0])
+    annotsAlreadyInUse.push(file.differential_expression_file_info.annotation_association)
   })
 
   // filter down the annotations to only allow choosing an annotation that hasn't been chosen already
   // each annotation is allowed to be associated with only one DE file
   const annotationOptions = annotationsAvailOnStudy?.map(
-    cf => ({ label: cf.name, value: cf.name })
+    cf => ({ label: cf.name, value: `${cf.name}--${cf.type}--${cf.scope}` })
   ).filter(
     annotObj => !annotsAlreadyInUse.includes(annotObj.value)
   )
 
   const associatedAnnotation = annotationOptions?.find(
-    opt => opt.value === file.differential_expression_file_info.annotation_association?.name
+    opt => opt.value === file.differential_expression_file_info.annotation_association
   )
 
   /* while mapping the computational methods constant to label/value pairs for the select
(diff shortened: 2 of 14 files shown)


app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+      <label className="labeled-select">Statistical test (computational-method)
+        {/* using CreateableSelect here so that users can add an option if their method isn't listed */}
+        <CreatableSelect
+          data-analytics-name="differential-expression-statistical-test-select"
+          options={compMethodOptions}
+          value={associatedCompMethod}
+          className="labeled-select"
+          isClearable
+          onChange={val => updateAssociatedCompMethod(file, val)}
+          placeholder="Start typing to select or add your method"
```
eweitz, line 130:
```suggestion
          placeholder="Start typing to select or add your method or statistical test"
```

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+          value={associatedAnnotation}
+          placeholder="Select one..."
+          onChange={val => updateAssociatedAnnotation(file, val)}/>
+      </label>
+    </div>
+    <div className="form-group">
+      <label className="labeled-select">Statistical test (computational-method)
+        {/* using CreateableSelect here so that users can add an option if their method isn't listed */}
+        <CreatableSelect
+          data-analytics-name="differential-expression-statistical-test-select"
```
eweitz, line 124:
```suggestion
          data-analytics-name="differential-expression-computational-method-select"
```

app/javascript/components/upload/DifferentialExpressionStep.jsx
```diff
@@ -0,0 +1,74 @@
...
+    if (deFiles.length === 0) {
+      addNewFile(DEFAULT_NEW_DE_FILE)
+    }
+  }, [deFiles.length])
+
+  return <div>
+    <div className="row">
+      <div className="col-md-12">
+        <p className="form-terra">
+          Upload a file with differential expression for a particular clustering and annotation
```
eweitz, line 51: Like elsewhere:

```suggestion
          Upload a file with differential expression for a particular clustering and annotation.
```

eweitz (approved):
Code looks good, overall!  I suggest several non-blocking maintainability and consistency refinements.

michelle-vel (approved):
These changes look great and I'm very excited to see this implemented, along with the code I've been working on to process DE uploads. Can't wait to see it in action

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+  updateFile,
+  saveFile,
+  deleteFile,
+  bucketName,
+  isInitiallyExpanded,
+  isAnnDataExperience,
+  annotationsAvailOnStudy,
+  menuOptions
+}) {
+  const validationMessages = validateFile({ file, allFiles, allowedFileExts })
```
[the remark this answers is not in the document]
ehanna4: Yeah that's correct we don't have DE specific CSFV, but `validateFile()` also checks for a file being chosen and so disables/enables the Save button appropriately. I created [SCP-5154](https://broadworkbench.atlassian.net/browse/SCP-5154) for CSFV and will add it as a TODO here

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+      <label className="labeled-select">Associated annotation
+        <Select options={annotationOptions}
+          data-analytics-name="differential-expression-associated-annotation-select"
+          value={associatedAnnotation}
+          placeholder="Select one..."
+          onChange={val => updateAssociatedAnnotation(file, val)}/>
+      </label>
+    </div>
+    <div className="form-group">
+      <label className="labeled-select">Statistical test (computational-method)
```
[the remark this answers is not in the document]
ehanna4: I'm totally good with "Computational method", especially as that's the name of the CONST that defines the options for that dropdown. The ticket had said it like: “statistical test” dropdown for the test method (aka computational-method) used" which is why I included both, but it makes sense to me to go with your suggestion

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+  ).forEach(file => {
+    annotsAlreadyInUse.push(file.differential_expression_file_info.annotation_association[0])
+  })
+
+  // filter down the annotations to only allow choosing an annotation that hasn't been chosen already
+  // each annotation is allowed to be associated with only one DE file
+  const annotationOptions = annotationsAvailOnStudy?.map(
+    cf => ({ label: cf.name, value: cf.name })
+  ).filter(
+    annotObj => {return !annotsAlreadyInUse.includes(annotObj.value)}
```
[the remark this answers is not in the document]
ehanna4: oh interesting, yeah looks like removing the curly-brackets you can remove the return too and that seems slightly more consistent with the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)
Screenshot 2023-05-26 at 2 28 07 PM

ehanna4:
Added Feature Flag around showing the upload UI for DE so can merge this PR while we get all the pieces in place (per discussion at standup 5/30)

app/controllers/api/v1/study_files_controller.rb
```diff
@@ -750,7 +750,10 @@ def study_file_params
           ann_data_file_info_attributes: [
             :_id, :reference_file, :has_clusters, :has_metadata, :has_expression, :has_raw_counts,
             data_fragments: AnnDataFileInfo::DATA_FRAGMENT_PARAMS
-          ]
+          ],
+          differential_expression_file_info_attributes: [
+            :_id, clustering_association: [], annotation_association: []
```
bistline, line 755: Non-blocking: I know that the Rails-integration part is scoped in SCP-5096, so this can be addressed there. 
 That being said, if we are adding these parameters here, then they should be defined in `differential_expression_file_info.rb`.  Also, according to SCP-4999, our current assumption is one annotation per file, which would also mean a single cluster as well.  In that case, they can be strings rather than arrays.  We only need arrays if there's a one-to-many relationship.  Also, I would favor naming the first one `cluster_group_id`.

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -0,0 +1,136 @@
...
+  allFiles.filter(differentialExpressionFileFilter).filter(
+    diffExpFile => diffExpFile.differential_expression_file_info.annotation_association?.length > 0
+  ).forEach(file => {
+    annotsAlreadyInUse.push(file.differential_expression_file_info.annotation_association[0])
+  })
+
+  // filter down the annotations to only allow choosing an annotation that hasn't been chosen already
+  // each annotation is allowed to be associated with only one DE file
+  const annotationOptions = annotationsAvailOnStudy?.map(
+    cf => ({ label: cf.name, value: cf.name })
```
bistline, line 48: Blocker: if we're only matching on the value of `name`, we run the risk of confusing cluster- and study-based annotations that have the same name.  We need to be matching on all three annotation attributes: `name`, `type`, and `scope`.  Since this is DE, the `type` will always be `group`, but we still need to match on that.  Since this is a select, we should use the URL request parameter format, like `cell_type--group--study`.

bistline (changes requested):
One blocker that I note on matching annotations by name, type, and scope - not just name.  Otherwise this looks good to me.

app/controllers/api/v1/study_files_controller.rb
ehanna4: Switched to strings and added the fields to `differential_expression_file_info.rb`
 I liked the naming to all match i.e. all ending with `_association` so I did not change the `clustering_assocation` to `cluster_group_id` but I'm not so married to it that if it gets changed later (like in [SCP-5096](https://broadworkbench.atlassian.net/browse/SCP-5096)) that's no problem

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
ehanna4: oh thanks for raising this, please see the latest commit and let me know if that addresses your concern: [652a709](https://github.com/broadinstitute/single_cell_portal_core/pull/1784/commits/652a70983472904e43a6ae2dc845be7078624fdf)

bistline (approved):

bistline:
Build failure looks like a red herring - unrelated to any changes here.

PR #1255, opened 2021-11-15
reference image upload tests and UX polish (SCP-3758)

devonbush:
This fixes a couple UX papercuts in reference image upload, and adds a test

TO TEST:
 1. open a study in the new upload wizard, and go to the reference image tab
 2. select an image to upload
 3. confirm the 'name' text field displays correctly, and is prepopulated with the filename
 4. upload the image
 5. confirm that a broken image tag does not appear while the file is in 'waiting for remote image' state.
 6. once the image is sent to firecloud, confirm the image is shown

PR #1786, opened 2023-05-30
Remove the image upload step from the upload wizard (SCP-5045)

ehanna4:
Remove the unsupported reference image upload step in the upload wizard and accompanying feature flag references. 
Also remove the units from the AnnData expression matrix tab in preparation for Raw Counts being added as a full option for AnnData upload in [SCP-4754 ](https://broadworkbench.atlassian.net/browse/SCP-4754?atlOrigin=eyJpIjoiZmFiYjA3YmQ0ZGYxNGFjYTgwMDEzMTgxOTU4ZmYzMmEiLCJwIjoiaiJ9)

To test
- pull this branch and run the db migration via `bin/rails db:migrate`
- then start up a local server and navigate to the feature flags page: https://localhost:3000/single_cell/feature_flags and see that `reference_image_upload` is not there.
- navigate to the upload wizard on any study and note that the image tab is not there.


eweitz (approved):
Code looks good!  If we ever resurrect this feature, here's some historical context.

Demo: 
* https://drive.google.com/file/d/1owzK8QgVjYlsJWpD3GFBs6F1pycEiNaO/view

PRs:
* https://github.com/broadinstitute/single_cell_portal_core/pull/1165
* https://github.com/broadinstitute/single_cell_portal_core/pull/1255
* https://github.com/broadinstitute/single_cell_portal_core/pull/1358

bistline (approved):
Looks good!

Should we also remove references to `Image` files from the references below?  It certainly doesn't break anything to leave them, but I just checked and we have don't have any `Image` files in production.  There are two in staging so we probably would need to do cleanup via a migration to be sure.  Just a thought!

https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/study.rb#L1027
https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/study_file.rb#L25
https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/lib/cluster_viz_service.rb#L66
https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/controllers/api/v1/visualization/explore_controller.rb#L139

PR #1787, opened 2023-05-31
Better error handling for checking user Terra terms of service compliance (SCP-5113)

bistline:
#### BACKGROUND & CHANGES
This update adds better error handling to checking an authenticated user's terms of service status in Terra.  Previously, any non-200 response from the `/register` [endpoint on orchestration](https://api.firecloud.org/#/Profile/getUserStatus) would cause SCP to thing the user needs to accept updated terms.  However, we have seen numerous instances of `503` responses, which do not contain any relevant information at all.  This causes SCP to redirect to a page containing erroneous information for the user, and is highly disruptive.

The new method `User#check_terra_tos_status` now handles the logic of checking this, and will now also emit the original HTTP status code as well.  This allows us better control over corner cases, as only a `200` response with `tosAccepted: false` or a `401` actually mean the user needs to accept updated terms (though in the latter case, the `/register` endpoint should not respond `401` according to the Swagger documentation, but as other API endpoints will, this is a reasonable guard clause).  `User#must_accept_terra_tos?` is preserved as a passthru calling the aforementioned method and only checking the `must_accept` value.

In addition, `FireCloudClient#process_firecloud_request` has been updated to emit the original `RestClient::Exception` error instead of reclassifying it as a `RuntimeError`.  This was necessary to preserve HTTP status codes.

#### MANUAL TESTING
To test normal usage:
1. Boot as normal and sign in with your account
2. Log in and confirm you are not redirected to the Terra ToS page, navigating around to various pages to confirm no redirect happens later
3. In a terminal window, get the password for the `single.cell.user1@gmail.com` account:
```
vault read secret/kdux/scp/development/$(whoami)/scp_config.json | grep TEST_USER_PASSWORD | awk '{ print $2 }'
```
4. Log out and log in with this account (note: you will have to accept the SCP terms of service if you've never used this account locally)
5. Once complete, load the home page and confirm you are redirected to the Terra ToS page
6. Navigate to the "My Profile" page (from the top right menu) and confirm you eventually get redirected back to the Terra ToS page

To test other states:
1. Log out and sign back in with your normal account
2. In `app/models/user.rb`, insert the following at line 387
```
raise RestClient::Exception.new(nil, 401)
```
3. Load the home page and confirm you are eventually redirected to the Terra ToS page
4. Back in `user.rb`, change the `401` to `404`
5. Load the home page again and confirm you do not get redirected
7. Try the same with `503` and any other valid HTTP status code to confirm you do not get redirected

### DIFF
--- a/app/javascript/lib/scp-api.jsx
+++ b/app/javascript/lib/scp-api.jsx
@@ -250,8 +250,11 @@ export function setupRenewalForReadOnlyToken(studyAccession) {
  */
 export async function checkTerraTosAcceptance() {
   const apiUrl = '/site/check_terra_tos_acceptance'
-  const [response] = await scpApi(apiUrl, defaultInit())
-  return response.mustAccept
+  await scpApi(apiUrl, defaultInit()).then(response => {
+    return { mustAccept: response.mustAccept, status: response.status }
+  }).catch(error => {
+    return { mustAccept: error.mustAccept, status: error.status }
+  })
 }
 
 /**
--- a/app/javascript/lib/scp-api.jsx
+++ b/app/javascript/lib/scp-api.jsx
@@ -250,11 +250,8 @@ export function setupRenewalForReadOnlyToken(studyAccession) {
  */
 export async function checkTerraTosAcceptance() {
   const apiUrl = '/site/check_terra_tos_acceptance'
-  await scpApi(apiUrl, defaultInit()).then(response => {
-    return { mustAccept: response.mustAccept, status: response.status }
-  }).catch(error => {
-    return { mustAccept: error.mustAccept, status: error.status }
-  })
+  const [response] = await scpApi(apiUrl, defaultInit())
+  return response.mustAccept
 }
 
 /**
@@ -927,26 +924,30 @@ export default async function scpApi(
     } else {
       return [response, perfTimes, true]
     }
-  } else if (response.status === 401 || response.status === 403) {
-    showMessage(
-      <div>
-        Authentication failed<br/>
-        Your session may have timed out. Please sign in again.<br/><br/>
-      </div>,
-      'api-auth-failure',
-      {
-        source: 'api',
-        url,
-        isError: true,
-        messageType: 'error-client',
-        statusCode: response.status
-      }
-    )
-    throw new Error(`Authentication error: ${response.status}`)
   }
   if (toJson) {
     const json = await response.json()
-    if (Array.isArray(json.errors)) {
+    // special handling for Terra terms of service checks
+    // don't throw error so we can pass back JSON response
+    if (typeof json.must_accept !== 'undefined') {
+      return [camelcaseKeys(json)]
+    } else if (response.status === 401 || response.status === 403) {
+      showMessage(
+        <div>
+          Authentication failed<br/>
+          Your session may have timed out. Please sign in again.<br/><br/>
+        </div>,
+        'api-auth-failure',
+        {
+          source: 'api',
+          url,
+          isError: true,
+          messageType: 'error-client',
+          statusCode: response.status
+        }
+      )
+      throw new Error(`Authentication error: ${response.status}`)
+    } else if (Array.isArray(json.errors)) {
       throw new ApiError(json, response.status, path)
     } else {
       throw new Error(json.error || json.errors)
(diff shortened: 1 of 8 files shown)


app/javascript/lib/scp-api.jsx
```diff
@@ -250,8 +250,11 @@ export function setupRenewalForReadOnlyToken(studyAccession) {
  */
 export async function checkTerraTosAcceptance() {
   const apiUrl = '/site/check_terra_tos_acceptance'
-  const [response] = await scpApi(apiUrl, defaultInit())
-  return response.mustAccept
+  await scpApi(apiUrl, defaultInit()).then(response => {
+    return { mustAccept: response.mustAccept, status: response.status }
+  }).catch(error => {
+    return { mustAccept: error.mustAccept, status: error.status }
+  })
```
eweitz, line 257: Blocker: this breaks when the user must accept updated Terra terms of service.

I just manually tested this locally, and `checkTerraTosAcceptance` now returns `undefined`.  Here's the downstream code that calls this, [in application.js](https://github.com/broadinstitute/single_cell_portal_core/blob/f7a1eba618fcfd35a114dd7b83be410dc679b3d4/app/javascript/vite/application.js#L57):

```
    ScpApi.checkTerraTosAcceptance().then(mustAcceptTerraTos => {
      if (mustAcceptTerraTos) {
        window.location = '/single_cell/terra_tos'
      }
    })
```

So with `mustAcceptTerraTos` always returning `undefined`, this new code will always give a negative (often false negative) result.  In other words, it will always effectively behave as if the user does not need to accept Terra ToS.

Technical issues:

1. The `return` is inside a callback function, and doesn't get assigned to anything in the containing scope.  That's why `checkTerraTosAcceptance` always returns `undefined`.

2. [`scpApi` returns an array](https://github.com/broadinstitute/single_cell_portal_core/blob/f7a1eba618fcfd35a114dd7b83be410dc679b3d4/app/javascript/lib/scp-api.jsx#L912) when the response is OK.

3. `scpApi` indeed throws an error upon 401 or 403.  But lines 932-948 in scp-api.jsx entail that if the response status is 401 or 403, then the return value of `scpApi` will be a string, not a JSON object with a `mustAccept` or `status` property.  The UI throws an error modal saying "Authentication failed Your session may have timed out. Please sign in again."  Now this API endpoint returns a more specific 401 or 403, that modal is misleading, and in this case we should ensure the frontend executes `window.location = '/single_cell/terra_tos'`. This is the bug.

A full-fidelity test of 401 / 403 handling is impractical, as you note.  However, you can easily and adequately test UI handling by setting [line 81 in `app/controllers/api/v1/site_controller.rb`](https://github.com/broadinstitute/single_cell_portal_core/pull/1787/files#diff-ef6cc11ca821ecf554c8d9a3c345231cf35421b2a16f30449d74eaa4a112b58fR81) to `render json: { must_accept: true }, status: 401`.

Here's an initial pass for part of a fix:

```suggestion
  return await scpApi(apiUrl, defaultInit()).then(([response]) => {
    return response.mustAccept
  }).catch(error => {
    return error.mustAccept
  })
```

Note that the `error` case would require further updates to the 401/403 handling in `scp-api.jsx`.

eweitz (changes requested):
This will be good when it's fixed!  I found a non-trivial bug, see [details](https://github.com/broadinstitute/single_cell_portal_core/pull/1787#discussion_r1212284060).

app/javascript/lib/scp-api.jsx
bistline: Ah yes, I see what you're saying now.  OK - I think this is still tractable but will require more work.

bistline: I just realized I'd had this working previously but I took it out in a later commit because I thought it was unnecessary given some other refinements.  I think I have a good idea on how to handle this now - thanks for catching it!

eweitz (approved):
Code looks good!  

I'm glad we'll finally eliminate false positives for the Terra ToS check -- which likely cause a substantial fraction of [SCP Terra ToS page views](https://mixpanel.com/s/9ToAc) -- while also preserving related performance optimizations from #1760.

ehanna4 (approved):
Looks good 👍

PR #1790, opened 2023-06-05
Ensure proper cleanup of `DataArray` entries on file deletion (SCP-5158)

bistline:
#### BACKGROUND & CHANGES
This update adds a secondary check on all file deletions to remove any `DataArray` documents that have been "orphaned" (defined as a document where the `study_file_id` no longer exists in the portal).  While the incidence of these are rare, and the direct cause is not completely understood, these documents can sometimes permanently block new ingests of the same file type, particularly in the case of metadata files.  The only indication that this is happening will be an ingest error with the message `batch ops error` and a code of `E11000` from MongoDB (unique index constraint violation).  The user is completely blocked in this case, unless they chose to create a completely new study.  Otherwise, it requires direct database intervention by an admin to fix.

Now, whenever a file is deleted in SCP, part of the `DeleteQueueJob` will be to check that study for any orphaned arrays and to delete them if found.  These queries will run in the background and complete very quickly, so load on the application is minimal.  In most cases, they will exit immediately.  A quick check on production found 110,663 orphaned arrays (out of almost 57M total documents, or less than 0.2% - details on the query are in the Jira ticket).  As of now, these documents are not causing any direct issues, and in the future if the study owner deletes any files (or has an ingest failure) they will automatically clean up.  This removes the need for a larger data migration, and will prevent the issue from getting larger moving forward.

#### MANUAL TESTING
1. Boot as normal (including DelayedJob workers) and sign in
2. Create a new study, and upload the metadata file at `test/test_data/alexandria_convention/metadata.v2-0-0.txt`
3. Once completed, open a Rails console session in a separate terminal window and manually delete the file and any `CellMetadatum` entries, but leave the `DataArray` documents in place (this simulates the error found by [this user](https://broadinstitute.zendesk.com/agent/tickets/310001)):
```
study = Study.last
study.metadata_file.delete
study.cell_metadata.delete_all
DataArray.where(study_id: study.id).exists?
=> true
```
4. Back in the upload wizard, refresh the page and re-upload the same file
5. Once the ingest fails with the `batch ops error`, wait ~30s for the `DeleteQueueJob` to run
6. In the Rails console, confirm that all `DataArray` documents from the prior upload have been removed:
```
DataArray.where(study_id: study.id).exists?
=> false
```

### DIFF
--- a/app/models/delete_queue_job.rb
+++ b/app/models/delete_queue_job.rb
@@ -78,7 +78,6 @@ class DeleteQueueJob < Struct.new(:object, :study_file_id)
         end
         delete_differential_expression_results(study: study, study_file: object)
         delete_parsed_data(object.id, study.id, CellMetadatum, DataArray)
-        delete_all_cells_arrays(study.id)
         study.update(cell_count: 0)
         reset_default_annotation(study:)
       when 'AnnData'
@@ -87,7 +86,6 @@ class DeleteQueueJob < Struct.new(:object, :study_file_id)
           # delete user annotations first as we lose associations later
           delete_user_annotations(study:, study_file: object)
           delete_parsed_data(object.id, study.id, ClusterGroup, CellMetadatum, Gene, DataArray)
-          delete_all_cells_arrays(study.id) if object&.ann_data_file_info&.has_metadata?
           delete_fragment_files(study:, study_file: object)
           # reset default options/counts
           study.reload
@@ -113,6 +111,9 @@ class DeleteQueueJob < Struct.new(:object, :study_file_id)
         end
         object.study_file_bundle.destroy
       end
+
+      # delete any "orphaned" DataArray that might have persisted due to some kind of earlier error
+      delete_orphaned_arrays(study.id)
 
       # overwrite attributes to allow their immediate reuse
       # this must be done with a fresh StudyFile reference, otherwise upload_file_name may not overwrite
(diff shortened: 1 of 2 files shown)


app/models/delete_queue_job.rb
```diff
@@ -171,6 +174,13 @@ def delete_parsed_data(object_id, study_id, *models)
     end
   end
 
+  # ensure there are no orphaned DataArray entries that will cause downstream index violations
+  def delete_orphaned_arrays(study_id)
+    study_file_ids = StudyFile.where(study_id:).pluck(:id)
+    cursor = DataArray.where(study_id:, :study_file_id.nin => study_file_ids)
```
eweitz, line 180: I somehow hadn't been aware of [`nin`](https://www.mongodb.com/docs/manual/reference/operator/query/nin/#mongodb-query-op.-nin), i.e. "not in".

eweitz (approved):
Code looks good!

Thanks for the [query details](https://broadworkbench.atlassian.net/browse/SCP-5158?focusedCommentId=79431).  For future reference, see [background technical discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1685722234247539).

app/models/delete_queue_job.rb
bistline: All of the MongoDB [query operators](https://www.mongodb.com/docs/manual/reference/operator/query-comparison/) are available using the above syntax of `Model.where(:field.operator => query_values`).

bistline:
Build failure had same false positive from [SCP-5159](https://broadworkbench.atlassian.net/browse/SCP-5159) and one true regression (fixed)

ehanna4 (approved):
Functional review - works as advertised ! Thanks for the thorough testing steps
Screenshot 2023-06-09 at 4 21 05 PM

PR #1794, opened 2023-06-06
Making the options panel more salient (SCP-5055, SCP-5056, SCP-5057, SCP-5058)

bistline:
#### BACKGROUND & CHANGES
This update makes the view options panel in the Explore tab more noticeable.  These changes include:

- Making the background color slightly darker
- Making the work "OPTIONS" slightly larger
- Adding a tooltip explanation for the "Annotations" select
- Replacing the gear cog icon with an eye 

Demo of all changes:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/e49acdc1-ddd5-4ad0-8910-0bf7512afa1f

These only apply to the normal options panel view.  When a user is interacting with differential expression results, the UX is unchanged.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Load any study with visualizations and differential expression results (like `Human milk - differential expression`)
3. Confirm the options panel behaves as shown in the video
4. Click the Differential Expression button and confirm that the background goes back to white, and returns to light blue when closed

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -333,6 +333,7 @@ export default function ExploreDisplayTabs({
   function getPanelWidths() {
     let main
     let side
+    const isSelectingDE = showDifferentialExpressionPanel || showUpstreamDifferentialExpressionPanel
     if (showViewOptionsControls) {
       if (showDifferentialExpressionTable) {
         // DE table is shown.  Least horizontal space for plots.
@@ -342,7 +343,7 @@ export default function ExploreDisplayTabs({
         // Default state, when side panel is "Options" and not collapsed
         main = 'col-md-10'
         // only set options-bg if we're outside the DE UX
-        side = showDifferentialExpressionPanel ? 'col-md-2' : 'col-md-2 options-bg'
+        side = isSelectingDE ? 'col-md-2' : 'col-md-2 options-bg'
       }
     } else {
       // When options panel is collapsed.  Maximize horizontal space for plots.
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -1,7 +1,7 @@
 import React, { useState, useEffect } from 'react'
 import _clone from 'lodash/clone'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faLink, faArrowLeft, faCog, faTimes, faUndo } from '@fortawesome/free-solid-svg-icons'
+import { faLink, faArrowLeft, faEye, faTimes, faUndo } from '@fortawesome/free-solid-svg-icons'
 
 import StudyGeneField from './StudyGeneField'
 import ClusterSelector from '~/components/visualization/controls/ClusterSelector'
@@ -354,6 +354,8 @@ export default function ExploreDisplayTabs({
   // Determine if the flag show_explore_tab_ux_updates is toggled to show explore tab UX updates
   const isNewExploreUX = getFeatureFlagsWithDefaults()?.show_explore_tab_ux_updates
 
+  const optionsLabel = <span className="options-label">OPTIONS</span>
+
   return (
     <>
       <div className="row">
@@ -435,7 +437,7 @@ export default function ExploreDisplayTabs({
               <button className="action view-options-toggle view-options-toggle-on"
                 onClick={toggleViewOptions}
                 data-analytics-name="view-options-show">
-                OPTIONS <FontAwesomeIcon className="fa-lg" icon={faCog}/>
+                {optionsLabel} <FontAwesomeIcon className="fa-lg" icon={faEye}/>
               </button>
             }
             { enabledTabs.includes('annotatedScatter') &&
@@ -573,7 +575,7 @@ export default function ExploreDisplayTabs({
           <div className="view-options-toggle">
             {!showDifferentialExpressionPanel && !showUpstreamDifferentialExpressionPanel &&
               <>
-                <FontAwesomeIcon className="fa-lg" icon={faCog}/> OPTIONS
+                <FontAwesomeIcon className="fa-lg" icon={faEye}/> {optionsLabel}
                 <button className="action"
                   onClick={toggleViewOptions}
                   title="Hide options"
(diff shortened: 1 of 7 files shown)


bistline:
Build failure tracked in SCP-5159

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -333,6 +333,7 @@ export default function ExploreDisplayTabs({
   function getPanelWidths() {
     let main
     let side
+    const isSelectingDE = showDifferentialExpressionPanel || showUpstreamDifferentialExpressionPanel
```
eweitz, line 336: Convenient!

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -432,10 +434,13 @@ export default function ExploreDisplayTabs({
               />
             }
             { !showViewOptionsControls &&
-              <button className="action view-options-toggle view-options-toggle-on"
+              <button className={showDifferentialExpressionPanel ?
+                "action view-options-toggle view-options-toggle-on" :
+                "action minified-options view-options-toggle view-options-toggle-on"
```
eweitz, line 439: Easier visual diff:

```suggestion
                "action view-options-toggle view-options-toggle-on" :
                "action view-options-toggle view-options-toggle-on minified-options"
```

For even better maintainability, I think this could be something conciser like:

```
className={
 'action view-options-toggle view-options-toggle-on ' +
 showDifferentialExpressionPanel ? '' : 'minified-options'
}
```

But I haven't actually tried that locally.

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -573,8 +578,8 @@ export default function ExploreDisplayTabs({
           <div className="view-options-toggle">
             {!showDifferentialExpressionPanel && !showUpstreamDifferentialExpressionPanel &&
               <>
-                <FontAwesomeIcon className="fa-lg" icon={faCog}/> OPTIONS
-                <button className="action"
+                <FontAwesomeIcon className="fa-lg" icon={faEye}/> <span className="options-label">OPTIONS</span>
+                <button className={showDifferentialExpressionPanel ? "action" : "action action-with-bg"}
```
eweitz, line 582: Perhaps:

```suggestion
                <button className={`action ${showDifferentialExpressionPanel ? '' : 'action-with-bg'}`}
```

app/javascript/components/visualization/controls/AnnotationSelector.jsx
```diff
@@ -53,7 +56,12 @@ export default function AnnotationControl({
 
   return (
     <div className="form-group">
-      <label className="labeled-select">Annotation
+      <label className="labeled-select">Annotation&nbsp;
+        <a className="action help-icon"
+           data-toggle="tooltip"
+           data-original-title="Use this menu to specify how the data are grouped or colored">
```
eweitz, line 62: Briefer, for consideration:

```suggestion
           data-original-title="Select how data are grouped or colored">
```

app/javascript/styles/_explore.scss
```diff
@@ -243,3 +246,24 @@
     border-bottom: 1px solid #ddd;
   }
 }
+
+.options-label {
+  font-size: 1.2em;
+}
+
+.right-margin-10 {
```
eweitz, line 254: Blocker: this too tightly couples selector names to style declarations. It negates the value of selectors by hardcoding the exact style into the name. It's like inline styling, but seems more complex to maintain.  It also switches the usual word order of our selectors.

Let's use less specific selector names, to better convey intent and give more flexibility in the style. E.g. `.margin-extra-left` conveys we something farther left than `.margin-left`, or if we think this warrants a unique styling for the element (this selector is only used once) then `.action-reset-view` would also work.

Regarding word order, the [convention](https://github.com/search?q=repo%3Abroadinstitute%2Fsingle_cell_portal_core%20.margin&type=code) is e.g. `.margin-left` or `.margin-right` or `.margin-extra-left`, rather than `.right-margin-<etc>`.

eweitz (approved):
Code and demo video look good!  The [earlier discussion](https://broadinstitute.slack.com/archives/C024HV2FANL/p1686075982633079) was fruitful.

I note one trivial blocker for maintainability, and suggest several non-blocking refinements.

app/javascript/styles/_colors.scss
```diff
@@ -9,3 +9,4 @@ $warning-icon-color: #C45500;
 
 $danger-color: rgb(219, 50, 20);
 $success-color: #75b041;
```
ehanna4, line 11: TOL:  The existing colors defined here `danger` and `success` are used pretty universally across the app, but I'm not sure `options` has the same general application. It seems we don't have a strong habit of adding colors to this file but considering it's titled `colors.scss` it makes sense to add here haha and I'm wondering if we ought to do more of that.. would it make sense to title this `options-background-color` though to be most specific?

ehanna4 (approved):
The changes look good and running the accessibility audit on the page there no new issues! 

I left a TOL and then have the following concern that is likely outside the scope of this ticket. We now have three info icons in the options panel that have different actions. Annotation-info shows tooltip on hover, DE-button-info shows tooltip on hover and clicking opens modal, Subsampling-info does nothing on hover and clicking opens a different style tooltip..

bistline:
> We now have three info icons in the options panel that have different actions. Annotation-info shows tooltip on hover, DE-button-info shows tooltip on hover and clicking opens modal, Subsampling-info does nothing on hover and clicking opens a different style tooltip

That's another good point - it would make sense to unify them at least in terms of the their "trigger" action.  Lemme see if I can find one that makes sense for all 3.

app/javascript/styles/_explore.scss
bistline: [4e68b70](https://github.com/broadinstitute/single_cell_portal_core/pull/1794/commits/4e68b70238f93feffbfb92dfb44e181f08d5ee89)

app/javascript/components/explore/ExploreDisplayTabs.jsx
bistline: [4e68b70](https://github.com/broadinstitute/single_cell_portal_core/pull/1794/commits/4e68b70238f93feffbfb92dfb44e181f08d5ee89)

app/javascript/components/explore/ExploreDisplayTabs.jsx
bistline: [4e68b70](https://github.com/broadinstitute/single_cell_portal_core/pull/1794/commits/4e68b70238f93feffbfb92dfb44e181f08d5ee89)

PR #1797, opened 2023-06-07
Raise number of study gene search autocomplete results to 50 (SCP-4979)

bistline:
#### BACKGROUND & CHANGES
This update raises the number of autocomplete suggestions from 8 to 50 for study-based gene searches.  This is in response to a user request to address a case where the study had custom gene scores with composite names that contained similar prefixes, like `.m.C1_sick.[gene]`.  Only showing 8 suggestions made it harder to discover these scores, whereas 50 will allow for greater flexibility.  These results scroll automatically in the suggestions div, so no additional UI changes are necessary.

#### MANUAL TESTING
1. Boot as normal and load any synthetic study with many genes, such as `Human all genes`
2. In the gene search field, enter `pt` and confirm you see a scrolling div of suggestions like below:
autocomplete

### DIFF
--- a/app/javascript/lib/search-utils.js
+++ b/app/javascript/lib/search-utils.js
@@ -12,7 +12,7 @@ export const NUM_SUGGESTIONS = 50
  * @param {Array<String>} targets List of strings to match against
  * @param {Integer} numSuggestions Number of suggestions to show
  */
-export function getAutocompleteSuggestions(inputText, targets, numSuggestions= NUM_SUGGESTIONS) {
+export function getAutocompleteSuggestions(inputText, targets, numSuggestions=NUM_SUGGESTIONS) {
   // Autocomplete when user starts typing
   if (!targets?.length || !inputText) {
     return []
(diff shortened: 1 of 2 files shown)


app/javascript/lib/search-utils.js
```diff
@@ -9,7 +12,7 @@ import stringSimilarity from 'string-similarity'
  * @param {Array<String>} targets List of strings to match against
  * @param {Integer} numSuggestions Number of suggestions to show
  */
-export function getAutocompleteSuggestions(inputText, targets, numSuggestions=8) {
+export function getAutocompleteSuggestions(inputText, targets, numSuggestions= NUM_SUGGESTIONS) {
```
eweitz, line 15:
```suggestion
export function getAutocompleteSuggestions(inputText, targets, numSuggestions=NUM_SUGGESTIONS) {
```

eweitz (approved):
Code looks good!  Please add Jira ticket to PR title for auditability.

PR #1799, opened 2023-06-12
Remove react-scripts (SCP-5143)

ehanna4:
We were alerted by App-sec that the dependency "NTH-check" had a vulnerability that required updating. Further investigation by the team determined NTH-check is occurring in react-scripts and it appears that react-scripts is no longer being maintained. Therefore we opted to remove react-scripts from our code. 

Since the team already did a huge chunk of work moving to Vite the lift was not too large to remove react-scripts now.

The changes in this PR were tested locally and on staging. Tests all passed locally, builds built on staging and locally, and running through the release engineering tests on staging all worked as expected. 

In removing react-scripts yarn required adding three dependencies around an existing dependency `postcss`

Testing locally you can pull this branch and run `bundle install` and `yarn install` and boot up vite
`bin/vite dev` 
and boot up a local server 
`./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails s
`
and confirm you can see localhost


eweitz (approved):
Nice!  Related upstream CRA issue: https://github.com/facebook/create-react-app/issues/13186.

Do things work if the new 3 packages are moved from `dependencies` to `devDependencies`?  If so, that'd enable faster build times, and (theoretically, due to less JS being downloaded and parsed) slightly faster page load times.

ehanna4:
@eweitz good idea, locally it appears moving the 3 new dependencies to `devDependencies` does work. I'll make that change and will double check staging once merged

eweitz (approved):

bistline (approved):

PR #1806, opened 2023-06-20
Selection UI for author-computed differential expression (SCP-4925)

eweitz:
This lets users select DE computed by authors, enabling new types of comparison on a broader set of annotations.

### Overview
Previously, DE was only available for cell-type-like groups, as one-vs-rest comparisons, and it was computed by SCP.  This is  popular with users when it's available, but still rather limited.

Now, study owners or editors ("authors" in viewer-facing text) can upload DE results that they have precomputed and refined.  These results can include only one-vs-rest comparisons like before, or only pairwise comparisons, or both.  They can also include DE results not only for just cell-type-like annotations, but any group-based annotation with > 1 group.

This author-computed DE also enables study owners to upload data that is non-exhaustive.  In other words, owners can upload DE for an incomplete set of comparisons.

These updates also make the DE dot plot call-to-action more salient.

### Scenarios
Author-computed DE enables the following the comparison profiles for each annotation group.  Here "all" means "for all potential comparisons in the annotation in this comparison type", and similarly for "some" and "none".

1. All one-vs-rest, all pairwise
2. All one-vs-rest, some pairwise
3. All one-vs-rest, no pairwise
4. Some one-vs-rest, all pairwise
4. No one-vs-rest, all pairwise
5. Some one-vs-rest, some pairwise
6. No one-vs-rest, no pairwise

### Video
Here's how it looks.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/22e003ef-30bf-4769-b132-ebc6ea530860

### Test
An automated test covers basic functionality.  If desired, to manually test:

1.  Go to "Human milk - differential expression" study.  If on staging after this is merged, that's [SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/human-milk-differential-expression).
2. Select annotation "General_Celltype"
3. Click "Differential expression"
4. Confirm two select drop-down menus appear
5. Select "B cells"
6. Click into 2nd menu, confirm it says "rest" then does _not_ say "CSN1S1 macrophages", then says "dendritic cells".  This verifies handling of incomplete pairwise comparisons.
7. Select "rest" in 2nd menu
8. Confirm DE table displays with content in rows
9. Confirm there is a button that says "Dot plot" with an icon at left of the text
10. In the 1st menu, select "CSN1S1 macrophages"
11. Confirm the DE table is removed from the DE panel
12. Click into the 2nd menu, confirm it does _not_ include "rest".  This verifies handling of incomplete one-vs-rest comparisons.  Also confirm it does not include "B cells", which verifies handling of naturally-sorted incomplete pairwise comparisons.
13. Open DevTools, go to Network panel
14. Select "dendritic cells" in 2nd menu
15. Note that no content loads in DE panel, which is expected
16. In Network panel, confirm a request is sent for a URL that includes "cluster_umap_txt--General_Celltype--CSN1S1_macrophages--dendritic_cells--study--wilcoxon.tsv".

Loading author-computed DE results into the DE table will be handling in separate work within the next two weeks.

### Further detail
More context and discussion is available in [this SCP demo video](https://drive.google.com/file/d/1t6oeMQnjgRiWUKkezzffy24_I4k288xC/view?t=3m24s).  

This satisfies SCP-4925.


app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -36,19 +36,29 @@ import Tooltip from 'react-bootstrap/lib/Tooltip'
...
+    selectedAnnot = annotList.default_annotation
   }
 
+  return [selectedCluster, selectedAnnot]
+}
+
+/** Determine if currently selected cluster has differential expression outputs available */
+function getClusterHasDe(exploreInfo, exploreParams) {
+  const flags = getFeatureFlagsWithDefaults()
+  if (!flags?.differential_expression_frontend || !exploreInfo) {return false}
```
bistline, line 58: Thinking out loud - I think we're fairly invested in differential expression visualizations.  Should we retire this feature flag?

app/models/differential_expression_result.rb
```diff
@@ -87,11 +87,72 @@ def filename_for(label, comparison: nil)
...
+        'eosinophils',
+        'fibroblasts',
+        'GPMNB macrophages',
+        'LC1',
+        'LC2',
+        'neutrophils',
+        'T cells'
+      ]
+
+      # Two important notes for pairwise comparisons:
```
bistline, line 116: Useful context!  Thanks for putting this in.

app/models/differential_expression_result.rb
```diff
@@ -87,11 +87,72 @@ def filename_for(label, comparison: nil)
...
+        'dendritic cells' => ['eosinophils', 'fibroblasts', 'GPMNB macrophages', 'LC1', 'LC2', 'neutrophils', 'T cells'],
+        'eosinophils' => ['fibroblasts', 'GPMNB macrophages', 'LC1', 'LC2', 'neutrophils', 'T cells'],
+        'fibroblasts' => ['GPMNB macrophages', 'LC1', 'LC2', 'neutrophils', 'T cells'],
+        'GPMNB macrophages' => ['LC1', 'LC2', 'neutrophils', 'T cells'],
+        'LC1' => ['LC2', 'neutrophils', 'T cells'],
+        'LC2' => ['neutrophils', 'T cells'],
+        'neutrophils' => ['T cells']
+      }
+
+      pairwise_comparisons.each_pair do |label, comparisons|
```
bistline, line 141: Slightly more concise, and also matches the implementation for `one_vs_rest_files`.
```suggestion
      pairwise_files = pairwise_comparisons.map do |label, comparisons|
        comparisons.map do |comparison|
          [label, comparison, filename_for(label, comparison:)]
        end
      end
```

app/models/differential_expression_result.rb
```diff
@@ -87,11 +87,72 @@ def filename_for(label, comparison: nil)
...
   # this is important as it sidesteps the issue of study owners renaming clusters, as cluster_name is cached here
+  #
+  # @return [Hash<String => Array<String, String>, Array<String, String, String>]
   def result_files
-    files = one_vs_rest_comparisons.map { |label| filename_for(label) }
-    Hash[one_vs_rest_comparisons.zip(files)]
+    pairwise_files = []
+    is_author_de = false
+
+    # TODO (SCP-5096): Productionize this block, remove example data
```
bistline, line 98: Just a note: while this is totally fine for the purposes of this work, I would advocate for instead having a setup block in the MANUAL TESTING section of this PR to populate the required data into the `DifferentialExpressionResult` objects that we should have in our local databases.  It pollutes the code less and removes the need to take this out in a successive PR.

Example:
```
result = study.differential_expression_results.find_by(annotation_name: 'General_Celltype')
result.pairwise_comparisons = {
 ... required data
}
result.save!
```

bistline (approved):

app/javascript/components/explore/ExploreDisplayTabs.jsx
eweitz: I had the same thought, then found some prior discussion that we intend to retain the flag so we can easily exempt public studies from displaying DE if the owner requests so.  A new [code comment](https://github.com/broadinstitute/single_cell_portal_core/pull/1806/files#diff-d56fa6c127e17ffa714e80c992b809c04b0dda9a45a93425adca8a56f0d607c2R203) notes this too.

bistline: Oh that's a great point - I agree with that.

app/models/differential_expression_result.rb
eweitz: Yeah, I think this is a kludge almost any way we slice it.  The current approach is noisy, but seems easier to reason about and easier to remove (delete a clear block of code).  The proposed approach is cleaner in code, but harder to reason about, and requires more effort to add and especially to remove and restore to its original local state.

bistline: Fair points, especially regarding cleanup.

eweitz: On second thought, having synthetic data seems like the conventional approach for this sort of thing.  I'll plan to incorporate such study data (and remove this block if it's still here) in my forthcoming PR for adding content to the DE table for author-computed DE data.

ehanna4 (approved):
Overall looks good! I hope to test more on Staging as my local DE set up is not the most reliable and I am still seeing only one dropdown vs rest (but again I think this must be an issue with my local DE study)

I noted a possible function name change in the code and then also I had some slightly less-than-ideal interactions with the dot-plot button in the DE panel, but I see these are also occurring on staging so I will make a new ticket for those things unless you wanted to address any here. I'll just list them here as well:

1. Clicking the dotplot button creates a dotplot and moves you to that tab, however upon changing the annotation in the dropdown the dotplot does not update which is a bit of a confusing experience
2. If I had clicked the dotplot button and then switched to the Heatmap or Genome tab, clicking the dotplot button again does not do anything, I was expecting it would jump me to the dotplot tab
3. Most critically the heatmap content overlaps the DE panel 

Again those last 3 things are all on Staging currently as well. I only mention them here in case they were simple to throw into this work, but I can definitely make a new ticket for those. Please let me know!

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/d13f31bd-5ad0-4557-89d2-8fc3539d7bb1

Screenshot 2023-06-22 at 1 15 21 PM

eweitz:
> Again those last 3 things are all on Staging currently as well. I only mention them here in case they were simple to throw into this work, but I can definitely make a new ticket for those.

Great finds.  All those pre-existing issues seem worth fixing.  I agree the overlapping heatmap issue is most important of the 3 you found.  They shouldn't be hard but also don't seem quite trivial to fix, so if you could open a ticket for those, I'd appreciate it!

app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
```diff
@@ -76,77 +80,174 @@ function getMatchingDeOption(deObjects, group, clusterName, annotation) {
...
-        <br/>
       </div>
-      }
+      {deGenes && <><br/><br/></>}
+    </>
+  )
+}
+
+/** Pick groups of cells for one-vs-rest-only differential expression (DE) */
+export function DifferentialExpressionGroupPicker({
```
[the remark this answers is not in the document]
eweitz: Agreed, I'll update this to something along those lines in my forthcoming PR for the author DE table, and link to that change from this thread when done.  (I merged just now before changing since it helps faster unblock Jon.)

ehanna4:
Ticket for the pre-existing issues here: https://broadworkbench.atlassian.net/browse/SCP-5195

app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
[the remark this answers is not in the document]
eweitz: [Done](https://github.com/broadinstitute/single_cell_portal_core/pull/1813/commits/52974b81b5e9dceae3f3be3e6ed881844b1d012c)!  I went with "OneVsRestDifferentialExpressionGroupPicker" -- longer, but clearer and more guessable.

PR #1816, opened 2023-06-29
Fix row-height bug in morpheus filtering row (SCP-5163)

ehanna4:
The row height of the filtering options in morpheus was previously a strange height making the rows cut off. I removed the height attribute `.slick-cell { height:"+t+"px; }"` to prevent this occurring.

If you search the diff for `.slick-cell { height:"+t+"px; }"` you can see it only appears on the left (old) and not in the new (right) side. 

broken:
Screenshot 2023-06-27 at 2 02 56 PM

Fixed:

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/3d6ef795-5c4b-444a-a2ed-9f0d63fb4965


eweitz (approved):
Demo looks good!

Nit: it looks like the bottom border of the first "Filter" view -- for "Rows" -- gets unintentionally preserved upon switching to "Columns".  See 0:04 in your helpful video.  But that's a slightly different issue, much less severe, and also likely a lot lower incidence.

bistline (approved):

PR #1822, opened 2023-07-13
Prevent clobbering differential expression output files (SCP-5222)

bistline:
#### BACKGROUND & CHANGES
This update moves the cleanup of SCP-calculated differential expression results to before the ingest of author-provided results, instead of after successful parsing.  This prevents the output file cleanup from erroneously removing newly-generated results that may have had the same name as previous results, which just clobbered the existing auto-calculated files.  This scenario can happen if a user selects a cluster/annotation combination that already had results, and they use `wilcoxon` as the computational method.  If the uploaded DE file contains any one-vs-rest comparisons, then these files will be deleted if results already existed for those annotation values.

Additionally, if a study has author-provided DE results, this study will no longer be eligible for automatic differential expression calculations.  This is for two reasons: 1) we want to favor author-provided results, and 2) automatic calculation could potentially clobber existing author result files for the same reason as above.

#### MANUAL TESTING
Note: if you have visualization caching turned on, this can cause problems later in the Explore tab as the response is cached.  You can turn this off temporarily by running the following command from project root (ensuring you see the corresponding message):
```
bin/rails dev:cache
=> Development mode is no longer being cached.
```
2. Select any study that has DE results and go to the upload wizard
3. In a separate browser tab/window, open the Explore tab for the same study
4. Note that the `Differential expression` button is visible, meaning results are available
5. Back in the upload wizard, upload the [test author DE file](https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/development/tests/data/author_de/author_de_test_data_human_milk_All_Cells_UMAP_General_celltype.csv) and wait 10-15 seconds for the ingest process to launch
6. Back in the Explore tab, refresh the page and note that the `Differential expression` is gone
7. Open a Rails console session and confirm that this study is no longer eligible for automatic DE:
```
study = Study.find_by(accession: '<accession of study>')
DifferentialExpressionService.study_eligible?(study)
=> false
```
8. Optional: if you wish to restore the old DE results:
    i. Wait for the ingest to complete, then delete the test DE file
    ii. From the Rails console, run the following command:
```
DifferentialExpressionService.run_differential_expression_on_all("<accession of study>")
```


eweitz (approved):
Code looks good!  Thanks for the clear explanation.

app/lib/differential_expression_service.rb
```diff
@@ -243,7 +244,9 @@ def self.results_exist?(study, annotation)
   def self.study_eligible?(study, skip_existing: false)
     begin
       validate_study(study)
-      find_eligible_annotations(study, skip_existing:).any? && study.has_raw_counts_matrices?
+      find_eligible_annotations(study, skip_existing:).any? &&
+        study.has_raw_counts_matrices? &&
+        !study_has_author_de?(study)
```
eweitz, line 249: When I see multi-clause conditionals broken out onto separate lines in Python or JavaScript, each clause vertically aligns with the others.  Here, the clauses after the first aren't aligned, which caused me to double-take.

Non-blocking, for consideration:

```suggestion
      find_eligible_annotations(study, skip_existing:).any? &&
      study.has_raw_counts_matrices? &&
      !study_has_author_de?(study)
```

(I haven't tested that locally.)

test/services/differential_expression_service_test.rb
```diff
@@ -265,4 +266,16 @@ class DifferentialExpressionServiceTest < ActiveSupport::TestCase
                                     test_array: @@studies_to_clean)
     assert_not DifferentialExpressionService.study_eligible?(empty_study)
   end
+
+  test 'should not mark studies with author de as eligible' do
```
eweitz, line 270: For consistent visual parsing:

```suggestion
  test 'should not mark studies with author DE as eligible' do
```

eweitz (approved):
Code looks good!  I note two subjective maintainability nits, neither is blocking.

app/lib/differential_expression_service.rb
bistline: I take your point about readability.  That being said, the above formatting is considered conventional in Ruby: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineOperationIndentation

My feeling would be to for us to retain idiomatic/conventional usage for each language in our stack.  It will make us all better at reading that code natively, which will pay dividends as we move forward, both in this project and others.

Another example that I know we've debated that seems very relevant: `unless` vs. `if !expression`...

eweitz: Reasonable!

app/lib/differential_expression_service.rb
```diff
@@ -303,6 +306,18 @@ def self.validate_study(study)
...
+  # determine if a study has author-uploaded DE results
+  # this will stop the automatic calculation of new DE results if they add more data
+  #
+  # * *params*
+  #   - +study+ (Study) => Study to validate
+  #
+  # * *returns*
+  #   - (Boolean)
+  def self.study_has_author_de?(study)
+    study.study_files.by_type('Differential Expression').any?
```
ehanna4, line 318: So would this prevent any further automatic calculation even in cases where they had kept the SCP DE results and just added more files after?

ehanna4:
I tested with 2 different studies locally and got this error after attempting to upload the DE file in the upload wizard

Screenshot 2023-07-14 at 4 02 11 PM

I'm always mildly skeptical that something is amiss in my local env for DE but even with refreshing and trying to remediate the error the console specified I can't get the error to go away, maybe we could pair/screen share this next week so we can figure out if it's a real issue or just a wierd thing in my env?

ehanna4 (approved):
Works! The error I was running into was indeed a local configuration issue and not any issue with this PR/code.

PR #1831, opened 2023-07-19
Renew token rather than modal (SCP-5164)

ehanna4:
The issue:

> I often see a modal upon accessing SCP, whether local or on production:
> 
> There appears to be an issue with your session - please sign out and back in again to continue using Single Cell Portal.
> 
> The message is misleading enough to be disruptive.  The issue typically resolves for me upon merely refreshing the page.  It happens to me a few times a day.

This modal was both misleading and failed to log anything to Mixpanel for future analysis or severity/incidence data. This PR removes that modal entirely and renews the token instead. Also adds logging to Mixpanel when this renewal occurs.

Testing is a bit tedious as you must wait an hour without doing anything with your local env 
- Pull this branch and boot up your local environment
- sign in 
- let page sit for 60min
- after 60min refresh or click a study link
- no modal will pop up and there should be an event in mixpanel :
- you can also check your userauthtoeken in the console with `window.SCP`


app/javascript/lib/scp-api.jsx
```diff
@@ -243,6 +243,19 @@ export function setupRenewalForReadOnlyToken(studyAccession) {
   }, renewalTime)
 }
 
+
+/**
+ * Renew userAccessToken
+ * @param {String} studyAccession Study accession, e.g. SCP123
+ */
+export async function renewUserAccessToken(studyAccession) {
```
bistline, line 251: Thinking out loud: Right now, there's always a study being used in context with this request. But what if the user is on the home page and their session is about to expire? It impacts search/download requests, so it's definitely a valid context. However, the workaround of refreshing the page works fine as they won't lose state, so I don't consider it blocking. It is something we should consider moving forward, though - the renewUserAccessToken method should likely be adapted to handle this case.

bistline (approved):
Code looks good, will have to leave a browser open for a while for the manual test.  Question below regarding non-study based scenarios.

ehanna4:
I'm not convinced by the results of this change, I think a more root cause fix would be better. I'm going to close this in favor of pursuing that.

eweitz (commented):
_This PR self-closed a little while ago, and Emily kindly let me know.  We agreed I'll leave this review since it was nearly done and might illuminate a complicated area of SCP IAM code._

---

Kudos for tackling a problem that might seem easy on the surface, but exists atop a hard-to-reason-about area of the code.  

As is, this PR conflates two kinds of token, which seems likely to have adverse effects.

The frontend uses two kinds of access token:

* User access token: authenticates the user for SCP API access (see [1](https://github.com/broadinstitute/single_cell_portal_core/blob/e3eea038934d6a1464ff955b31612cfffdbb486d/app/javascript/providers/UserProvider.jsx#L11), [2](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/javascript/lib/scp-api.jsx#L43)), and also determines [whether the user is anonymous](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/javascript/providers/UserProvider.jsx#L95).  The SCP API access is more of a concern than the "isAnonymous", since the latter seems like dead code.

* Read-only access token (a.k.a. "GCS access token", a.k.a. "read access token"): authenticates the user for GCP / GCS bucket access, _even when they are not signed in_.

The `/renew_token` SCP API endpoint currently just renews the read-only access token, not the user access token.  However, this PR calls `/renew_token` for the user access token, which is the root of the conflation.  The diagrams and "Authentication timeout scenarios" section in #1747 summarize logic around `/renew_token`.  The same expiry and renewal logic for read-only access tokens in `renew_token` seems applicable to user access tokens, it's just that we should have a way to return the proper kind of token.

Here's a possible effect of using read-only access token as a user token:

If the user is signed in and their user access token is an empty string (which might be mostly a side-effect of [increasing session length from 1 hour to 24 hours](https://github.com/broadinstitute/single_cell_portal_core/pull/1707)), and the user is in a study-specific view (i.e. `studyAccession` is available), then their "user access token" would be overwritten with a "read-only access token".  In that case, if there study is public, then they would get errors for SCP API calls that involve writes.

Possible solution: 
* Extend `/renew_token` to accept a parameter like "type", which accepts "user_access_token" or "read_only_access_token", and then calls existing downstream code.
* Alternatively, I could also see renaming `/renew_token` to `/renew_read_only_access_token` and adding a new SCP API endpoint `/renew_user_access_token`.

PR #1820, opened 2023-07-11
Preserving all test logs (SCP-5028)

bistline:
#### BACKGROUND & CHANGES
This update persists all test logfiles along with every CI run as workflow artifacts.  These logs are automatically persisted for up to 90 days, and are available at the bottom of the workflow summary page (under `Artifacts`).

Example:
workflow_logs

#### MANUAL TESTING
1. Go to the [workflow summary page](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/5521868919) for the last run in this PR
2. Confirm you can download the log files at the bottom of the page and that both `test.log` and `delayed_job.test.log` are present

PR #1840, opened 2023-07-26
Always save logs in CI runs (SCP-5028)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where failed CI runs do not save logfiles.  Now, logs are always retained, regardless of status.


eweitz (approved):
Thanks for this quick fix re #1820!

ehanna4 (approved):
Nice!

PR #1843, opened 2023-07-31
Multi-gene support and match mode for differential expression search (SCP-5227)

eweitz:
This lets users easily assess differential expression for multiple genes, e.g. from a pathway.

### Multi-gene DE find
Previously (#1823), the "Find a gene" feature only allowed users to search the DE table for one gene name at a time.  This was useful but rather limited.    Now, the feature is upgraded to "Find genes", and lets users type in or paste multiple genes as a list, e.g. from a biochemical pathway or gene module.

Related improvements complement this new multi-gene support for the "DE find" feature.

### Unfound genes ###
Before, when a searched gene symbol wasn't found, the table was simply blank.  Now, it explicitly tells the user no genes were found, and how to address that.  This "unfound genes" handling also exists for the multi-gene case.  That's necessarily more complex: often, _some_ searched genes are found, and some are not.  The UI shows which genes weren't found.  It's a compact visual summary, with intuitive ways to copy or simply see the the full list of unfound genes, and to clear (only) the unfound genes from the gene name search input box.

### Match mode ###
By default, DE find returns partial substring matches.  This is how "Find" works in most applications, and is a decent starting point here.  However, just because you're searching "CD4" doesn't mean you're also interested in "CD44".  Such partial matches aren't a significant issue for single-gene DE find, but it adds _a lot_ of noise for a main use case of multi-gene DE find -- assessing DE in a list of genes from a pathway.  To mitigate this, a new bullseye icon appears at right of the search box.  Hovering over it briefly explains its purpose: changing gene name search to match full names.  Clicking the bullseye toggles between full- and partial-string matching on each gene name token in the search box.

### Responsive layout, etc.
With more going on in the DE find feature, it's worth ensuring the layout doesn't get mangled.  Specific UI elements now more robustly adjust and resize to keep the experience smooth even in narrower or shorter viewports.

A handful of small bugs in related analytics instrumentation is fixed now, as well.

### Video
Here's how it looks, in the context of a realistic use case.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/69963e01-8cc5-4047-8fa2-9bf802ec012c

### Test
Automated tests verify behavior.  To manually test:

1.  Go to DE pilot study
2. Click "Differential expression" button, select "B cell"
3. In "Find genes" search box, type in "CD4"
4. Confirm CD4 and CD44 show in table, among other genes
5. Click bullseye icon at right of DE "Find genes" search box
6. Confirm _only_ CD4 appears in DE table
7. In "Find genes", add " FASN" to your query -- i.e. making it "CD4 FASN"
8. Confirm CD4 and FASN (and only those genes) appear in DE table
9. Add " asdf qwer zxcv" to your DE search box, making the query "CD4 FASN asdf qwer zxcv"
10. Confirm message below search box says "Genes not found: asdf qwer and 1 more"
11. Hover over "1 more", confirm tooltip includes "zxcv"
12. Click copy icon at right of message, paste new clipboard content somewhere
13. Confirm clipboard content is "Genes not found (3): asdf, qwer, zxcv"
14. Click clear ("x") icon at right of message
15. Confirm "asdf", "qwer", and "zxcv" disappear from search box, leaving only found genes ("CD4 FASN")

### Further detail
More context and discussion are available in this [SCP demo video](https://drive.google.com/file/d/1G-ylcHhWOCYkAUc3Msq-kiu1lKwPYY9v/view?t=0m24s).  This satisfies SCP-5227.


bistline (approved):
Very slick!  Works as advertised, and is surprisingly full-featured for such a small control.

I am _slightly_ concerned about the amount of features we've added in given the amount of screen real estate - it's a lot of icons with tooltips that do some somewhat non-standard things - but the associated instrumentation should give us insight into how well users are handling all of that.

eweitz:
I share your mild concern around density and novelty in this UI.  

To mitigate overcrowding, I've tried to carefully use _conditional_ display, e.g. only showing the match mode icon when there's a gene-name query, or only showing the copy icon when not all unfound genes are displayed.  To mitigate confusion, I'm hoping the tooltips shown on hover and the slightly-thematic icons will be easily discovered.  

I just added an analytics report on "[Buttons atop DE table](https://mixpanel.com/s/3ijcZc)" to the "[Differential expression](https://mixpanel.com/s/3ijcZc)" board in Mixpanel.  To your point, that'll help evaluate how much users engage with these auxiliary controls.  If we see they get less usage than expected, then we can refine their design, our onboarding, and / or perhaps simply remove some.

app/javascript/components/explore/DifferentialExpressionPanel.jsx
```diff
@@ -312,68 +332,91 @@ function DifferentialExpressionTable({
   return (
     <>
       <div className="de-table-buttons">
-        <DotPlotButton dotPlotGenes={dotPlotGenes} searchGenes={searchGenes} />
-        <DownloadButton bucketId={bucketId} deFilePath={deFilePath} />
+        {genesToShow.length > 0 &&
```
ehanna4, line 335: Not a blocker:
I see `genesToShow.length` 4 times I think in this first section, I wonder if these could be pulled into one const? Also might be beneficial to add a check for `genesToShow` so maybe something like:
`const genesToShowLength = genesToShow?.length`

app/javascript/components/explore/DifferentialExpressionPanel.jsx
```diff
@@ -409,20 +452,137 @@ function rangeFilterGenes(deFacets, deGenes, activeFacets) {
...
+
+/** Copy unfound genes to user's system clipboard */
+function copyUnfoundGenes(unfoundGenes) {
+  const numUnfound = unfoundGenes.length
+  const unfound = unfoundGenes.join(', ')
+  copyToClipboard(`Genes not found (${numUnfound}): ${unfound}`)
+}
+
+/** Clear gene names that haven't been found in multi-gene DE search */
+function clearUnfoundGeneNames(unfoundGenes, searchedGenes, setSearchedGenes) {
```
ehanna4, line 530: This would be so cool to have for the general study gene search too

ehanna4 (approved):
The attention to detail on all the UX nice-ities on this is really nice. The clearing of unfound genes, toggling between partial and full match, copy unfound genes (including the hidden from view ones), appropriately showing a portion of the unfound genes with '...' and tooltips, etc. etc. it feels really smooth from a user perspective.

I would love to see some of those features in the general study gene search, what do others think of if we make a ticket to look into pulling any of those features out and up to be used more widely in the app?

overall really nice 👍

PR #1844, opened 2023-07-31
Prevent upload of author DE files without clustering & annotation associations (SCP-5234)

bistline:
#### BACKGROUND & CHANGES
This update fixes an issue where study owners were able to upload a differential expression file to be parsed before adding any metadata or clustering files.  This resulted in a file that was stuck and could not be ingested as is.  Now, the save button is disabled until those associations are set, which can only happen once both the metadata and cluster files have been ingested.

New validation messages:
Screenshot 2023-07-31 at 2 44 36 PM

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to the `Feature flags` admin page and confirm that the `show_de_upload` feature flag is turned on (or is enabled for your user account if not)
3. Go to the upload wizard for any study that has both clustering and metadata
4. Click the nav entry for `Differential expression` and confirm that the save button is disabled with the above messages
5. Select a clustering file and annotation, then select any text file to upload (doesn't matter as we're not saving)
6. Confirm that the save button enables once all required fields are set

### DIFF
--- a/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
+++ b/app/javascript/components/upload/DifferentialExpressionFileForm.jsx
@@ -8,6 +8,10 @@ import { clusterFileFilter } from './ClusteringStep'
 import CreatableSelect from 'react-select/creatable'
 
 const allowedFileExts = FileTypeExtensions.plainText
+const requiredFields = [
+  { label: 'Associated annotation', propertyName: 'differential_expression_file_info.annotation_name'},
+  { label: 'Associated clustering file', propertyName: 'differential_expression_file_info.clustering_association'},
+]
 
 /** renders a form for editing/uploading a differential expression file */
 export default function DifferentialExpressionFileForm({
@@ -23,7 +27,7 @@ export default function DifferentialExpressionFileForm({
   menuOptions
 }) {
   // TODO (SCP-5154) Add DE specific clientside validation
-  const validationMessages = validateFile({ file, allFiles, allowedFileExts })
+  const validationMessages = validateFile({ file, allFiles, allowedFileExts, requiredFields })
 
   const fragmentType = isAnnDataExperience ? 'cluster' : null
 
--- a/app/javascript/components/upload/DifferentialExpressionStep.jsx
+++ b/app/javascript/components/upload/DifferentialExpressionStep.jsx
@@ -49,7 +49,8 @@ export function DifferentialFileUploadForm({
     <div className="row">
       <div className="col-md-12">
         <p className="form-terra">
-          Upload a file with differential expression for a particular clustering and annotation.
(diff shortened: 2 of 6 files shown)


app/javascript/components/upload/DifferentialExpressionFileForm.jsx
```diff
@@ -8,6 +8,10 @@ import { clusterFileFilter } from './ClusteringStep'
 import CreatableSelect from 'react-select/creatable'
 
 const allowedFileExts = FileTypeExtensions.plainText
+const requiredFields = [
+  { label: 'Associated annotation', propertyName: 'differential_expression_file_info.annotation_name'},
+  { label: 'Associated clustering file', propertyName: 'differential_expression_file_info.clustering_association'},
```
eweitz, line 13: To refine casing from what the screenshot shows is

> You must specify Associated annotation
> You must specify Associated clustering file

to

> You must specify associated annotation
> You must specify associated clustering file

```suggestion
  { label: 'associated annotation', propertyName: 'differential_expression_file_info.annotation_name'},
  { label: 'associated clustering file', propertyName: 'differential_expression_file_info.clustering_association'},
```

eweitz (approved):
Code and screenshot look good!  I suggest a refinement that's out-of-scope, but germane.

When uploading a processed matrix, specifying an associated raw matrix requires reloading the page .  I imagine that also affects specifying associated annotation and clustering, right?  If so, that makes SCP-5221 higher value to fix.

bistline:
> Code and screenshot look good! I suggest a non-blocking refinement in user-facing text.
> 
> When uploading a processed matrix, specifying an associated raw matrix requires reloading the page . I imagine that also affects specifying associated annotation and clustering, right? If so, that makes [SCP-5221](https://broadworkbench.atlassian.net/browse/SCP-5221) higher value to fix.

That's a great point - Let me try that out locally and see if the behavior is the same.  I think it depends on what level the necessary data is rendered into the necessary component.

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
ehanna4: I actually like the exact word usage so I can map it directly to a spot in the form, though in that case it might make sense to utilize semi-colons to make the sentence structure better like:

> You must specify: Associated annotation
> You must specify: Associated clustering file

ehanna4 (approved):
Works as advertised 👍

app/javascript/components/upload/DifferentialExpressionFileForm.jsx
bistline: I like this better - I agree about matching the form case to the message, and this seems like a good middle ground that will benefit all forms.

eweitz: Excellent suggestion, Emily!  I agree.

PR #1845, opened 2023-08-02
Remove modal and renew user access token (SCP-5164)

ehanna4:
This updates the handling for a user's access token expiring. Previously we popped open a modal telling users to re-sign-in if they were logged in but the `window.SCP.userAccessToken` was blank, this modal was not helpful because the user didn't actually need to re-sign-in they would just need to refresh the page. This change follows the logic utilized for the `readOnlyToken` and renews the `userAccessToken` before it expires and removes the modal entirely so the user should not be bothered or notice the token getting refreshed as appropriate.

To test:
You can check out the new API endpoint in swagger - https://localhost:3000/single_cell/api/v1#/Site/site_renew_user_access_token 

If you don't authorize you won't get a token, but if you authorize before trying it you will see the returned token and expiration info.

You can do a semi-proper test locally by going to line 268 in `scp-api.jsx` and change the renewalTime to something much quicker like 60000 (1 minute). 
Then:
- boot up your localhost
- watch the rails log for something like after the allotted time you set for the timeout above:

> Started GET "/single_cell/api/v1/site/renew_user_access_token" for 127.0.0.1 at 2023-08-02 14:20:58 -0400
Processing by Api::V1::SiteController#renew_user_access_token as JSON
Renewing user access token via SCP API for #<User:0x00000002853f5e80>
Completed 200 OK in 1550ms (Views: 0.7ms | MongoDB: 0.8ms | Allocations: 61357)


app/javascript/lib/scp-api.jsx
```diff
@@ -243,6 +242,33 @@ export function setupRenewalForReadOnlyToken(studyAccession) {
...
+    const userAccessTokenObject = camelcaseKeys(window.SCP.userAccessTokenObject)
+    // Set the renewal time to be 5 minutes before expiration
+    renewalTime = userAccessTokenObject.expiresIn * 1000 - FIVE_MINUTES
+  }
+
+  setTimeout(async () => {
+    const apiUrl = `/site/renew_user_access_token`
+    const [response] = await scpApi(apiUrl, defaultInit())
+    const userAccessTokenObj = response
+    window.SCP.userAccessTokenObject = userAccessTokenObj
```
eweitz, line 265: This seems to be the only place `window.SCP.userAccessTokenObject` is initialized.  That entails this API call will be issued at page load, on every page load, for signed-in users.  Such load seems unnecessary.

Both `readOnlyAccessTokenObject` and the original `window.SCP.userAccessToken` are set on page load, server-side, in a Rails template.  The latter is set like so in [line 36 of /app/views/layouts/application.html.erb](https://github.com/broadinstitute/single_cell_portal_core/blob/850f7da6db0aa961087392a5df0af5e79e69b562/app/views/layouts/application.html.erb#L36):

```
window.SCP.userAccessToken = '<%= get_user_access_token(current_user) %>';
```

Note how the analogous read-only access token is initialized around [line 5 in /app/views/site/_study_visualize.html.erb](https://github.com/broadinstitute/single_cell_portal_core/blob/850f7da6db0aa961087392a5df0af5e79e69b562/app/views/site/_study_visualize.html.erb#L5):

```
    if (typeof window.SCP.readOnlyToken === 'undefined') {
      window.SCP.readOnlyTokenObject = <%= raw get_read_access_token(@study, current_user).to_json %>
      window.SCP.readOnlyToken = window.SCP.readOnlyTokenObject.access_token
    }
```

Let's apply that pattern in line 36 of `/app/views/layouts/application.html.erb`.  

Also, importantly, note how in `setupRenewalForReadOnlyToken` we do:

```
    const [response] = await scpApi(apiUrl, defaultInit())
    const readOnlyTokenObject = response
    window.SCP.readOnlyToken = readOnlyTokenObject.accessToken
```

We need to similarly set `window.SCP.userAccessToken` in this function.  That's [used](https://github.com/search?q=repo%3Abroadinstitute%2Fsingle_cell_portal_core%20userAccessToken&type=code) in a few critical places, which are ultimately used to set the `Bearer` token in e.g. `scp-api.jsx`, `metrics-api.jsx`, and elsewhere.

eweitz (changes requested):
This improves things considerably over #1831!  I note a subtle but significant bug -- details below.

app/javascript/lib/scp-api.jsx
ehanna4: ah okay yeah I'll make the update in application.html.erb to set up the tokenObject

For `window.SCP.userAccessToken` it is set in the `generate_user_access_token()` function that is called to get the token in scp-api so it shouldn't need to be set again here. I mentioned that in the comment right below this line, but if that doesn't seem clear I can look at moving things around. Since `generate_user_access_token()` was an existing function I didn't want to change it up unnecessarily which is why I didn't pull that out, but I agree it's not necessarily as clear.

eweitz: I was curious about that comment, i.e.:

> `// the userAccessToken is set in the window in the generate_access_token in the user model`

A bit more specifically, `userAccessToken` is set in the `window` in the `application` view via `generate_access_token`, which is defined in the `user` model.

> For window.SCP.userAccessToken it is set in the generate_user_access_token() function that is called to get the token in scp-api so it shouldn't need to be set again here.

Right now, as you allude to, `userAccessToken` is only set in `application.html.erb`.  The problem I'm drawing attention to here is that when `userAccessToken` expires while a page is in view, the current PR's code will not renew it, even though the function says it will in its docstring.  

The problem the `renewFooToken` functions solve is that access tokens expire in 60 minutes after being granted, but a page can be viewed (i.e., not reloaded) for longer and the user expects its features to still work at, say, minute 75.  This is described more fully in #1747.  

More concretely, the problem is that it's `userAccessToken` which is used as the bearer token for client-side access to e.g. the SCP API and Bard API (via `getAccessToken()`, which resolves to `window.SCP.userAccessToken`), but `userAccessToken` is not getting its value renewed here.  The variable `userAccessTokenObject` is not directly used for the bearer token (nor should it).  To fix the expiry-during-page-view issue this function is intended to fix, the `window.SCP.userAccessToken` needs to be assigned the renewed access token in `setUpRenewalForUserAccessToken`.

eweitz: (To keep track of blocking issues, I updated my original comment in this thread to mark this as a place I think a change needs to be made.)

app/views/layouts/application.html.erb
```diff
@@ -32,7 +32,7 @@
           localStorage.removeItem('previous-search-url');
           window.location = searchUrl;
       }
-
+      window.SCP.userAccessTokenObject = <%= raw get_user_access_token(current_user).to_json %>
       window.SCP.userAccessToken = '<%= get_user_access_token(current_user) %>';
```
eweitz, line 36: This may well work!  However, if it does, it'd be through a surprising mechanism that would confuse maintainers.

I just pulled this down, and see that `window.SCP.userAccessTokenObject` is set to a string value (`"ya29...`) here, not JSON as the code assumes.  That's because this `get_user_access_token` returns a string, not JSON.  This causes `renewalTime` over in `setUpRenewalForUserAccessToken` to get set to `NaN` on the first evaluation, which triggers an immediate request to `/site/renew_user_access_token`.  That endpoint then returns JSON as expected, and seems to set things right.

So this seems like it'll cause `/site/renew_user_access_token` to always be called soon (but not immediately) on page load, which is unexpected.  

I see other calls to `/search/facets`, `/search?type=gene&page=1`, and `/search?type=study&page=1` and `/site/check_terra_tos_acceptance` all precede that.  Per my [notes yesterday](https://github.com/broadinstitute/single_cell_portal_core/pull/1845#discussion_r1282309732), that initial 4-call volley will use the value of `userAccessToken` as set in line 36 here.  I'm curious if `userAccessToken` as set in line 36 still gets initialized to an empty string now.  If so, then the initial call volley seems likely to return unexpected results, since they depend in part on a user's authentication / identity, e.g. search results partly depend on what you're authorized to see.

I realize this is a lot of detail; I note it here to hopefully help clarify complicated auth flows.  I'm happy to chat or pair if that seems useful!

bistline (commented):
Couple small style nits, I'll wait until the changes requested are done to submit the review.

app/views/layouts/application.html.erb
ehanna4: Yep definitely committed that too soon, please see: [a6eaa2c](https://github.com/broadinstitute/single_cell_portal_core/pull/1845/commits/a6eaa2cf3af35447792ea6c602c715deafd0dcc5) for better handling

app/javascript/vite/application.js
```diff
@@ -61,7 +60,10 @@ document.addEventListener('DOMContentLoaded', () => {
     })
   }
 
-  setTimeout(checkMissingAuthToken, 1000);
+  if (window.SCP.userSignedIn) {
+    ScpApi.setUpRenewalForUserAccessToken()
+  }
```
eweitz, line 65: Blocker: I'm concerned that we silently drop a clause from the [original check](https://github.com/broadinstitute/single_cell_portal_core/pull/1845/files#diff-154a7bd91be253d5383f8b76ec079079cff01ce9b80c97a633ebbaae1b5e47b3L3), which was:

```
if (window.SCP.userAccessToken === '' && window.SCP.userSignedIn) {
```

Two pre-existing problems:

1.`window.SCP.userAccessToken` is an empty string if the user stays signed in but loads a new page after only an hour.  Proper handling would be to refresh the user access token server-side, _before_ the very first HTTP response in the page load (i.e., before the HTML document is returned).

2. When `window.SCP.userAccessToken` is an empty string, stuff still seems to basically work.  This is unsettling, and a sign we don't understand something important.  SCP API calls ought to fail when `window.SCP.userAccessToken === '' && window.SCP.userSignedIn`.  The fact that SCP API calls work when the bearer token is empty means that token isn't working like we expect, which is a problem.

Fixing those underlying issues is out-of-scope here.  However, this PR degrades our observability into this issue.  Before, we developers would at least get annoyed by the modal telling us to refresh the page.  We could dismiss the modal and use the page with SCP API working, but it was still a valid warning that _something is awry_.  We knew when those 2 pre-existing problems manifested.  Now, we lack even that informal observability.  We'll now only be able to detect cases of that problematic `window.SCP.userAccessToken === ''` scenario via actively querying DevTools Console.

If we don't fix problem #1 here -- and I think that'd be reasonable as a separate task -- then we at least need to not degrade our observability into the incidence of pre-existing problem #2.  I could go either way on keeping or removing the modal -- it's very annoying, but seems to report a deep true positive.  In either case, I think we need to log to Mixpanel pages that load and have `window.SCP.userAccessToken === '' && window.SCP.userSignedIn`.

Perhaps:

```suggestion
  if (window.SCP.userSignedIn) {
    ScpApi.setUpRenewalForUserAccessToken()
    if (window.SCP.userAccessToken === '') {
       log('error', 'user-access-token-empty-string')
    }
  }
```

eweitz (approved):
I note a trivial blocker, but other than that this code looks good!

app/javascript/vite/application.js
ehanna4: Good call, done in [4333211](https://github.com/broadinstitute/single_cell_portal_core/pull/1845/commits/4333211d2e96cff20c982cdc2b44a646acb645db)

bistline (approved):
Looks good - I suppose it's worth mentioning that this won't address the corner case where we don't get the user token in the DOM and the user never refreshes the page.  There's not really a solution to be found there - just calling it out.  Otherwise I think this refinement will be a _huge_ improvement over the current UX.

PR #1849, opened 2023-08-07
Fix regression from userAccessToken renewal PR (SCP-5164)

ehanna4:
This fixes the issue caused by the [PR #1845](https://github.com/broadinstitute/single_cell_portal_core/pull/1845)

Previously before that PR the userAccessToken itself was being returned via `get_user_access_token()` but that got refactored to `get_user_access_token_hash` to get the full object with the expirary info as well. There wasn't sufficient checks done then on the resulting object and so that broke the home page search when a user was signed out. This fixes that so the token is set appropriately either to a token or ""

Error:
Screenshot 2023-08-07 at 4 16 34 PM

Fixed - not signed in
Screenshot 2023-08-07 at 4 45 28 PM

Signed in
Screenshot 2023-08-07 at 4 46 12 PM

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -354,6 +354,8 @@ module ApplicationHelper
   def get_user_access_token_hash(user)
     if user.present?
       user.valid_access_token
+    else
+      {}
     end
   end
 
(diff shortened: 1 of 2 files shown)


bistline (approved):
Thanks for the super speedy turnaround!  Confirmed that this fixes the issue (can reproduce by switching to `development` and back while not signed in).

eweitz (approved):

PR #1850, opened 2023-08-09
Fix false positive validation failures due to non-ASCII data (SCP-5271)

eweitz:
This fixes a bug where characters like "γδ" in files would cause validation failures, even though all rules were satisfied.

Previously, client-side file validation (CSFV) used _string length_ in its chunked file reader.  This works when characters are all ASCII, as each is 1 byte.  But it often fails for non-ASCII characters, as each is > 1 byte.  This caused CSFV to fail for files that exceed the size of a chunk (which is common) and have non-ASCII characters (which is rare).

Now, CSFV handles large files that contain non-ASCII data, by using _byte length_ of the string.  It does so via `TextEncoder`.  (We could coincidentally also use `Buffer`, but that Node construct is only available in our stack due to a dependency that's somewhat unstably written to SCP's global scope via igv.js.)

To test:
* Go to upload wizard
* In "Clustering" tab, choose the previously false-positive-case at `test/test_data/cluster_PR_1850_includes_non-ascii.txt`
* Confirm no validation error is thrown
* Choose the (pre- and post-change) true-negative-case at `test/test_data/cluster_PR_1850_ascii_only.txt`
* Confirm no validation error is thrown

This satisfies SCP-5271.


bistline (approved):
Nice and neat - works as advertised.

ehanna4 (approved):
Nice fix and thanks for adding the test file cases!

PR #1854, opened 2023-08-10
Mass assign studies to collections as curator (SCP-2624)

bistline:
#### BACKGROUND
As a collection curator, you can currently only add studies individually to collections by editing each study directly.  For large collections, this can be painful to manage, as well as very slow.

#### CHANGES
This update adds the ability to mass assign studies directly on the collection itself.   An new form field called `Included studies` will allow the curator to enter a comma/space delimited list of study accessions.  This list is sanitized and validated before all valid studies are assigned to the collection.  Any studies that do not pass validation (either don't exist or curator does not have permission to view them) will be removed, and the curator will be notified which studies were not added.

Updated form:
Screenshot 2023-08-10 at 4 05 41 PM

#### MANUAL TESTING
1. Boot as normal and sign in
2. Click `Browse collections` and click `Edit` next to any collection
3. In the `Included studies` field, enter a selection of valid study accessions, using a mix of commas/spaces, with extra space as well between accessions
4. Click `Update collection` at the bottom of the page
5. After the collection saves, you should see list of buttons to view each study from the edit page
Screenshot 2023-08-10 at 4 45 15 PM
6. Clicking any of the links should load that study in the corresponding collection view (with `?scpbr=` in the URL)
7. Go back to the edit page, and enter a bogus accession like `SCP00000`
8. Confirm the collection still save, but that you get a message saying `SCP00000 could not be added to this collection`.

### DIFF
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -38,6 +38,8 @@ class BrandingGroupsController < ApplicationController
     clean_params = branding_group_params.to_h
     clean_params[:users] = self.class.merge_curator_params(params[:curator_emails], nil, current_user)
     clean_params.delete(:user_ids)
+    clean_params[:studies] = self.class.merge_study_params(params[:study_accessions], current_user)
+    clean_params.delete(:study_ids)
     @branding_group = BrandingGroup.new(clean_params)
 
     respond_to do |format|
@@ -67,11 +69,15 @@ class BrandingGroupsController < ApplicationController
         # delete the param since it is not a real model param
         clean_params.delete("reset_#{image_name}")
 
-        # merge in curator params
+        # merge in curator and study params
         clean_params[:users] = self.class.merge_curator_params(
           params[:curator_emails], @branding_group, current_user
         )
         clean_params.delete(:user_ids)
+        clean_params[:studies] = self.class.merge_study_params(
+          params[:study_accessions], current_user
+        )
+        clean_params.delete(:study_ids)
       end
 
       if @branding_group.update(clean_params)
@@ -109,6 +115,13 @@ class BrandingGroupsController < ApplicationController
     users
   end
 
+  # allow mass assignment of studies from collection edit view
+  def self.merge_study_params(study_list, user)
+    accessions = study_list.split(',').map(&:strip)
+    # skip checking group shares for performance reasons
+    Study.where(:accession.in => accessions).select { |study| study.can_view?(user, check_groups: false) }
+  end
+
   private
 
   # Use callbacks to share common setup or constraints between actions.
@@ -120,8 +133,8 @@ class BrandingGroupsController < ApplicationController
   def branding_group_params
     params.require(:branding_group).permit(:name, :tag_line, :public, :background_color, :font_family, :font_color,
                                            :splash_image, :banner_image, :footer_image, :external_link_url, :external_link_description,
-                                           :reset_splash_image, :reset_footer_image, :reset_banner_image,
-                                           user_ids: [])
+                                           :reset_splash_image, :reset_footer_image, :reset_banner_image, :user_ids,
+                                           :study_ids)
   end
 
   def authenticate_curator
--- a/app/controllers/branding_groups_controller.rb
+++ b/app/controllers/branding_groups_controller.rb
@@ -94,7 +94,7 @@ class BrandingGroupsController < ApplicationController
       )
 
       if @branding_group.update(clean_params)
-        notice = "Collection '#{@branding_group.name}' was successfully updated."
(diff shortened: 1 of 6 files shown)


app/controllers/branding_groups_controller.rb
```diff
@@ -36,16 +36,27 @@ def edit
...
+    studies = self.class.find_studies_from_accessions(params[:study_accessions], current_user)
+    clean_params[:study_ids] = studies.map(&:id)
+    missing_studies = self.class.get_missing_studies(
+      self.class.param_to_array(params[:study_accessions]), studies
+    )
     @branding_group = BrandingGroup.new(clean_params)
 
     respond_to do |format|
       if @branding_group.save
+        notice = "Collection '#{@branding_group.name}' was successfully updated."
```
eweitz, line 50: Non-blocking: we conventionally use double quotes in user-facing text and put the variable part last.  

Putting variable parts last makes a text message more robust for visual layout in the app and debugging contexts like Mixpanel, Sentry or grep; and makes it easier to search in those debugging contexts.  I don't recall off-hand if these messages get logged via our modal logging to e.g. Mixpanel, but it seems like good practice nonetheless.

```suggestion
        notice = "Successfully updated collection \"#{@branding_group.name}\""
```

test/controllers/branding_group_controller_test.rb
```diff
@@ -72,8 +82,13 @@ class BrandingGroupControllerTest < ActionDispatch::IntegrationTest
...
     assert @collection.can_edit?(@user)
     assert @collection.can_edit?(new_user)
+    assert_equal 3, @collection.study_list.size
+    assert_equal accessions.sort, @collection.study_list.sort
+    accessions = [studies.first.accession]
+    # test sanitizer logic by using blank space and extra commas
     curator_params = {
-      curator_emails: "#{@user.email}",
+      curator_emails: " #{@user.email}  , ,",
+      study_accessions: "      #{studies.first.accession},,, ",
```
eweitz, line 91: You know a comment like line 88 is useful when a reviewer overlooks it at first, and almost requests exactly what you wrote. :)

app/controllers/branding_groups_controller.rb
```diff
@@ -109,6 +134,23 @@ def self.merge_curator_params(curator_list, collection, user)
...
   end
 
+  # allow mass assignment of studies from collection edit view
+  def self.find_studies_from_accessions(study_list, user)
+    accessions = param_to_array(study_list)
+    # skip checking group shares for performance reasons
+    Study.where(:accession.in => accessions).select { |study| study.can_view?(user, check_groups: false) }
+  end
+
+  # covert a comma- or space-delimited string to an array of strings, removing empty values
```
eweitz, line 144:
```suggestion
  # convert a comma- or space-delimited string to an array of strings, removing empty values
```

app/controllers/branding_groups_controller.rb
```diff
@@ -109,6 +134,23 @@ def self.merge_curator_params(curator_list, collection, user)
...
+  def self.param_to_array(param)
+    param.split(/[,\s]/).map(&:strip).reject(&:blank?)
+  end
+
+  # detect which studies could not be saved (either don't exist or curator does't have permission)
+  def self.get_missing_studies(original_accessions, studies)
+    original_accessions - studies.map(&:accession)
```
eweitz, line 151: Such an elegant construct in Ruby, `array - array`.  I didn't know this was possible.

For reference, JavaScript and Python don't support this (to their discredit, IMO).

app/controllers/branding_groups_controller.rb
```diff
@@ -66,17 +77,31 @@ def update
...
+      )
+      clean_params[:study_ids] = studies.map(&:id)
+      missing_studies = self.class.get_missing_studies(
+        self.class.param_to_array(params[:study_accessions]), studies
+      )
+
       if @branding_group.update(clean_params)
-        format.html { redirect_to merge_default_redirect_params(branding_group_path(@branding_group), scpbr: params[:scpbr]),
-                                  notice: "Collection '#{@branding_group.name}' was successfully updated." }
+        notice = "Collection '#{@branding_group.name}' was successfully updated."
```
eweitz, line 97:
```suggestion
        notice = "Successfully updated collection \"#{@branding_group.name}\""
```

eweitz (approved):
Code looks generally good!  I suggest some non-blocking maintainability refinements.

app/controllers/branding_groups_controller.rb
bistline: I find that surprising - I would have thought at least Python would have that.  I use `-` and `&` a lot when doing array "math" (difference and intersection, respectively).

ehanna4 (approved):
Works as advertised

PR #1856, opened 2023-08-14
Restoring source maps for Sentry (SCP-5265)

bistline:
#### BACKGROUND & CHANGES
This update restores Javascript sourcemaps for Sentry integration.  This will allow viewing un-minified source code when triaging events in Sentry, letting developers see the exact line/file that emitted the error.  This is being done through the `@sentry/vite-plugin` and an authentication token configured in vault.

#### MANUAL TESTING
1. Pull branch and initialize your environment with `./rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Start vite with `bin/vite dev`
3. In a separate terminal, boot the server as normal and load the home page
4. In Sentry, go to the `Releases` page for `single-cell-portal`: https://broad-institute.sentry.io/releases/?project=1424198
5. If you click into the latest release (which should be [82c2700e95ea](https://broad-institute.sentry.io/releases/82c2700e95ea5e41baac54062b3f3bb8d3db44bc/?project=1424198), but there may be a newer one), you should see that there is a new session adopted on the right like so:
Screenshot 2023-08-14 at 3 17 17 PM

6. In the bottom righthand corner, you should see that there are 10 artifacts listed for `Source Maps`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@reach/router": "^1.3.3",
     "@sentry/react": "^7.54.0",
     "@sentry/tracing": "^7.54.0",
+    "@sentry/vite-plugin": "^2.6.2",
     "@single-cell-portal/igv": "2.6.4-beta.1",
     "@tanstack/react-table": "^8.8.5",
     "@vitejs/plugin-react": "^1.2.0",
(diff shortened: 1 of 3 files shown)


ehanna4 (approved):
Code looks good 👍

eweitz (approved):
Thanks for this!  In addition to restoring actionable stack traces as you note, I'm also looking forward to A) error groups and B) incidence data that will persist beyond the most recent 1-3-ish releases.

PR #1857, opened 2023-08-14
Add Slack notifications for orchestration smoke test (SCP-5232)

bistline:
#### BACKGROUND
Currently, our Terra orchestration API smoke test job runs on a fixed schedule: M/W/F at 8AM.  Scheduled Github actions only automatically notify the last user to edit the job.  If that person is away from work, potential failures will go unnoticed until their return.  Furthermore, Github only sends notifications on failures.  If the job stops running for whatever reason, then this also could go unnoticed until someone checks manually.

#### CHANGES
This update adds Slack notifications to the `#scp-alerts` channel at the conclusion of every run for the smoke test.  This way, all subscribed team members will be able to see the status of jobs without having to go to the actions overview page in this repository.  Additionally, if a job fails, it adds the `@channel` notifier to ping all members.  Successful jobs will still post, but no alert is sent.  This gives a quick and easy way to confirm that jobs are running.  All messages will contain a link to the corresponding run for further debugging (like downloading log files, for instance).

Note: some of the previous messages in this channel have incorrect formatting - there's no way to delete messages from app integrations, so please disregard these.

#### MANUAL TESTING
1. Visit the `#scp-alerts` channel to see previous test messages, including one from 2023-08-15 1:36PM that failed and set the `@channel` notification
2. Click any of the links to corresponding jobs and confirm you are take to the correct place

### DIFF
--- a/.github/workflows/run-orch-smoketest.yml
+++ b/.github/workflows/run-orch-smoketest.yml
@@ -29,11 +29,12 @@ jobs:
           ORCH_SMOKE_TEST: true
           CI: true
         run: |
-          export VAULT_TOKEN=$( vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID )
-          bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json \
-                                  -s secret/kdux/scp/staging/scp_service_account.json \
-                                  -e test -v $DOCKER_IMAGE_TAG -n single-cell-portal-test \
-                                  -c bin/run_orch_smoke_test.sh
+          exit 1 # manual testing for faster iteration
+#          export VAULT_TOKEN=$( vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID )
+#          bin/load_env_secrets.sh -p secret/kdux/scp/staging/scp_config.json \
+#                                  -s secret/kdux/scp/staging/scp_service_account.json \
+#                                  -e test -v $DOCKER_IMAGE_TAG -n single-cell-portal-test \
+#                                  -c bin/run_orch_smoke_test.sh
       - name: Get current date
         id: date
         run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
@@ -41,9 +42,9 @@ jobs:
         id: slack-message
         run: |
           if ${{ steps.run-smoke-test.outcome == 'failure' }}; then
-            echo "::set-output name=message::\"@channel :x Orchestration smoke test ${{ steps.date.outputs.date }}: ${{ job.status }}\""
+            echo "::set-output name=message::@channel :x Orchestration smoke test ${{ steps.date.outputs.date }}: ${{ job.status }}"
           else
-            echo "::set-output name=message::\":green_check_mark Orchestration smoke test ${{ steps.date.outputs.date }}: ${{ job.status }}\""
+            echo "::set-output name=message:::green_check_mark Orchestration smoke test ${{ steps.date.outputs.date }}: ${{ job.status }}"
           fi
       - name: Notify Slack
         id: slack-notify
--- a/.github/workflows/run-orch-smoketest.yml
+++ b/.github/workflows/run-orch-smoketest.yml
@@ -13,11 +13,11 @@ jobs:
     runs-on: self-hosted
 
     steps:
-      - name: Check out repository code
-        uses: actions/checkout@v2
-      - name: Build image and setup env
-        id: build-image-setup-env
-        uses: ./.github/actions/docker-build-env-setup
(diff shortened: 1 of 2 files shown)


.github/workflows/run-orch-smoketest.yml
```diff
@@ -30,6 +31,46 @@ jobs:
                                   -s secret/kdux/scp/staging/scp_service_account.json \
                                   -e test -v $DOCKER_IMAGE_TAG -n single-cell-portal-test \
                                   -c bin/run_orch_smoke_test.sh
+      - name: Notify Slack failure
+        id: notify-slack-on-fail
+        uses: slackapi/slack-github-action@v1.24.0
+        if: failure()
+        with:
+          channel-id: "C05JHJ0UA1M" #scp-alerts
```
ehanna4, line 39: Thanks for the comment here clarifying the name of the channel lol

ehanna4 (approved):
Looks good, thanks for the thoughtfull-ness on only pinging `@channel` for failures 🙃

.github/workflows/run-orch-smoketest.yml
```diff
@@ -30,6 +31,46 @@ jobs:
...
+              "blocks": [
+                { 
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": ":green_check: Orchestration smoke test: ${{ job.status }}\nhttps://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+                  }
+                }
+              ]
+            }
```
eweitz, line 71:
```suggestion
```

Non-blocking: notifying on success seems likely to cause [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue).  Only alerting on failure seems best.

As is, this alert is significantly noisier than other alerts.  The other alerts in [#scp-alerts](https://broadinstitute.slack.com/archives/C05JHJ0UA1M), sourced from Mixpanel, only trigger on _positive_ cases.  That is, the alert is designed to detect an anomaly, and when it's detected (i.e. is positive), _only then_ send an alert.  It is very simple: to quickly notice a problem, people merely need to be members of #scp-alerts and not have the channel muted.  This matches the design of other DSP Slack alerting channels, e.g. [#workplace-resilience](https://broadinstitute.slack.com/archives/CA1481PRN).

This PR's design makes alerting more complex.  To ensure high signal-to-noise and thus avoid alarm fatigue, it requires each member of #scp-alerts to mute the channel.  It also requires all other alerts to include the `@channel` mention to ensure they get noticed by people.  The benefit is that it might mitigate false negatives by surfacing a canary -- i.e., if the 3-times-weekly negative alert doesn't post, maybe we're likelier to notice that #scp-alerts no longer includes these posts than we are to notice the no-longer-running job in GitHub Actions.  The cost is likelier alarm fatigue, per above.

I think cost exceeds benefit in this design, but I don't feel too strongly about it.  If we keep this design, let's at least update all Mixpanel alerts for #scp-alerts to include `@channel`.

eweitz (approved):
Nice integration!  

I'm mildly concerned about alert fatigue as described below, but I don't consider that blocking.  I added "Alert design, observability, and fatigue" as an [engineering strategy topic](https://docs.google.com/document/d/11eNlwMjwKoUcx4TQflzxw4DuN9q03fNNLphUuAVNIc4/edit#bookmark=id.wmcdb32csd9n).

.github/workflows/run-orch-smoketest.yml
```diff
@@ -30,6 +31,46 @@ jobs:
...
+        with:
+          channel-id: "C05JHJ0UA1M" #scp-alerts
+          payload: |
+            {
+              "blocks": [
+                { 
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "@channel :x: Orchestration smoke test: ${{ job.status }}\nhttps://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
eweitz, line 47: If we go with [failure-only alerts](https://github.com/broadinstitute/single_cell_portal_core/pull/1857/files#r1298605498):

```suggestion
                    "text": ":x: Orchestration smoke test: ${{ job.status }}\nhttps://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```

.github/workflows/run-orch-smoketest.yml
bistline: That's a great point, and one I thought about for a while.  I had it notify in all cases (and alert on failure) because I wanted a way to ensure that the job was running.  That being said, we can get the same information from the Actions tab in our repository - and GH Actions only notify on failure, or on success after repeated failures.  This PR as coded would change that behavior.  I think your concern is totally valid - I'll go ahead and remove this block.  Thanks for thinking this through!

PR #1858, opened 2023-08-16
Truncate list of genes for consensus scatter plots (SCP-5252)

bistline:
#### BACKGROUND & CHANGES
This update fixes a UI bug where long lists of queried genes in consensus scatter plots (i.e. collapsing gene expression by mean/median) will push the scatter plots down significantly on the page.  Now, only the first 3 genes queried will display in the plot header, and a final badge will show `and {n} more` with a popover containing the remaining gene names.  Additionally, this new styling has been applied to violin plot titles as well, which currently display the entire gene list underneath the plot.  While this doesn't have quite the same usability issue, it is harder to read, and this way the styling is unified.

An example of a 50-gene query on production:
Screenshot 2023-08-16 at 11 03 02 AM

Updated UI with truncated list:
Screenshot 2023-08-17 at 4 45 12 PM

Existing violin plots titles:
Screenshot 2023-08-17 at 4 55 55 PM

Updated violin plot titles:
Screenshot 2023-08-17 at 4 55 02 PM

#### MANUAL TESTING
1. Boot as normal and load any study with gene expression, such as the `Human milk - differential expression`
2. Query for 4 or more genes, then select "Mean" from the "Collapse by" menu
3. Confirm you get the popover badge containing the gene names not displayed
4. Click to the Distribution tab and confirm the same plot title
5. Go back to the default view, query for less than 4 genes and confirm that the popover is not shown

### DIFF
--- a/app/javascript/components/visualization/PlotTitle.jsx
+++ b/app/javascript/components/visualization/PlotTitle.jsx
@@ -18,7 +18,7 @@ export function formatGeneList(genes) {
     return formattedGenes
   }
   const hiddenGenes = <Popover id="genes-tooltip" className="tooltip-wide">{makeGeneBadges(hidden)}</Popover>
-  const hiddenOverlay = <OverlayTrigger trigger={['hover', 'focus']} key='hidden-genes' rootClose placement="right"
+  const hiddenOverlay = <OverlayTrigger trigger={['hover','focus']} key='hidden-genes' rootClose placement="right"
                                         overlay={hiddenGenes}>
     <span className='badge'>and {hidden.length} more</span>
   </OverlayTrigger>
--- a/app/javascript/components/visualization/PlotTitle.jsx
+++ b/app/javascript/components/visualization/PlotTitle.jsx
@@ -3,20 +3,22 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faInfoCircle } from '@fortawesome/free-solid-svg-icons'
 import { Popover, OverlayTrigger } from 'react-bootstrap'
 
+function makeGeneBadges(genes) {
+  return genes.map(gene => {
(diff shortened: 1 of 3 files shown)


app/javascript/components/visualization/PlotTitle.jsx
```diff
@@ -1,6 +1,30 @@
...
 import { faInfoCircle } from '@fortawesome/free-solid-svg-icons'
+import { Popover, OverlayTrigger } from 'react-bootstrap'
+
+function makeGeneBadges(genes) {
+  return genes.map(gene => {
+    return <span className="badge popover-badge" key={gene}>{gene}</span>
+  })
+}
+
+// divide gene list into two parts, only showing first 3
```
eweitz, line 12: For hover tooltip in references, and less linter noise:

```suggestion
/** Divide gene list into two parts, only showing first 3 */
```

eweitz (approved):
A neat, useful refinement!  I suggest one non-blocking maintainability refinement.

ehanna4 (approved):
Functional review - looks good and much nicer having the genes less in the way

Out of scope for this PR but wanted to mention is that the distribution tab doesn't seem to update correctly if you add more genes after it's drawn. This is also true on Prod so out of scope here but found while testing this. I can make a ticket if that seems the right move?

To reproduce search for genes, choose collapse by mean then click over to the Distribution tab and then add another gene in the search. The distribution tab doesn't change, but if you click over to the Scatter tab you can see the gene search did work and update there

Screenshot 2023-08-22 at 4 56 46 PM
Screenshot 2023-08-22 at 4 57 44 PM

ehanna4:
Update - I made a ticket for the out-of-scope bug I found. Going to move the Jira ticket for this to Done as Erics review request looks minimal.

PR #1860, opened 2023-08-21
Enlarge plots and space for content (SCP-5186, SCP-5378)

eweitz:
This enlarges visualizations and refines vertical placement for other content.  It gives more space to analyze and explore.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/410ff08e-7189-455b-b9c9-ff6afe9e86fc

### Overview
Previously, the study title appeared below the navigation bar in Study Overview.  The navbar was sparsely used.  This ensured study titles were always prominently visible, but it also required shrinking plots and pushing content outside the area visible by default.  It also threw off certain vertical placements when study titles were especially long, as the title would wrap over two lines and push down all content -- without that content resizing to stay in view by default.  Other UI elements like scatter plot legends, scatter plot x-axis titles, violin plot titles, and the "Differential expression" button also had vertical positioning oddities that led to layout misalignment, lost space for content, or both.

Now:
* Study title is in the navbar, which makes scatter plots taller by 60px
* Scatter plot legends have 90px more vertical real estate, enough to display another 2-3 legend entries without scrolling
* Violin plots also start 90px higher: more space for plots and labels
* DE table shows 10 rows, up from 8

### Long study titles
**Problem**: Moving the title into the navbar also makes the main UI more robust to long titles.  Various vertical positioning logic had specific assumptions about title _height_, and a title that wrapped over _two_ lines broke those assumptions.  Certain controls and plot labels would shift down by 30-40 pixels.  This subtly broke their layout by pushing them off the screen.  Users could often recover these by scrolling, but that's a UX bug in a layout that otherwise always puts that content above the fold.

**Solution**: Study titles now always appear on one line.  Title font is also smaller, consistent in size with many popular products.  To give more space for titles, copy has and icons have been trimmed from controls in the right end of this global header, considerably narrowing it (but not overcrowding).  

[Our P25 viewport width is ~1400px](https://mixpanel.com/project/2120588/view/19059/app/boards#id=5442864), and easily 90% of study titles display in the resulting available navbar width in those narrower-than-normal viewports.  For even thinner viewports or exceptionally long titles, the text is truncated at right.  An ellipsis indicates more is available, and hovering over the text shows the full title in a tooltip.  When presented with the options to A) wrap across lines or B) show title on one line with some cases truncated, product folks chose the latter.

Here's how the responsive layout looks, with a focus on long study title handling.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/92d9b61d-e501-4d87-83f4-6f0719d55ac8

### Related fixes
Violin plot titles and the DE button vertical placement were both incidentally refined as part of this work.  

#### Violin plot titles
Violin plot titles used a long-SCP-outdated, layout-inefficient style.  The violin plot title also oddly noted the _clustering_ -- which is mainly relevant for scatter plots.  The violin plots are much more about expression in a gene across groups _in an annotation_.  The violin plot now refers to the annotation, which is more meaningful.  The plot title display styling is also more consistent with scatter plots.  Similarly, violin plots also handle consensus statistics / "Collapse by" is a display styling like scatter plots.

#### "Differential expression" button
The "Differential expression" button also had a vertical positioning bug.  Previously, the button oddly moved up ~15px upon clicking into the "Distribution" tab.  (This was due to a complication with the "+" / "Create annotation" button.)  Now the DE button stays in the same spot.  Subtly inconsistent padding around the DE button in its default state is also fixed.

### Further details
Richer context is available in the 2023-08-22 [SCP demo recording](https://drive.google.com/file/d/1whqhirP8ciQx4tkKnQ5No8CB9RzyBGbY/view?t=27s).  This work was done as part of Developer's Choice Days.  It relates to SCP-5186 and SCP-5378.


eweitz:
I've converted this back to a draft to account for the refactoring entailed by #1867, and a bug found by Jon.

jlchang (changes requested):
Blocker (still reviewing, but thought I'd communicate this blocker early)

I don't think we have a synthetic study that has more than one "experiment" that would have surfaced the need for clustering information in violin plot titles:
> The violin plot title also oddly noted the clustering -- which is mainly relevant for scatter plots. The violin plots are much more about expression in a gene across groups in an annotation. 
The violin plot reflects the expression of a gene across groups in a annotation **relative to the selected clustering**. Therefore both clustering and annotation are relevant to the violin plot. An example would be SCP44 in production (We use this one for our manual smoke tests):
The default clustering "Atlas of epithelial cells" is derived from normal mice so the [violin plot for `Reg3b`](https://singlecell.broadinstitute.org/single_cell/study/SCP44/small-intestinal-epithelium?genes=Reg3b&cluster=Atlas%20of%20epithelial%20cells%20%28Figure%201%29&spatialGroups=--&annotation=Mouse--group--study&subsample=all&tab=distribution#study-visualize) doesn't show striking differences between the mice in the clustering. In contrast, a separate experiment in the study, visualized with the clustering "Infection with enteric pathogens", shows that mice infected with salmonella (and one H. polygyrus infected mouse) have increased expression of `Reg3b`, an anti-microbial gene. The clusterings have different mice - normal mice for the altas and infected + non-infected mice for the second experiment.

Two clusterings may share an annotation but the violin plot from that shared annotation will be different if the clusterings are derived from different cells. In studies set up with multiple experiments, the selected clustering provides important context for the annotation being visualized and should be included in the violin plot title.

eweitz:
> Two clusterings may share an annotation but the violin plot from that shared annotation will be different if the clusterings are derived from different cells. In studies set up with multiple experiments, the selected clustering provides important context for the annotation being visualized and should be included in the violin plot title.

Insightful catch!  I've updated as you suggest -- here's how it looks.

Violin_plot_title_refined__SCP_2023-08-31

app/javascript/lib/layout-utils.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  const jqStudyHeader = $('.study-header')
+  if (jqStudyHeader.tooltip) {jqStudyHeader.tooltip('hide')}
+}
+
+/**
+ * Set global header end width, and mitigate long study titles on narrow screens
+ */
+export function adjustGlobalHeader() {
```
bistline, line 137: FWIW I checked on production, and the longest study title we have is `Intra-tumoral injection of caerin 1.1 and 1.9 peptides in vaccinated TC-1 tumour bearing mice with PD-1 blockade modulates macrophage heterogeneity and the activation of CD8+ T cells in the tumour microenvironment`.  I renamed one of my local studies to that and this dealt with that just fine.

bistline (approved):
Much nicer overall!  The usability gains are very nice, and we gain a lot of vertical real estate back.

bistline:
Test failures are due to Terra release happening at the same time...

jlchang (approved):
Functional review - works as expected. Looking forward to the additional real estate!

Two non-blocking suggestions -  

1. Would coloring the title text help to make the title easier to catch the eye? 

Looking at the title in the nav bar, because so much of our UI in Explore is white on dark blue AND because the Title text is pretty small (to accommodate long titles), it feels a little "lost in the crowd". I'm not suggesting a color (I have below-average color sense) but I'm including a screenshot to show how color might make the title easier to pick out for someone unfamiliar with our layout:

Screenshot 2023-10-24 at 8 43 32 AM


2. Should we make sure that the info icon for DE doesn't wrap to the next line?

Screenshot 2023-10-24 at 8 53 24 AM

eweitz:
The most recent build failure seems like a false positive.  The exit code indicates an out-of-memory (OOM) error.

```
.search_controller_test.rb:179 test_should_return_search_results_using_numeric_facets starting
search_controller_test.rb:179 test_should_return_search_results_using_numeric_facets completed (9.341 sec)
.search_controller_test.rb:296 test_should_run_inferred_search_using_facets starting
Error: Process completed with exit code 137.
```

We have [1 prior recorded case of that](https://github.com/broadinstitute/single_cell_portal_core/pulls?q=is%3Apr+%22Error%3A+Process+completed+with+exit+code+137%22+is%3Aclosed+).  I'll re-run the build.

eweitz:
Thanks for the useful reviews.  I ended up increasing study header font size slightly, and fixing a few other issues.  I deferred refining the info icons as that's a pre-existing issue; these changes don't change its incidence.

PR #1861, opened 2023-08-22
Execute fewer searches on quick gene search updates (SCP-5264)

ehanna4:
Previously the gene search in the explore tab was updated to make the search more user friendly/less cumbersome removing the need to "double submit" (aka hitting enter twice) in order to execute a search. However, we encountered some trouble with this when users did lots of gene searches in quick succession resulting in a whole bunch of data fetch calls being made that weren't even needed.

This PR improves upon this by reducing the number of executed data fetches on gene search updates. 

1. A debounce is added to the gene search bar so that there is a slight delay before attempting to execute on the search that is canceled if a user changes their search quickly. This reduces the data fetches and unnecessary re-renders without adding noticeable lag (delay here is 300ms)
2. A timeout is added around the `navigate` call in `updateExploreParams` for gene search updates. This adds a 1 second delay to actually executing the `navigate` and subsequent data fetches in case changes have made it through the search debounce but more changes come through, to again reduce the calls to fetch.

The following videos show the way the changes impact the calls and UX

Before:
See how quick adds to gene search shoots off fetches for all the updates:

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/7c1e4d54-6684-4357-9724-744eb3ab3377

There was also a laggy UX for Dotplots and Heatmaps where the render for the first gene search updates would go and then silently change when the next fetch completed without a spinner, and could be jarring.

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/eb453d7a-abea-467c-9776-5438c2702080

After:
The gene updates don't each trigger a fetch, only after a delay and so only a few fetches are done, also the spinner remains until the last fetch so the Dotplot doesn't have that strange laggy update behavior
https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/99927f09-db08-4790-9d65-6bbeaa447f6b


eweitz (commented):
This is a tough client-side performance engineering problem!  [Here are details](https://docs.google.com/document/d/113bQoyx7YqamZqRDZJpBREP-NXaTlDufsNe_3l0Dtho/edit) for our chat this afternoon.

ehanna4:
Chatting with Eric the benefit of the fewer fetches is not being outweighed by the slowdown from the timeouts and with the work done by Jon of merging the data calls for Heatmap and Dotplot we are seeing a high improvement already in this space. Going to close this PR for now and potentially revisit at a later time to prioritize higher priority work at this time.

PR #1867, opened 2023-08-28
Create panel manager for explore page (SCP-5215)

ehanna4:
Part one for the Facet Filter component - just a refactor of existing code no new functionality and very little new code. I wanted to get this out so any other work being done in these files can work within this refactor and also because this is already a large PR so wanted to split the actual new feature into its' own PR.

This PR restructures the code a little bit to make the addition of the Filter Facet component easier.

The Explore tab consists of (among other things) the Plot (left side) the Panel (right side) and tabs (Scatter, Dot Plot, etc.) The code in `ExploreDisplayTabs` had managed which tabs, plots and panels to show and was almost 900 lines long. Now that we want to add a third right-side Panel option (Facet Filter) it would make the logic more complicated and make this file even longer. So instead I have separated out a bunch of the logic for managing the right-side Panel into a file called `ExploreDisplayPanelManager`. This will then figure out which of the soon to be three panels (Options/Regular, DE, Facet Filter) to display. 

To test this PR:
- There is no new functionality and all the looks should be the same - so pull down this branch and start up a local server.
- Navigate to a study (one with DE will have more cases to test)
- Then compare to Staging or Prod, clicking around, clicking the DE button, changing sizes etc. 
- There should not be a difference between local and Staging or Prod.

Upcoming work - the new Facet Filter panel will be added and the `ExploreDisplayPanelManager` will sort out when to show this.

Notes - I haven't added new tests yet, will plan to incorporate in future PR.

Showing that staging and local are the same looking:
https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/e4e90e92-db8f-49c2-9106-9b13c492b618

Screenshot 2023-08-28 at 12 26 42 PM

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
+++ b/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
@@ -1,0 +1,473 @@
+import React, { useState, useEffect } from 'react'
+import _clone from 'lodash/clone'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faLink, faEye, faTimes, faUndo } from '@fortawesome/free-solid-svg-icons'
+
+import ClusterSelector from '~/components/visualization/controls/ClusterSelector'
+import AnnotationSelector from '~/components/visualization/controls/AnnotationSelector'
+import SubsampleSelector from '~/components/visualization/controls/SubsampleSelector'
+import { ExploreConsensusSelector } from '~/components/visualization/controls/ConsensusSelector'
+import SpatialSelector from '~/components/visualization/controls/SpatialSelector'
+import CreateAnnotation from '~/components/visualization/controls/CreateAnnotation'
+import PlotDisplayControls from '~/components/visualization/PlotDisplayControls'
+import GeneListSelector from '~/components/visualization/controls/GeneListSelector'
+import InferCNVIdeogramSelector from '~/components/visualization/controls/InferCNVIdeogramSelector'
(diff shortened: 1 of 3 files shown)


app/javascript/components/explore/ExploreDisplayPanelManager.jsx
```diff
@@ -0,0 +1,473 @@
...
+
+/**
+ * Manages the right panel section of the explore view of a study. We have three options for the right side
+ * panel with different controls to adjust the plots: Options (or default), Differential Expression, and
+ * Facet Filtering (currently in progress)
+ *
+ * This manager will return the correct content and header for the appropriate panel to display.
+ *
+ *  */
+export default function ExploreDisplayPanelManager({
```
bistline, line 167: I totally get why we have to do this, but I get slightly worried at the amount of arguments for this function.  30 seems unwieldy, but having said that I don't see how we get around that without more refactoring.

bistline (approved):

app/javascript/components/explore/ExploreDisplayPanelManager.jsx
ehanna4: Yeah 100% agree. I hope to reduce this in my next PR as I keep working out the stuff actually necessary to be shared between the components.

eweitz (approved):
Good call refactoring here!  

Code looks broadly good; I agree with the discussion on parameters.  My local manual tests work as expected.

PR #1868, opened 2023-08-29
Updating to Terra orchestration V2 billing endpoints (SCP-5284)

bistline:
#### BACKGROUND && CHANGES
This update changes the `FireCloudClient` class to point at the updated [V2 billing API endpoints](https://api.firecloud.org/#/BillingV2) in the Terra orchestration API.  This is in response to the planned retirement of said endpoints.  No user-facing functionality will change, only the internal implementation of using the new endpoints.  This update also filters out any billing projects for which the user has no defined role (meaning they cannot be used for study creation), and fixes an issue on error handling when workspace creation fails that prevented the page from redirecting and showing the error message.  Lastly, this update removes the unused `create_billing_project` and `get_billing_accounts` methods from `FireCloudClient`, as we no longer allow either action since we removed the billing OAuth scope.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Select "My billing projects" from the profile nav menu
3. Confirm the page loads and that you see billing projects, and that those you own also list the members
4. Click the "Create a study" link at the top of the page
5. Confirm you see the same billing projects listed in the select menu for `Terra billing project`.

### DIFF
--- a/app/controllers/billing_projects_controller.rb
+++ b/app/controllers/billing_projects_controller.rb
@@ -38,9 +38,9 @@ class BillingProjectsController < ApplicationController
     billing_projects.each do |project|
       project_name = project['projectName']
       @projects[project_name] = {
-          status: project['creationStatus'],
-          role: project['role'],
-          members: project['role'] == 'Owner' ? @fire_cloud_client.get_billing_project_members(project_name) : nil
+        status: project['status'],
+        roles: project['roles'],
+        members: project['roles'].include?('Owner') ? @fire_cloud_client.get_billing_project_members(project_name) : nil
       }
     end
   end
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -1070,7 +1070,7 @@ class StudiesController < ApplicationController
       if current_user.registered_for_firecloud && client.registered?
         available_projects = client.get_billing_projects
         available_projects.each do |project|
-          if project['status'] == 'Ready'
+          if project['status'] == 'Ready' && project['roles'].any?
             @projects << [project['projectName'], project['projectName']]
           end
         end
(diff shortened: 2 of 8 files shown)


ehanna4 (approved):
Functional review - can see the billing projects page and can see a list of options for billing project when creating a study

eweitz (approved):
Code looks good!  Further context is [here](https://broadinstitute.slack.com/archives/C9YSEMSFR/p1692896882771359), [here](https://broadinstitute.slack.com/archives/CBEHTH601/p1692897261528529), and [here](https://broadinstitute.slack.com/archives/CBEHTH601/p1692976718009489).

jlchang (approved):
I'm glad the billing endpoint switchover was the simple change we hoped it was!

One question: The dropdown under "Terra Billing Project" lists all the options viewable under "My billing projects".

I seem to remember that there is an issue with study creation if the study creator is only a user on the selected Terra Billing project (ie. not an owner)? **Should we only be listing the Terra billing projects where the study creator is an owner**? (When I go to "Create a study", I see billing projects like "shalek-lab-firecloud", "help-gatk", "broad-firecloud-dsde-methods" etc that I shouldn't try to create SCP studies under...

bistline:
> One question: **Should we only be listing the Terra billing projects where the study creator is an owner?**

Oh yeah... there's some funkiness with the permission model where there are some billing projects you can "see", but don't have permission to use to create workspaces. They changed something along the years and not all the projects that show up can be used.  Also, some have no active billing, but we can't know that until we try to use it.  But if you're a billing project `User` then you have permission to create workspaces, just not see other workspaces in that project.

There's apparently another issue where if the workspace creation fails, it's not handling the redirect correctly.  I'm going to fix that as well.  Thanks for spotting this!

PR #1870, opened 2023-08-30
Upgrade to N2D instances for ingest, allow AnnData extraction to scale VM dynamically (SCP-5291)

bistline:
#### BACKGROUND
When a user uploads an AnnData file, regardless of whether or not this is a "reference" file, SCP will still launch an ingest process to validate that the file can be opened, and that data can be read.  Since all ingest VMs are `n1-highmem-4` instances (4 cores, 26 GB RAM), this process will fail with a `137` exit code (OOM exception) if the files is too large as AnnData files are not streamed line-by-line by our ingest code like plain text files.  Additionally, while these instances are cost-effective and reliable, there are newer generations of virtual machines available in GCE.

#### CHANGES
This update switches all ingest VMs to use the [`N2D`](https://cloud.google.com/compute/docs/general-purpose-machines#n2d_machines) family, which is the latest generation of general-purpose VMs in GCE.  These machines are based on a newer AMD platform and offer better performance and cost savings over previous N-generation machines.  Also, like the `n2` family, they offer 8 GB of RAM per CPU as opposed to 6.5 in `n1` instances.

Additionally, AnnData ingest processes now dynamically scale the machine type based off of the size of the file, up to 256 GB (`n2d-highmem-32`).  This is only true for the initial extraction phase - all downstream jobs will still use `n2d-highmem-4` instances.  The reason for the capping on extraction size is purely for cost reasons as we pay per CPU + GB RAM hourly for VMs.  For reference, this is the cost per hour for all `n2d` instances, not counting persistent disks:
```
n2d-highmem-2: $0.11                                                                                  
n2d-highmem-4: $0.23                                                                                                                                                         
n2d-highmem-8: $0.46                                                                           
n2d-highmem-16: $0.91                                                                            
n2d-highmem-32: $1.82                                                                           
n2d-highmem-48: $2.73                                                                           
n2d-highmem-64: $3.64                                                                         
n2d-highmem-80: $4.55                                                                 
n2d-highmem-96: $5.47
```

It is worth noting that this places an effective limit on the size of AnnData files at 256 GB, but given that the largest in the portal as of right now is only ~12.5 GB, this seems like a reasonable cap. The file that failed to ingest that prompted this work was 27 GB, which would open on the standard `n2d-highmem-4` instance.  Lastly, this fixes a bug in `IngestJob` where the `gene_count` was not set correctly following processed expression extraction.

#### MANUAL TESTING
1. Start all services and log in
2. Create a new study, or go to an existing study and upload a new parseable file
3. In `development.log`, note that `machine_type` is changed:
```
...
:resources:
  :regions:
  - us-central1
  :virtual_machine:
    :boot_disk_size_gb: 300
    :labels:
      :study_accession: scp101
      :user_id: 63e27b72f39faa00560c1e25
      :filename: cluster_example_txt
      :action: ingest_pipeline
      :docker_image: scp-ingest-pipeline
      :docker_tag: '1_27_3'
      :environment: development
      :file_type: cluster
      :machine_type: n2d-highmem-4
      :boot_disk_size_gb: '300'
    :machine_type: n2d-highmem-4
...
```
4. To test VM scaling, enter a Rails console session and confirm you see the same output from the following commands:
```
params = AnnDataIngestParameters.new
=>
#<AnnDataIngestParameters:0x0000000109658920
...

params.machine_type
=> "n2d-highmem-4"

params.file_size = 50.gigabytes
=> 53687091200

params.machine_type
=> "n2d-highmem-8"

params.file_size = 100.gigabytes
=> 107374182400

params.machine_type
=> "n2d-highmem-16"
```

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -107,11 +107,13 @@ class FileParseService
           # obsm_key is only set for parsing a new singular clustering
           if obsm_key.present?
             params_object = AnnDataIngestParameters.new(
-              anndata_file: study_file.gs_url, extract: %w[cluster], obsm_keys: [obsm_key]
+              anndata_file: study_file.gs_url, extract: %w[cluster], obsm_keys: [obsm_key],
+              file_size: study_file.upload_file_size
             )
           else
             params_object = AnnDataIngestParameters.new(
-              anndata_file: study_file.gs_url, obsm_keys: study_file.ann_data_file_info.obsm_key_names
+              anndata_file: study_file.gs_url, obsm_keys: study_file.ann_data_file_info.obsm_key_names,
+              file_size: study_file.upload_file_size
             )
           end
           # TODO extract and parse Raw Exp Data (SCP-4710)
@@ -119,7 +121,7 @@ class FileParseService
           # setting attributes to false/nil will omit them from the command line later
           # values are interchangeable but are more readable depending on parameter type
           params_object = AnnDataIngestParameters.new(
-            anndata_file: study_file.gs_url, extract: nil, obsm_keys: nil
+            anndata_file: study_file.gs_url, extract: nil, obsm_keys: nil, file_size: study_file.upload_file_size
           )
         end
         job = IngestJob.new(
--- a/app/models/ann_data_ingest_parameters.rb
+++ b/app/models/ann_data_ingest_parameters.rb
@@ -40,12 +40,30 @@ class AnnDataIngestParameters
     matrix_file_type: nil,
     gene_file: nil,
     barcode_file: nil,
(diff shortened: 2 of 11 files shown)


eweitz (approved):
Code looks good!  More context is [here](https://broadinstitute.slack.com/archives/CBEHTH601/p1693330132774289) and [here](https://broadinstitute.slack.com/archives/CBEHTH601/p1693488370568509).

jlchang (approved):
Functional review - confirmed that local AnnData --extract ingest requests `n2d`series machine_type and machine_type logic (per rails console testing) scales with file size info. Looking forward to seeing this change in play!

PR #1878, opened 2023-09-09
Cell Facet Filter Component (SCP-5215)

ehanna4:
Initial work for the cell facet filtering UI work. 

This builds on the library work for cell facet filtering and adds controls to the UI. Starting with 3 dropdown menus listing the 5 prioritized annotations/facets returned by the library built [here](https://github.com/broadinstitute/single_cell_portal_core/pull/1862). Pre-populated but allows for changing facet choice. Allowing additional, or removal entirely of facets is work to be done in [TICKET](https://broadworkbench.atlassian.net/browse/SCP-5297?atlOrigin=eyJpIjoiNzY4NGRlOWIyNzU2NDQzNWI1NmZmYWJhMTE0OGUxYTciLCJwIjoiaiJ9). This is gated by a feature flag so won't be GA until the team is satisfied will not be visible to the users.

Todos that still need to be done, I can address in this PR or as subsequent work:
- Loading spinners are not aligned properly
- Needs tests
- Some general UI touch ups like colors, border edge rounding, spacing, etc. 
-
I wanted to get this out for review ASAP but recognize these ^ things still need to be addressed.

To test:
- pull this branch
- run the migration to ensure the new feature flag is available
- start up a local portal and log in
- flip the feature flag to true for yourself from the feature flags dashboard
- navigate to a study
- you should see the button for facet filtering, similar to the button for DE in the options panel
- clicking the button will show the facets and their filters, clicking the check boxes next to the filters will add or remove points from the plot as appropriate
- you can change the facets using the drop downs, you can also change the "color by" annotation using that dropdown
- clicking the back arrow to return to the default options panel should reset the plot to showing all the points

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/368cbcc9-d153-47e7-bd4d-a58ffd5ad718

https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/28b1ff43-8a80-455d-b54d-633693b035f9


db/migrate/20230907050143_add_flag_for_cell_facet_filtering.rb
```diff
@@ -0,0 +1,11 @@
+class AddFlagForUserDeUpload < Mongoid::Migration
```
bistline, line 1: Blocker: I can't test the functionality in this PR because this migration won't run.  The name of the file and the class of the migration have to be the same.  E.g. `20230907050143_add_flag_for_cell_facet_filtering.rb` would need to have a class name of `AddFlagForCellFacetFiltering`.  Since the class name is actually `AddFlagForUserDeUpload` it barfs with:
```
NameError: uninitialized constant AddFlagForCellFacetFiltering
```

ehanna4: oh geeze sorry - fixed [70de49a](https://github.com/broadinstitute/single_cell_portal_core/pull/1878/commits/70de49a11166461740cf9e9fab050ed6bb81d690)

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -203,8 +165,11 @@ export default function ExploreDisplayTabs({
     }
   }
 
+  console.log('in xplordisplaybtbs:', cellFaceting)
```
bistline, line 168: debug statement

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -203,8 +165,11 @@ export default function ExploreDisplayTabs({
...
+
   /** Update filtered cells to only those that match annotation group value filter selections */
   function updateFilteredCells(selection) {
+    console.log('here:', selection)
```
bistline, line 172: more debug

app/javascript/components/explore/FacetFilteringModal.jsx
```diff
@@ -0,0 +1,52 @@
...
+import Modal from 'react-bootstrap/lib/Modal'
+import { closeModal } from '~/components/search/controls/SearchPanel'
+
+/** Show modal box with information on differential expression in SCP */
+export default function FacetFilteringModal() {
+  const [showCellFacetModal, setShowCellFacetModall] = useState(false)
+  const cellFacetModalContent = (
+    <div>
+      <p> Cell filtering provides the ability to view cell annotations within the context of other annotations.
+        You can visualize the expression of particular cells in a subset of the dataset through the facet filters.
```
bistline, line 13: I hadn't fully realized that this will extend to expression-based visualizations, but it does totally work!  Very cool.

bistline (approved):
Overall, really great initial implementation!  Very exciting to see this starting to work.  And doubly-exciting that it works for single gene expression visualizations.  I note two UI papercuts that are CSS related.  

First, when you switch between annotations to color by, the Options panel blows out to nearly half the page:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/d59342db-8b13-4b7a-a274-c1b28885773f

Secondly, if you click to hide the options panel, you can still see part of it in the background:

bg_bleed

Also, some likely out-of-scope questions below:

1) I noticed that as you select filter values to include/exclude, we're making an API request for plot data.  However, I don't see any change in the data - it's just asking for the same cluster again w/o any new params.  This might be because I have caching turned off locally, but I suspect that some trigger is firing to ask for data again.  Is there a way to disable this?   My understanding was that we'd be doing all the filtering on the front-end without having to request new data, unless the user adds new facets to the list.

2) It's very cool to see that this works on the single-gene expression scatter plot.  However, when I add a second gene to go to dotplot/heatmaps, the functionality stops working (which I would assume is expected).  Should we add disabling this feature to our "remaining work" ticket so that users aren't confused?  And perhaps add a ticket to explore extending this work to include multi-gene expression views?

ehanna4:
Improved the CSS/UI papercuts noted by @bistline 
The options panel no longer gets giant when loading and actually hides fully when "X-ed" out. I think I'll need to chat with @eweitz about some of the CSS stuff for the DE panel as it's gotten a little bit funky with these changes.


https://github.com/broadinstitute/single_cell_portal_core/assets/54322292/192cb498-2aa1-4bbe-99e0-4b8d60972c3b

eweitz (approved):
It's fantastic to see a prototype UI on this!  

Some areas need refinement before general availability, as you note, but this seems directionally good.

I note a handful of non-blocking suggestions below.  

Beyond those granular things, the UX for changing facets caught my eye.  The dropdowns add a fair amount of visual complexity.  That said, they're also a neat way to enable reordering facets, and a clear affordance that it's possible.  Are there other examples of UIs that offer dropdowns for changing facets that each have a list of filters?

Among faceted search UIs, I see two broad patterns to view more facets.  The first pattern is to show all facets by default and require the user to scroll; see e.g. [Amazon](https://www.amazon.com/camera/s?k=camera), [CELLxGENE](https://cellxgene.cziscience.com/e/fa3e2a80-1c9a-4097-a02f-bf47e143fca7.cxg/), [Google Shopping](https://www.google.com/search?tbm=shop&hl=en&psb=1&q=camera&oq=camera&sclient=products-cc), [Terra](https://app.terra.bio/#library/showcase), [UniProt](https://www.uniprot.org/uniprotkb?query=app), and [Variation Viewer](https://www.ncbi.nlm.nih.gov/variation/view).  The second is to show only some facets by default, and have a button like "Additional filters" to enable the user to add or remove facets; see e.g. [PubMed](https://pubmed.ncbi.nlm.nih.gov/?term=pubmed+cancer), [ClinVar](https://www.ncbi.nlm.nih.gov/clinvar/?term=cancer), and [Genome Reference Consortium](https://www.ncbi.nlm.nih.gov/grc/human/issues).  

Almost all of those UIs sort facets non-alphabetically, and none support facet reordering.  These notes are food for thought; the novel dropdown UI for changing facets doesn't strike me as a blocker.

bistline:
> Improved the CSS/UI papercuts noted by @bistline The options panel no longer gets giant when loading and actually hides fully when "X-ed" out. I think I'll need to chat with @eweitz about some of the CSS stuff for the DE panel as it's gotten a little bit funky with these changes.
> 
>  better.css.mov

Looks great - thanks!

PR #1881, opened 2023-09-15
Check for cluster group before trying to show annotation (SCP-5307)

ehanna4:
There is a bug that only seems to affect admins with not being able to load the User Annotations page. The [corresponding error in Sentry](https://broad-institute.sentry.io/issues/4218900801/?environment=production&project=1424198&query=is%3Aunresolved+level%3Aerror+%21stack.abs_path%3A%2Aideogram%2A&referrer=issue-stream&sort=freq&statsPeriod=1h&stream_index=0) indicates there is a missing `cluster_group` that the code is trying to make calls on. This PR adds a check to make sure `cluster_group` is present before trying to build up the list of annotations to show.

This only affects Prod, - Staging and Local don't appear to have this issue. My assumption is there must be some old/corrupt study annotation only admins have access to that is missing a `cluster_group` which is why the page won't load. I can't test that this fixes the issue until it hits prod, but it doesn't change any behavior on local in my testing. This also doesn't change data, it only changes what is displayed in this extremely rarely used page in the UI so it seems a safe guess to try out to see if it alleviates the issue.

### DIFF
--- a/app/views/user_annotations/index.html.erb
+++ b/app/views/user_annotations/index.html.erb
@@ -17,27 +17,29 @@
           </thead>
           <tbody>
             <% @user_annotations.each do |user_annotation| %>
-              <tr>
-                <td class="annotation-name"><%= user_annotation.name %></td>
-                <td><%= user_annotation.values.map{|x| "<span class='label label-#{ x == 'Undefined' ? 'warning' : 'default'} #{user_annotation.name_as_id}'>#{x}</span>"}.join(' ').html_safe %></td>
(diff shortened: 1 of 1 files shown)


ehanna4:
formatting makes this change appear more substantial - the change is just line 20 `  <% if user_annotation.cluster_group.present? %>` and the corresponding `<% end %>` at line 42, otherwise it's just formatting since everything had to be indented one more

eweitz (approved):
Code looks good!  [Slack discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1694794532211369) has additional detail.

bistline (approved):

PR #1884, opened 2023-09-20
Re-plot filtered cells from cell faceting as mostly transparent (SCP-5277)

bistline:
#### BACKGROUND & CHANGES
This update changes the behavior of the cell faceting feature to re-plot filtered groups as mostly transparent, light grey points instead of completely removing them.  This retains the original shape of the plot while also clearly demonstrating which cells have been filtered.  In addition, these cells are added to a new trace called `--Filtered--`, which can also be hovered over in the legend to bring them to the front for closer inspection.

Also, this update refactors the `intersect` and `reassignFilteredCells` functions outside of the `ScatterPlot` component to allow them to be tested directly, while still preserving HMR in local development.

Demo of new plotting behavior:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/e13b2e13-e3a3-48ba-8e3d-57c4502c578c

Note: If you are performing cell faceting on anything other than the default annotation, selecting groups will cause a full reload of the scatter plot, including calling the API.  This has been tracked in SCP-5313.

#### MANUAL TESTING
1. Boot as normal and sign in, then ensure the `show_cell_facet_filtering` is enabled for your user account
3. Load the `Human milk - differential expression` study
4. If you haven't already done so, set the default annotation to `General_Celltype` (this makes testing a little easier)
5. Click the button to start cell faceting once it is enabled
6. Confirm you can add/remove groups and that they are moved to the `--Filtered--` trace which can then be hovered over

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -23,6 +23,7 @@ const { defaultScatterColor } = PlotOptions
 import LoadingSpinner from '~/lib/LoadingSpinner'
 import { formatFileForApi } from '~/components/upload/upload-utils'
 import { successNotification, failureNotification } from '~/lib/MessageModal'
+import { FILTERED_TRACE_NAME, FILTERED_TRACE_COLOR } from '~/lib/cluster-utils'
 
 window.Plotly = Plotly
 
@@ -325,9 +326,12 @@ function RawScatterPlot({
     const keys = Object.keys(dataArrays)
     keys.forEach(key => intersectedData[key] = [])
     window.dataArrays = dataArrays
+    // array of cell indexes that are being plotted
+    const plotted = []
     for (let i = 0; i < filteredCells.length; i++) {
       const filteredCell = filteredCells[i]
       const allCellsIndex = filteredCell.allCellsIndex
+      plotted.push(allCellsIndex)
       for (let j = 0; j < keys.length; j++) {
         const key = keys[j]
         const dataArray = dataArrays[key]
@@ -335,8 +339,47 @@ function RawScatterPlot({
         intersectedData[keys[j]].push(filteredElement)
       }
     }
-    scatter.data = intersectedData
-    return scatter
+    return [intersectedData, plotted]
+  }
+
+  /**
+   * Reassign plot data to change cells filtered by cell faceting to --Filtered-- trace (light grey, transparent)
+   * this keeps the shape of the original plot but clearly shows what cells are excluded
+   *
+   * @param plotted {Array} Indices of points that are to be plotted
+   * @param originalData {Object} original scatter data object
+   * @param filteredData {Object} filtered scatter data object from cell faceting
+   * @returns {Object} reorganized scatter data object with new --Filtered-- trace
+   */
+  function reassignFilteredCells(plotted, originalData, filteredData) {
+    const allIndexes = [...Array(originalData['x'].length).keys()]
+    const originalSet = new Set(allIndexes)
+    const plottedSet = new Set(plotted)
+    const reassignedIndices = [...originalSet].filter(i => !plottedSet.has(i))
+    const newPlotData = {}
+    const keys = Object.keys(originalData)
+    keys.forEach(key =>  {
+      newPlotData[key] = []
+    })
+    for (let idx = 0; idx < reassignedIndices.length; idx++) {
+      for (let k = 0; k < keys.length; k++) {
+        const key = keys[k]
+        if (key === 'annotations') {
+          newPlotData[key].push(FILTERED_TRACE_NAME)
+        } else if (key === 'expression') {
+          newPlotData[key].push(FILTERED_TRACE_COLOR)
+        } else {
+          const dataArray = originalData[key]
+          const replottedElement = dataArray[idx]
+          newPlotData[key].push(replottedElement)
+        }
+      }
+    }
+    for (let j = 0; j < keys.length; j++) {
+      const key = keys[j]
+      newPlotData[key].push(...filteredData[key])
+    }
+    return newPlotData
   }
 
   /** Process scatter plot data fetched from server */
@@ -347,7 +390,9 @@ function RawScatterPlot({
     const layout = scatter.layout
 
     if (filteredCells) {
-      scatter = intersect(filteredCells, scatter)
+      const originalData = scatter.data
+      const [intersected, plottedIndexes] = intersect(filteredCells, scatter)
+      scatter.data = reassignFilteredCells(plottedIndexes, originalData, intersected)
     }
 
     const plotlyTraces = updateCountsAndGetTraces(scatter)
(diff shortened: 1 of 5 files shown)


eweitz (approved):
Code and demo video look good!  

I note a few granular non-blocking suggestions for prose and maintainability.

[Discussion in Slack](https://broadinstitute.slack.com/archives/C024HV2FANL/p1695142919252829) has more context for our future reference.

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -941,3 +923,70 @@ function getDragMode(isCellSelecting) {
...
+ * @param plotted {Array} Indices of points that are to be plotted
+ * @param originalData {Object} original scatter data object
+ * @param filteredData {Object} filtered scatter data object from cell faceting
+ * @returns {Object} reorganized scatter data object with new --Filtered-- trace
+ */
+export function reassignFilteredCells(plotted, originalData, filteredData) {
+  const allIndexes = [...Array(originalData['x'].length).keys()]
+  const originalSet = new Set(allIndexes)
+  const plottedSet = new Set(plotted)
+  const reassignedIndices = [...originalSet].filter(i => !plottedSet.has(i))
```
eweitz, line 965: A trivial-but-definite performance optimization: `allIndexes` is an array, then gets transformed into a set (`originalSet`), then gets transformed back to an array `[...originalSet]`.  

However, `originalSet` is never actually used as a set.  Transforming the `plotted` array to `plottedSet` makes sense at a glance due to `!plottedSet.has(i)`, but applying the same transformation to `allIndexes` seems unnecessary.

```suggestion
  const plottedSet = new Set(plotted)
  const reassignedIndices = allIndexes.filter(i => !plottedSet.has(i))
```

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -941,3 +923,70 @@ function getDragMode(isCellSelecting) {
...
 
 ScatterPlot.getPlotlyTraces = getPlotlyTraces
+
+/** Intersect filtered cells with dataArrays results */
+export function intersect(filteredCells, scatter) {
+  const intersectedData = {}
+  const dataArrays = scatter.data
+  const keys = Object.keys(dataArrays)
+  keys.forEach(key => intersectedData[key] = [])
+  window.dataArrays = dataArrays
```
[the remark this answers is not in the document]
bistline: Oh sure - that was actually from [here](https://github.com/broadinstitute/single_cell_portal_core/pull/1862/files#diff-5cf73749539bbea95f82b8026e93c27be8fbb4401b22b336ff1fe612fcdbb7caR327) but I'm happt to do that.

app/javascript/components/visualization/ScatterPlot.jsx
bistline: Ah that's a great catch - will do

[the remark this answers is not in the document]
bistline: That's a great point - I did some testing with https://jsbench.me/, and while it does say that the `for` loop is more than twice as fast, it's only about a 30ms difference in wall time, even at 1M entries in the array.  So, I'd agree that it is both true and not relevant.

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -941,3 +923,70 @@ function getDragMode(isCellSelecting) {
...
+  const plottedSet = new Set(plotted)
+  const reassignedIndices = [...originalSet].filter(i => !plottedSet.has(i))
+  const newPlotData = {}
+  const keys = Object.keys(originalData)
+  keys.forEach(key =>  {
+    newPlotData[key] = []
+  })
+  for (let idx = 0; idx < reassignedIndices.length; idx++) {
+    for (let k = 0; k < keys.length; k++) {
+      const key = keys[k]
```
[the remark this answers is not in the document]
bistline: Normally I'd agree, but in this particular case the outer loop is using `idx` as the counter variable, and I didn't want the inner one to use `i`.  I went for `k` since it's the iterator for the `keys`, whereas the outer loop is going through an array of indices.

app/javascript/components/visualization/ScatterPlot.jsx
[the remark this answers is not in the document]
eweitz: Thanks, that's helpfully concrete!

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -941,3 +923,71 @@ function getDragMode(isCellSelecting) {
...
+ *
+ * @param plotted {Array} Indices of points that are to be plotted
+ * @param originalData {Object} original scatter data object
+ * @param filteredData {Object} filtered scatter data object from cell faceting
+ * @returns {Object} reorganized scatter data object with new --Filtered-- trace
+ */
+export function reassignFilteredCells(plotted, originalData, filteredData) {
+  const allIndexes = [...Array(originalData['x'].length).keys()]
+  const plottedSet = new Set(plotted)
+  const reassignedIndices = allIndexes.filter(i => !plottedSet.has(i))
```
ehanna4, line 966: These couple of lines seem a bit unnecessarily complex. It took me a while to sort out what it was doing, and it's technically making 3 extra arrays that aren't really needed (via the spread, the .keys() and the [] brackets) I think this could be accomplished using just a forloop like:

`const reassignedIndices = []
 const plottedSet = new Set(plotted)
for (let i = 0;  i < originalData['x'].length; i++)
    if (!plottedSet.has(i)) reassignedIndices.push(i)`

ehanna4 (approved):
Overall approved, left one NIT and one suggestion that I think would be good to consider to simplify a couple of lines

app/javascript/components/visualization/ScatterPlot.jsx
bistline: Oh that's a great idea - let me do some benchmarking to see how that performs but I have a feeling that will be both simpler and faster.

bistline: Yup - twice as fast and definitely easier to read.  FWIW this has the same performance as the earlier comment (only about 120ms vs 60) but the readability improvement alone is worth it.

jlchang (commented):
@bistline I thought I saw something odd with the filtering on Friday. Trying to reproduce it now... Could you hold off on merging this for a bit?

bistline:
> @bistline I'm trying to reproduce a viz bug, but first, I'll describe a filtering UI bug. I wanted to to plot points by `cell_type__ontology_label` and then filter by `General_Celltype`. Each "Filter plotted points by" dropdown seems limited to two options? I was about to give up when I found a way to do what I wanted but it the steps to reproduce were pretty wonky. I'm starting with a completely fresh load of the "Human milk - differential expression" study and choose "color by" `General_Celltype`, but none of the "filter by" dropdowns have `General_Celltype`. If I set both "color by" and "filter by" to `cell_type__ontology_label`, then "Reset view" and look at cell filters again, THEN I'm in a position to do what I want (I suspect this is because the dropdowns don't request updates when "color by" is changed?
> 
>  Cell_filter_UI.bug.mov

@jlchang I think a couple things might be happening here.  First, I _think_ the reason you're only seeing those two options is because of logic from #1862, which will prioritize cell- or disease-like facets over other ones.  But I think you're correct in that those dropdown menus don't update when a different annotation is selected (which it should).  And in retrospect, it might be nice to see other options in those menus, but have the most appropriate one promoted?  @eweitz did that original work and he can better describe the logic/reasoning behind it.

Either way, both seem out of scope for this PR, but certainly the update issue should be addressed in downstream work.

PR #1880, opened 2023-09-14
Restore manual Appcue pin for visualization tour of SCP viz demo study (SCP-5254)

bistline:
#### BACKGROUND & CHANGES
This update adds in a manual button to link to the [Demo visualization study](https://singlecell.broadinstitute.org/single_cell/study/SCP2271/scp-visualizations-demo) on production.  Visibility of this flag is governed by the feature flag `show_appcue_viz_tour`.  

Examples of button on home page:
appcue_button_home

And study overview page:
appcue_button_study

#### MANUAL TESTING
1. Run the new migration to create the feature flag
2. Boot as normal and note there is no button on the home page or on any study overview page
3. Sign in, and go to the "[Feature flags](https://localhost:3000/single_cell/feature_flags)" admin page, then set the flag override on your user account for `show_appcue_viz_tour` to `true`
4. Go back to the home page and confirm the button is shown
5. Load any study and confirm the button is present there as well
6. Open Chrome Dev Tools > Network
7. Click the link, confirm you are redirected to SCP2271 on production
8. Back in the Network tab, confirm you see a Mixpanel event for `click:tab` with the `text` value of `appcue-viz-demo` (the `tab` event is because these links are inside tabs on both pages)

PR #1885, opened 2023-09-21
Remove manual Appcue pin for visualization tour (SCP-5323)

bistline:
#### BACKGROUND & CHANGES
This update removes the manual Appcue pin for the SCP visualization tour.  Not using the client integration through Appcue resulting in the tour not behaving correctly.  As such, we are removing the manual pin.

#### MANUAL TESTING
1. Run the new migration to retire the feature flag with `./rails_local_setup.rb && source config/secrets/.source_env.bash && bin/rails db:migrate`
2. Boot as normal and sign in
3. Confirm you do not see the pin on the home page or any study overview
4. Go to the [Feature flags](https://localhost:3000/single_cell/feature_flags) admin console and confirm that `show_appcue_viz_tour` is no longer listed

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -165,7 +165,6 @@ class SiteController < ApplicationController
     # normally we would do this in React but the tab display is in the Rails HTML view
     # we need to check server-side since we have to account for @study.can_visualize? as well
     @explore_tab_default = @study.can_visualize?
-    @show_appcue_pin = FeatureFlaggable.feature_flags_for_instances(current_user).[](:show_appcue_viz_tour)
   end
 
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good overall -- a nice tech debt resolution.

Per PR #1880, it'd also help to remove the now-unused `.appcue-pin ` in `app/javascript/styles/_global.scss`.

bistline:
> Confirmed the absence of manual Appcue pin in both locations and removal of show_appcue_viz_tour feature flag.
> 
> Is there anything we need to do on the Appcues end?

No - it's already turned off there, so no further work is needed.

PR #1887, opened 2023-09-26
Disable sentry plugin during local development (SCP-5327)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with the `sentryVitePlugin` where breakpoints do not work locally.  The cause is unclear, and disabling the plugin appears to fix the problem.  Since we do not report to Sentry in non-production environments, this seems preferable to digging in further.  It will default to _not_ disabling the plugin so as not to interfere with deployed instances.

This can be done in a local environment when starting the Vite dev server with:
```
DISABLE_SENTRY=true bin/vite dev
```

#### MANUAL TESTING
1. Boot vite with `DISABLE_SENTRY=true`, and start the Rails server as normal
2. Open Chrome DevTools, and go to the `Sources` tab
3. Press Cmd+P and open `metrics-api.js`
4. Add a breakpoint at line 406:
```
export function log(name, props = {}) {
==>  let isDifferentialExpressionEnabled // track differential expression visibility globally
```
5. On the homepage, click any link and confirm that the breakpoint halts execution when attempting to log to Mixpanel
6. (Optional) Boot in Dockerized mode with `bin/docker-compose-setup.sh`
7. Run the same test and confirm the breakpoint works

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -6,7 +6,14 @@ import { readFileSync } from 'fs'
 
 // Match latest non-draft at https://github.com/broadinstitute/single_cell_portal_core/releases
 const version = readFileSync('version.txt', { encoding: 'utf8' })
-const disableSentry = !!process.env.DISABLE_SENTRY
+
+// sentryVitePlugin should be disabled in local development as it prevents using breakpoints
+// to turn off locally, run:
+//
+// DISABLE_SENTRY=true bin/vite dev
+//
+// otherwise, this evaluates to false and leaves plugin enabled in all other scenarios
+const disableSentry = typeof process.env.DISABLE_SENTRY !== 'undefined' && process.env.DISABLE_SENTRY === 'true'
 
 export default defineConfig({
   'define': {
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -6,6 +6,7 @@ import { readFileSync } from 'fs'
 
 // Match latest non-draft at https://github.com/broadinstitute/single_cell_portal_core/releases
 const version = readFileSync('version.txt', { encoding: 'utf8' })
+const disableSentry = !!process.env.DISABLE_SENTRY
 
 export default defineConfig({
   'define': {
@@ -23,7 +24,8 @@ export default defineConfig({
       'org': 'broad-institute',
       'project': 'single-cell-portal',
       'authToken': process.env.SENTRY_AUTH_TOKEN,
-      'telemetry': false
+      'telemetry': false,
+      'disable': disableSentry
     })
   ],
   'build': {
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -29,6 +29,7 @@ services:
     entrypoint: ./vite-dev-entrypoint.sh
     environment:
       RAILS_ENV: development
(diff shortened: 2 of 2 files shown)


vite.config.ts
```diff
@@ -6,6 +6,7 @@ import { readFileSync } from 'fs'
 
 // Match latest non-draft at https://github.com/broadinstitute/single_cell_portal_core/releases
 const version = readFileSync('version.txt', { encoding: 'utf8' })
+const disableSentry = !!process.env.DISABLE_SENTRY
```
eweitz, line 9: For maintainability:

```suggestion
// Enabling Sentry on localhost causes Chrome DevTools breakpoints to not work,
// but might occasionally be useful to debug logging to Sentry.
//
// To enable Sentry logging, you must omit DISABLE_SENTRY in your Vite command, e.g.:
//
//    DISABLE_SENTRY="true" bin/vite dev # This is what you usually want for local dev
//
//    DISABLE_SENTRY="false" bin/vite dev # Doesn't actually enable Sentry
//
//    bin/vite dev # Enables Sentry, but causes localhsot DevTools breakpoints to fail
//
// Details: https://github.com/broadinstitute/single_cell_portal_core/pull/1887
const disableSentry = !!process.env.DISABLE_SENTRY
```

eweitz (approved):
Thanks for fixing local breakpoints!  

Ideally, rather than `DISABLE_SENTRY`, I think we'd have an `ENABLE_SENTRY` argument / flag, since we want to disable Sentry in localhost _by default_ (and thus keep breakpoints working).  I elaborate on this a bit below in a suggestion to add a code comment.  But I'm not sure off-hand how to get that working nicely with `docker-compose-dev.yaml`, so I don't consider it a blocker.

I tested the code locally, and it works as expected.

vite.config.ts
bistline: That's a great point - as written, _any_ value passed here gets interpreted as a "turn off Sentry" option.  Let me rethink this a little bit.  I need to ensure that it always is turned on in deployed environments since this config applies to everything, so the logic likely needs to be stricter.

bistline: [7679252](https://github.com/broadinstitute/single_cell_portal_core/pull/1887/commits/76792526a23070c3a4be8c9c8ccf808d0468b8e2)
As per our offline discussion, I'm keeping the `DISABLE` over `ENABLE` nomenclature so we don't have to thread this through the rest of the stack since this config is shared for all environments.  But the logic now works correctly:

```
# sentryVitePlugin enabled, breakpoints don't work locally
bin/vite dev 

# sentryVitePlugin disabled, breakpoints restored
DISABLE_SENTRY=true bin/vite dev

# value disregarded, sentryVitePlugin enabled
DISABLE_SENTRY="anything other than 'true'" bin/vite dev
```

I also added a comment per your suggestion.  Lastly, I tested this in Docker and it works correctly, and I will check after this is merged to ensure sourcemaps are still being served to Sentry.

eweitz: Nice!

jlchang (approved):
Now that I've learned how to set breakpoints to do debugging, I'm very glad this change will be going out soon!

PR #1893, opened 2023-10-03
Preserve color assignments during cell filtering (SCP-5310)

bistline:
#### BACKGROUND
When using the cell filtering feature, if a user deselects any label that results in an entire group being removed from the plot, all color assignments will shift up by one value.  This in essence changes the meaning of the plot, as the color assignments do not persist.  This is confusing and visually jarring to the end user.

#### CHANGES
Now, original color assignments are persisted when using the cell filtering feature.  Any completely filtered groups will move to the `--Filtered--` trace, but the colors will not shift in accordance.

Existing bug:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/804d5309-f784-4761-b63c-f2e9a3fae1ae

Fixed UX:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/cc889cea-0945-4574-9293-056baf9e20d8

#### MANUAL TESTING
1. Boot as normal, sign in and load the `Human milk - differential expression` study
2. Once enabled, click the `Cell filtering` button for the `All Cells UMAP` clustering
3. Deselect `macrophage` from the `cell_type__ontology_label` annotation
4. Confirm that the color for `LC2` remains grey, as pictured above

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -598,6 +598,7 @@ function RawScatterPlot({
             updateIsSplitLabelArrays={updateIsSplitLabelArrays}
             externalLink={scatterData.externalLink}
             originalLabels={originalLabels}
+            setOriginalLabels={setOriginalLabels}
           />
         }
       </div>
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -333,7 +333,7 @@ function RawScatterPlot({
       const originalData = scatter.data
       const [intersected, plottedIndexes] = intersect(filteredCells, scatter)
       scatter.data = reassignFilteredCells(plottedIndexes, originalData, intersected, annotIsNumeric, setOriginalLabels)
-    } else {
+    } else if (!annotIsNumeric) {
       setOriginalLabels(getPlottedLabels(scatter.data))
     }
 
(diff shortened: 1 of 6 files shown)


jlchang (approved):
Functional review - confirmed annotation group colors no longer shift when entire traces are filtered.

Looking forward to playing with Cell filtering in production!

app/javascript/components/visualization/ScatterPlot.jsx
```diff
@@ -324,12 +326,15 @@ function RawScatterPlot({
...
+    const annotIsNumeric = scatter.annotParams.type === 'numeric'
     const layout = scatter.layout
 
     if (filteredCells) {
       const originalData = scatter.data
       const [intersected, plottedIndexes] = intersect(filteredCells, scatter)
-      scatter.data = reassignFilteredCells(plottedIndexes, originalData, intersected)
+      scatter.data = reassignFilteredCells(plottedIndexes, originalData, intersected, annotIsNumeric, setOriginalLabels)
+    } else if (!annotIsNumeric) {
+      setOriginalLabels(getPlottedLabels(scatter.data))
```
eweitz, line 337: Would this be be simpler and faster?

```suggestion
      setOriginalLabels(scatter.annotParams.values)
```

I haven't tested that, but it looks like `getPlottedLabels` recomputes on the client what's already available at hand from the server response.

eweitz (approved):
Code looks generally good!

I note one non-blocking, possible refinement for speed and maintainability.  I could also see merely noting it as a code comment (unless it's clearly inapplicable) if you want this to go out tomorrow and want more bandwidth for testing.

app/javascript/components/visualization/ScatterPlot.jsx
bistline: Yes, it does - if you have a suggestion for a simpler way to compute the unique entries from a long array, I'm all ears!

eweitz: Agreed, spread/Set seems ideal to uniquify an array, if needed.  Rather, the optimization that crossed my mind was avoiding that need, e.g. maybe using `scatter.annotParams.values` only for cluster-wide annotations, or adjusting color assignment to e.g. always start with red even for study-wide annotations with filtering.  Neither of those seem obviously worthwhile right now, I just note it as food for thought; your approach here is good!

PR #1899, opened 2023-10-11
Disable cell filtering if study has spatial clustering (SCP-5362)

bistline:
#### BACKGROUND & CHANGES
This update disables cell filtering for studies that have spatial clustering.  The reasoning is that the cells in the spatial mappings may not be the same cells from the embeddings (e.g. UMAP), and so filtering by cell name will not work reliably across the two.  Having incomplete or non-existent filtering would be confusing, so as we evaluate potential workarounds or messaging cell filtering will be disabled for any studies with spatial clustering.

#### MANUAL TESTING
1. Boot as normal and load the `Chicken raw counts` study (or any other study with spatial clustering)
2. Confirm the `Cell filtering` button does not appear


eweitz (approved):

jlchang (approved):
Looks good!

bistline:
Build failure addressed in #1900.

PR #1900, opened 2023-10-11
Fix SearchFacetPopulator and associated tests re: OLS updates (SCP-5371)

bistline:
#### BACKGROUND & CHANGES
This update fixes a regression with `SearchFacetPopulator` and associated tests due to changes upstream to the Ontology Lookup Service API.  Specifically, the value for an ontology title no longer lives at `ontology.config.title` but has moved to `ontology.config.localizedTitles.en` to support translations.  We will now look first in the original place, and if no title is found, we will look in the localized section.  Fallback title remains the URL.

#### MANUAL TESTING
1. Pull branch and load secrets with `./rails_local_setup.rb && source config/secrets/.source_env.bash`
2. Run `bin/rails test test/lib/search_facet_populator_test.rb` and confirm all tests pass

### DIFF
--- a/app/lib/search_facet_populator.rb
+++ b/app/lib/search_facet_populator.rb
@@ -56,7 +56,7 @@ class SearchFacetPopulator
       urls.zip(browser_urls).each do |url, browser_url|
         ontology = fetch_json_from_url(url)
         # check if response has expected keys; if not, default to URL for name value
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Thanks for the quick fix!

jlchang (approved):
Looks good.

PR #1897, opened 2023-10-10
Collapsible cell filters, human-friendly names (SCP-5297)

eweitz:
Users can now see more categories for filtering cells, and richer data, in a more inviting design.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/aeb62c92-b01f-4bfc-ad4c-3632deb54b78

### Collapsible filters
Previously (#1895), filter lists could be very long, making it especially difficult to visually scan across categories to search.  Now, long filter lists are tailored, with quick ways to expand and collapse.  By default, filter lists are _partly_ collapsed only if they are > 5 filters.  This eases interaction with short lists, and makes long lists usefully previewable.  

Beyond each filter list now being _partly_ collapsible, each list can also be _fully_ collapsed, and _all filter lists_ can also be collapsed.  Fully collapsing all filter lists trades off easy interactivity for easy scannability -- supporting the use case of most easily seeing which facets (i.e., categories, annotations) are available to search.

### Human-friendly names, cleaner layout
Also previously, facet names showed exactly what the author uploaded; these names often had technical artifacts like underscores and SCP-internal artifacts like `__ontology_label`.  Now, facet names are human-friendly, so e.g. "cell_type__ontology_label" becomes "Cell type", and a tooltip reports it as coming from the SCP metadata convention, among other useful supplementary data about the annotation.  

These facet names are shorter, yet new layout robustness ensures salient parts of a name like "Reported infant medical events description" are never truncated to e.g. "Reported infant me...".  Supplementary usage notes have been moved from salient info icons into subtler tooltips on relevant section-header text.  The clean default look is cleaner, and surfaces new features contextually.

This also prepares the layout for more facets.  Facets (and filters) that haven't yet loaded are _visible_, but not interactive.  Clean visuals clearly indicate that new state.  

Altogether, these frontend refinements align the prototype with popular UX conventions for faceted search, as seen in e.g. [Amazon](https://www.amazon.com/camera/s?k=camera), [Google Shopping](https://www.google.com/search?tbm=shop&hl=en&psb=1&q=camera&oq=camera&sclient=products-cc), and [Terra](https://app.terra.bio/#library/showcase).  By aligning with common faceted search UI patterns, refining visual signal-to-noise, and surfacing things contextually, the design becomes more familiar and inviting for users to engage.

## Test
A new automated test verifies facet collapse functionality.  To manually test:
1. Go to "Human milk - differential expression" with cell filtering enabled
2. Click "Cell filtering"
[...]

PR #1901, opened 2023-10-11
Enable cell filtering by all facets, not just top 5 (SCP-5361)

eweitz:
This lets users filter plotted cells by _all_ eligible categories, not just the top 5.  It vastly broadens search capability.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/b641c7ba-ea08-488d-b7ad-ff800908311b

## Overview
Previously (#1897), only up to 5 facets (i.e. categories, annotations) were shown for cell filtering.  These initial facets were sorted by relevance, which helps surface interesting ways to search.  The initial set was limited to 5 facets to enable fast time-to-interactive (TTI), and also to not overwhelm the server.  However, only showing 5 out of often 25+ eligible facets severely narrowed users' search capabilities.  User frustration around that was foreseeable.

Now, all eligible facets are shown, and remaining facets after the initial load become interactive as data is retrieved.  To keep UX TTI fast and server saturation lower, data is fetched in batches / chunks of <= 5 facets at a time, rather than all remaining facets at once.  Even after 25+ facets are fully loaded, cell filtering is still instant in a real, representative dataset of 48K cells.

## Known issues
To get user feedback sooner, this enhancement achieves the main user-facing goal of quickly loading all facets, but ships with some rough edges.  Examples aimed for refinement in a subsequent PR for this ticket:

* If you click "Cell filtering" _while data is loading_, then unloaded facets won't become interactive while the panel is in view.  You need to exit the panel then re-enter for newly-loaded facets to become interactive.

* Too many requests are sent.  This is a pre-existing issue, but it's exacerbated in this enhancement.  Duplicative requests should be eliminated.

* Requests should be fully serialized.  The `/facets` requests sometimes begin in parallel, and sometimes _begin_ serially. 
 Each batched request _should not start_ until the previous one in the stream has completed.

These issues seem fine for an allow-listed beta, but warrant refinement before general availability.

## Test
To manually test:
1. Go to "Human milk - differential expression" with cell filtering enabled
2. Soon after the page loads, click "Cell filtering"
3. Confirm the first 5 facets appear and are interactive
4. Scroll fully down in panel, check if any facets are loaded
5. Confirm any loading facets appear grey, with loading icon, and filters disabled
6. If any facets are loading, click the "<-" to go back to "Options" panel
7. Wait a bit, click "Cell filtering" again
8. Confirm facets have finished loading (repeat 6-7 if so)
9. Scroll down to bottom of panel
10. Click any filter in "Antibiotics during delivery"
11. Confirm plot updates

This relates to SCP-5361.


bistline (approved):
First off - very slick!  This makes things much more useful overall.  I don't see any issues code-wise.  Apologies for the lengthly response below:

To your point about the multiple requests - I share the view that for a closed beta, this is fine, but for GA would need to be addressed.  While I was testing locally, most of the requests for downstream facets were issued 2 or 3 times.  My suspicion is that our liberal use of `useEffect` in the Explore tab is having unintended consequences that are becoming more apparent here.  I'm reminded of previous debugging sessions when dealing with gene search load where we saw multiple identical requests issued within milliseconds of each other from the same client.  The point I'm trying to make here is that I don't think the issue is limited to (or derived from) the cell faceting feature - I believe that it is an underlying issue that is becoming more apparent with the release of this feature, and may be more complicated than we anticipate.

To that end - the fact that these responses will be cached will help greatly with load & rendering speed.  We could consider pre-caching this data if we find that reducing duplicate requests proves difficult to resolve.  I think ideally we would defer caching until the regular cluster visualization caches are generated so as to limit server load.

jlchang (approved):
I'm thrilled that our small group of beta testers will be able to test out filtering by all facets. 

One non-blocking UI suggestion (possibly piggyback for a follow-on ticket) - in our legends, the annotation labels are alphabetized, can we also alphabetize the filter options? For label findability, having the filter names sorted should be generally helpful and any `_YN` options would be listed in a predictable ordering.

Another `Known issue` (I don't think this is addressed in a different ticket for this release or in a follow on ticket), filter group availability after `Color by` has been changed:
If the `Color by` selection is changed (for example, from General_Celltype to infant_sick_YN), the original label selected for color by (General_Celltype) cannot be found as a "Filter by" option unless you exit the cell filtering panel then re-enter.
Video of issue in case it isn't already described in a ticket:

https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/03e7432e-08f2-42da-9d1d-014498f511e2

One minor bug: If I test with a different study "Mouse stomach cells and GERD, with over 200 cell types", which has 5 group-based annotations, only three "Filter by" groups are in the cell filtering panel when I would expect 4. (If I choose the numeric annotation for "color by", then I see four "Filter by" groups, but I would have expected 5):

https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/fa9bda0f-d876-4427-8072-054ad2b432ba


I'd consider it blocking IF we're consistently not getting expected filter groups but NON-blocking if it is an edge case for studies with either cluster-based group-type annotations or group-type annotations with >200 labels. 

My gut feeling is that the issue is an edge case but I don't have enough local studies to test to be sure. Testing with "Mouse stomach cells and GERD" suggests that cluster-based group-type annotations do not cause missing filter groups (caveat: this study has two fewer annotations).

eweitz:
Thanks for the thorough functional review, Jean!  I've created a few tickets to track the issues you found:

* SCP-5372: Changing color-by should enable facet for that annotation
* SCP-5373: Refine filter list ordering
* SCP-5374: Not all eligible facets available

I believe all of these are not caused by these changes, i.e. I think the causes are pre-existing, but most of them are newly apparent.  The "changing color-by" issue is also seen on production now, although it manifests in the lack of an expected dropdown menu option; the first ticket has a new further clarifying video.

eweitz:
The [build failure](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/6487491857/job/17617915891?pr=1901#step:5:3965) is a false positive, fixed by #1900.

```
Total elapsed time: 26 minutes, 53 seconds

### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  SearchFacetPopulatorTest#test_populate_facets_from_alexandria_convention_data [/home/app/webapp/test/lib/search_facet_populator_test.rb:20]:

  SearchFacetPopulatorTest#test_updates_existing_facets [/home/app/webapp/test/lib/search_facet_populator_test.rb:52]:
Exiting with code: 1
```

PR #1903, opened 2023-10-12
updating to 6.1.7.6 re: permission issue

bistline:
Fixes issue with `activesupport` in `6.1.7.5` where a library is not readable to users other than `root`.


eweitz (approved):
Thanks for the fix!  Here's [more context in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1697117644760229), for future reference.

jlchang (approved):
Makes sense to me!

PR #1904, opened 2023-10-13
Stream, smoothen cell facet loading (SCP-5361)

eweitz:
This streams the facet loading UX for a smoother start to cell filtering.

## Overview
Previously (#1901), if you clicked "Cell filtering" before all facets had loaded, then facets beyond the top 5 would be stuck in a loading UI.  Those facets would remain disabled even after data was loaded.  Making non-top facets interactive required closing and re-opening the panel.

Now, facets become interactive as new batches of data are loaded, i.e. in a streaming fashion.  Here's how it looks.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/16efe849-bd94-42fa-ab53-177fec28afeb

## Test
1.  Go to "Human milk - differential expression" with cell filtering enabled
2.  Soon after the page loads, click "Cell filtering"
3.  Confirm the first 5 facets appear and are interactive
4. Scroll fully down in panel, check if any facets are loaded
5.  Confirm any loading facets appear grey, with loading icon, and filters disabled
6. Wait for a while to let local `/facets` requests complete; don't exit "Cell filtering" panel
7. Confirm facets that were loading and disabled are now loaded and interactive, without any effort needed from you

This satisfies SCP-5361.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -199,6 +199,7 @@ export default function ExploreDisplayTabs({
 
   // if the exploreParams update need to reset the initial cell facets
   useEffect(() => {
+    if (!cellFaceting) {return}
     const [newCluster, newAnnot] = getSelectedClusterAndAnnot(exploreInfo, exploreParams)
     getCellFacetingData(newCluster, newAnnot)
   }, [exploreParams?.cluster, exploreParams?.annotation])
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -202,7 +202,7 @@ export default function ExploreDisplayTabs({
     if (!cellFaceting) {return}
     const [newCluster, newAnnot] = getSelectedClusterAndAnnot(exploreInfo, exploreParams)
     getCellFacetingData(newCluster, newAnnot)
-  }, [exploreParams?.cluster, exploreParams?.annotation])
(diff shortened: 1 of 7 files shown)


jlchang (approved):

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -193,15 +201,11 @@ export default function ExploreDisplayTabs({
...
-  if (!cellFaceting) {
-    getCellFacetingData(selectedCluster, selectedAnnot)
-  }
-
   // if the exploreParams update need to reset the initial cell facets
   useEffect(() => {
     const [newCluster, newAnnot] = getSelectedClusterAndAnnot(exploreInfo, exploreParams)
     getCellFacetingData(newCluster, newAnnot)
-  }, [exploreParams?.cluster, exploreParams?.annotation])
+  }, [exploreInfo?.cluster, exploreInfo?.annotation])
```
bistline, line 208: Can you elaborate why this change was necessary?  I ask because I wonder if it explains some of the other "duplicate request" behavior we've seen other places, and if this paradigm could also be used there (like for intra-study gene search).

bistline (approved):
Nice!  I love it when the solution is to _remove_ code...

app/javascript/components/explore/ExploreDisplayTabs.jsx
eweitz: Insightful question.  Per our chat, it turns out this fixed one problem, and created another.  I've opened SCP-5379 to track that fix.  For now, I've reverted back to `exploreParams`, which is noisier at a technical level but avoids user-facing confusing.  The ticket has further details and steps to reproduce.

PR #1905, opened 2023-10-17
Show cell count beside each filter, update on selection (SCP-5378)

eweitz:
This shows adaptive cell counts beside each filter.  It surfaces category distributions, and how they change upon filtering.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/5796e260-187b-4688-b4a0-451514707d91

### Show cell count beside each filter
Previously (#1904), seeing the distribution of cells across groups in an annotation required changing the "Annotation" (a.k.a. "Color by") menu.  Distributions were only reported in the scatter plot, via the legend.  This made it tedious to assess cell distributions across annotations.

Now, cell counts are reported not only beside each entry in the scatter plot legend, but also beside each filter in all facets.  This makes it easy to determine how cells are distributed across _all_ annotated categories.

### Update on selection
The new filter counts are dynamic.  Each time you check or uncheck a filter, counts for all filters in all facets update to reflect the number of cells matched by the new selection.  These adaptive counts are a powerful way to summarize results of slicing and dicing single-cell datasets across annotations.

## Test
1.  Go to "Human milk - differential expression" with cell filtering enabled
2. Click "Cell filtering"
3. Note numbers beside each filter
4. In "Cell type" facet, click "More..."
5. Confirm "animal cell" does not appear (it has 0 matching cells in the clustering; its prior display was a bug)
6. Click a filter to deselect it
7. Confirm the count beside the deselected filter is 0
8. Confirm counts beside filters in other facets update

This relates to SCP-5378, a Developer's Choice Days project.


Gemfile
```diff
@@ -4,7 +4,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '3.1.3'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '6.1.7.5'
+gem 'rails', '6.1.7.6'
```
bistline, line 7: I got confused as to why this was appearing here until I realized the merge target wasn't `development`.

app/javascript/lib/cell-faceting.js
```diff
@@ -86,62 +86,47 @@ function sortAnnotationsByRelevance(annotList) {
...
     results = filterableCells
   } else {
-    for (i = 0; i < facets.length; i++) {
+    for (let i = 0; i < facets.length; i++) {
       facet = facets[i]
       if (facet in selection) { // e.g. 'infant_sick_YN'
         const friendlyFilters = selection[facet] // e.g. ['yes', 'NA']
 
-        const filter = {}
+        const filter = new Set()
```
bistline, line 107: Much nicer overall

app/javascript/lib/cell-faceting.js
```diff
@@ -150,7 +135,8 @@ export function filterCells(
   const numFacetsSelected = Object.keys(selection).length
   const numFiltersSelected = Object.values(selection).length
   const filterLogProps = {
-    perfTime: filterPerfTime,
+    'perfTime': filterPerfTime,
+    'perfTime:counts': perfTimeCounts,
```
bistline, line 139: Is this an "I'm curious how long it takes" or an "I'm concerned about how long it may take"?

app/javascript/lib/cell-faceting.js
```diff
@@ -182,6 +168,110 @@ function mergeFacetsResponses(newRawFacets, prevCellFaceting) {
   return mergedRawFacets
 }
 
+/** Omit any filters that match 0 cells in the current clustering */
+function trimNullFilters(cellFaceting) {
```
bistline, line 172: On first pass, I think I agree with this approach (removing empty entries) but I wonder if there's a scientific value for seeing that a combination of filters results in populations being reduced to 0?

bistline (approved):
Very cool enhancement.  As per our offline discussion, I am seeing the following rendering issue on my setup:
Screenshot 2023-10-17 at 1 54 06 PM
It's unclear if this is a larger issue, or just a result of my particular multi-monitor setup and screen resolutions.  I don't consider it to be blocking as it is not reproducible anywhere other than my machine (so far).

app/javascript/lib/cell-faceting.js
eweitz: Yes, I think there's definitely scientific value in seeing when a selection yields a 0-count for a given filter.  This "null filter" trimming addresses something a bit different.  

It turns out that _some filters have a 0-count without any selection having been made_.  For example, if you go to [SCP1671 on production](https://singlecell.broadinstitute.org/single_cell/study/SCP1671/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation), there's a filter shown for "animal cell".  Selecting and deselecting the "animal cell" filter does nothing, unlike all the other filters in that facet.  That's because while "animal cell" is an annotation group value in the metadata file, no cell in that clustering has the group value "animal cell".

That's a pre-existing issue, but it becomes severe once filter counts are shown.  Such "null groups" would also affect the scatter plot legend and DE panel, but get omitted by code that builds `countsByLabel`.  This measure in `trimNullFilters` applies a similar approach to omitting those null groups in filters.

app/javascript/lib/cell-faceting.js
eweitz: Mostly the former.  In my local testing, getting counts was typically < 50 ms.  Since I wrote the trivial timing code while benchmarking and it doesn't cost us anything to log the extra property, I thought I'd add it in for richer insight towards possible optimizations later.

app/javascript/lib/cell-faceting.js
bistline: Ah - thank you for that explanation.  That makes total sense!

eweitz:
Good find on the response layout bug, Jon and Jean.  That's fixed now; see video.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/86604f7e-0a7d-42b6-89da-ba7d18b17d1c

eweitz:
> One possibly-blocking observation - if I try to look at this feature on a larger dataset (using the dataset from staging SCP323 set up for PR#1906), I can reproducibly get "RangeError: Maximum call stack size exceeded"

Nice catch here, @jlchang!  I think that's actually a new symptom of a pre-existing bug that previously only manifested as a _subtly_ incorrect / confusing UX (which might have been worse).  In brief, your finding reveals that cell filtering sometimes shows filters for wrong clustering.  

I've opened SCP-5380 to track that, including steps to reliably reproduce the pre-existing bug, and put it in the upcoming sprint.  Since the underlying cause of the bug seems separate from cell filter counts, it seems best to scope for a separate ticket / PR.

jlchang:
> Good find on the response layout bug, Jon and Jean. That's now fixed by [300b307](https://github.com/broadinstitute/single_cell_portal_core/commit/300b3074ff582de4235fb7a2571c899361e75585).   

The checkboxes line up with the filter name but the cell count location is still off for me
filter_panel_layout_still_has_issues

jlchang (approved):
Approving with the understanding that [SCP-5380](https://broadworkbench.atlassian.net/browse/SCP-5380) is GA-blocking and not in scope for this PR.

I'm still seeing some issues with filter panel layout (see image in [comment above](https://github.com/broadinstitute/single_cell_portal_core/pull/1905#issuecomment-1768733186)), while they decrease readability, I'm comfortable with acknowledging the problem as a known issue for the invite-only beta.

eweitz:
Good insights!  I've now fixed this more robustly, with finer tailoring and less code, in c82da16.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/ed1bdeb1-5411-4a66-938b-cfb9e3150b97

PR #1912, opened 2023-10-23
Cell filtering: ensure correct facets on clustering change (SCP-5380)

eweitz:
This fixes a bug in cell filtering sometimes seen upon switching clusterings.

### Overview
Previously, if you changed the "Clustering" menu, and the new clustering was not study-wide and had annotations or groups different than the previous clustering, then a few problems would manifest:

1. If the new clustering had an annotation _not_ in the previous clustering, then the new facet would display, but all its filters would be unexpectedly not selected.  

2. If the new clustering had an annotation that _was_ in the previous clustering, but that annotation had only 1 group, then the facet would display, even though facets are only eligible for cell filtering if their annotations have > 1 group in the current clustering.

Now, cell filtering works as expected in those cases: filters in new facets are selected by default, and new facets that are ineligible are not shown.  Clusterings get changed [~660 times / day](https://mixpanel.com/s/37lHyd), so although it's unclear how often the other needed conditions are met, this is plausibly a high-incidence scenario, worth having fixed while the beta is closed.

This includes some related refinements.  It fixes the `iconColor` bug mentioned in #1907 -- it manifested reliably for me in some studies with Vite SW caching in development.  Also, tooltips in the cell filtering panel that display upon hovering over text are now slightly delayed.

### Video
Here's a video showing the problem on production, and its fix on development.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/b967e5a4-2061-4baa-8361-11f09fe3e31c

### Test
A new automated test adds verification for this fix.  To manually test, if desired:

1.  Go to "Human milk - differential expression" DE pilot, with cell filtering enabled
2. Ensure your default clustering is "All Cells UMAP"
3. Switch clustering to "Epithelial Cell Subclusters"
4. Confirm that no "Cell type" facet is shown (i.e., case 1 described above is fixed)
5. Scroll down, confirm that "Epithelial Cell Subclusters" is shown, and that its filters are selected (case 2)

This satisfies SCP-5380.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -65,9 +65,6 @@ function getHasComparisonDe(exploreInfo, exploreParams, comparison) {
 
 /** Handle switching to a new clustering that has annotations (i.e., facets) not in previous clustering */
 export function handleClusterSwitchForFiltering(cellFilteringSelection, newCellFaceting, setCellFilteringSelection) {
-  console.log('cellFilteringSelection', cellFilteringSelection)
-  console.log('newCellFaceting', newCellFaceting)
-  console.log('setCellFilteringSelection', setCellFilteringSelection)
   if (cellFilteringSelection) {
     const existingSelectionFacets = Object.keys(cellFilteringSelection)
     const updatedSelectionFacets = newCellFaceting.facets.filter(nf => !existingSelectionFacets.includes(nf.annotation))
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -63,6 +63,71 @@ function getHasComparisonDe(exploreInfo, exploreParams, comparison) {
   return hasComparisonDe
 }
 
+/** wrapper function with error handling/state setting for retrieving cell facet data */
+function getCellFacetingData(cluster, annotation, setterFunctions, context, prevCellFaceting) {
+  const [
+    setCellFilteringSelection,
+    setClusterCanFilter,
+    setFilterErrorText,
+    setCellFilterCounts,
+    setCellFaceting
+  ] = setterFunctions
+
+  const {
+    exploreInfo,
+    studyAccession,
+    cellFilteringSelection
+  } = context
+
+  const showCellFiltering = getFeatureFlagsWithDefaults()?.show_cell_facet_filtering
+  if (showCellFiltering) {
+    const allAnnots = exploreInfo?.annotationList.annotations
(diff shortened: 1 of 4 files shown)


app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -63,6 +63,79 @@ function getHasComparisonDe(exploreInfo, exploreParams, comparison) {
   return hasComparisonDe
 }
 
+/** Handle switching to a new clustering that has annotations (i.e., facets) not in previous clustering */
+export function handleClusterSwitchForFiltering(cellFilteringSelection, newCellFaceting, setCellFilteringSelection) {
+  console.log('cellFilteringSelection', cellFilteringSelection)
```
bistline, line 68: are these debug statements, or functional?

bistline (commented):
I'm still seeing the filters being unselected after changing clustering (following the steps from manual testing):
Screenshot 2023-10-23 at 10 46 11 AM

app/javascript/components/explore/ExploreDisplayTabs.jsx
eweitz: Ah, debug -- removed in 1694be2.

bistline:
> I'm still seeing the filters being unselected after changing clustering (following the steps from manual testing): Screenshot 2023-10-23 at 10 46 11 AM

After conferring, this appears to be a result of local development + uncached responses + competing requests.  This has been categorized in other [tickets](https://broadworkbench.atlassian.net/browse/SCP-5379), and is likely to be very hard to reproduce in a deployed environment (though not impossible - with larger studies it could be easier).

bistline (approved):

jlchang (changes requested):
Manual review instructions behave as described but interacting with the FilterPlottedCells feature shows some odd behavior:
@20s deselected `LC2` of `General CellType` annotation; @24s  top "Filter by" annotation flips back and forth from `General CellType` to `Cell Type` (I'm not doing anything to trigger this change); when in `General CellType` state, `LC2` cell count shows 35398 and not zero.
If Clustering switched to "Epithelial Cell Subclusters", while `LC2` is still deselected @50s, the reendered plot shows LC1 and LC2 cells, deselecting of `LC1`removes all cells (correctly reflecting that neither LC1 or LC2 are selected but after a few seconds the cell counts self-restore as seen earlier). Clicking `LC2` replots the LC2 cells but clicking `LC1` does not add back LC1 cells (and the cell count for LC1 sometimes becomes 4427 or sometimes doesn't exist).



https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/3de37300-c723-48ef-95df-db2f6e8e56db

PR #1913, opened 2023-10-23
Encode annotation names for faceting API (SCP-5377)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with the cell filtering UX where annotations with non-word characters in their names (like `Viral+`) get improperly transformed to spaces leading to lookup errors.  Now, all annotation names in the cell faceting UX are encoded correctly before being transmitted. 

#### MANUAL TESTING
This will require editing the name of an existing annotation to replicate the behavior as we do not have any that currently fit this scenario in our synthetic/example data.  In addition, you will want to ensure that caching is turned off as this will impact the content of the `/explore` response.
1. Enter a Rails console session, and load the `Human milk - differential expression` study:
```
```
2. Update the `infant_sick_YN` annotation to include a `+` character (along with the underlying `DataArray` entries):
```
metadata = study.cell_metadata.by_name_and_type('infant_sick_YN', 'group')
data_arrays = DataArray.where(study_id: study.id, study_file_id: metadata.study_file_id, linear_data_type: 'CellMetadatum',
 linear_data_id: metadata.id)
new_name = metadata.name + '+'
metadata.update(name: new_name)
data_arrays.update_all(name: new_name)
```
3. In a separate terminal, ensure caching is turned off by running `bin/rails dev:cache` until you see `Development mode is no longer being cached.`
4. Boot as normal and sign in, then load the example study
5. Once the clustering loads, click the `Filter plotted cells` button and ensure that the `Infant sick YN+` annotation has loaded and is useable.
6. Back in the Rails console, undo the renaming of the annotation:
```
metadata.update(name: 'infant_sick_YN')
data_arrays.update_all(name: 'infant_sick_YN')
```

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -372,7 +372,7 @@ function RawScatterPlot({
     }
 
     scatter.hasArrayLabels =
-      scatter.annotParams.type === 'group' && scatter.data.annotations.some(annot => annot.includes('|'))
+      scatter.annotParams.type === 'group' && scatter.data.annotations.some(annot => annot?.includes('|'))
 
     if (clusterResponse) {
       concludeRender(scatter)
--- a/app/javascript/lib/scp-api.jsx
+++ b/app/javascript/lib/scp-api.jsx
@@ -158,7 +158,7 @@ export async function createUserAnnotation(
  * @param {String} cluster Name of requested cluster
  */
 export async function fetchAnnotationFacets(studyAccession, annotations, cluster) {
-  annotations = annotations.join(',')
+  annotations = encodeURIComponent(annotations.join(','))
   const params = `annotations=${annotations}&cluster=${cluster}`
   const apiUrl = `/studies/${studyAccession}/annotations/facets?${params}`
   const [annotationFacets] = await scpApi(apiUrl, defaultInit())


jlchang (approved):
Glad to know that the `Viral+` annotation won't get left out of the party!

Manual review works as expected. Code change seems straightforward enough :)

I suspect unrelated to the change in this PR and probably due to turning cacheing off for the testing,  I can trigger `TypeError: Cannot read properties of undefined (reading 'includes')` IF 1. `Epithelial Cells UMAP` is the selected clustering and 2. I DON'T wait for all the filters to load before I select a filter.

bistline:
> Glad to know that the `Viral+` annotation won't get left out of the party!
> 
> Manual review works as expected. Code change seems straightforward enough :)
> 
> I suspect unrelated to the change in this PR and probably due to turning cacheing off for the testing, I can trigger `TypeError: Cannot read properties of undefined (reading 'includes')` IF 1. `Epithelial Cells UMAP` is the selected clustering and 2. I DON'T wait for all the filters to load before I select a filter.

Yes, that does appear to be unrelated.  I think it has to do with SCP-5379 or SCP-5355, possibly both.  There's some inconsistency with state when facets are still being loaded.  This appears to only be the case as you pointed out when you change clusters and then interact with the cell filtering menu while things are still loading.  I expect this will be less of a problem on production, but still possible.  As I previously suspected, it does have to do with checking for `scatter.hasArrayLabels` and some data not being fully present.

I've added in a null-safe operator to prevent that error from being thrown, though that can lead to this scenario:
Screenshot 2023-10-24 at 9 56 32 AM

Luckily, this effect is only temporary, and goes away once facets finish loading and you select another filter.  I'm not going to attempt to address that problem as I think it will be covered in either of the above tickets.

PR #1914, opened 2023-10-24
Cell filtering: fix scalability in large datasets (SCP-5351)

eweitz:
This fixes a `RangeError: Maximum call stack size exceeded` error seen in ~ 150k+ cell clusterings, and slightly speeds up legend color when filtering.

To test:
* If on staging, go to [SCP367](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP367/neuronal-study-scp2247-on-staging) with filtering enabled.  If local, go to an especially large study that previously manifested the "RangeError: Maximum call stack size exceeded" error.
* Click "Cell filtering"
* Click any filter to deselect it
* Click it again to select it
* Confirm no "RangeError: Maximum call stack size exceeded" error message appears above the scatter plot

This satisfies SCP-5351.  Further context is available in today's [SCP demo recording](https://drive.google.com/file/d/1jZRJptJNqwiJyERKl3NCtBVXBwiDW4R8/view?t=4m8s).

### DIFF
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -49,8 +49,6 @@ function RawScatterPlot({
   countsByLabel, setCountsByLabel, hiddenTraces=[], isSplitLabelArrays, updateExploreParams,
   filteredCells
 }) {
-  console.log('')
-  console.log('start RawScatterPlot')
   const [isLoading, setIsLoading] = useState(false)
   const [bulkCorrelation, setBulkCorrelation] = useState(null)
   const [labelCorrelations, setLabelCorrelations] = useState(null)
@@ -63,12 +61,8 @@ function RawScatterPlot({
   const [editedCustomColors, setEditedCustomColors] = useState({})
   const [customColors, setCustomColors] = useState({})
 
-  console.log('in RawScatterPlot, 1')
-
   const isRefGroup = getIsRefGroup(scatterData?.annotParams?.type, genes, isCorrelatedScatter)
   const [originalLabels, setOriginalLabels] = useState([])
-
-  console.log('in RawScatterPlot, 2')
 
   const flags = getFeatureFlagsWithDefaults()
 
@@ -125,8 +119,6 @@ function RawScatterPlot({
 
     return scatter
   }
-
-  console.log('in RawScatterPlot, 3')
 
   /** redraw the plot when editedCustomColors changes */
   useEffect(() => {
@@ -329,8 +321,6 @@ function RawScatterPlot({
     concludeRender()
   }
 
-  console.log('in RawScatterPlot, 4')
-
   /** Process scatter plot data fetched from server */
   function processScatterPlot(clusterResponse=null, filteredCells) {
     let [scatter, perfTimes] =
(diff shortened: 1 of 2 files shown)


bistline (approved):
Looks good - and I like the simple solution.  Non-blocking comments:

I tried this locally on a 1.46M cell study.  It works as expected, though there is some lagginess.  In looking at the Mixpanel events for `filter-cells`, I was seeing around 800ms for filtering and about 500ms for counting, meaning it took on average ~1.3s to replot everything.  Not amazing, but also totally usable.

The only issue I ran into was that after selecting several filters and letting the UI think for a bit, I got the "Aw snap" page:
Screenshot 2023-10-24 at 1 57 46 PM

It's unclear if this is just because of the local dev setup, or because now that we can filter on very large datasets, we are more prone to this issue.  For now, completely non-blocking, but worth keeping an eye one.

jlchang (approved):
Thanks for the explanation of the fix at demo. I no longer get the `RangeError` when i select a filter on a local copy of production SCP795 (aka Staging SCP323) but that study (766K cells) does take a looooooong time to load and to load the cell filters.

eweitz:
Thanks for the reviews, and more helpful diagnostics.

> 800ms for filtering and about 500ms for counting, meaning it took on average ~1.3s to replot everything

Surfacing these measurements is useful.  I appreciate it -- I hadn't tested in such a big study!  Below are clarifying notes.

The `perfTime` metric includes the time spent in `filterCells` for counting.  This means those measurements convey 500 ms for filtering = 500 for counting + 300 ms for other filtering logic = 800 ms to return the newly filtered cells and their new counts.  

Downstream time spent to plot the graph is tracked in `perfTime:frontend:plot` in the `plot:scatter` event.  Even that more graphical phase for [just plotting typically takes ~2200 ms for 1-1.5M cells](https://mixpanel.com/project/2120588/view/19059/app/boards#id=3496145&editor-card-id=%22report-34908019%22). That excludes backend times, and frontend times in network transfer and JSON parse.

So, in that 1.46M cell study, I estimate that each filter selection would take 3-4 seconds between clicking the checkbox and seeing an updated scatter plot, likely comprising 0.8 - 1.0 s to filter with counts, and 2 - 3 s to plot.  Waiting 3-4 s is indeed sluggish, but also likely around the 99th percentile of slowness for what we can expect for that interaction, given that dataset size is in our top ~1%.

> The only issue I ran into was that after selecting several filters and letting the UI think for a bit, I got the "Aw snap" page

Crashing concerns me more than slowness.  It's due to using too much memory.  I have some promising leads to optimize that in the `cell-faceting.js` library.  That will likely also help with speed.  I'll fold some memory optimizations into my ongoing work in SCP-5389.

> a looooooong time to load and to load the cell filters.

I sympathize!  Running `VITE_FRONTEND_SERVICE_WORKER_CACHE=true DISABLE_SENTRY=true bin/vite dev` makes it faster.

Cavaets are worth noting.  First, loading with typical-production speeds won't uncover edge cases as frequently.  (Testing the UI in development without frontend service worker caching resembles a user loading a study that hasn't been accessed by anyone else since the most recent deployment.)  Second, it requires manually [clearing the frontend SW cache](https://github.com/broadinstitute/single_cell_portal_core/pull/1565#user-content-how-to-clear) upon updating backend data.  

That said, such caching makes testing and development iteration _much_ faster for cell filtering.

PR #1915, opened 2023-10-26
Speed up, improve scalability of cell filtering initialization (SCP-5389)

eweitz:
This improves time-to-interactive for cell filtering, makes the Explore UI snappier, and enables filtering bigger datasets.

### Overview
This often decreases memory usage by > 25% and speeds up processing time by 130-150% when initializing filters.  These scale and speed improvements are achieved through a few micro-optimizations, and one larger optimization.

**Decrease memory usage by > 25%.**  `initCrossfilter` now passes around an array of indexes, rather than structuring those indexes into an object.  This is basically how they're received from the `/facets` API.

**Speed up processing by 30-50%**.  Previously, `trimNullFilters` would always call `getCellsByFacet`.  That's a relatively expensive operation in setup, and not needed for all studies.  Now, calls to `getCellsByFacet` only happen when needed, i.e. we detect a "null filter" -- a filter that has a 0-count by default in a clustering.

**Speed up processing by 100%**.  The initial volley of requests to `/facets` was sent twice upon page load.  That's because an upstream component explicitly set state twice, as part of a long-standing unrelated optimization.  Now, that double-state-set doesn't cause a duplicate `/facets` fetch volley.  This cuts in half the time to fully initialize facets, which otherwise contributes to noticeable UI latency in very large datasets.

This also improves observability by newly logging metrics about the initialization of cell faceting, and refining filter-cell logs.

### Test
1.  Open Network panel in DevTools
2.  Go to "Human milk - differential expression" study.  If on staging, that's [SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303).
3.  In DevTools Network panel, filter requests to "facets".  Confirm only 5 requests were sent to `/facets`, not 10.  (If local with service worker cache, go to DevTools Console, change "Default levels" menu to also show "Verbose" output, and enter "facets" into Console "Filter" box", and confirm 5 not 10 such messages are logged.)
4. Change "Clustering" from "All Cells UMAP" to "Epithelial Cells Subclusters"
5. Click "Filter plotted cells"
6. Confirm no "Cell type" facet is shown
7. Scroll down, confirm "Epithelial Cells Subclusters" facet is shown, with all filters selected
8. Deselect a filter, confirm plot updates as expected
9. Select the filter, confirm plot updates as expected

This satisfies SCP-5389.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -230,8 +230,19 @@ export default function ExploreDisplayTabs({
   const isCorrelatedScatter = enabledTabs.includes('correlatedScatter')
 
 
-  // if the exploreParams update need to reset the initial cell facets
+  // If clustering or annotation changes, then update facets shown for cell filtering
   useEffect(() => {
+    if (!exploreInfo) {return}
+    if (exploreInfo.skipFetchFacets) {
+      // The loadStudyData in ExploreView updates exploreParams _twice_ upon
+      // loading the page.  To avoid doubling requests to the `/facets` API
+      // endpoint, this special `skipFetchFacets` prop is set to true in the
+      // 2nd upstream update.  This block _skips_ that 2nd volley of /facets
+      // requests triggered by that pageload-time double state update
+      // in loadStudyData.
+      exploreInfo.skipFetchFacets = false
+      return
+    }
     const [newCluster, newAnnot] = getSelectedClusterAndAnnot(exploreInfo, exploreParams)
     const setterFunctions = [
       setCellFilteringSelection,
--- a/app/javascript/lib/cell-faceting.js
+++ b/app/javascript/lib/cell-faceting.js
@@ -292,7 +292,7 @@ function initCrossfilter(facetData) {
     // An array of integers, e.g. [6, 0, 7, 0, 0]
     // Each element in the array is the index-offset of the cell's group value assignment
     // for the annotation facet at that index.
-    const facetIndex = new Uint8Array(cells[i])
+    const facetIndex = cells[i]
     filterableCell.facetIndex = facetIndex
     filterableCells.push(filterableCell)
   }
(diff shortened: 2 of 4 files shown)


jlchang (approved):
Manual review - works as described.

Can't wait to see these performance improvements in production. Also looking forward to using the new logging metrics to see how the filterPlottedCells functionality performs for portal users.

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -230,8 +230,19 @@ export default function ExploreDisplayTabs({
   const isCorrelatedScatter = enabledTabs.includes('correlatedScatter')
 
 
-  // if the exploreParams update need to reset the initial cell facets
+  // If clustering or annotation changes, then update facets shown for cell filtering
   useEffect(() => {
+    if (!exploreInfo) {return}
+    if (exploreInfo.skipFetchFacets) {
+      // The loadStudyData in ExploreView updates exploreParams _twice_ upon
```
bistline, line 237: I think this highlights a lurking issue that we have throughout the explore UX - we tie a lot of processing/rendering off of `useEffect` calls that look for changes to `exploreParams`.  I wonder if consolidation of these calls into a more centralized function would help.

app/javascript/lib/cell-faceting.js
```diff
@@ -276,7 +294,7 @@ function getCellsByFacet(filterableCells, annotationFacets) {
   const cellsByFacet = {}
   for (let i = 0; i < annotationFacets.length; i++) {
     const facet = annotationFacets[i]
-    const facetCrossfilter = cellCrossfilter.dimension(d => d[facet])
+    const facetCrossfilter = cellCrossfilter.dimension(d => d.facetIndex[i])
```
bistline, line 297: FWIW this is still the line where I get the "pre-crash" pause in the DevTools debugger when looking at very large studies in the cell filtering UX.

bistline (approved):
Looks good - I confirm the duplicate requests have been eliminated.

app/javascript/lib/cell-faceting.js
eweitz: Thanks, this is useful to know.  The containing function here is still probably being called too frequently.  So although this function is a lot lighter, it's still not light enough to deal with many quick filter clicks in our biggest 1% of studies.

PR #1920, opened 2023-10-31
Enable hierarchical cell filtering, and reset all filters (SCP-5386)

eweitz:
This adds easy high-level filtering, and a way to reset filters.  Common tasks that were often 25+ clicks are now 2 clicks.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/cb9ba195-6634-4a49-86c3-6fa15ab9941b

### Overview
Previously, if you wanted to deselect all filters in a facet, you needed to click every filter.  That can be a lot!  This kind of task is common if you want to filter _additively_.  For example, if the "Cell type" (or "Donor ID" or "Biosample ID") facet has 26 filters and you _only_ want to see 1 cell type, then you'd need to click 25 times to deselect all the other cell types.  

Now, that task is only 2 clicks.  Click the new facet-level checkbox beside any facet label to deselect _all_ filters in the facet, then click the 1 cell type you want to see.  This optimizes the interaction cost from O(n + 1) to O(2) -- substantial savings in human time and working memory.

You can now also click an intuitive icon atop the filter section to reset all filters to their default state.

### Test
A new automated test verifies some of the new behavior.  To manually test:

1.  Go to "Human milk - differential expression" study ([SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/human-milk-differential-expression) if on staging)
2. Click "Filter plotted cells"
3. Scroll down to "Donor ID" facet
4. Hover over facet, click checkbox to left of "Donor ID"
5. Confirm all "Donor ID" filters are deselected, and plot shows only filtered cells in grey
6. Click grey undo icon at right of "Filter by" section header
7. Confirm all "Donor ID" filters are selected, and plot shows no filtered cells in grey

This satisfies SCP-5386.

PR #1922, opened 2023-11-03
Sort cell filters by count, enable sorting by label (SCP-5373)

eweitz:
This shows more relevant filters at a glance, through useful ordering.  It also lets researchers better group related filters.

### Overview
Previously (#1920), filters were ordered unintuitively, as determined by their incidental order in original data and unrelated SCP processing.  That confused a beta tester.

Now, filters are ordered more usefully.  Filters are sorted by cell count, by default.  This makes it effortless to discern the _distribution_ of the various categories that filters represent.  The physical size difference of more digits helps visualize spreads.  Having the most numerous category at top also makes it easiest to create a big visual difference in the plot, because the group with the most cells -- sometimes _a lot_ more cells than the others -- becomes most easily hidden.

Users can also easily change the default sort-by-count to sort-by-label.  This can display related filters together, e.g. in the biosample ID facet, where A) cell counts are often fairly evenly distributed across filters and B) there is often _internal structure_ in the filter labels ("BM03_13wk_r1", "BM03_13wkpp_r2192", etc.).  Once you see these related filters side-by-side, you see how their counts change as you update filters in other facets.  That uncovers meaningful patterns.

### Video

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/b3e9507c-692f-49f4-9915-b80d1a38e30e

### Possible work future
In the future, it might be worth refining a few things:
* The "Reset filters" button could reset not just filter selections, but also filter sorting.  I briefly tried this, and it ended up being more complex than I anticipated, so this seems sensible to defer.

* The sort-by-label implementation is sometimes awkward.  The rudimentary [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order) is often only natural in a technical sense; in the more relevant and common sense, it sometimes orders things unnaturally.  This is particularly apparent in biosample IDs for SCP1671.  Likewise, annotations are sometimes Boolean-esque with values like "yes", "no", "NA".  These cases could probably be easily inferred and cast with more useful orderings as seen in publications, to avoid less conventional orders like "no", "NA", "yes".

### Test
A new automated test verifies this new functionality.  To manually test:

1.  Go to "Human milk - differential expression" study
2. Click "Filter plotted cells"
3. Click "More..." in the "Cell type" facet
4. Confirm filters are ordered by count
5. Click the sort icon at top right in facet
6. Confirm filters are ordered by label

This satisfies SCP-5373.

### DIFF
--- a/app/controllers/api/v1/visualization/annotations_controller.rb
+++ b/app/controllers/api/v1/visualization/annotations_controller.rb
@@ -277,7 +277,8 @@ module Api
             study_file_id = scope == 'study' ? @study.metadata_file.id : cluster.study_file_id
             array_query = {
               name: annotation[:name], array_type: 'annotations', linear_data_type: data_obj.class.name,
-              linear_data_id: data_obj.id, study_id: @study.id, study_file_id:
+              linear_data_id: data_obj.id, study_id: @study.id, study_file_id:, subsample_annotation: nil,
+              subsample_threshold: nil
             }
             annotation_arrays[identifier] = DataArray.concatenate_arrays(array_query)
             facets << { annotation: identifier, groups: annotation[:values] }
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -80,15 +80,14 @@ function parseAnnotationName(annotationIdentifier) {
 }
 
 /** UI control to update how filters are sorted */
-function SortFiltersIcon({ facet, handleSort, sortKey }) {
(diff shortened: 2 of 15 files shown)


jlchang (approved):
Functional review - works great!

Approving and mentioning a UI papercut, for filters with long names in narrow browser window settings, the filter name and the counts can collide:
Screenshot 2023-11-06 at 4 37 06 PM

For timing on addressing the UI papercut, I defer to you. If not resolved in this PR, please make a follow-up ticket because not being able to read count info makes the filtering panel less helpful.

bistline (approved):

PR #1923, opened 2023-11-07
Fix DE table columns (SCP-5402)

eweitz:
This fixes a bug in which the table for SCP-computed DE and one-vs-rest author DE reported values for the wrong metric.  

The values are usually clearly awry upon basic scientific inspection, but it doesn't break the UI, so users would likely suspect something is broken with SCP computation even though the problem is confined to front-end parsing.

### Test
Two new automated tests protect against regressions.  To manually test:

1.  Go to any study with SCP-computed DE
2. Open Network panel in Chrome DevTools
3. Click "Differential expression"
4. Select a group
5. Note the first few genes in DE table, and their data for log2(FC) and adj. p-value.
6. In Network panel, filter to "googleapis", and find the file from the bucket
7. Confirm the same genes appear atop the values, and that their data logfoldchanges and pvals_adj match the DE table

This satisfies SCP-5402.

### DIFF
--- a/app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
+++ b/app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
@@ -35,8 +35,6 @@ export function parseDeFile(tsvText, isAuthorDe=false) {
 
     if (tsvLine === '') {continue}
     const row = tsvLine.split('\t')
-    console.log('row')
-    console.log(row)
 
     let deGene
 
@@ -253,7 +251,7 @@ export function PairwiseDifferentialExpressionGroupPicker({
 /** Pick groups of cells for one-vs-rest-only differential expression (DE) */
 export function OneVsRestDifferentialExpressionGroupPicker({
   bucketId, clusterName, annotation, deGenes, deGroup, setDeGroup, setDeGenes,
-  countsByLabel, deObjects, setDeFilePath, isAuthorDe, sizeMetric, significanceMetric
+  countsByLabel, deObjects, setDeFilePath, isAuthorDe
 }) {
   let groups = getLegendSortedLabels(countsByLabel)
   groups = groups.filter(group => {
@@ -272,7 +270,7 @@ export function OneVsRestDifferentialExpressionGroupPicker({
 
     setDeFilePath(deFilePath)
 
-    const deGenes = await fetchDeGenes(bucketId, deFilePath)
+    const deGenes = await fetchDeGenes(bucketId, deFilePath, isAuthorDe)
 
     setDeGroup(newGroup)
     setDeGenes(deGenes)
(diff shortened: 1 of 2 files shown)


eweitz:
The [build failure](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/6789293482/job/18456146317#step:5:4069) is a false positive, as a Rails test passed but not Rails code changed.

```
Total elapsed time: 25 minutes, 43 seconds

### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  SearchFacetTest#test_should_find_all_filter_matches:
  NoMethodError: undefined method `[]' for nil:NilClass
Exiting with code: 1
```

That false-positive seems caused by [degraded HCA Azul reliability](https://broadinstitute.slack.com/archives/CBEHTH601/p1699017232383669).

app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
```diff
@@ -36,37 +36,55 @@ function parseDeFile(tsvText, isAuthorDe=false) {
...
+        // TODO (SCP-5201): Show significant zeros, e.g. 0's to right of 9 in 0.900
+        name,
+        size,
+        significance
+      }
+    } else {
+      // names  scores  logfoldchanges  pvals pvals_adj pct_nz_group  pct_nz_reference
+      const [
+        index, // eslint-disable-line
+        name, score, size, altSignificance,	significance
```
bistline, line 71: It looks like there's a stray tab character here.

bistline (approved):

bistline:
> That false-positive seems caused by [degraded reliability in HCA Azul](https://broadinstitute.slack.com/archives/CBEHTH601/p1699017232383669).

That would track - we're sourcing some of that info from Azul, so it could cause failures.

app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
eweitz: Thanks, tidied in 36a7cc6.

jlchang (approved):
Functional review - results now match DE data in bucket.

Thanks for catching this bug!

eweitz:
The [latest build failed](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/6790325446/job/18459480234?pr=1923#step:5:4029), also a false positive as it's a Rails test failure and no Rails code changed.  However, the immediate cause differs.  For easier diagnostics, here it is:

```
Total elapsed time: 26 minutes, 56 seconds

### FAILURES and ERRORS ###
*** Yarn test failures ***

*** Rails test failures ***

  StudySyncServiceTest#test_should_get_files_in_batches_and_give_counts:
  Mongoid::Errors::Validations: 
Exiting with code: 1
```

PR #1925, opened 2023-11-09
Adding ImportService prototype (SCP-5339)

bistline:
#### BACKGROUND
As part of our work on importing data from the NeMO Identifiers API, we need a service that can programmatically source external data and create the necessary study & file entries in the database with minimal interaction from a user or admin.  This will remove the need for curators to upload files and fill out forms through the SCP UI as this data is publicly available via an API.

#### CHANGES
This update adds the prototype `ImportService`, as well as the `ImportServiceConfig` module and the `ImportServiceConfig::Nemo` and `ImportServiceConfig::Hca` classes.  The architectural overview is as follows:

`ImportService`
* Handles API calls using embedded clients in configuration classes
* Top-level error handling and reporting
* Pulls remote files from external sources and pushes to workspace buckets
* Calls `FileParseService` to ingest new files

`ImportServiceConfig`
* Module that contains common methods for accessing remote services, including getters and setters
* Handles populating and saving `Study` and `StudyFile` entries by querying APIs and traversing associations, where appropriate

`ImportServiceConfig::Nemo`, `ImportServiceConfig::Hca`
* Contain service-specific logic for accessing data, such as attribute/term conversions and association chains
* Hold reference to underlying service API clients to pass along to `ImportService`

Some notes/caveats:

We do not have exemplar datasets yet for either service, so some features are not fully implemented:
* Calling ingest pipeline (currently commented out)
* Mixpanel reporting for study creation
* File downloads can point to external services, but there is additional work to hook this in (currently saving in `study_file.external_link_url` but this is not hooked into actual downloads yet)
* Setting `ann_data_file_info.data_fragments` for extracting cluster embeddings from `AnnData` files
* There is now a mechanism to prevent importing duplicates for studies/files  - `:external_identifier` - but this is not enforced yet (not required until Q1 2024)
* We are waiting on ontology terms from NeMO, so there may be future work on term harmonization

#### MANUAL TESTING
1. Enter a Rails console session
2. Create a NeMO configuration:
```
file_id = 'nemo:alc-t6a5pxv'
user_id = User.find_by(email: <your email>).id
branding_group_id = BrandingGroup.first.id
config = ImportServiceConfig::Nemo.new(file_id:, user_id:, branding_group_id:)
```
3. Confirm you can load the specified file (this is a small BAM index in a GCP bucket, which makes the copy much quicker later):
```
config.load_file
=> 
{"id"=>"nemo:alc-t6a5pxv",
 "file_name"=>"BI006_marm028_Munchkin_M1_rxn1.4.bam.bai",
 "data_type"=>"align",
 ...
```
4. Populate other data by loading associations:
```
config.traverse_associations!
=> 
{"file_id"=>"nemo:alc-t6a5pxv'",           
 "study_id"=>"nemo:col-f3yvj88",          
 "project_id"=>"nemo:grn-avkxj1y",
...  
```
5. Import the data and create a new study & file (this should take ~20-30s to complete):
```
study, study_file = config.import_from_service
...

Importing Nemo study: A Molecular and Cellular Atlas of the Marmoset Brain, 10x v3 from nemo:col-f3yvj88
Importing Nemo file: BI006_marm028_Munchkin_M1_rxn1.4.bam.bai from nemo:alc-t6a5pxv
=> [#<Study _id: 654d62ce94ec8f0694a74c47, created_at: 2023-11-09 22:53:30.138787 UTC, updated_at: 2023-11-09 22:53:30.138787 UTC, user...
```
6. Confirm that `external_identifier` has been set and that the file is in the bucket by checking for a generation tag (yours will be different):
```
study.external_identifier
=> "nemo:col-hwmwd2x"

study_file.external_identifier
=> "nemo:der-ah1o5qb"

study_file.generation
=> "1699569281122025"
```
7. Clean up study & file
```
study_file.destroy
study.destroy_and_remove_workspace
```


jlchang (approved):
Thanks for the architectural overview at demo last week.

FYI, I tried to view the created study in browser saw it listed in the correct collection but got an error trying to load the study.
`undefined method `html_safe' for nil:NilClass`
```
<div id="study-description-container">     
  <div id="study-description-content" class="ck-content">       
    <%= @study.full_description.html_safe %>
  </div>
</div>
```
Looks like `@study.full_description` was not set and I wasn't sure if that needs to be addressed in this PR.

bistline:
> Thanks for the architectural overview at demo last week.
> 
> Functional review - works as described.
> 
> FYI, I tried to view the created study in browser saw it listed in the correct collection but got an error trying to load the study. `undefined method `html_safe' for nil:NilClass`
> 
> ```
> <div id="study-description-container">     
>   <div id="study-description-content" class="ck-content">       
>     <%= @study.full_description.html_safe %>
>   </div>
> </div>
> ```
> 
> Looks like `@study.full_description` was not set and I wasn't sure if that needs to be addressed in this PR.

Ah yes - great catch!  That's something that is handled at the controller level.  I can change that so it always happens regardless of how the study is created.

app/models/import_service_config.rb
```diff
@@ -0,0 +1,220 @@
...
+    'tar' => 'application/x-gtar',
+    'h5' => 'application/x-hdf',
+    'h5ad' => 'application/x-hdf'
+  }.freeze
+
+  ALLOWED_SCP_MODELS = [Study, StudyFile].freeze
+
+  attr_accessor :client, :user_id, :file_id, :study_id, :branding_group_id
+
+  # name of importing service (e.g. NeMO, Hca)
```
eweitz, line 20: For clarity, this seems like it should be either the conventional name of the upstream service (NeMO, HCA), or the class name of the service (Nemo, Hca), but not a mix.

```suggestion
  # name of importing service (e.g. Nemo, Hca)
```

app/models/import_service_config.rb
```diff
@@ -0,0 +1,220 @@
...
+    study
+  end
+
+  def to_study_file(study_id, taxon_common_name, taxon_attribute: :common_name, format_attribute: :file_format)
+    file_info = load_file.with_indifferent_access
+    study_file = to_scp_model(StudyFile, study_file_default_settings, study_file_mappings, file_info)
+    # assign study, taxon, and content_type
+    study_file.study_id = study_id
+    study_file.taxon_id = taxon_from(taxon_common_name, attribute: taxon_attribute)&.id
+    ext = file_info[format_attribute].gsub(/^\./, '') # trim leading period, if present
```
eweitz, line 130: Out of curiosity, do you recall encountering any compound extensions (e.g. ".tar.gz") in NeMO or HCA?

eweitz (approved):
Code looks good!

app/models/import_service_config.rb
bistline: I've seen `.h5ad.tar` in NeMO, but not `.tar.gz`.  HCA I haven't seen any yet with compound names, but I haven't looked super hard for any.

app/models/import_service_config.rb
bistline: I agree - this has more to do with how Ruby maps class names to file names.  If we went with `NeMO`, the name of the file would actually be `ne_mo.rb`.  Since this is only used for logging, I think it's fine just just have a hard-coded return for `service_name` that maintains the correct formatting.

PR #1926, opened 2023-11-13
Fix cell filtering when data has "--Unspecified--" group (SCP-5398)

eweitz:
This fixes a rare case of overselection in cell filtering, and a problem in unreleased code merged last week.

### Overview
Previously, filtering would occasionally remove more cells than were in the deselected group.  It turns out this only occurs if a clustering had an annotation with cells that were _not_ assigned to any group by the author.  Such cells are typically marked with the special label "--Unspecified--" at request-time.

Now, filtering in a cluster that has an "--Unspecified--" group returns the expected number of cells.  Although the diagnosis was complex, the fix simply required refining how the ungrouped cells are detected and explicitly marked as unspecified.

These changes also fix a previously unreported bug in changes from last week (#1922).  While fixing the main issue, I found that filtering then searching a gene broke the Explore UI.  That now works as expected.

### Video
Here's how the bug and its fix look on copies of [SCP2407](https://singlecell.broadinstitute.org/single_cell/study/SCP2407/single-cell-atlas-of-the-liver-myeloid-compartment-before-and-after-cure-of-chronic-viral-hepatitis), a public study that motivated this work.

#### Bug
https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/bac360e1-4756-431f-ac69-a4b3df6b4689

#### Fix
https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/2463b851-f9fe-4cea-8326-574b1db5f7a4

### Test
An automated test was updated to prevent regression in "--Unspecified--" handling.  To manually test:

* If you're on staging and this PR is [deployed](https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-deploy-hotfix-to-staging/) there, go to [SCP374](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP374/hepatitis-and-unspecified-scp2407-on-staging) ("Hepatitis and --Unspecified--: SCP2407 on staging") with filtering enabled
* Click "Filter plotted cells" button
* Note the counts in the default filters for the "Celltype initial" facet, e.g. 3124 for "CD14+ Monocyte"
* Deselect the "CD14+ Monocyte" checkbox
* Confirm the "--Filtered--" scatter plot legend entry reports 3124 cells (not e.g. 5643)
* Select the "CD14+ Monocyte" checkbox
* Confirm the "--Filtered--" legend entry disappears

I've also added a new [manual smoke test for cell filtering](https://docs.google.com/document/d/1rKcOQPtkbhZCbTqxTnaT-4E29VLzkJrm0kAoWjLx3QA/edit#heading=h.o66s4o4su847) to the release engineering process.  That'll help ensure any breaks in the happy path for cell filtering don't reach production.

This satisfies SCP-5398.  It may well also fix SCP-5390, another bug related to "--Unspecified--".

### DIFF
--- a/app/controllers/api/v1/visualization/annotations_controller.rb
+++ b/app/controllers/api/v1/visualization/annotations_controller.rb
@@ -296,6 +296,9 @@ module Api
               else
                 label = annotation_arrays[annotation][index] || '--Unspecified--'
               end
+              if label == ''
+                label = '--Unspecified--'
+              end
               facet[:groups].index(label)
             end
           end
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -253,7 +253,7 @@ function CellFacet({
 
   const [sortKey, setSortKey] = useState('count')
 
-  const unsortedFilters = facet.unsortedGroups
+  const unsortedFilters = facet.unsortedGroups ?? []
   let filters
 
   //  Naturally sort groups (see https://en.wikipedia.org/wiki/Natural_sort_order)
@@ -264,7 +264,7 @@ function CellFacet({
   } else {
     // Sort categorical filters (i.e., groups)
     const filterCounts = facet.originalFilterCounts
-    const sortedGroups = facet.unsortedGroups.sort((a, b) => {
+    const sortedGroups = unsortedFilters.sort((a, b) => {
       if (filterCounts[a] && filterCounts[b]) {
         return filterCounts[b] - filterCounts[a]
       }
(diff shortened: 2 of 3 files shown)


bistline (approved):
Looks good - nice work on the diagnosis.

jlchang (approved):
Functional test: works as described.

I notice that a filter of `--Unspecified--` now shows up in the filtering panel for facets like `Celltype subcluster` and `HCV RNA IU ml SVRX`. It seems useful, especially for facets like `HCV RNA IU ml SVR24` which only have >1 choice if `--Unspecified--` is taken into account. It is a little odd to see `--Unspecified--` as the majority tag for `Celltype subcluster` <- that caught my eye as unexpected but in thinking on it further may be more useful than confusing.

eweitz:
Thanks for the insightful review!  I agree -- seeing the special value is striking in the rare applicable cases, but useful.

PR #1929, opened 2023-11-15
Prevent loading very large collections for admins (SCP-5356)

bistline:
#### BACKGROUND & CHANGES
This update now paginates very large collections on the "My annotations" and "Admin control panel".  Both these pages load all annotations/users when viewed by admins, which leads to extremely slow rendering.  The latter is an admin-only page, but is used frequently to schedule deployments banners or to disable a user's daily download quota.  The "My annotations" is low-traffic (~80 views/month), but suffers from the same issue when loading user annotations.  Additionally, a search feature has been added to both pages to replace the datatables "filter" box.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to the Admin control panel
3. Click the "User roles" tab, and enter your email in the search box
4. Click 'Find' and confirm that it loads your account
5. To test the pagination, change `per_page` to `1` on `admin_configurations_controller.rb#41`
6. Reload the admin control panel, and confirm you now see pagination controls in the User tab
7. Click 'My annotations' from the profile menu, and confirm the same functionality works there as well (you can edit `per_page` on `user_annotations_controller.rb#24`


jlchang (commented):

Thank you so much for this quality of (admin) life improvement!

jlchang (approved):

PR #1930, opened 2023-11-15
Using GitHub Actions to build portal Docker image (SCP-4944)

bistline:
UPDATE: This can't be merged until we either implement deployments through Github Actions, or we can figure out a trigger for Jenkins to watch for the completion of these builds.  Otherwise, Jenkins will deploy to staging before the requested tag has been built.

#### BACKGROUND
Starting next May, DevOps will no longer maintain any Jenkins servers in DSP.  Since we use Jenkins for our CD pipeline (building Docker images, running deployments), we will need to migrate these tasks over to Github Actions, or we will have to maintain our Jenkins server ourselves.

#### CHANGES
This update reimplements the tasks for building the portal Docker image as a single workflow, and can be run from any Github-hosted runner (not just self-hosted).  Since these runners do not have `gcloud` installed or configured, this is now done via the `google-cloud-cli:latest` Docker image.  This is how the `build_image.sh` script determines what tags are available, and runs checks/cleanups before and after building and publishing the image.  In addition, since the `runner` user does not have publish permissions in our GCR repository, we can use the `gcloud` container to generate an access token to authenticate into GCR via `docker login`.

It will be possible to trigger this build manually like we do in Jenkins, but this will not be enabled until this feature branch has been merged into the default `development` branch.  Afterwards, it can be triggered against any desired branch.  Running against `master` will set the tag to the most recent release.  Selecting any other branch will tag the image with the name of that branch.

Additionally, we will need to migrate to Artifact Registry by May 15, 2024.  This work will be compatible with that planned migration - the only updates needed will be in the `docker push` and any `gcloud container` calls.  The rest of this workflow should still function correctly after the migration.

NOTE: before merging this PR, we should disable the corresponding Jenkins jobs:
Development tag: https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-build-portal-docker-image-from-development/
Release tag: https://scp-jenkins.dsp-techops.broadinstitute.org/job/scp-build-portal-docker-image-from-master/

#### MANUAL TESTING
Since this relies on pushes to target branches, manual testing is not recommended as it may clobber changes in GCR.  Instead, please see an instance of a successful run [here](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/6869970091), as well as evidence of a successful preemptive check when [an existing tag was found](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/6879663548/job/18712093386#step:6:16).

### DIFF
--- a/.github/workflows/build-and-publish-development-image.yml
+++ b/.github/workflows/build-and-publish-development-image.yml
@@ -36,7 +36,7 @@ jobs:
           # this will create a Docker volume called 'gcloud-config' that contains credentials for use with gcloud
           export VAULT_TOKEN=$( vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID )
           vault read -format=json $VAULT_SECRET_PATH | jq .data > $SERVICE_ACCOUNT_PATH
-          docker run -v "$SERVICE_ACCOUNT_PATH:$SERVICE_ACCOUNT_PATH" -it --name gcloud-config $GCLOUD_DOCKER_IMAGE gcloud \
+          docker run -v "$SERVICE_ACCOUNT_PATH:$SERVICE_ACCOUNT_PATH" --name gcloud-config $GCLOUD_DOCKER_IMAGE gcloud \
                  auth activate-service-account --key-file=$SERVICE_ACCOUNT_PATH
       - name: Build development image
         env:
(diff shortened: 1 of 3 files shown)


bistline:
Last failed Docker build was testing prevention of clobbering a tag other than `development`, so failure is the correct outcome.

eweitz (approved):
Thanks for getting SCP ahead of this big infrastructural shift.  Code looks good, and the detailed PR description conveys things well.

jlchang (approved):
I get the general gist, it'll have to do.

bistline:
Closing temporarily while working on other task migrations to avoid clogging up CI queue.

PR #1931, opened 2023-11-16
Add sparkbars to show proportion of cells filtered (SCP-5406)

eweitz:
This adds tiny bar charts to each filter.  These reveal how groups _proportionally_ change upon cell filtering.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/acd09f75-360b-41bb-a3c6-9f8bc9002aee

### Overview
Previously, filter counts reported how many cells were highlighted in the plot, and these counts changed as the filter selection was updated.  However, it was unclear how many cells in each group were _not_ highlighted in the plot, i.e. how many in each group were plotted in faint grey as part of the general "--Filtered--" legend entry.  

This meant it was hard for researchers to know what _proportion_ of e.g. diseased cells -- or cells from a given cell type, or cells from that replicate -- were in the filtered result.

Now, small bar charts below each filter make such insights easy, and available at a glance.  These "sparkbars" apply the technique of [sparklines](https://en.wikipedia.org/wiki/Sparkline): miniaturizing a visualization so that the graphics tightly integrate with and complement text.  

The sparkbars come in two types:

* Not all passed: these bars have soft blue and grey segments.  Each segment has a width that represents the proportion of cells that are highlighted in the plot in blue, and cells that are _not_ highlighted in the plot in grey.  A new tooltip reports the exact counts and percentages represented by each segment.  Cells highlighted in the plot are labeled "Passed" in blue, and cells _not_ highlighted in the plot are labeled "Filtered" in grey.  The "Filtered" label and faint grey color aligns with the "--Filtered--" legend entry and how these cells are faintly plotted in grey.  The term "Passed" is common in filtering contexts, from [signal processing](https://en.wikipedia.org/wiki/Passband) to [genomics](https://www.google.com/search?q=filtered+reads+pass).  Applying the term here reuses a general concept that researchers are familiar with, in a relevant new context.

* All passed: these soft green bars indicate when all cells in a group are among those highlighted in the plot, i.e. when their baseline count matches their filtered count.

### Test
1.  Go to the "Human milk - differential expression" study locally, or [SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/human-milk-differential-expression) on staging
2. Click the "Filter plotted cells" button
3. Deselect the "BM03" checkbox in the "Donor ID" facet
4. Confirm that other rows in the "Donor ID" facet have a green bar below the count, meaning all cells in that group passed filtering
[...]

PR #1935, opened 2023-11-21
Fix filter-then-search bug in sparkbar tooltips (SCP-5406)

eweitz:
This fixes a bug in which filtering then searching a gene broke the Explore UI.  

This bug in #1931 was caught before release, while doing manual tests as part of release engineering.

### Test
1. Go to an initialized study
2. Apply a cell filter
3. Search a gene
4. Confirm page does not show a large error message

You might notice an issue in filter counts upon step 3.  It turns out that's pre-existing; see [context in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1700585748862709?thread_ts=1700585742.808839&cid=CBEHTH601).

This relates to SCP-5406.

### DIFF
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -209,12 +209,13 @@ function BaselineSparkbar({ baselineCount, passedCount }) {
 
   const maxWidthPx = `${maxWidth}px`
   const passedWidthPx = `${selectedWidth}px`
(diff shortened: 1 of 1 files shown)


bistline (approved):

jlchang (approved):

eweitz:
My local `yarn-test` and `yarn ui-test` both pass.  So I'll merge before this build passes, and we'll check the release build.

PR #1938, opened 2023-11-27
Skip frequently failing tests in smoke test (SCP-5405)

bistline:
#### BACKGROUND & CHANGES
This update adds 3 tests to the list of skipped tests for the Orchestration smoke test that runs 3x weekly in an effort to increase stability and improve signal on these builds while not reducing global coverage. The reasoning behind this change is that the referenced tests fail often for non-reproducible reasons in the staging instance of Sam, and don't cover "critical user journey" codepaths.  These tests will still run on normal CI builds.

#### MANUAL TESTING
Since this is not a smoke test build, you should be able to look at the output of the CI run below and see that the following tests ran: `FireCloudClientTest#test_create_and_mange_user_groups`, `FireCloudClientTest#test_update_billing_project_members`, and `FireCloudClientTest#test_get_methods`.

### DIFF
--- a/test/integration/external/fire_cloud_client_test.rb
+++ b/test/integration/external/fire_cloud_client_test.rb
@@ -262,6 +262,7 @@ class FireCloudClientTest < ActiveSupport::TestCase
 
   # get available workflows
   def test_get_methods
+    skip if @smoke_test
     # get all available methods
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Nice signal-to-noise improvement!  Thanks for describing further at standup.

jlchang (approved):

PR #1939, opened 2023-11-27
Make Viridis default color profile for expression scatter plots (SCP-3545)

bistline:
#### BACKGROUND & CHANGES
This update changes the default expression color profile to `Viridis` from `Reds`.  We had previously done [research](https://docs.google.com/document/d/1U9ikJOxe4Fb9fieX3Rse-591xGD_KMZP5uxAml8YA3w/edit#heading=h.vgsws97lyzyo) to support this change, and had never followed through on switching it.  Since the release of the cell filtering feature, this has become more pertinent as "filtered" cells are now difficult to distinguish from cells that have low or no observed expression when using single-color profiles.

#### MANUAL TESTING
1. Boot as normal and load any study with expression data
2. Query for a gene, and confirm that the default profile is `Viridis`
3. Select `Reds` from the color profile, and confirm it is still flipped with red being the maximum value

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,7 @@
 module ApplicationHelper
 
+  DEFAULT_COLOR_PROFILE = 'Viridis'.freeze
+
   # overriding link_to to preserve branding_group params
   def scp_link_to(name, url, html_options={}, &block)
     if url.start_with?('https') || url.start_with?('/')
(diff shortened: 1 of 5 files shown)


eweitz (approved):
Nice -- a timely and useful change.

jlchang (approved):
Woo hoo! Looking forward to this :)

PR #1942, opened 2023-11-29
Enable links to cell filtering; fix legend-filter interaction (SCP-5395, SCP-5417)

eweitz:
This URL-parameterizes cell filtering, to let users easily share and reproduce their findings.  It also fixes a legend-filter bug.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/c8be0fe1-ee58-498f-8560-750bff5c4927

And here's how it looked before.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/08ef7c48-271c-453f-ad73-d6680e9b6e9e

Previously, users could not preserve their cell filtering state.  We have frontend infrastructure to register state in the URL, and route from URL parameters to application state, but it simply wasn't configured for cell filtering.  

Now, users can get easily back to where they were in filtering, or send their colleagues or other audience a URL that instantly reproduces their cell filtering results.  This initial implementation for sharing and bookmarking is crude but effective.  Later refinements could include e.g. A) opening the cell filtering panel on page load if `facets` is specified, B) indicating filters are being applied after the initial plot loads and before filters are applied, and C) using ontology IDs instead of labels when possible.

Incidentally, URL parameterization fixes a bug that caused filters (and their counts) to reset when a legend entry was clicked.

### Test
* Apply a filter, or multiple
* Refresh page
* Confirm filters get applied to plot automatically
* Click legend entry while filters are applied
* Confirm filters don't get cleared

This satisfies SCP-5395.

### DIFF
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -209,13 +209,12 @@ function BaselineSparkbar({ baselineCount, passedCount }) {
 
   const maxWidthPx = `${maxWidth}px`
   const passedWidthPx = `${selectedWidth}px`
-  const leftPx = -1 * `${passedWidthPx}px`
 
   const fullClass = baselineCount === passedCount ? ' full' : ''
   const baseTop = passedCount === 0 ? '0' : '2px'
 
   const passedStyle = { width: passedWidthPx }
-  const filteredStyle = { width: maxWidthPx, left: leftPx, top: baseTop }
+  const filteredStyle = { width: maxWidthPx, top: baseTop }
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


bistline (approved):
Works as advertised.  I agree that there are potential refinements to be made.  I think the biggest will be some sort of visual indication that the filtering is being reapplied - at first I thought it wasn't working, but then the UI eventually caught up and the state restored.

jlchang (approved):
There's a pre-existing caching issue that had me befuddled for a bit but i should be relatively rare. Worth checking on in staging perhaps.

Everything look good for this first-pass at sharable bookmarks.

eweitz:
> I think the biggest will be some sort of visual indication that the filtering is being reapplied - at first I thought it wasn't working, but then the UI eventually caught up and the state restored.

I've opened SCP-5434 to track that and related refinements.

> There's a pre-existing caching issue

Thanks for walking me through this.  Based on my understanding, I think SCP-5355 might mostly solve that.

PR #1944, opened 2023-11-30
Updating to new ToS endpoint, fixes to support testing (SCP-5298)

bistline:
#### BACKGROUND & CHANGES
This update uses the new Sam endpoints for sourcing user Terra terms of service information.  The `/register` endpoint still contains the `tosAccepted` flag, but this information is apparently stale (at best) or completely incorrect.  The new endpoint contains information including acceptance date, version, and a boolean flag detailing whether or not the user is permitted system access (`permitsSystemUsage`).  This is the new flag that SCP will check for terms of service status.

Note: service accounts are exempt from this new ToS infrastructure, so attempting to check the ToS status for a service account will result in a `404`.

This also fixes a bug where overriding the default Terra billing project did not work from the constructor.  Now, calling `FireCloudClient.new(project: <value>)` will correctly set the value.

#### MANUAL TESTING
1. Pull this branch and enter a Rails console session
2. Load your user account:
```
user = User.find_by(email: <email address>)
```
3. Confirm you can get your terms of service status from the helper:
```
user.must_accept_terra_tos?
=> false
```
4. Use the API endpoint to get more detailed information:
```
client = FireCloudClient.new(user:)
=> 
#<FireCloudClient:0x0000000113df6650                                       
...

client.get_terms_of_service
=> {"acceptedOn"=>"1970-01-01T00:00:00Z", "isCurrentVersion"=>true, "latestAcceptedVersion"=>"1", "permitsSystemUsage"=>true}
```

### DIFF
--- a/app/models/fire_cloud_client.rb
+++ b/app/models/fire_cloud_client.rb
@@ -88,7 +88,7 @@ class FireCloudClient
   #
   # * *return*
   #   - +FireCloudClient+ object
-  def initialize(user: nil, project: nil, service_account: self.class.get_primary_keyfile, api_root: BASE_URL)
+  def initialize(user: nil, project: PORTAL_NAMESPACE, service_account: self.class.get_primary_keyfile, api_root: BASE_URL)
     # when initializing without a user, default to base configuration
     if user.nil?
       # instantiate Google Cloud Storage driver to work with files in workspace buckets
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good!  My manual tests pass.

Like your test output, mine also output `"acceptedOn"=>"1970-01-01T00:00:00Z"`.  Using [Unix epoch time](https://en.wikipedia.org/wiki/Unix_time) for what seems like it ought to be `null` is a bit unintuitive.  But it's an upstream artifact, unlikely to have impact, and presumably derives from Sam doing the JVM equivalent of [`new Date(null)`](https://stackoverflow.com/questions/49339980/why-new-datenull-is-returning-this-date-1970-01-01t000000-000z).

jlchang (approved):
Manual tests behave as described.

PR #1950, opened 2023-12-05
Refine UX for cell filter linking (SCP-5434)

eweitz:
This more cleanly and clearly loads the Explore UI for cell filtering links -- helping orient researchers to subsetted views.

Previously (#1942), cell filters applied upon page load were reflected in scatter plots, but only after an _unfiltered_ plot was displayed.  Also, the cell filtering panel wasn't shown, so the user would need to click "Filter plotted cells" to see which filters had been applied.  These rough edges made the cell filter link loading UX a bit disorienting.

Now when Explore loads with any cell filters specified, the scatter plots aren't displayed _until filters are applied_ to the cells, and the cell filtering panel shows by default.  This makes the UX smoother, and the filtered plots easier to understand.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/820683fc-9ecb-40ef-82c6-fdea6d949b82

### Test
1.  Apply filters
2. Reload page
3. Confirm initial scatter plot reflects applied filters
4. Confirm cell filtering panel is shown

For more context, see the [SCP demo recording](https://drive.google.com/file/d/1qG5TLahIXXfeiOt7C-H9mu_edtfahbTk/view?t=7s).  This satisfies SCP-5434.

### DIFF
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -376,7 +376,6 @@ function CellFacet({
     setIsFullyCollapsed(isAllListsCollapsed)
   }, [isAllListsCollapsed])
 
-
   let facetStyle = {}
   if (!facet.isLoaded) {
     facetStyle = {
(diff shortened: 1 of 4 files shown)


bistline (approved):

jlchang (approved):
Functional review - changes work as described.

Happy to see an MVP get some polish - we haven't been able to!

PR #1952, opened 2023-12-06
Skip extraneous check for 200 from home page on deployments (SCP-5447)

bistline:
#### BACKGROUND & CHANGES
This removes an extraneous check after a successful deployment where upon receiving a `200` response from the home page, the `remote_deploy.sh` script then checks again for the same status.  While rare, this occasionally leads to "failed" deployments when that extra check doesn't return `200`.  It is unclear why this secondary check fails - whether it is a race condition, or residual "unhealthy" status from the load balancer not fully clearing - but in all cases the deployment does in fact succeed, but the job is marked as a failure nonetheless.  Now, the HTTP status is stored in a variable, and upon successfully receiving a `200`, the checks will cease and the while loop exits.  Deployments that do not receive a `200` response inside 6 minutes are still marked as failures.

Examples of recent "failed" deployments:
- staging: [Deploy to staging SCP instance · broadinstitute/single_cell_portal_core@a82e2f2](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/7091702112/job/19301410539) 
- production: [Deploy to production SCP instance · broadinstitute/single_cell_portal_core@84eb97a](https://github.com/broadinstitute/single_cell_portal_core/actions/runs/7118126717/job/19380347135#step:8:245)

#### MANUAL TESTING
WIP: I'm going to try deploying this feature branch to staging a few times to ensure that it still deploys correctly and I don't see the failure.  It's difficult to prove a negative, so I'm instead going to rely on the fact that the remote deployment still works.

### DIFF
--- a/bin/remote_deploy.sh
+++ b/bin/remote_deploy.sh
@@ -14,7 +14,7 @@ function main {
     VERSION_TAG="development"
     echo "### USER: $(whoami) ###"
 
-    while getopts "v:h" OPTION; do
+    while getopts "v:" OPTION; do
       case $OPTION in
         v)
           VERSION_TAG="$OPTARG"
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Thanks for fixing the "false failed deployment" situation, Jon!

PR #1955, opened 2023-12-11
Extending `ImportService` (SCP-5409)

bistline:
#### BACKGROUND & CHANGEs
This extends work started in #1925 to include the following:
* Automatic file ingests via `FileParseService`
* Support for downloading files directly from external services
* Setting default or custom `obsm_key_names` for AnnData cluster extraction
* Prevent duplication of studies/files created via `ImportService`
* Human-readable `imported_from` tag for finding all studies/files imported from a given service

Note: Currently, `ImportService` will only work with HCA due to the lack of parseable files in NeMO.

#### MANUAL TESTING
1. Start DelayedJob and enter a Rails console session
2. Set up parameters for job (using `Spatial and single-cell transcriptional landscape of human cerebellar development` and the AnnData contributor matrix `aldinger20.processed.h5ad`) 
```
study_id = "85a9263b-0887-48ed-ab1a-ddfa773727b6"
file_id = 'b0517500-b39e-4c7a-b2f0-794ddc725433'
user_id = User.find_by(email: '<your email>').id
branding_group_id = BrandingGroup.first.id
params = { study_id:, file_id:, user_id:, branding_group_id: }
```
3. Call `ImportService` to create study/file and call ingest:
```
ImportService.import_from(ImportServiceConfig::Hca, **params)
...
Importing HCA study: Spatial and single-cell transcriptional landscape of human cerebellar development from 85a9263b-0887-48ed-ab1a-ddfa773727b6
Ingesting file: aldinger20.processed.h5ad (b0517500-b39e-4c7a-b2f0-794ddc725433) from imported study SCP130 (85a9263b-0887-48ed-ab1a-ddfa773727b6)
=> [#<Study _id: 65773cf094ec8f692d0a3f68, created_at: 2023-12-11 16:47:05.7065 UTC, updated_at: 2023-12-11 16:47:05.7065 UTC, user_id:  ...
```
4. Validate that the entities created correctly and that the file is parsing (should take ~15 minutes to ingest all data, though clustering/metadata is available quickly)
```
study = Study.find_by(external_identifier: params[:study_id])
=> #<Study _id: 65773cf094ec8f692d0a3f68, created_at: 2023-12-11 16:47:05.706 UTC, updated_at: 2023-12-11 16:47:06.846 UTC, user_id: BS...

study.imported_from
=> "HCA"

study_file = StudyFile.find_by(external_identifier: params[:file_id])
=> #<StudyFile _id: 65773d2994ec8f692d0a3f6d, created_at: 2023-12-11 16:47:38.22 UTC, updated_at: 2023-12-11 16:47:38.562 UTC, upload_f...

study_file.imported_from
=> "HCA"

study_file.download_path
=> "https://service.azul.data.humancellatlas.org/repository/files/b0517500-b39e-4c7a-b2f0-794ddc725433?catalog=dcp33&version=2021-11-15T11%3A23%3A19.351000Z"

study_file.parsing?
=> true
```

### DIFF
--- a/app/lib/import_service.rb
+++ b/app/lib/import_service.rb
@@ -44,7 +44,6 @@ class ImportService
 
     begin
       study, study_file = configuration.import_from_service
-      # TODO: uncomment this block after file parsing is enabled for NeMO and SCP-5400 is complete
       # extra work will be required but is unknown until we have the dataset (e.g. populating AnnData data_fragments)
       identifier = "#{study.accession} (#{study.external_identifier})"
       log_message "Ingesting file: #{study_file.upload_file_name} (#{study_file.external_identifier}) from imported study #{identifier}"
--- a/app/lib/import_service.rb
+++ b/app/lib/import_service.rb
@@ -52,12 +52,9 @@ class ImportService
       [study, study_file]
     rescue RuntimeError, RestClient::Exception, Google::Apis::ClientError => e
       log_message("Error importing from #{config_class}: #{e.class} - #{e.message}", level: :error)
-      ErrorTracker.report_exception(e, configuration.user)
-      study = Study.find_by(external_identifier: configuration.study_id)
-      study_file = StudyFile.find_by(external_identifier: configuration.file_id)
-      [study, study_file].compact.each do |instance|
-        DeleteQueueJob.new(instance).delay.perform
-      end
+      ErrorTracker.report_exception(e, configuration.user, configuration)
+      # don't run cleanup as it potentially could delete an existing study/file
+      # this is handled in ImportServiceConfig
       nil
     end
   end
@@ -145,6 +142,9 @@ class ImportService
   # * *params*
   #   - +study+ (Study) => study from which to remove Terra workspace
   def self.remove_study_workspace(study)
+    # first check if a study already exists using this external_identifer to prevent accidental workspace deletion
+    return false if Study.where(external_identifier: study.external_identifier, :id.ne => study.id).exists?
+
     if ApplicationController.firecloud_client.workspace_exists?(study.firecloud_project, study.firecloud_workspace)
       ApplicationController.firecloud_client.delete_workspace(study.firecloud_project, study.firecloud_workspace)
     end
--- a/app/lib/import_service.rb
+++ b/app/lib/import_service.rb
@@ -1,7 +1,7 @@
 # service for loading data into SCP from external APIs, such as the NeMO Identifiers API or HCA Azul service
 class ImportService
   extend ServiceAccountManager
-  include Loggable
+  extend Loggable
 
   # API clients that can use ImportService
   ALLOWED_CLIENTS = [NemoClient, HcaAzulClient].freeze
@@ -38,6 +38,10 @@ class ImportService
     raise "unsupported config: #{config_class}" unless defined?(config_class) && ALLOWED_CONFIGS.include?(config_class)
 
     configuration = config_class.new(...)
+    unless configuration.valid?
+      raise configuration.errors.full_messages.join(', ')
+    end
+
     begin
       study, study_file = configuration.import_from_service
       # TODO: uncomment this block after file parsing is enabled for NeMO and SCP-5400 is complete
@@ -48,7 +52,7 @@ class ImportService
       [study, study_file]
     rescue RuntimeError, RestClient::NotFound, Google::Apis::ClientError => e
       log_message("Error importing from #{config_class}: #{e.class} - #{e.message}", level: :error)
-      ErrorTracker.report_exception(e, config.user)
+      ErrorTracker.report_exception(e, configuration.user)
       nil
     end
   end
--- a/app/lib/import_service.rb
+++ b/app/lib/import_service.rb
@@ -47,7 +47,7 @@ class ImportService
       # TODO: uncomment this block after file parsing is enabled for NeMO and SCP-5400 is complete
       # extra work will be required but is unknown until we have the dataset (e.g. populating AnnData data_fragments)
       identifier = "#{study.accession} (#{study.external_identifier})"
-      log_message "Ingesting file: #{study_file.upload_file_name} from imported study #{identifier}"
+      log_message "Ingesting file: #{study_file.upload_file_name} (#{study_file.external_identifier}) from imported study #{identifier}"
       FileParseService.run_parse_job(study_file, study, study.user)
       [study, study_file]
     rescue RuntimeError, RestClient::Exception, Google::Apis::ClientError => e
--- a/app/lib/import_service.rb
+++ b/app/lib/import_service.rb
@@ -50,9 +50,14 @@ class ImportService
       log_message "Ingesting file: #{study_file.upload_file_name} from imported study #{identifier}"
       FileParseService.run_parse_job(study_file, study, study.user)
       [study, study_file]
-    rescue RuntimeError, RestClient::NotFound, Google::Apis::ClientError => e
+    rescue RuntimeError, RestClient::Exception, Google::Apis::ClientError => e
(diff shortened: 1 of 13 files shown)


jlchang (commented):
Functional review - study created as described.

Questions:
- If the datasets are consistently processed to have the same metadata content, it'd be great to set the appropriate default annotation when the study is created. Do we expect to have a configuration file of "defaults" for each collaborating group that would hold a value for `default annotation`?
- If I go to upload wizard, should I expect Species to have a value?
- If I go to study details, when does the parse status for the h5ad file become a green checkmark, when the first (anndata --extract) completes? or when the last of the fragment files has completed parsing? (The latter would seem more accurate, especially for large h5ad files where fragment ingest might take some time)
- When we run the job, we specify the "user" who becomes the study owner. Thinking ahead, do we want these studies to be owned by a collaboration-specific email address that the collaborating group (eg. HCA or NeMO) has access to? Or the email of an individual within the collaboration?

app/models/import_service_config.rb
```diff
@@ -15,7 +16,9 @@ module ImportServiceConfig
 
   ALLOWED_SCP_MODELS = [Study, StudyFile].freeze
 
-  attr_accessor :client, :user_id, :file_id, :study_id, :branding_group_id
+  DEFAULT_OBSM_KEYS = %w[X_umap].freeze
```
jlchang, line 19: the default key in ingest_pipeline.py is X_tsne. This default would just override that one but it might be clearer to have the same default in both places. I'll make a ticket to update ingest_pipeline.py to X_umap.

app/models/import_service_config/nemo.rb
```diff
@@ -109,6 +111,9 @@ def populate_study_file(scp_study_id)
         library = load_study&.[]('technique')
       end
       study_file.expression_file_info.library_preparation_protocol = find_library_prep(library)
+      http_url = file_access_info(protocol: :http)&.[]('url')
```
jlchang, line 114: In the study created from an HCA h5ad called via API, when I went to the upload wizard, library_preparation_protocol was not populated, do we think we'll have an analogous query in app/models/import_service_config/hca.rb that we'll use to autopopulate `study_file.expression_file_info.library_preparation_protocol` in a later PR? And if NeMO uses a different ontology, some amount of mapping would occur here and in ingest?

jlchang (approved):
Great to see water flowing. Now we just need NeMO-flavored water...

bistline:
> Functional review - study created as described.
> 
> Questions:
> 
> * If the datasets are consistently processed to have the same metadata content, it'd be great to set the appropriate default annotation when the study is created. Do we expect to have a configuration file of "defaults" for each collaborating group that would hold a value for `default annotation`?

That's a great suggestion.  We don't have anything for that yet, but it would be a simple add-on.

> * If I go to upload wizard, should I expect Species to have a value?

Yes, and it looks like the info out of `ExpressionFileInfo` isn't getting passed back properly.  I'll look into it.

> * If I go to study details, when does the parse status for the h5ad file become a green checkmark, when the first (anndata --extract) completes? or when the last of the fragment files has completed parsing? (The latter would seem more accurate, especially for large h5ad files where fragment ingest might take some time)

Technically it's after the first extraction job finishes.  We've never really gone back to refine this.  I could see waiting until all the downstream parses are done, but that gets a little messy.  But it probably would be more correct.

> * When we run the job, we specify the "user" who becomes the study owner. Thinking ahead, do we want these studies to be owned by a collaboration-specific email address that the collaborating group (eg. HCA or NeMO) has access to? Or the email of an individual within the collaboration?

Technically, we're using a user account that is also the curator of the collection, since these will all be added to collections.  I can definitely see using a group account, but it still has to be a Google account at the end of the day.  So right now, it's the individual, but having more distributed access could be very useful.

app/models/import_service_config/nemo.rb
bistline: Yeah it looks like there's a bug with this - I'm looking into it.

app/models/import_service_config.rb
bistline: Oh right, I'd forgotten about that.  I don't have a vested interest in it being `X_umap`, but that was what I used in testing for the HCA file.  I didn't even check if it had `X_tsne`.  I agree they should be the same.

app/models/import_service_config/nemo.rb
bistline: @jlchang [c227474](https://github.com/broadinstitute/single_cell_portal_core/pull/1955/commits/c2274748e3beb6052d9fd100ecc7d6af708e89b5)

jlchang (changes requested):
Screenshot 2023-12-19 at 8 29 10 AM. 

Upon re-review, noticed the study name was incorrect and the study workspace was **`scp-development-jlchang/DELETE-1ffda9b82fdb0a6673c79f7f449db2e3d651bef645372f3d80cce8e43f4c895a`**, indicating that some cleanup activity is happening ~~before study creation~~ (correction, the cleanup process will rename Terra workspaces so the study owner can create a new study re-using the name they want).  

Also, on initial review I saw a visualizable study but the study now has no data and no files - the Terra workspace is inaccessible to me:  
`You cannot access this workspace because it either does not exist or you do not have access to it.`

Similarly the bucket is gone:   
`BucketNotFoundException: 404 gs://fc-5e088ec6-02be-4c74-907b-39dbd3bb5fe7 bucket does not exist.`

I did receive successful email notifications for Expression, Cluster and Metadata ingest on Dec 13 and don't see any failure emails for "SCP66" (my local ImportService study).

Requesting changes to ensure that `ImportService`-created studies persist ~~and have the correct study name~~ (Update: study name issue is likely due to testing that re-import of an existing HCA study is blocked).

bistline:
> Requesting changes to ensure that `ImportService`-created studies persist ~and have the correct study name~ (Update: study name issue may be due to testing that re-import of an existing HCA study is blocked. Open question - what caused the study to be queued for deletion?).

I can confirm there is a lurking bug here with accidental deletion on any kind of validation fail.  The initial import works correctly, but if someone tries to re-use an `external_identifier` it will correctly prevent using that, but then the cleanup logic will _also_ delete the original, valid study.

bistline:
@jlchang [4f14a90](https://github.com/broadinstitute/single_cell_portal_core/pull/1955/commits/4f14a90f820473d9200a953290dab4e26d92588c)

bistline:
Looking at log output from CI run shows upstream instability led to test failure:
```
FireCloud API request (PATCH) https://api.firecloud.org/api/workspaces/single-cell-portal-test/test_create_and_manage_workspace-b30b4040-35a3-4753-9ae5-c52062464a95/acl?inviteUsersNotFound=true with tracking identifier: 839419950053-compute@developer.gserviceaccount.com
Timed out reading data from server: ;  encountered when requesting 'https://api.firecloud.org/api/workspaces/single-cell-portal-test/test_create_and_manage_workspace-b30b4040-35a3-4753-9ae5-c52062464a95/acl?inviteUsersNotFound=true', attempt #1
```

jlchang (approved):
Re-reviewed and the expression file info is now present - thanks Jon!

eweitz (approved):

app/lib/import_service.rb
```diff
@@ -38,17 +38,26 @@ def self.import_from(config_class, ...)
...
 
     configuration = config_class.new(...)
+    unless configuration.valid?
+      raise configuration.errors.full_messages.join(', ')
+    end
+
     begin
       study, study_file = configuration.import_from_service
       # TODO: uncomment this block after file parsing is enabled for NeMO and SCP-5400 is complete
       # extra work will be required but is unknown until we have the dataset (e.g. populating AnnData data_fragments)
```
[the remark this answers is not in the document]
bistline: Great catch - I'll take that comment out.

PR #1956, opened 2023-12-11
Only read 1 megabyte of log for email delivery (SCP-5454)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where the user-facing error email sometimes will not deliver in the case where the `user_error.log` file is very large, resulting in a `broken pipe` error in the logs.  This is due to the portal attempting to read the entire contents into memory and then adding it to the email body.  Now, this uses the [same logic as the admin error email](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/models/ingest_job.rb#L1036) where only 1 megabyte of the log is read.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -1029,7 +1029,7 @@ class IngestJob
   def generate_error_email_body(email_type: :user)
     case email_type
     when :user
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Thanks for fixing the "email bug", Jon!

PR #1962, opened 2024-01-04
Speed up cell filtering for non-default annotations (SCP-5355)

eweitz:
This uses the plot data cache for cell filtering in non-default annotations, speeding up a UX flow and decreasing server load.

Previously, applying any cell filter in an eligible non-default annotation would send a request to the `/clusters` SCP API endpoint.  Such requests were always server-side file cached -- as that `/cluster` endpoint would have been previously requested upon changing annotations.  So while it was quite fast, a loading icon was still shown, and the network transfer plus JSON parse times made things feel slower than snappy.  And while serving essentially static files from the server cache caused negligible extra load, it's still good to reduce that superfluous work.

Now, the `/cluster` response is retrieved from the client-side, in-memory plot data cache, even for non-default annotations.  That makes cell filtering faster!

### Test
Automated tests were updated to account for this change.  To manually test:
1.  Go to a study with multiple annotations eligible for filtering
2. Switch to a non-default annotation
3. Apply a filter
4. Confirm filter applies, but no SCP API request is sent

This satisfies SCP-5355.


jlchang (approved):
Functional review - confirmed additional requests are not sent to the /clusters endpoint

bistline (approved):

PR #1963, opened 2024-01-04
Bypass file upload by supplying GS URL or bucket path (SCP-5392)

bistline:
#### BACKGROUND
Currently, if a user doesn't want to upload a file directly through the SCP UI, they can use the `sync` feature.  This allows them to register files in SCP that have already been pushed to the workspace bucket.  However, there are certain file types and actions that are not supported via `sync`, most notably parsing AnnData files, or author-generated differential expression outputs.  To provide these files, the user is forced to upload them through the UI, which is time-consuming and prone to errors for very large files (as AnnData files tend to be).

#### CHANGES
This update now gives users the ability to specify a path (or GS URL) to a file that already exists in the workspace bucket directly through the upload wizard and bypass the upload process completely.  This can be used for _any_ file form, including the "AnnData UX".  The form will accept either a relative path to the file from the bucket root, or a fully-qualified GS URL.  This value is then used to access the remote file and validate that it exists before saving the new `StudyFile` record.  Additionally, some helper features have been added to browse the study bucket (for copying paths or URLs), and to link to our [documentation](https://singlecell.zendesk.com/hc/en-us/articles/360061006011) on how to upload large files directly to buckets. 

Additionally, client-side file validation (CSFV) is still enabled for uploads using this feature.  This was achieved by adapting the same paradigm used in `sync` to read the remote file.  If the file fails validation, the normal error message UX shows, and the save button is disabled.  Example:

Screenshot 2024-01-10 at 12 20 32 PM

In the case where the file cannot be found - either it does not exist, or if the user does not have permission to view the file client-side (will be true in private studies for non-Terra users) - instead of showing an error, a warning will be issued noting that the file cannot be accessed, and will be validated completely on save.  This allows this feature to be used by the large cross section of private study, non-Terra users, and mimics the same functionality of very large uploads in CSFV, where validation is either too expensive, or impossible.  Example message:

Screenshot 2024-01-10 at 12 20 12 PM

The user can still save the file, and normal server-side validations will enforce, which include validating that the file exists in the bucket.  If the file still cannot be found, the save will fail and the following message displayed to the user:

Screenshot 2024-01-10 at 2 51 47 PM

Demo of the entire UX, including documentation/bucket links and CSFV:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/f4d40ed8-846c-42a4-aa8a-7c426827c2d2

Lastly, if a user deletes a file through the UI that used this feature, the file will not be removed from the bucket.  This mirrors the same functionality in `sync`, and prevents removing potentially large files that failed ingest for non-deterministic reasons.  The following message is displayed instead:

new_delete_confirm

#### MANUAL TESTING
2. Go to the upload wizard for any study (use `Classic` mode for new studies)
3. Go to the Clustering tab, and click the `Use bucket path` button next to the upload control
4. Click the `Browse bucket` link, then add the files `test/test_data/cluster_example.txt` and `test/test_data/cluster_bad.txt` to the bucket
5. Back in the upload wizard, supply the value `cluster_bad.txt` in the field for the bucket path
6. Confirm you see the same error message as shown above, and that the file save is disabled
7. Next, provide the name `no_file.txt`, and confirm that the warning `Unable to access the requested file` is displayed
8. Give the cluster a name and then click `Save`, confirming that the save fails with `no file found at no_file.txt` error
9. Now specify the fully qualified GS URL for `cluster_example.txt` like shown in the demo and then click `Save`, confirming that the action succeeds, and there is now a link to download the file
10. Once the ingest completes, click the `Delete` button and confirm you see the same dialog as above, then delete the file
11. Back in the bucket window, refresh the page and confirm the file is still present
12. (Optional) Try the same with any other file type and confirm you are able to validate & save files once they have been added to the bucket

### DIFF
--- a/app/javascript/components/upload/ExpandableFileForm.jsx
+++ b/app/javascript/components/upload/ExpandableFileForm.jsx
@@ -6,7 +6,6 @@ import { Popover, OverlayTrigger } from 'react-bootstrap'
 import LoadingSpinner from '~/lib/LoadingSpinner'
 import FileUploadControl from './FileUploadControl'
 import Button from 'react-bootstrap/lib/Button'
-import { FormatDeleteConfirmation } from '~/components/upload/form-components'
 
 /** renders its children inside an expandable form with a header for file selection */
 export default function ExpandableFileForm({
@@ -136,7 +135,12 @@ export function SaveDeleteButtons({
     </Button>
   }
 
-  const deleteText = FormatDeleteConfirmation(file)
+  let deleteText
+  if (file?.remote_location) {
+    deleteText = 'will remain in the bucket because you provided a remote path to an existing file.'
+  } else {
+    deleteText = 'will be removed from the bucket.'
+  }
 
   return <div className="flexbox-align-center button-panel">
     <SaveButton
--- a/app/javascript/components/upload/ExpandableFileForm.jsx
+++ b/app/javascript/components/upload/ExpandableFileForm.jsx
@@ -6,6 +6,7 @@ import { Popover, OverlayTrigger } from 'react-bootstrap'
 import LoadingSpinner from '~/lib/LoadingSpinner'
 import FileUploadControl from './FileUploadControl'
 import Button from 'react-bootstrap/lib/Button'
+import { FormatDeleteConfirmation } from '~/components/upload/form-components'
 
 /** renders its children inside an expandable form with a header for file selection */
 export default function ExpandableFileForm({
@@ -135,12 +136,7 @@ export function SaveDeleteButtons({
     </Button>
   }
 
-  let deleteText
-  if (file.remote_location) {
-    deleteText = ' remain in the bucket because you provided a remote path to an existing file.'
-  } else {
-    deleteText = ' be removed from the bucket.'
-  }
+  const deleteText = FormatDeleteConfirmation(file)
 
   return <div className="flexbox-align-center button-panel">
     <SaveButton
--- a/app/javascript/components/upload/ExpandableFileForm.jsx
+++ b/app/javascript/components/upload/ExpandableFileForm.jsx
@@ -135,6 +135,12 @@ export function SaveDeleteButtons({
     </Button>
   }
 
+  let deleteText
+  if (file.remote_location) {
+    deleteText = ' remain in the bucket because you provided a remote path to an existing file.'
+  } else {
+    deleteText = ' be removed from the bucket.'
+  }
 
   return <div className="flexbox-align-center button-panel">
     <SaveButton
@@ -147,8 +153,8 @@ export function SaveDeleteButtons({
       onHide={() => setShowConfirmDeleteModal(false)}
       animation={false}>
       <Modal.Body className="">
-        Are you sure you want to delete {file.name}?<br />
-        <span>The file will be removed from the workspace and all corresponding database records deleted.</span>
+        Are you sure you want to delete {file.name}?<br /><br />
+        <span>All corresponding database records will be deleted. The file will <strong>{deleteText}</strong></span>
       </Modal.Body>
       <Modal.Footer>
         <button className="btn btn-md btn-primary" onClick={() => {
--- a/app/javascript/components/upload/FileUploadControl.jsx
+++ b/app/javascript/components/upload/FileUploadControl.jsx
@@ -7,6 +7,9 @@ import { StudyContext } from '~/components/upload/upload-utils'
 import ValidateFile from '~/lib/validation/validate-file'
 import ValidationMessage from '~/components/validation/ValidationMessage'
 import { TextFormField } from './form-components'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons'
+import { Popover, OverlayTrigger } from 'react-bootstrap'
 
 // File types which let the user set a custom name for the file in the UX
 const FILE_TYPES_ALLOWING_SET_NAME = ['Cluster', 'Gene List']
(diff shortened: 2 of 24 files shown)


jlchang:
Tried a "happy path" test to upload a misc file and got an unexpected error (see screenshot)
Screenshot 2024-01-08 at 8 36 15 AM

bistline:
> Tried a "happy path" test to upload a misc file and got an unexpected error (see screenshot) Screenshot 2024-01-08 at 8 36 15 AM I get the same error even if I supply just "sampled_genes.tsv" as the path to file in GCP bucket.

Ah I think this is because we're not setting some required values now client-side.  I'll look into this.

app/javascript/components/upload/form-components.jsx
```diff
@@ -90,6 +98,13 @@ export function SaveDeleteButtons({ file, updateFile, saveFile, deleteFile, vali
     </button>
   }
 
+  let deleteText
+  if (file.remote_location) {
+    deleteText = ' will remain in the bucket because you provided a remote path to an existing file.'
+  } else {
+    deleteText = ' will be removed from the bucket.'
```
eweitz, line 105:
```suggestion
    deleteText = 'remain in the bucket because you provided a remote path to an existing file.'
  } else {
    deleteText = 'be removed from the bucket.'
```

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -135,6 +135,12 @@ export function SaveDeleteButtons({
     </Button>
   }
 
+  let deleteText
+  if (file.remote_location) {
+    deleteText = ' remain in the bucket because you provided a remote path to an existing file.'
+  } else {
+    deleteText = ' be removed from the bucket.'
```
eweitz, line 142: No space needed:

```suggestion
    deleteText = 'remain in the bucket because you provided a remote path to an existing file.'
  } else {
    deleteText = 'be removed from the bucket.'
```

app/javascript/components/upload/form-components.jsx
```diff
@@ -100,8 +115,8 @@ export function SaveDeleteButtons({ file, updateFile, saveFile, deleteFile, vali
       onHide={() => setShowConfirmDeleteModal(false)}
       animation={false}>
       <Modal.Body className="">
-        Are you sure you want to delete { file.name }?<br/>
-        <span>The file will be removed from the workspace and all corresponding database records deleted.</span>
+        Are you sure you want to delete {file.name}?<br /><br />
+        <span>All corresponding database records will be deleted. The file <strong>{deleteText}</strong></span>
```
eweitz, line 119: Aligning with the simpler duplicate of this in `ExpandableFileForm.jsx`:

```suggestion
        <span>All corresponding database records will be deleted. The file will <strong>{deleteText}</strong></span>
```

app/javascript/components/upload/ExpandableFileForm.jsx
```diff
@@ -147,8 +153,8 @@ export function SaveDeleteButtons({
       onHide={() => setShowConfirmDeleteModal(false)}
       animation={false}>
       <Modal.Body className="">
-        Are you sure you want to delete {file.name}?<br />
-        <span>The file will be removed from the workspace and all corresponding database records deleted.</span>
+        Are you sure you want to delete {file.name}?<br /><br />
+        <span>All corresponding database records will be deleted. The file will <strong>{deleteText}</strong></span>
```
eweitz, line 157: Non-blocking, but the fact that we:

* needed to change this in 2 places (here and in `form-components.jsx`), and
* we observed drift in this PR

suggests that extracting this into one simple component would ensure efficient and consistent future changes.

eweitz (approved):
Code and demo look good!  This makes sync drastically more discoverable.  I look forward to seeing it get closer to feature parity with our existing sync UI, especially client-side file validation.

app/javascript/components/upload/ExpandableFileForm.jsx
bistline: Agreed - I'm not sure why this is in two places but I didn't feel like dealing with it at the time.  Lemme see if I can refactor this without breaking anything.

bistline: @eweitz [e5b0899](https://github.com/broadinstitute/single_cell_portal_core/pull/1963/commits/e5b0899373d71df80fce943c062355fcb3c7bc7e)

app/javascript/components/upload/UploadWizard.jsx
```diff
@@ -135,6 +136,10 @@ export function RawUploadWizard({ studyAccession, name }) {
     })
   }
 
+  if (!window.SCP.readOnlyToken) {
+    fetchReadOnlyToken(studyAccession)
+  }
```
eweitz, line 141: See comment beginning "Do callers of this need".  I suspect we'd want to call `setupRenewalForReadOnlyToken` (with the slight extension noted there) unconditionally here.

```suggestion
  setupRenewalForReadOnlyToken(studyAccession)
```

eweitz (commented):
Kudos on integrating CSFV!  That'll be a big improvement for usability.

I think this might overlook a common scenario with read-only access token renewal.  See comments below.

app/javascript/lib/scp-api.jsx
```diff
@@ -235,6 +235,20 @@ export async function fetchStudyFileInfo(studyAccession, includeOptions=true, mo
   return response
 }
 
+/**
+ * Returns read-only service account access token value & object and places in window.SCP object
+ * for use in fetching data from GCS buckets client-side
+ *
+ * @param studyAccession Study accession, e.g. SCP123
+ */
+export async function fetchReadOnlyToken(studyAccession) {
```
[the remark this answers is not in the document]
bistline: That is an excellent point - and yes, I would agree.  We should set up the renewal process, and changing the name makes sense.

As an aside, I think we should look at this entire token renewal chain as we have [some evidence](https://broadworkbench.atlassian.net/browse/SCP-5294) that we are renewing these tokens WAY too often.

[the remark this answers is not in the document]
bistline: Update - I've close SCP-5294 as I've been unable to prove that there actually is a quota for issuing OAuth tokens.  There's quotas on request rates, and issuing OAuth _clients_, but not _tokens_.  As such, I don't think there's a problem anymore with our reissuing the read-only service account token every ~10s or so.

app/javascript/components/upload/UploadWizard.jsx
bistline: [a6ff9d9](https://github.com/broadinstitute/single_cell_portal_core/pull/1963/commits/a6ff9d91c1e2f85f97f3c250c3d107772e62262e)

app/javascript/lib/scp-api.jsx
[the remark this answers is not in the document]
bistline: [a6ff9d9](https://github.com/broadinstitute/single_cell_portal_core/pull/1963/commits/a6ff9d91c1e2f85f97f3c250c3d107772e62262e)

app/javascript/components/upload/FileUploadControl.jsx
```diff
@@ -48,11 +79,43 @@ export default function FileUploadControl({
...
+      if (issues.errors.length === 0) {
+        updateFile(file._id, {remote_location: trimmedPath})
+      } else {
+        // prevent saving if there are errors
+        updateFile(file._id, {remote_location: ''})
+      }
+    }).catch( () => {
+      // just catch error and allow user to proceed - validation will be handled server-side or in ingest
+      setFileValidation({ validating: false, issues: {}, fileName: trimmedPath })
+    })
```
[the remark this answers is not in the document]
bistline: [a6ff9d9](https://github.com/broadinstitute/single_cell_portal_core/pull/1963/commits/a6ff9d91c1e2f85f97f3c250c3d107772e62262e)

bahill (approved):
Approved with the following to be fixed later:

- [SCP-5479](https://broadworkbench.atlassian.net/browse/SCP-5479)
- [SCP-5480](https://broadworkbench.atlassian.net/browse/SCP-5480)

bistline:
> 3ef4a7c

@bahill [3ef4a7c](https://github.com/broadinstitute/single_cell_portal_core/pull/1963/commits/3ef4a7c495616e2d1276e923814b79a78ace9027)

bistline:
Build failure looks like a test-only failure - the `aldinger20.processed.h5ad` matrix is apparently no longer available via [Azul](https://explore.data.humancellatlas.org/projects/85a9263b-0887-48ed-ab1a-ddfa773727b6/project-matrices).

PR #1965, opened 2024-01-16
Cell filtering for violin plots (SCP-5275)

eweitz:
This extends cell filtering to violin plots, highlighting how expression distributions change upon subsetting.

### Overview
Previously (#1962), filtering only worked for cells in scatter plots.  That let users see how removing various combinations of cell annotations affected cluster scatter plots or similar gene expression scatter plots.  However, those visualizations are very dense.  So it was often difficult to discern certain important high-level changes by looking at scatter plots.

Now, with cell filtering applied to violin plots, it's much easier to discern how filtering out various cell annotations affects the _distribution_ of a given gene's expression within each group of a cell annotation.  For example, in JUN gene expression for the study "Cellular and transcriptional diversity over the course of human lactation" (SCP1671), filtering out cells with "Infant sick: no" noticeably affects expression distribution in neutrophils and other groups in the "General cell type" cell annotation.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/2ea88ca6-3958-404e-bb22-2aebf7e30993

### Next steps

This is intended as a minimally viable new feature.  Limitations and potential future work include:

1. Traces with no matching cells disappear, like cell filtering in scatter plots did initially.  It might help to keep these.
2. Distribution changes in some groups seem dramatic, but this may well be due to small sample size for the given group.  
3. The spinning icon still appears upon cell filtering.  Ideally, filtering would update things faster, and no loader would appear.  Enhancing the violin plot legend like the scatter plot legend would ameliorate that.
4. Animating distribution shifts upon filtering would aid object continuity, and perception of changes.

### Test
Automated tests have been updated to account for these changes.  To manually test:

1.  Go to study "Cellular and transcriptional diversity over the course of human lactation" ([SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation) on staging)
2. Search for gene "JUN"
3. Click "Distribution" tab
4. Deselect "Infant sick: no"
5. Confirm update occurs in traces showing distributions in each group of this cell annotation

This satisfies SCP-5275.


app/javascript/lib/web-worker.js
```diff
@@ -0,0 +1,78 @@
+/** Web worker that wraps a CPU-intensive function */
+function setViolinCellIndexesWorker() {
```
bistline, line 2: Very cool to see this setup - took me a minute to grok but I _think_ I understand how this works.  Question: are there other things in the UI where this paradigm could be applied to improve UI performance?

test/js/setup-tests.js
```diff
@@ -24,6 +24,10 @@ global.scrollTo = jest.fn()
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder
 
+global.SCP = {
```
bistline, line 27: Can you add a comment as to why this is necessary?  Seeing this sprinkled across the codebase triggers my 'code-smell' warnings...

bistline (approved):
Works as advertised, and pretty slick for an MVP!

Question: the loading icon appearing is apparently due to the expression data being re-fetched each time, similar to how non-default annotations didn't use the browser cache for cell filtering and re-requested the scatter data each time.  This is totally fine as an MVP, but I'm curious how you envision addressing this moving forward.

app/javascript/lib/web-worker.js
eweitz: Potentially!  Web workers enable client-side multithreading.  Generally, I think, that's is most valuable when A) there's a compute-intensive task, B) it takes a while to complete, and C) completing it as fast as possible isn't a high priority for the UX.

* **Client-side file validation for big local files**.  Right now the UI doesn't freeze, but it does sputter; web workers would make it smooth.  In this case, the cost of the slow-down for web workers is probably worth the more complete validation we could do for CSFV.

* **Cell filtering**.  Applying filters to large numbers of cells can sometimes take a while, and that can make the UI less responsive than usual.  Computing filter updates in a non-UI thread would help that.  However, while it would keep the UI responsive, it might actually make filtering itself a bit slower than doing it on the main UI thread.  That's at least what I observed when experimenting with web workers and CPU-intensive work in other context a year or two ago.

test/js/setup-tests.js
eweitz: Good suggestion, done in 52108d8.  Those checks sprinkled through the codebase were previously rather implicit / opaque ways to detect if the code was running in a test.  This makes those checks more explicit and clearer.

eweitz:
> the loading icon appearing is apparently due to the expression data being re-fetched each time, similar to how non-default annotations didn't use the browser cache for cell filtering and re-requested the scatter data each time. This is totally fine as an MVP, but I'm curious how you envision addressing this moving forward.

Nice catch -- I've included this atop a list of future refinements in a new ticket, SCP-5490.

jlchang (approved):
Manual test - can see violin plot updates occur after changes to plot filters. Changes to violin plots seem to show changes that might be expected from scatter plots.

It might be good to have informal scientific review from friends and family seeing as the scatter plot feedback for plot filtering caught some  edge cases before we made the feature GA.

PR #1968, opened 2024-01-17
Changing HCA project to 'Mapping the developing human immune system across organs' (SCP-5481)

bistline:
#### BACKGROUND & CHANGES
This update changes the HCA project referenced in `test/integration/external/import_service_config/hca_test.rb` to `1.3 Million Brain Cells from E18 Mice`.  This is the [reference dataset](https://support.10xgenomics.com/single-cell-gene-expression/datasets/1.3.0/1M_neurons) from 10X.  This was due to the previous contributor matrix file of `aldinger20.processed.h5ad` not being found in `Spatial and single-cell transcriptional landscape of human cerebellar development` for a protracted period of time due to a new catalog release.  The reason for selecting the 10X dataset is that this is the first project that is always indexed when a new catalog is released.  As such, the amount of time that it might be "unavailable" will be very short.  Also, a similar approach of "skipping" tests due to a reindex used in `hca_azul_client_test.rb` has been adopted for `import_service_config/hca_test.rb`.

#### MANUAL TESTING
1. Pull branch and initialize environment
2. Run `bin/rails test test/integration/external/import_service_config/hca_test.rb` and confirm it passes
3. OR refer to the test run in the Checks section below and confirm it passed

### DIFF
--- a/test/integration/external/import_service_config/hca_test.rb
+++ b/test/integration/external/import_service_config/hca_test.rb
@@ -9,14 +9,14 @@ module ImportServiceConfig
       @user_id = @user.id
       @branding_group_id = @branding_group.id
       @attributes = {
-        file_id: 'b0517500-b39e-4c7a-b2f0-794ddc725433',
-        study_id: '85a9263b-0887-48ed-ab1a-ddfa773727b6',
+        file_id: '9f43771e-588c-4cd8-b8f3-3ce81fdea00d',
+        study_id: 'fcaa53cd-ba57-4bfe-af9c-eaa958f95c1a',
         obsm_key_names: %w[X_tsne X_umap],
         user_id: @user.id,
         branding_group_id: @branding_group.id
       }
       @configuration = ImportServiceConfig::Hca.new(**@attributes)
-      @project_name = 'Spatial and single-cell transcriptional landscape of human cerebellar development'
+      @project_name = "Mapping the developing human immune system across organs"
     end
 
     after(:all) do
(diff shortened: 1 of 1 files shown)


jlchang (changes requested):
I'm guessing that the file we test against wasn't available for a while? The changes make sense in that context and the tests pass BUT, today, the original test also passes...

In the HCA data portal, [the original project](https://explore.data.humancellatlas.org/projects/85a9263b-0887-48ed-ab1a-ddfa773727b6) is available and querying the Azul API shows that the file ID for the original test case has not changed. 
Screenshot 2024-01-18 at 9 25 59 AM

Given the restoration of the original test case, is this change still needed and/or do we need a different strategy to address this test when it fails?

Discussed with @bistline in Slack about picking a project "early" in the catalog and perhaps checking with DSP's team that works with HCA DCP to see if a specific project can be consistently placed at the head of new HCA data catalogs.

bistline:
> Discussed with @bistline in Slack about picking a project "early" in the catalog and perhaps checking with DSP's team that works with HCA DCP to see if a specific project can be consistently placed at the head of new HCA data catalogs.

After talking with Jean, we've decided to switch to the `1.3 Million Brain Cells from E18 Mice` project.  This is always the first project returned (checked in both the current catalog and also the original `dcp1` catalog).  Since we only look at DCP catalogs, this seems a safe assumption.

jlchang (approved):
Thanks for seeking out a more robust test case for `hca_test.rb`! Hopefully the combo of "1.3 Million Brain Cells from E18 Mice" and judicious test skipping will do the trick!

PR #1970, opened 2024-01-22
Update UI to suggest gcloud storage instead of gsutil (SCP-5477)

jlchang:
Moving away from `gsutil cp` and using `gcloud storage cp` has the [potential for faster data transfer](https://medium.com/tag-techblog/from-gsutil-to-gcloud-storage-introducing-simplicity-and-improved-performance-4946a1299786) for our study owners, making data upload less time-consuming. This shift from `gsutil` to `gcloud storage` has already been applied to [SCP documentation](https://singlecell.zendesk.com/hc/en-us/articles/360061006011) and this PR implements SCP UI updates to support the transition away from `gsutil cp`.

Most of the changes in this PR are linting whitespace updates ( eg. to change `<br/>` tags to `<br />` etc). 

The user-facing changes are:
* [Update of copy command for bulk download](https://github.com/broadinstitute/single_cell_portal_core/blob/42d24aa082b01c9f6bf1ff2e90273711fc3c6296/app/views/site/_study_download_data.html.erb#L19) in /app/views/site/_study_download_data.html.erb line 19
> To download all files using gcloud CLI tools, run the following command:
> gcloud storage cp -r gs://<%= @study.bucket_id %> [target path]

  
* [Update copy command for sequence file upload](https://github.com/broadinstitute/single_cell_portal_core/blob/42d24aa082b01c9f6bf1ff2e90273711fc3c6296/app/javascript/components/upload/SequenceFileStep.jsx#L54) in /app/javascript/components/upload/SequenceFileStep.jsx line 54
>  If you already have the gcloud CLI installed (Sept 2022 release or later) you can upload files directly using the following command:
> 
> `gcloud storage cp /path/to/files gs://{formState.study.bucket_id}`

  
* [Updated button text](https://github.com/broadinstitute/single_cell_portal_core/blob/42d24aa082b01c9f6bf1ff2e90273711fc3c6296/app/views/studies/sync_study.html.erb#L35) (`gsutil` -> `gcloud CLI`; line 35) for documentation, REMOVED button for survey (lines 39-51 deleted) in sync/app/views/studies/sync_study.html.erb  
[lines 101 - 106](https://github.com/broadinstitute/single_cell_portal_core/blob/42d24aa082b01c9f6bf1ff2e90273711fc3c6296/app/lib/synthetic_study_populator.rb#L101)

  
Non-user-facing changes: 

* Updated synthetic study populator to use gcloud storage functions] in app/lib/synthetic_study_populator.rb 

  
* Updated comment text containing "gsutil" to use "gcloud storage" instead:
/app/lib/study_sync_service.rb 
/image-pipeline/expression-scatter-plots.js
/test/services/study_sync_service_test.rb

This PR supports SCP-5477.

### DIFF
--- a/app/javascript/components/upload/SequenceFileStep.jsx
+++ b/app/javascript/components/upload/SequenceFileStep.jsx
@@ -43,29 +43,29 @@ function SequenceForm({
     <div className="row">
       <div className="col-md-12">
         <div className="form-terra">
-          Primary sequence information, such as BAM, BAM Index, and FASTQ files<br/>
+          Primary sequence information, such as BAM, BAM Index, and FASTQ files<br />
           <p>
-            <b>Non-human Data</b><br/>
+            <b>Non-human Data</b><br />
             If you have a few, small (under 2 GB) non-human sequence files, they can be uploaded here.
             For uploading many or larger files, please refer to the instructions in
-            our <a href="https://singlecell.zendesk.com/hc/en-us/articles/360061006011-Uploading-Large-Files-Using-Gsutil-Tool" target="_blank" rel="noopener noreferrer">documentation</a>.
+            our <a href="https://singlecell.zendesk.com/hc/en-us/articles/360061006011" target="_blank" rel="noopener noreferrer">documentation</a>.
           </p>
           <p>
-            If you already have gsutil installed you can upload files directly using the following command:
+            If you already have the gcloud CLI installed (Sept 2022 or later) you can upload files directly using the following command:
           </p>
           <pre>
-            gsutil -m cp /path/to/files gs://{formState.study.bucket_id}
+            gcloud storage cp /path/to/files gs://{formState.study.bucket_id}
           </pre>
           <p>
-            <b>Primary Human Data</b><br/>
+            <b>Primary Human Data</b><br />
             Primary sequence data derived from humans should be stored in other biological databases and can be linked here
             by selecting &apos;Yes&apos; for &apos;Primary Human Data&apos; and then providing a link in the text field.
           </p>
         </div>
       </div>
     </div>
-    { sequenceFiles.length > 1 && <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_SEQUENCE_FILE}/> }
-    { sequenceFiles.map(file => {
+    {sequenceFiles.length > 1 && <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_SEQUENCE_FILE} />}
+    {sequenceFiles.map(file => {
       const associatedBaiFile = findBundleChildren(file, formState.files)[0]
       return <SequenceFileForm
         key={file.oldId ? file.oldId : file._id}
@@ -79,8 +79,8 @@ function SequenceForm({
         fileMenuOptions={serverState.menu_options}
         associatedBaiFile={associatedBaiFile}
         bucketName={formState.study.bucket_id}
-        isInitiallyExpanded={sequenceFiles.length === 1}/>
+        isInitiallyExpanded={sequenceFiles.length === 1} />
     })}
-    <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_SEQUENCE_FILE}/>
+    <AddFileButton addNewFile={addNewFile} newFileTemplate={DEFAULT_NEW_SEQUENCE_FILE} />
   </div>
 }
--- a/app/javascript/components/upload/SequenceFileStep.jsx
+++ b/app/javascript/components/upload/SequenceFileStep.jsx
@@ -51,7 +51,7 @@ function SequenceForm({
             our <a href="https://singlecell.zendesk.com/hc/en-us/articles/360061006011" target="_blank" rel="noopener noreferrer">documentation</a>.
           </p>
           <p>
-            If you already have the gcloud CLI installed (Sept 2022 or later) you can upload files directly using the following command:
+            If you already have the gcloud CLI installed (Sept 2022 release or later) you can upload files directly using the following command:
           </p>
           <pre>
             gcloud storage cp /path/to/files gs://{formState.study.bucket_id}
--- a/app/javascript/components/upload/SequenceFileStep.jsx
+++ b/app/javascript/components/upload/SequenceFileStep.jsx
@@ -45,7 +45,7 @@ function SequenceForm({
         <div className="form-terra">
           Primary sequence information, such as BAM, BAM Index, and FASTQ files<br />
           <p>
-            <b>Non-human Data</b><br />
+            <b>Non-human data</b><br />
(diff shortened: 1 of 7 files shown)


app/lib/synthetic_study_populator.rb
```diff
@@ -148,7 +147,7 @@ def self.add_files(study, study_config, study_folder, user)
     end
   end
 
-  def self.process_expression_file_params(study, file_info)
+  def self.process_expression_file_params(_study, file_info)
```
eweitz, line 150: `_study` seems unused in this method, so ideal to remove here and in calling code.

```suggestion
  def self.process_expression_file_params(file_info)
```

app/lib/synthetic_study_populator.rb
```diff
@@ -185,52 +184,51 @@ def self.process_mtx_file_params(study, file_info)
...
-        StudyFile.find_by!(study: study, name: cluster_file_name).id.to_s
+        StudyFile.find_by!(study:, name: cluster_file_name).id.to_s
       end
     end
     params
   end
 
   # process species/annotation arguments, return a hash of params suitable for passing to a StudyFile constructor
-  def self.process_genomic_file_params(study, file_info)
+  def self.process_genomic_file_params(_study, file_info)
```
eweitz, line 231: And similarly in callers, since this parameter is unused.

```suggestion
  def self.process_genomic_file_params(file_info)
```

app/javascript/components/upload/SequenceFileStep.jsx
```diff
@@ -43,29 +43,29 @@ function SequenceForm({
...
-            If you already have gsutil installed you can upload files directly using the following command:
+            If you already have the gcloud CLI installed (Sept 2022 release or later) you can upload files directly using the following command:
           </p>
           <pre>
-            gsutil -m cp /path/to/files gs://{formState.study.bucket_id}
+            gcloud storage cp /path/to/files gs://{formState.study.bucket_id}
           </pre>
           <p>
-            <b>Primary Human Data</b><br/>
+            <b>Primary Human Data</b><br />
```
eweitz, line 60: Since it's in the changeset:

```suggestion
            <b>Primary human data</b><br />
```

app/javascript/components/upload/SequenceFileStep.jsx
```diff
@@ -43,29 +43,29 @@ function SequenceForm({
     <div className="row">
       <div className="col-md-12">
         <div className="form-terra">
-          Primary sequence information, such as BAM, BAM Index, and FASTQ files<br/>
+          Primary sequence information, such as BAM, BAM Index, and FASTQ files<br />
           <p>
-            <b>Non-human Data</b><br/>
+            <b>Non-human Data</b><br />
```
eweitz, line 48: Since it's in the changeset:

```suggestion
            <b>Non-human data</b><br />
```

eweitz (approved):
Code looks good!  It'll help to have users adopt this faster (and not deprecated) new approach.

I suggest a few incidental refinements for maintainability and user-facing consistency; none are blocking.

app/lib/synthetic_study_populator.rb
bistline, line 150: Non-blocking: since the `study` parameter is actually unused, rather than prefacing with the `_` I'd simply remove it here and on line 124.  These are the only two places it's referenced.

app/lib/synthetic_study_populator.rb
bistline, line 231: See above comment re: `process_expression_file_params` - same thing here.

app/views/site/_study_download_data.html.erb
```diff
@@ -16,20 +16,20 @@
...
+            <p class="lead">To download all files using gcloud CLI tools, run the following command:</p>
+            <pre>gcloud storage cp -r gs://<%= @study.bucket_id %> [target path]</pre>
               <% if @directories.any? %>
                 <p class="lead">To download all data files in a specific folder, use the following commands:</p>
                 <table class="table table-condensed table-striped">
                   <thead>
                   <tr>
                     <th>Folder</th>
-                    <th>gsutil command</th>
+                    <th>gcloud CLI command</th>
```
bistline, line 27: Non-blocking: this is a little "department of redundancy department" - `CLI` means "command line interface", so I don't think we need the final "command".

bistline (approved):
Looks good!  Some non-blocking style comments.

Somewhat related: I was testing this out locally, and everything works as expected.  However, I unknowingly tried it on a study with an AnnData file and received this error:

```
ERROR: This download was skipped since fast hash calculation tools
are not installed. You can change this by running:
	$ /Library/Developer/CommandLineTools/usr/bin/python3 -m pip install google-crc32c --upgrade --target /Users/bistline/google-cloud-sdk/lib/third_party
You can also modify the "storage/check_hashes" config setting.
```

It gives me a workaround, so I don't consider this blocking at all, but we might consider adding info about this to our documentation (if it doesn't already exist).

jlchang:
> Somewhat related: I was testing this out locally, and everything works as expected. However, I unknowingly tried it on a study with an AnnData file and received this error:
> 
> ```
> ERROR: This download was skipped since fast hash calculation tools
> are not installed. You can change this by running:
> 	$ /Library/Developer/CommandLineTools/usr/bin/python3 -m pip install google-crc32c --upgrade --target /Users/bistline/google-cloud-sdk/lib/third_party
> You can also modify the "storage/check_hashes" config setting.
> ```

Will add info about this to our documentation.

app/views/site/_study_download_data.html.erb
jlchang: b4e8c64a2

app/lib/synthetic_study_populator.rb
jlchang: 5d5213403

app/lib/study_sync_service.rb
```diff
@@ -364,7 +362,7 @@ def self.gzipped?(file)
       first_two_bytes = file.download(range: 0..1, skip_decompress: true)
       first_two_bytes.rewind
       first_two_bytes.read == StudyFile::GZIP_MAGIC_NUMBER
-    rescue => e
+    rescue StandardError => e
```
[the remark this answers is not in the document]
jlchang: Good to know! I'll leave the change in place on the assumption that my linter would do it again ;)

app/javascript/components/upload/SequenceFileStep.jsx
jlchang: 5d5213403

app/javascript/components/upload/SequenceFileStep.jsx
jlchang: 5d5213403

app/lib/synthetic_study_populator.rb
jlchang: 5d5213403

PR #1972, opened 2024-01-23
Cell filtering by numeric annotations: client library (SCP-5385)

eweitz:
This adds a client library for filtering cells by quantitative dimensions.

### Overview
Cell annotations can be categorical or numeric.  Cell filtering currently supports categorical (i.e. group-based) annotations, where users check and uncheck boxes for each group in an annotation.

Now, we can programmatically filter by numeric annotations as well.  This enables users to easily answer important questions, for example "what is this gene's expression in samples collected very soon after this experimental milestone", or "very soon _or_ very late after it".

The technical design of this cell filtering extension enables rich quantitative queries.  You can apply a variety of operators in any filter, and all apply multiple filters in the same numeric facet.

### Technical examples
Here are some ways you can filter with numbers; see "Tests" for more technical context.

#### Simple numeric filters
* `["equals", 1.3]`
* `["not equals", 1.3]`
* `["greater than", 6]`
* `["greater than or equal to", 6]`
* `["less than", 6]`
* `["less than or equal to", 6]`
* `["between", [5, 42]]` -- inclusive, i.e. 5 <= d <= 42
* `["not between", [5, 42]]` -- inclusive, i.e. !(5 <= d <= 42)
 
#### Compound numeric filters
* `[["between", [95, 100]], ["between", [1200, 1300]]` -- (95 <= d <= 100) or (1200 <= d <= 1300)

Compound numeric filters could be used to e.g.:
* Isolate the peaks of multimodal distributions (as in above concrete example)
* Isolate the tails of distributions

### Video
Here's how it looks!  (My next PR will include a UI for numeric filters controls.)

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/188da161-0d5f-4f1b-9ac1-0790a25e6e9f

### Test
A new automated test verifies this library extension, and other tests have been updated accordingly.  

Here's how you can manually test it:
1.  Go to the "Cellular and transcriptional diversity in human lactation" study
2.  Ensure Vite is running
3.  In line 430 of `ExploreDisplayTabs.jsx`, uncomment this line:
```
// window.SCP.updateFilteredCells = updateFilteredCells
```
4.  In line 228 of `cell-faceting.js`, uncomment this line:
```
// console.log('# filtered results', results.length)
```
5.  Open Console panel in DevTools and run the following
```
window.SCP.updateFilteredCells({'time_post_partum_days--numeric--study': [['between', [0, 9]]]}) // 0 <= days <= 9
[...]

PR #1976, opened 2024-02-05
Fix occasional Explore UI break due to cell filtering (SCP-5508)

eweitz:
This fixes a bug where clicking "Filter plotted cells" sometimes breaks the Explore UI.

On Friday morning, I noticed a distinct increase in "Cannot read properties of undefined (reading 'length') in t" [errors in Mixpanel](https://mixpanel.com/s/4pgbDa).  The increase coincided with the [deployment of SCP 1.65.0](https://github.com/broadinstitute/single_cell_portal_core/releases/tag/1.65.0) around 1:00 PM last Wednesday.  So the bug is likely caused by a change to cell filtering code in 1.65.0, i.e. either #1965 or #1972.  I suspect the latter.

### Test
1.  Go to "Cellular and transcriptional diversity over the course of human lactation"
2. In "Clustering", select "Epithelial Cell UMAP"
3. Click "Filter plotted cells"
4. Confirm Explore UI does not replace scatter plots with red error message

I plan to write a regression test in this PR or the next, but given the impact of this bug I think it'd be worth reviewing and deploying today.

While manually testing this I encountered a few odd UI quirks due to overzealous updates related to `useEffect` or standard component state propagation.  I suspect that exists on production and will be fixed by SCP-5490.

This satisfies SCP-5508.

### DIFF
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -595,7 +595,7 @@ export function CellFilteringPanel({
   // Needed to propagate facets from URL to initial checkbox states
   useEffect(() => {
     setCheckedMap(defaultCheckedMap)
-  }, Object.values(defaultCheckedMap).join(','))
+  }, [Object.values(defaultCheckedMap).join(',')])
 
   /** Top header for the "Filter" section, including all-facet controls */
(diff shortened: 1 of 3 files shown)


bistline (approved):

jlchang (approved):
Functional review: confirmed that error no longer occurs.

Thanks so much for catching the issue and putting out a quick fix!

PR #1979, opened 2024-02-08
Adding user bookmarks (SCP-5431)

bistline:
#### BACKGROUND & CHANGES
This update adds the ability for users to bookmark selected views in the Explore tab.  This applies to all Explore-based visualizations that persist state into the URL bar (e.g. scatter, violin, dot plot, differential expression).  Along with a name, users can provide a description to annotate their bookmarks, which can also be updated and deleted from the same form.  Clicking "Save" without prompting a name will pre-fill the current URL path + query string as the name.  Additionally, users can see all their bookmarks in a list and go directly to each visualization - whether in the same study or a different one.

Here's a 1 min video showing the basic functionality (note - this only covers normal scatter plots, but all Explore tab visualizations are supported):

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/8b968c9c-1b03-43fe-bcfb-54e131333108

#### MANUAL TESTING
2. Click the bookmark (star) icon and confirm you are prompted to sign in
3. After signing in, click the bookmark icon again and then save without providing any values
4. Confirm the save succeeds, then click "See bookmarks" to pull up the list, confirming that the new bookmark is listed with the path as the name
5. Close the list and then provide a name/description for this bookmark and save, confirming it succeeds
6. Load different visualization types (violin, dotplot, etc.) and confirm you can bookmark/load then without issue
7. Delete any bookmark, then re-open the bookmarks list and confirm it no longer appears


app/controllers/api/v1/api_docs_controller.rb
```diff
@@ -58,12 +58,12 @@ class ApiDocsController < ActionController::Base
...
         end
         tag do
-          key :name, 'MetadataSchemas'
-          key :description, 'Metadata Convention schema definitions'
+          key :name, 'Bookmarks'
+          key :description, 'Bookmarked/saved views for users'
         end
         tag do
-          key :name, 'Reports'
-          key :description, 'Reports/Portal Stats'
```
eweitz, line 66: Can we keep these endpoints?  

They're a bit buggy right now, but were essential for some [times-series analytics not in Mixpanel](https://docs.google.com/spreadsheets/d/1FSpP2XTrG9FqAqD9X-BHxkCZae9vxZA3cQLow8mn-bk/edit#gid=0).  Not deleting them would make the fix easier.  (If we do delete them, we should probably do it holistically.)

app/controllers/api/v1/api_docs_controller.rb
```diff
@@ -125,20 +126,16 @@ class ApiDocsController < ActionController::Base
           Api::V1::Visualization::ExpressionController,
           Api::V1::Visualization::ExploreController,
           Api::V1::UserAnnotationsController,
-          Api::V1::ReportsController
```
eweitz, line 128: Let's keep this, per above.

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,200 @@
...
+      swagger_path '/bookmarks' do
+        operation :get do
+          key :tags, [
+            'Bookmarks'
+          ]
+          key :summary, 'Get my Bookmarks'
+          key :description, 'Returns all Bookmarks for the given User'
+          key :operationId, 'bookmarks_path'
+          response 200 do
+            key :description, 'Array of Bookmarks objects'
```
eweitz, line 20:
```suggestion
            key :description, 'Array of Bookmark objects'
```

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  /** copies the url to the clipboard */
+  function copyLink() {
+    navigator.clipboard.writeText(location.href)
+  }
+
+  /** Find a matching bookmark using the current URL params, or return default empty bookmark  */
+  function findExistingBookmark() {
+    const bookmark = allBookmarks.find(bookmark => { return bookmark.path === getBookmarkPath() })
```
eweitz, line 39: More conventional, simpler:

```suggestion
    const bookmark = allBookmarks.find(bookmark => bookmark.path === getBookmarkPath())
```

For reference:
```
['a', 'b', 'c'].find(letter => { return letter === 'b' })
'b'
['a', 'b', 'c'].find(letter => letter === 'b')
'b'
```

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+      setBookmarkSaved(false)
+      return DEFAULT_BOOKMARK
+    }
+  }
+
+  const [currentBookmark, setCurrentBookmark] = useState(findExistingBookmark)
+
+  /** get bookmark ID from BSON object */
+  function getBookmarkId(bookmark) {
+    return bookmark?._id || bookmark?.id
```
eweitz, line 53: Could you remind me what the difference is between `_id` and `id`?

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  /** Add a bookmark to the list of user bookmarks, or update ref of existing */
+  function updateBookmarkList(bookmark) {
+    setBookmarkSaved(true)
+    const existingIdx = allBookmarks.findIndex(bookmark => { return bookmark.path === getBookmarkPath() })
```
eweitz, line 59: Per above:

```suggestion
    const existingIdx = allBookmarks.findIndex(bookmark => bookmark.path === getBookmarkPath())
```

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+    }
+    setServerBookmarksLoaded(false)
+    setCurrentBookmark(bookmark)
+  }
+
+  /** remove a deleted bookmark from the list */
+  function removeBookmarkFromList(bookmarkId) {
+    const remainingBookmarks = allBookmarks.filter(bookmark => {
+      return getBookmarkId(bookmark) !== bookmarkId
+    })
```
eweitz, line 76: More conventional, simpler:

```suggestion
    const remainingBookmarks = allBookmarks.filter(
      bookmark => getBookmarkId(bookmark) !== bookmarkId
    )
```

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+      } catch (error) {
+        setShowBookmarksModal(false)
+        setServerBookmarks([])
+        setServerBookmarksLoaded(false)
+      }
+    }
+  }
+
+  const loginPopover = <Popover data-analytics-name='login-bookmark-popover' id='login-bookmark-popover'>
+    You must sign in to bookmark this view
```
eweitz, line 208: Non-blocking: to lower interaction cost and thus boost engagement, can we make this easier for the user like so?

```suggestion
    Click to sign in, then bookmark this view
```

In other words, clicking the star would bring the user to the sign-in view, then they'd sign in, then they'd be brought back here, then they could click to bookmark.  It avoids needing to find the "Sign in" button (which new users might not notice, even though it's prominent) and clicking it.

That elides some detail of course, but such copy keeps it light and I think users would get it.  We could perhaps pave this road more later -- by e.g. enabling bookmarks via a fancy callback URL so the bookmark would be set automatically, with a transient "Saved bookmark" tooltip on page load -- but I suspect even this initial lower-interaction-cost UX would help more users would engage with bookmarks.

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+               value={currentBookmark.description}
+               onChange={handleFormUpdate}
+        />
+      </div>
+      <div className="form-group">
+      <button
+          type="button"
+          className="btn btn-primary"
+          aria-label="Save"
+          data-analytics-name="bookmark-submit"
```
eweitz, line 237: Aligning this name more with the UI would make analytics queries more intuitive.  

Also, there's a local word-order inconsistency in these analytics names.  It's currently (pending above suggestion) `bookmark-save`, `bookmark-delete`, and `manage-bookmarks`.  It should be either `bookmark-save`, `bookmark-delete`, and `bookmarks-manage`, or `save-bookmark`, `delete-bookmark`, and `manage-bookmarks`.  I'd slightly lean towards the latter, as verb-noun is more aligned with the UI and also more intuitive for analytics queries.

```suggestion
          data-analytics-name="save-bookmark"
```

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -0,0 +1,286 @@
...
+          disabled={saveDisabled}
+          onClick={handleSaveBookmark}
+      >{saveText}</button>
+      { bookmarkSaved &&
+        <button
+          type="button"
+          className="btn btn-danger pull-right"
+          aria-label="Delete"
+          data-analytics-name="bookmark-delete"
```
eweitz, line 246:
```suggestion
          data-analytics-name="delete-bookmark"
```

app/javascript/components/bookmarks/BookmarksList.jsx
```diff
@@ -0,0 +1,51 @@
...
+    return currentStudy[0] === bookmarkStudy[0]
+  }
+
+  function loadBookmark(bookmark) {
+    toggleModal()
+    if (isSameStudy(bookmark)) {
+      navigate(bookmark.path)
+    } else {
+      window.location = bookmark.path
+    }
```
eweitz, line 21:
```suggestion
    log('load-bookmark')
    }
```

app/javascript/components/bookmarks/BookmarksList.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import { Modal } from 'react-bootstrap'
+import LoadingSpinner from '~/lib/LoadingSpinner'
+import { navigate } from '@reach/router'
```
eweitz, line 4:
```suggestion
import { navigate } from '@reach/router'
import { log } from '~/lib/metrics-api'
```

app/javascript/components/bookmarks/BookmarksList.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+export default function BookmarksList({serverBookmarks, serverBookmarksLoaded, showModal, toggleModal}) {
+  const ACCESSION_MATCHER = /SCP\d{1,4}/
```
eweitz, line 7: More robust:

```suggestion
  const ACCESSION_MATCHER = /SCP\d+/
```

app/views/profiles/show.html.erb
```diff
@@ -148,6 +148,7 @@
         <% end %>
       </div>
     <% end %>
+    </div>
```
eweitz, line 151: Surprising!

eweitz (approved):
Code and demo look good!  This is a solid feature, and a foundation to potentially build on.

I suggest a handful of maintainability refinements.  I also suggest way to lower interaction cost in this UX, which might be trivial to implement.  None of that is blocking.

app/controllers/api/v1/api_docs_controller.rb
bistline: I took them out of the main Swagger controller because (while listed here) the referenced controller/endpoints were never "swaggerized", so they couldn't actually be used via Swagger.  You can confirm this by looking at [production](https://singlecell.broadinstitute.org/single_cell/api/v1#/Reports) right now.  If we want them added in, that's totally fine, but out-of-scope for this work.

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: Very nice!

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: Technically, MongoDB IDs are referenced as `_id` in the documents.  However, the Rails convention is to simply call it `id`, though both will return the same thing.  This is fine in the Ruby codebase, but in React we're just dealing with the JSON returned.  While working on this PR I noticed inconsistency in API responses, and I wasn't sure why, so I put that in as a helper.  This was many commits ago, so it may not still be happening.  I'll check it out and if possible refine this code.

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: Makes sense - good refinement.

app/views/profiles/show.html.erb
bistline: Oh that's actually a vestigial holdover - shouldn't be there!

app/controllers/api/v1/api_docs_controller.rb
eweitz: Ah, thanks for helping me understanding!  These changes make sense.

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: Some testing has confirmed that API responses will give back `_id`, but our React code is camel-casing this into `id` (not 100% sure where but I think in `scp-api.jsx`) so there's no guarantee of having one over the other at any given time.

app/views/profiles/show.html.erb
bistline: [309f02a](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/309f02ac5945957b77172d2b36feef535c4b063b)

app/javascript/components/bookmarks/BookmarksList.jsx
bistline: [309f02a](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/309f02ac5945957b77172d2b36feef535c4b063b)

app/javascript/components/bookmarks/BookmarksList.jsx
bistline: [309f02a](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/309f02ac5945957b77172d2b36feef535c4b063b)

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: [309f02a](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/309f02ac5945957b77172d2b36feef535c4b063b)

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: [309f02a](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/309f02ac5945957b77172d2b36feef535c4b063b)

jlchang (approved):

Suggestion for future refinements:
1. an X in the bookmark creation/deletion pop-up. (Once you create a bookmark, the star is filled in. In Chrome, you re-click the star to delete the bookmark. In our case, clicking the star merely opens/closes the pop-up. I hesitated to click the filled star because I wasn't sure if it would delete the bookmark ;) I'll admit, I was looking for an X to dismiss the pop-up
2. For visualization views, does it make sense to have an autopopulated field that holds the accession for the study? (I realized if I called my first dotplot bookmark "dot" and switch to a second study, "dot" would be a confusing bookmark but I may not remember all my previous bookmark names in other studies...). For search results, maybe the proposed `accession field` would just say "search"? [FWIW, with an accession field, we can potentially get away with showing a shorter path for default bookmark names]

Paper cuts
1. If you navigate to a different tab while your bookmark is still saving, it might fail to save (race condition) but the bookmark pop-up will retain the chosen bookmark name while in the new tab OR it might save and retain still the chosen bookmark name in the persisted pop-up. (probably if we navigate away, the pop-up goes away?)
https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/293d2c07-f15f-4e3f-b36c-2026fc47f777


Screenshot 2024-02-13 at 5 49 53 PM

3.  If you navigate to a different tab while the bookmark pop-up is open, the pop-up may no longer line up with the star icon

PS. I love that the tool-tip for the bookmark name in My Bookmarks is the URL path - nice touch!

bistline:
> Suggestion for future refinements:
> 
> 1. an X in the bookmark creation/deletion pop-up. (Once you create a bookmark, the star is filled in. In Chrome, you re-click the star to delete the bookmark. In our case, clicking the star merely opens/closes the pop-up. I hesitated to click the filled star because I wasn't sure if it would delete the bookmark ;) I'll admit, I was looking for an X to dismiss the pop-up

To confirm - an X like this one?  Our UX for adding/removing bookmarks mirrors the Chrome UX to a fair extent, so that seems like a good addition.
chrome-edit-bookmark


> 2. For visualization views, does it make sense to have an autopopulated field that holds the accession for the study? (I realized if I called my first dotplot bookmark "dot" and switch to a second study, "dot" would be a confusing bookmark but I may not remember all my previous bookmark names in other studies...). For search results, maybe the proposed `accession field` would just say "search"? [FWIW, with an accession field, we can potentially get away with showing a shorter path for default bookmark names]

Agreed - I nearly added it before, so this is good confirmation that it's useful.

> 
> Paper cuts
> 
> 1. (may be local-only artifact) If you navigate to a different tab while your bookmark is still saving, there is a race condition (The bookmark might save OR it might fail to save) but the bookmark pop-up will persist and retain the previously chosen bookmark name while in the new tab. 

I'll check that out - I hadn't tried that before. 

> 
> 2. If you navigate to a different tab while the bookmark pop-up is open, the pop-up may no longer line up with the star icon. If we change tabs, the pop-up should go away? (see above screenshot)
> 

I did experience this - it can also happen if the form changes size (like if there's an error message).  There's unfortunately no mechanism for us to move it once it's opened the first time.  This is all controlled by the Bootstrap framework.  I can see if closing/reopening the form would work, but other than that I don't think there's anything we can do without patching Bootstrap directly.

> PS. I love that the tool-tip for the bookmark name in My Bookmarks is the URL path - nice touch!

bistline:
@jlchang [7d25022](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/7d25022e5410e223cb980e9b102d0ad874ba003b)

jlchang:
Thanks for the speedy updates @bistline!

> > Suggestion for future refinements:
> > 
> > 1. an X in the bookmark creation/deletion pop-up. 
> To confirm - an X like this one?

Yup, just checked out [7d25022](https://github.com/broadinstitute/single_cell_portal_core/pull/1979/commits/7d25022e5410e223cb980e9b102d0ad874ba003b), looks great.

> > 2. an autopopulated field that holds the accession for the study
Actually, I don't know we need the autopopulated field in the pop-up, it'd be great if the study accession info was visible in My Bookmarks...

> 
> > Paper cuts
> > 
> > If you navigate to a different tab while your bookmark is still saving, there is a race condition (The bookmark might save OR it might fail to save) but the bookmark pop-up will persist and retain the previously chosen bookmark name while in the new tab.
> 
> UPDATE: this is expected, because we're actually changing the query string parameters. Technically, it's a completely new bookmark as `tab` is part of `exploreParams`.

I guess the thing that surprised me was that (not shown in my screen capture), when I had the alternate outcome of the race condition where the heatmap bookmark saved but the persisted bookmark pop-up in the correlation tab still said "heat2" in the bookmark name field. I assumed the save failed and was surprised to see that I had a "heat2" bookmark in my Bookmarks and it looked like I could save another bookmark called "heat2" for the correlation plot... This may just be a race condition you can only catch on local. I'll see if I can do it on Staging...

> 
> > 2. If you navigate to a different tab while the bookmark pop-up is open, the pop-up may no longer line up with the star icon. 
> 
> I did experience this - it can also happen if the form changes size (like if there's an error message). There's unfortunately no mechanism for us to move it once it's opened the first time. 

That's what I figured but thought I'd ask. Is there a way to auto-close the form if the size of page or change of tab occurs?

PR #1982, opened 2024-02-14
Fix error content not showing for bookmarks, Mixpanel logging (SCP-5431)

bistline:
This fixes a bug where error messages do not show render into the form when returned from the server.  This also adds basic Mixpanel logging for creating/updating/deleting bookmarks, which was missing from the original PR.

Note: this does not address known paper cuts, which are tracked in SCP-5516.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to any study with visualizations
3. Bookmark a view (providing a name), then select a different annotation and reuse the name
4. Confirm you see an error message in the form like below:
bookmark-error
5. Correct the error and save, update, and then delete the bookmark, confirming you see corresponding `event` requests in the network tab of Chrome DevTools for `create-bookmark`, `update-bookmark`, and `delete-bookmark`.


eweitz (approved):

jlchang (approved):
Functional review. Now I see that my old bookmarks just didn't have study accession populated. Thanks so much for catching the bug and making this fix!

PR #1985, opened 2024-02-16
Bookmarks followup work (SCP-5516)

bistline:
#### BACKGROUND & CHANGES
This update addresses several small paper cuts with using the bookmarks feature:

* Changing top-level tabs doesn't dismiss the open bookmarks form
* Open form can lose connection to star icon in certain scenarios
* Cannot close bookmark form or modal using escape key
* Users cannot reuse names of bookmarks across different studies 

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to any study with visualizations
3. Open the form, and then resize the page, confirming the form moves with the star icon (there's a small lag, this is to prevent the function from firing continuously on resizing)
4. Keeping the form open, click the Download tab and confirm the form closes
5. Go back to the Explore tab, reopen the form, and press escape, confirming the form closes
6. Create a bookmark with a specific name, and the go to a different study and create a bookmark with the same name, confirming it completes

(P.S. I know I'm technically out but I didn't want to leave this until I got back... feel free to make any updates to this branch as necessary!)

### DIFF
--- a/app/javascript/components/bookmarks/BookmarkManager.jsx
+++ b/app/javascript/components/bookmarks/BookmarkManager.jsx
@@ -63,6 +63,7 @@ export default function BookmarkManager({bookmarks, studyAccession, clearExplore
       setAllBookmarks(prevBookmarks => [...prevBookmarks, bookmark])
     }
     setServerBookmarksLoaded(false)
+    setCurrentBookmark(bookmark)
   }
 
   /** remove a deleted bookmark from the list */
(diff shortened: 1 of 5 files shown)


app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -86,11 +89,64 @@ export default function BookmarkManager({bookmarks, studyAccession, clearExplore
...
+  /** close open form if changing tabs */
+  function closeBookmarkForm() {
+    if (isUserLoggedIn() && formRef.current?.state?.show) {
+      formRef.current.handleHide()
+    }
+  }
+
+  /** wrapper to close both bookmark form & modal */
+  function closeAllComponents() {
+    closeBookmarkForm()
```
eweitz, line 101: I wonder if it'd be possible to pass in a triggering event for logging, e.g. "escape", "outside-keydown".  This is merely food for thought for later, and not blocking.

eweitz (approved):
Code looks good!

PR #1988, opened 2024-03-05
Cell filtering by numeric annotations: refined UI (SCP-5500)

eweitz:
This makes it easier and more robust to filter cells by quantitative dimensions.

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/1980), cell filtering by numeric annotations had basic functionality, but was somewhat tedious and brittle.  Input had to be manually typed in, resetting was hard, non-default operators weren't integrated, etc.

Now, the numeric cell filtering UX is more intuitive and well-rounded, and ready for general availability (GA).  

Refinements include:
* **Range slider**.  Users expect to be able to select numeric ranges by conveniently resizing and moving a slider.  The new custom slider widget enables that with a UI that has clear affordances, economizes vertical real estate, and supports special (but common for SCP) cases like "not applicable" values, a.k.a. "N/A".

* **Reset**.  To clear a numeric selection, you either had to remember your initial values or reload the page.  Now you can simply click a reset / refresh icon beside the facet.  Global reset now works for numeric cell facets, too.  

* **Operators beyond "between"**.  Non-default operators -- i.e. "not between", =, !=, >, >=, <, and <= from the client library are now integrated into the numeric cell filtering UI.  Operators that take 1 parameter (i.e. =, !=, >, >=, <, and <=) show tooltips on histogram bars, but no slider.  The 2-parameter operators show the slider but no histogram bar tooltips.

* **Full values always visible**.  Numeric values in the input fields were previously truncated if they were too large.  A custom fade affordance slightly mitigated that UI accuracy problem, but remained very unintuitive and error-prone.  Now, input fields widen and shrink font to ensure these important values are never even partly hidden from view.  This also helps keep the paired inputs (i.e. range min. and max.) on the same line in typical viewports.  For narrow viewports, input pairs now linebreak together, rather than oddly having min. input on one line and max. input on another.

* **More robust handling for invalid values**.  Typing a values less than the minimum, greater than maximum, or non-numbers changes the input border color, and otherwise reasonably handled.

### Video
Here's how it looks! 

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/1b8e3a7f-a5cc-4e44-805c-d566696002b2

_^ Video 1: Slider resize and reset, and facet-level reset; facets with "N/A"_

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/96019545-32cd-4576-873c-e9c99bc31146

[...]

PR #1989, opened 2024-03-08
Fix overshot slider for numeric cell filtering (SCP-5500)

eweitz:
This prevents misdisplay if numeric filter crosshair-select moves out-of-bounds, a bug fix that regressed mid-PR in #1988.

Here's how the fix looks.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/a300f33f-2330-45f7-b4c0-8c5d1a708647

### Test
Automatically testing drag interactions for React components in Jest is remarkably difficult, and turns out to be infeasible here.  To manually test:

1.  Go to a study with numeric annotations, e.g. "Cellular and transcriptional diversity over the course of human lactation" ([SCP303](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP303/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation) on staging).
2.  Click "Filter plotted cells".
3.  Find a numeric annotation, e.g. "BMI during study".
4. Drag the right handlebar substantially to the left
5. In an area of the histogram not covered by the slider, click down and drag your cursor beyond the right edge of the slider
6. Confirm handlebar stop at right edge of slider, and value in right input does not exceed maximum (nor say "NaN")

This relates to SCP-5500.

### DIFF
--- a/app/javascript/components/explore/NumericCellFacet.jsx
+++ b/app/javascript/components/explore/NumericCellFacet.jsx
@@ -677,7 +677,10 @@ export function NumericCellFacet({
     const [newValue, newValue2] =
       parseValuesFromBrushSelection(brushSelection, xScale, precision)
 
-    if (newValue > max || newValue < min || newValue2 > max || newValue2 < min) {
+    if (
+      newValue > max || newValue < min || newValue2 > max || newValue2 < min ||
+      Number.isNaN(newValue) || Number.isNaN(newValue2)
+    ) {
       // Prevent handlebar misdisplay if crosshair-select moves out-of-bounds
       // See "overshot_slider.mov"
       // in https://github.com/broadinstitute/single_cell_portal_core/pull/1988#pullrequestreview-1920037002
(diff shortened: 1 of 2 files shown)


jlchang (approved):
Manual review - crosshair selection of slider handles no longer overshoots.

When testing on my local instance, because the breast milk study has a lot of facets, it takes a while for the full set of facets to load. If I try to interact with the sliders, the loading of each new set of facets resets the numeric sliders to their default settings.

I do not see this effect on staging (may only be seen locally due to the slow page render times). Do we think this may be a problem for numeric facets for larger datasets?

Screencapture of the issue:
https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/c8bf521f-4aa8-4dce-b1d7-9fa2ce38bdb0

Approving this PR on the assumption that numeric slider reset during facet loading is confined to developer's local instances.

bahill:
@jlchang I do think it is local slowness. I was not able to move my mouse faster than the facets were loading on staging.

bahill (approved):

eweitz:
Thanks for the reviews!

Good find, Jean.  The issue you note seems independent of these changes, and likely low severity and very low incidence.  Its incidence is much higher on local SCP instances, especially without development frontend service worker caching enabled.

Regarding incidence, outside some local development scenarios, this issue would most likely only manifest in uncommon user flows on the very first Study Overview load for each study after a deployment, or rarely for very large studies.  As you describe, it only occurs between when A) a numeric cell facet became visible and B) all facets were fully loaded -- and the user would need to interact with the facet in that brief window.  It might happen among our top 1% of studies somewhat more frequently.  But I suspect the incidence among real users would be low even in that rare study cohort, as the timing of the triggering interaction would need to occur in a narrow window.  We could likely quickly assess incidence analytically.

Regarding severity, the unintentional reset of slider and inputs seems noticeable by the user, and easy for the user to work around by changing the values again.

I've added this issue to a list of seemingly similar-priority ones over in SCP-5534.

PR #1991, opened 2024-03-13
Release 1.67.0

jlchang:


jlchang (commented):
Ugh, our new false positive build failure persists

> ImportServiceTest#test_should_call_API_client_method:
  RuntimeError: Wrapped undumpable exception for: RestClient::InternalServerError: 500 Internal Server Error
Exiting with code: 1
Unlike other runs, this one hit a single `500` error and exited (other CI jobs persisted through multiple `500` and occasionally `403` errors to reach the `### FAILURES and ERRORS ###` summary section.

eweitz (approved):

app/javascript/components/explore/ExploreDisplayTabs.jsx
```diff
@@ -176,26 +176,31 @@ function getFacetsParam(initFacets, selection) {
   const minimalSelection = {}
 
   const initSelection = {}
-  initFacets.filter(f => !f.isSelectedAnnotation).forEach(facet => {
+  initFacets.filter(f => !f.isSelectedAnnotation)?.forEach(facet => {
```
bahill, line 179: Just curious - what does the "?" do?

bahill (approved):

app/javascript/components/explore/ExploreDisplayTabs.jsx
eweitz: `?.` is the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), which:

> accesses an object's property or calls a function. If the object accessed or function called using this operator is undefined or null, the expression short circuits and evaluates to undefined instead of throwing an error.

At a high level, `?` is sometimes a useful quick fix for a UI-breaking error of trying to access a property in a non-existent object.  However, due to it being somewhat opaque and suppressing exceptional states, IMO it's best to use sparingly.

PR #1993, opened 2024-03-18
Fix slider when min, max have long decimal values (SCP-5555)

eweitz:
This rounds long decimal values, to fix two related bugs in the slider of numeric cell facets:

### Overview
1.  Long decimal values get truncated from view.  This was the case previously for all long numbers, but the recent layout robustness improvement did not cover numbers that include long fractional segments, e.g. full-precision floats like 16.167313586858402.

2.  In cases of bug 1, slider resize doesn't work.  Values can still be changed via the manual input fields, but broken input-by-slider is quite inconvenient in these cases.

Rounding the minimum and maximum values to 2 decimal places (e.g. 0.345678912 -> 0.35) fixes both issues.  For context, we do somewhat similarly for gene expression values -- rounding to 3 decimal places at ingest.

### Video
Here's how it looks, before and after.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/4fd34520-092a-470f-a96b-70a112c85113

(The fact that the "--Filtered--" scatter plot legend doesn't appear is a separate bug, with a fix under development.  This dataset presents two previously untested scenarios, with this one being more prevalent.)

### Test
Automated tests have been updated to confirm this fix.  To manually test:
1.  Go to a study that has a numeric annotation with a maximum or minimum that has long decimal values.  [SCP390 on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP390/multi-species-asos-reproducing-scp2126-on-staging) is an example.
2. Click "Filter plotted cells".
3. Find the numeric cell facet that has long decimal values.
4. Confirm the long values are rounded to 2 decimal places.
5. Resize the slider.
6. Confirm slider resizes, and that new selection boundaries are shown in the two input fields.  ("--Filtered--" may not appear in the legend due to a separate bug.)

This satisfies SCP-5555.

### DIFF
--- a/app/javascript/lib/cell-faceting.js
+++ b/app/javascript/lib/cell-faceting.js
@@ -9,7 +9,7 @@ import crossfilter from 'crossfilter2'
 import { getIdentifierForAnnotation } from '~/lib/cluster-utils'
 import { fetchAnnotationFacets } from '~/lib/scp-api'
 import { log } from '~/lib/metrics-api'
-
+import { round } from '~/lib/metrics-perf'
 
 const CELL_TYPE_REGEX = new RegExp(/cell.*type/i)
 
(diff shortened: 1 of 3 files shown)


eweitz:
The failing Rails integration tests are being resolved now, separately.  They're independent of these changes.

bahill (approved):
code and manual review lgtm - noting other existing bugs in play

jlchang (approved):
Rounding seems like a low-effort way to mitigate this bug!

PR #1996, opened 2024-03-22
Fix cell filtering when numeric annotation has same value for all cells (SCP-5554)

eweitz:
This fixes a rare bug that breaks cell filtering when a numeric facet has a constant value.

Previously in that case, the plot would not update when the cell filtering selection changed.  No points changed to the filtered faint-grey color, and the "--Filtered--" scatter plot legend did not appear.  The bug seems very uncommon, but essentially breaks cell filtering when it occurs.

These simple changes fix the most severe problem, but an upstream change would A) fix related minor issues elsewhere, and B) speed up the exploration UX.  That potential future work is detailed in SCP-5572.  

A Mixpanel property name was also refined.  I had explored a more ambitious refinement of cell filtering analytics, as part of a more complex frontend fix for this bug, but it became apparent the changes would be more involved and out-of-scope.  SCP-5573 tracks those analytics refinements.

### Test
A unit test verifies this fix.  To manually test:
1.  Go to a study that has a numeric annotation with the same value for all cells.  [SCP390 on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP390/multi-species-asos-reproducing-scp2126-on-staging) is an example.
2. Deselect any categorical filter
3. Confirm the plot updates, and "--Filtered--" appears at bottom of the scatter plot legend

This satisfies SCP-5554.

### DIFF
--- a/app/javascript/lib/cell-faceting.js
+++ b/app/javascript/lib/cell-faceting.js
@@ -215,6 +215,11 @@ export function filterCells(
           const numericFilters = selection[facet] // e.g. [0, 20]
 
           if (numericFilters === undefined) {
+            // Some numeric annotations can have 1 (and only 1) value repeated
+            // for every cell.  Such annotations are not eligible as facets,
+            // because filtering requires > 1 value.
+            //
+            // TODO (SCP-5513): Screen numeric facets with constant value
             continue
           }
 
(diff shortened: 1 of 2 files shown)


jlchang (approved):
Manual testing works as described. Code changes seem logical to bypass the "single-numeric value" triggering condition.

Thanks for finding this rare, but feature-breaking failure mode!

kilonzi (approved):

PR #2002, opened 2024-04-01
Prevent clearing bookmark form on save

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with saving new bookmarks where the action completes and the bookmark is created, but the form empties and does not properly store the state unless the page is refreshed.  This appears to be a problem with multiple code paths simultaneously trying to set the currently loaded bookmark (both `updateBookmarkList` and `resetForm` call `setCurrentBookmark`) which ends up clobbering the value with the "default" bookmark.  Now, the form is left unchanged after a successful save to prevent this from happening.

#### MANUAL TESTING
1. Boot as normal and sign in, then load any study with visualizations
2. Bookmark the default view, and confirm the form does not reset on save
3. Load a different visualization and confirm the form empties and the star icon shows nothing is saved
4. Click "Reset view" and confirm that the star icon shows the default view is already bookmarked, and that the values shown in the form are correct

### DIFF
--- a/app/javascript/components/bookmarks/BookmarkManager.jsx
+++ b/app/javascript/components/bookmarks/BookmarkManager.jsx
@@ -20,6 +20,7 @@ import { log } from '~/lib/metrics-api'
 export default function BookmarkManager({bookmarks, studyAccession, clearExploreParams}) {
   const location = useLocation()
   const [allBookmarks, setAllBookmarks] = useState(bookmarks)
+  console.log(`allBookmarks: ${allBookmarks.length}`)
   const [saveText, setSaveText] = useState('Save')
   const [saveDisabled, setSaveDisabled] = useState(false)
   const [bookmarkSaved, setBookmarkSaved] = useState(null)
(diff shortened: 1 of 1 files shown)


bistline:
Build failure is known issue with user groups test.

jlchang (approved):
Functional review - fix as described. Code change makes sense given context from provided background.


Observation of slightly ambiguous UX: because of URL-based parameters persisting the state of the explore tab, `reset view` vs reloading the page from using the browser's `reload` functionality does two distinctly different things. It might be helpful to provide some way to have those two functionalties in physical proximity with language that would be more explicit about the distinction.

"restore default view" vs "reset page for current view" 
`restore default view` <- used to "start over" and reset study to original view (currently `reset view`)
`reset page (current view)` (aka. browser reload equivalent) <- use when explore tab isn't behaving properly (currently need to know to use browser reload)

Even though I know when to use which, even I'm sometimes surprised/confused when I use the browser reload and the gene search doesn't clear (whereas browser reload when DE is displayed always clears DE data because DE has not been URL parameterized)

If you agree the two resets are a little ambiguous/confusing for the occasional portal user AND worth addressing given resource constraits, please create a follow-on ticket.

PR #2005, opened 2024-04-04
Generalize global search footer note (SCP-5577)

eweitz:
This refines the footer with text that is less specific, but more correct and lower effort to maintain.

### Before
Old_advanced_search_footer_text__SCP_2024-04-04

### After
New_advanced_search_footer_text__SCP_2024-04-04

This satisfies SCP-5577.


bistline (approved):

jlchang (approved):
Change looks good to me. Thanks for identifying a low-maintenance, long-term solution!

PR #2008, opened 2024-04-09
Fix TypeError due to missing function on formRef (SCP-5585)

bistline:
#### BACKGROUND & CHANGES
This fixes an issue where resizing the screen while viewing a study when a user is not logged in throws a `TypeError` due to a missing function in the `BookmarkManager` component.  Since the user is not signed in, the ref is never rendered into the DOM.  Now, a state check is performed before calling `handleToggle` to ensure that the form is in fact open and can be hidden or toggled.  While this error had no user-facing impacts, it was very frequent, comprising ~18% (3.7K) of errors and ~10% (51) of session replays over the last month in Sentry.

#### MANUAL TESTING
2. Open the JS console
3. Resize the screen, and confirm that no `TypeError` is thrown
4. Log in, then open the bookmark form by clicking the star icon
5. Resize the screen again, confirming that the form repositions after you finish resizing

### DIFF
--- a/app/javascript/components/bookmarks/BookmarkManager.jsx
+++ b/app/javascript/components/bookmarks/BookmarkManager.jsx
@@ -89,9 +89,14 @@ export default function BookmarkManager({bookmarks, studyAccession, clearExplore
     setCurrentBookmark(findExistingBookmark())
   }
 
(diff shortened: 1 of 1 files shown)


eweitz (approved):
Thanks for this, and for preserving replay quota!

PR #2009, opened 2024-04-09
Enable linking to numeric cell facets, restore for categorical (SCP-5531)

eweitz:
This updates the URL as a user adjusts numeric cell facets, so they can link to results.  

It also fixes URL parameter handling for categorical cell facets.

Here's how it looks:

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/2b5c2a0d-f272-4096-bbba-6bd1f9f17db2

### Test
A new automated tests verifies behavior.  To manually test:
1.  Apply numeric and categorical cell filters
2.  Refresh page
3.  Confirm previously-selected filters are automatically applied

This satisfies SCP-5531.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -26,7 +26,9 @@ import ExploreDisplayPanelManager from './ExploreDisplayPanelManager'
 import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger'
 import Tooltip from 'react-bootstrap/lib/Tooltip'
 import PlotTabs from './PlotTabs'
-import { initCellFaceting, filterCells } from '~/lib/cell-faceting'
+import {
+  initCellFaceting, filterCells, getFacetsParam, parseFacetsParam
+} from '~/lib/cell-faceting'
 
 /** Get the selected clustering and annotation, or their defaults */
 export function getSelectedClusterAndAnnot(exploreInfo, exploreParams) {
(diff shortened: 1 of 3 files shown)


bistline (approved):

jlchang (changes requested):
Facet filtered URLs do not seem to restore state when the view is toggled out of the filter plotted cells panel (leaving the panel restores the view to all cells not filtered by the legend but the facet filtered URL remains) then returned to the filter plotted cells panel (The filter plotted cells panel reflects the facet filtered URL [epithelial deselected] but the plot shows the unfiltered state [epithelial plotted]).


https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/a0490189-80d3-42e0-83bd-d8db033c3d7e

Note: early in the video there is a race condition glitch that resets a deselected filter - likely a non-issue on production where filter facet loading is much quicker.

We also have an issue with an [ambiguous UX ](https://github.com/broadinstitute/single_cell_portal_core/pull/2002#pullrequestreview-1972035242) with respect to `reset view` vs reloading the page from using the browser's `reload` functionality (please read linked comment). The toggle from filtered to unfiltered further complicates this ambiguous UX. With respect to this PR I propose that we revisit the decision to restore the "all cells not filtered by the legend" view. For the rest of the ambiguous UX, we could ticket future work to bring "restore default view" and "reset page for current view" functionality in physical proximity <- is there additional complexity vis-a-vis bookmarks?

Once the post-toggle inconsistency between URL state (with filter plotted cells panel open) and view is addressed, I'll be happy to do additional testing. I have not tested the different flavors of numeric filtering yet.

eweitz:
Thanks for reporting those issues, Jean!

> Facet filtered URLs do not seem to restore state when the view is toggled out of the filter plotted cells panel (leaving the panel restores the view to all cells not filtered by the legend but the facet filtered URL remains) then returned to the filter plotted cells panel (The filter plotted cells panel reflects the facet filtered URL [epithelial deselected] but the plot shows the unfiltered state [epithelial plotted]).

This is a separate, pre-existing issue.  Once SCP-5553 is fixed, this effect will be resolved.

> We also have an issue with an ambiguous UX with respect to reset view vs reloading the page from using the browser's reload functionality (please read linked comment). The toggle from filtered to unfiltered further complicates this ambiguous UX and I propose that, with the parametrized URLs, perhaps we no longer need/want to restore the "all cells not filtered by the legend" view?

I agree that's an issue, and read the linked comment.  But it is also beyond the scope of the ticket, which is enabling URL parameters for numeric cell facets, and restoring URL parameters for categorical cell facets (i.e., restoring categorical cell facet URL parameterization to the state it was in as of PR #1942).

If you'd like to iterate on "Reset view" vs. browser refresh, in the context of bookmarks, cell filtering, and beyond, then please create a follow-on ticket and I'd be happy to collaborate there.

jlchang (approved):
Thanks for pointing out that the toggle glitch issue is already ticketed.

I'll make a ticket for the other suggestion [SCP-5590](https://broadworkbench.atlassian.net/browse/SCP-5590)

PR #2013, opened 2024-04-16
Filter global gene search using existing search parameters (SCP-5428, SCP-5511)

bistline:
#### BACKGROUND
Advanced search offers a powerful tool for discovering data in SCP.  Additionally, global gene search allows users to browse gene expression profiles across multiple studies at once.  While both are available from the home page, and use the same area of the application, their results are not tied together.  This makes refining the results for global gene search impossible.

#### CHANGES
This update now applies all existing search criteria to global gene searches and executes a parallel gene-scoped search alongside the study-scoped one.  Additionally, any queries that would also result in a search in Azul are now dropped from the gene-specific search.  The result is that the study-based search remains the same, but the gene-based search now includes all SCP-based results that match that query.  Additionally, any studies that do not contain the requested gene(s) are filtered as usual.   There are some added bug fixes as well (SCP-5511) to improve the overall UX.  Lastly, help text was updated for keyword search to note that studies can be directly loaded by entering their study accession.

#### KNOWN ISSUES
There is an existing issue where clicking the clear button on the gene search bar does not clear out any results, or update the search query parameters.  This is identical to the behavior of study-based gene search, and is expected.  Also, using the "Clear All" button does not empty the gene search bar.  Additionally, clicking over to the study search results and then back to the gene search will result in re-running the gene search.  This is because the violin plots have been dynamically removed from the DOM as a result of the routing and need to be re-rendered.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Run a faceted search for `species: Homo sapiens` and `cell type: B cell`
3. Confirm that you see results from Azul as well as the `Human milk - differential expression` example study (and potentially `Human all genes` if you have that in your local instance as well)
4. Click "Search genes" and confirm you see the same SCP studies as in the normal search results
5. Search for `CLDN4` and confirm you see a violin plot in the `Human milk - differential expression` study with high expression in the `LC1` cell type, and that the annotation dropdown has `General_Celltype` set already
6. Change the selected cluster/annotation and confirm no other plots are affected
7. Click "Clear All" and confirm that the results revert back to a normal empty search

### DIFF
--- a/app/javascript/components/search/controls/KeywordSearch.jsx
+++ b/app/javascript/components/search/controls/KeywordSearch.jsx
@@ -35,7 +35,7 @@ export default function KeywordSearch({ keywordPrompt }) {
       <p>&quot;single cell&quot; Smith</p>
     </ul>
     <p>* Structured data for authors is new in SCP, and many studies lack it, so author search results may be limited.</p>
-    <p>You can find studies directly by entering their accessions as terms: "SCP1 SCP101 SCP1234"</p>
+    <p>You can find studies directly by entering their accessions as terms: SCP1 SCP101 SCP1234</p>
   </div>)
 
   const textSearchLink = <a className="action advanced-opts"
--- a/app/javascript/components/search/controls/KeywordSearch.jsx
+++ b/app/javascript/components/search/controls/KeywordSearch.jsx
@@ -35,6 +35,7 @@ export default function KeywordSearch({ keywordPrompt }) {
       <p>&quot;single cell&quot; Smith</p>
     </ul>
     <p>* Structured data for authors is new in SCP, and many studies lack it, so author search results may be limited.</p>
+    <p>You can find studies directly by entering their accessions as terms: "SCP1 SCP101 SCP1234"</p>
   </div>)
 
   const textSearchLink = <a className="action advanced-opts"
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -170,6 +170,7 @@ module Api
 
       def index
         @viewable = Study.viewable(current_api_user)
+        @search_type = params[:type].to_sym
 
         # filter results by branding group, if specified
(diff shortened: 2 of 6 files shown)


bistline:
> Useful enhancement! Code looks good, and my local manual tests passed.
> 
> I see a few user-facing issues. Both are low severity.
> 
> 1. **No "text match only" badge**.  For _inferred_ study results for queries that include facets _and_ genes, the "text match only" badge is unexpectedly not shown.  

Nice catch - that should be easy enough.

> 2. **Axis label collides with horizontal tick marks**.  Also at time 0:30, notice how the "Expression" label overlaps with the numbers that mark left-right grid lines in the violin plot, e.g. "0.4848k".  

This also happens on the study-based gene search as well, so I agree that it is both existing and out-of-scope.  I'll open a ticket for this.

bistline:
@eweitz https://github.com/broadinstitute/single_cell_portal_core/pull/2013/commits/ddb84c15ec2452636264245ae228f3352f526ce2

app/javascript/components/search/controls/KeywordSearch.jsx
```diff
@@ -35,6 +35,7 @@ export default function KeywordSearch({ keywordPrompt }) {
       <p>&quot;single cell&quot; Smith</p>
     </ul>
     <p>* Structured data for authors is new in SCP, and many studies lack it, so author search results may be limited.</p>
+    <p>You can find studies directly by entering their accessions as terms: SCP1 SCP101 SCP1234</p>
```
jlchang, line 38: Thank you for adding this! Should we include "study accession" in the first sentence of this infobox as well?

jlchang (approved):
Worked great in functional testing. Can't wait to have folks try it out!

app/javascript/components/search/controls/KeywordSearch.jsx
bistline: Actually that's a great idea - I'll do that.

PR #1379, opened 2022-02-17
Adding Author and Publication information to Study objects (SCP-2980, SCP-4053)

bistline:
Currently, the only way for users to add information about authors and publications associated with their study is to do so inside the study description.  While this does make the information searchable, there is no standardized format, and discovering that information is haphazard at best.  This update adds two new models (`Author` and `Publication`) to track that information directly.  In addition, the Study Settings tab has now been refactored to split out into individual tabs for updating various aspects of the study.  Below is a screen shot of the new Settings UX (showing the author/publication tab):

study_settings_authors_pubs
 
If a study adds any publications/external resources, then a collapsable sidebar will render on the study description tab with links to the relevant papers/resources.  Here is a screenshot of that view (the tab is open by default, but can be closed):

study_description_pubs_open

In addition, this update adds contact links for any corresponding authors to render in the same sidebar.  If an author is marked as corresponding, then a button will render at the top of the sidebar.  Clicking on this button opens a popover with a `mailto:` link inside.  This popover is dynamically rendered into the page, and the email address is protected behind `base64` encoding.  This is to alleviate concerns about screen-scraping email addresses.  Below is a screenshot of what that popover looks like:

Screen Shot 2022-02-17 at 3 02 37 PM

MANUAL TESTING
1. Boot as normal and sign in
2. Load any study you have edit access to and open the Study Settings tab
3. Confirm that each sub-tab can be opened, and contains the correct information
4. Go to the authors/publications tab, and enter information about a publication, and add yourself as a corresponding author (you must click `Allow email?` to do so)
5. Click Update settings, and go back to the description view
6. Confirm your paper shows there, there is a button to contact you as an author that opens the popover on click, and that the sidebar can collapse and reopen

This PR continues work outlined in SCP-2980 and satisfies SCP-4053.

PR #2016, opened 2024-04-17
Add study publications to SCP API (SCP-5595)

eweitz:
This integrates studies' journal publications into SCP API, enabling better programmatic accessibility.

### Overview
[Previously](#1379), study editors and owners could add their publication's structured data, and it conveniently displayed at right in the Summary tab of the Study Overview page.  However, the data was only accessible via the UI, and not in a RESTful format.

Now publication data is available via two sets of [SCP API](https://singlecell.broadinstitute.org/single_cell/api/v1) endpoints: `/site/studies/{accession}` and a new group of CRUD endpoints anchored around `/studies/{accession}/publications`.  Implementation reuses patterns from "external resources".  These changes help decouple the frontend and backend, and enable more flexible ways to process publication data.  That promotes FAIR data.

### Limitations
There's a lot that could be improved about publications on SCP!  Some studies link to their publications in the study description, some do via "external resources"; such cases could likely be inferred and better structured in a publication object.  The sidebar UI display for publications could be refined.  Capturing DOI specifically might be nice.  "Citation" values could probably usually be automatically populated.  

This PR only intends to make existing structured data for publications accessible by API.

### Test
Automated tests verify expected behavior.  If you'd like to manually confirm:
1.  Go to a study, e.g. your local "Cellular and transcriptional diversity over the course of human lactation"
2.  Settings -> Authors/Publications
3. Add structured data for publication, e.g.:
- Title: Cellular and transcriptional diversity over the course of human lactation
- Journal: PNAS
- Citation: https://www.pnas.org/doi/10.1073/pnas.2121720119
- URL: https://doi.org/10.1073/pnas.2121720119
4. Click "Update settings"
5. Note your study's accession
6. Go to https://localhost:3000/single_cell/api/v1/site/studies/(your-study-accession)
7. Confirm a `publications` field exists in the response with the data you supplied

This satisfies SCP-5595, a Developer's Choice Days project.


jlchang (approved):
Love this step towards FAIR data! Publication object accessible by API as described.

app/controllers/api/v1/publications_controller.rb
```diff
@@ -0,0 +1,314 @@
...
+        rescue => e
+          ErrorTracker.report_exception(e, current_api_user, @publication, params)
+          MetricsService.report_error(e, request, current_api_user, @study)
+          render json: {error: e.message}, status: 500
+        end
+      end
+
+      private
+
+      def set_study
```
bistline, line 284: We have a concern called `StudyAware` that is included in all `ApiBaseController` descendants.  This provides built-in methods for `:set_study`, `:check_study_view_permission`, and `check_study_edit_permission`.  As this concern is already included in this controller, you can remove `:set_study` and `: check_study_permission` here and just rely on the `before_action` filters you've set up (though you'll need to rename `:check_study_permission` to `:check_study_edit_permission`.

Again, this seems to be a side effect of copying from `ExternalResoucesController`.  You can make the same edits there as well.  That controller predated the `StudyAware` module.

bistline (changes requested):
Great addition, and love that we're still being FAIR.  I note two code simplification edits that will reduce the amount of redundancy and improve latency by removing irrelevant status checks.

Also, unrelated - I initially wasn't able to add the publication because I ran into the "Study files is invalid" [bug](SCP-4142).  This may be worth addressing, either by improving the error message or (even better) avoid re-validating study file entries when all we're trying to do is update study settings.

app/controllers/api/v1/publications_controller.rb
```diff
@@ -0,0 +1,314 @@
+module Api
+  module V1
+    class PublicationsController < ApiBaseController
+      include Concerns::FireCloudStatus
```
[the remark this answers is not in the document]
eweitz: Done in 62d0940.

app/controllers/api/v1/publications_controller.rb
eweitz: Helpful!  Done in 254af54.  I also widened accessibility for public studies in 02b4b1b.

jlchang:
Functional review - confirmed that manual test still succeeds after latest refinements.

bistline (approved):

PR #2017, opened 2024-04-17
Adding bookmarks to home page search results (SCP-5604)

bistline:
#### BACKGOUND & CHANGES
With recent enhancements to [filter global gene search using search criteria](#2013), adding the ability to easily recall searches becomes more important.  This update adds the bookmark feature to the home page search, and can be used to save both study- and gene-based searches.  While the exact results will _not_ be cached, as this is data-dependent, the search parameters will be, allowing for users to easily find updates to existing queries.

#### MANUAL TESTING
1. Boot as normal
2. Run a study-based search
3. Note the bookmark icon appears inline with the query summary
4. Open the bookmark form and then save
5. Click over to `Search genes` and run a gene search, also bookmarking that result
6. From the bookmark for, click `All bookmarks`, and then use the entries to navigate back and forth between both searches

### DIFF
--- a/app/javascript/components/HomePageContent.jsx
+++ b/app/javascript/components/HomePageContent.jsx
@@ -10,7 +10,7 @@ import ResultsPanel from '~/components/search/results/ResultsPanel'
 import StudyDetails from '~/components/search/results/StudySearchResult'
 import StudySearchProvider, { StudySearchContext } from '~/providers/StudySearchProvider'
 import SearchFacetProvider from '~/providers/SearchFacetProvider'
-import UserProvider, { isUserLoggedIn } from '~/providers/UserProvider'
+import UserProvider from '~/providers/UserProvider'
 import BookmarkManager from '~/components/bookmarks/BookmarkManager'
 import ErrorBoundary from '~/lib/ErrorBoundary'
 
@@ -35,25 +35,19 @@ const LinkableSearchTabs = function(props) {
   // the queryParams object does not support the more typical hasOwnProperty test
   return (
     <div>
-      <nav className="nav search-links" data-analytics-name="search" role="tablist">
-        <Link to={`${basePath}/app/studies${location.search}`}
-          className={showGenesTab ? '' : 'active'}>
-          <span className="fas fa-book"></span> Search studies
-        </Link>
-        <Link to={`${basePath}/app/genes${location.search}`}
-          className={showGenesTab ? 'active' : ''}>
-          <span className="fas fa-dna"></span> Search genes
-        </Link>
-        { isUserLoggedIn() &&
-          <span className='margin-left'>
-            <BookmarkManager eagerLoadBookmarks={true} />
+        <nav className="nav search-links" data-analytics-name="search" role="tablist">
+          <Link to={`${basePath}/app/studies${location.search}`}
+                className={showGenesTab ? '' : 'active'}>
+            <span className="fas fa-book"></span> Search studies
+          </Link>
+          <Link to={`${basePath}/app/genes${location.search}`}
+                className={showGenesTab ? 'active' : ''}>
+            <span className="fas fa-dna"></span> Search genes
+          </Link>
+          <span id='home-page-bookmark'>
+            <BookmarkManager eagerLoad={true} />
           </span>
-        }
-        { !isUserLoggedIn() &&
-          <BookmarkManager eagerLoadBookmarks={true} />
-        }
-
-      </nav>
+        </nav>
       <div className="tab-content top-pad">
         <Router basepath={basePath}>
           <GeneSearchView path="app/genes"/>
--- a/app/javascript/components/HomePageContent.jsx
+++ b/app/javascript/components/HomePageContent.jsx
@@ -10,7 +10,8 @@ import ResultsPanel from '~/components/search/results/ResultsPanel'
 import StudyDetails from '~/components/search/results/StudySearchResult'
 import StudySearchProvider, { StudySearchContext } from '~/providers/StudySearchProvider'
 import SearchFacetProvider from '~/providers/SearchFacetProvider'
-import UserProvider from '~/providers/UserProvider'
+import UserProvider, { isUserLoggedIn } from '~/providers/UserProvider'
+import BookmarkManager from '~/components/bookmarks/BookmarkManager'
 import ErrorBoundary from '~/lib/ErrorBoundary'
 
 /** include search controls and results */
@@ -43,6 +44,15 @@ const LinkableSearchTabs = function(props) {
           className={showGenesTab ? 'active' : ''}>
           <span className="fas fa-dna"></span> Search genes
         </Link>
+        { isUserLoggedIn() &&
+          <span className='margin-left'>
+            <BookmarkManager eagerLoadBookmarks={true} />
+          </span>
+        }
+        { !isUserLoggedIn() &&
+          <BookmarkManager eagerLoadBookmarks={true} />
+        }
+
       </nav>
       <div className="tab-content top-pad">
         <Router basepath={basePath}>
--- a/app/javascript/components/HomePageContent.jsx
+++ b/app/javascript/components/HomePageContent.jsx
@@ -33,8 +33,8 @@ const LinkableSearchTabs = function(props) {
   const [bookmarks, setBookmarks] = useState([])
   const [hasLoadedBookmarks, setHasLoadedBookmarks] = useState(null)
 
-  async function loadUserBookmarks() {
-    setHasLoadedBookmarks(true) // short-circuit multiple calls to load bookmarks
+  // reload bookmarks from server when user switches tabs
+  async function reloadBookmarks() {
     try {
       const userBookmarks = await fetchBookmarks()
       setBookmarks(userBookmarks)
@@ -46,7 +46,8 @@ const LinkableSearchTabs = function(props) {
 
   useEffect(() => {
     if (isUserLoggedIn() && !hasLoadedBookmarks) {
-      loadUserBookmarks()
+      setHasLoadedBookmarks(true) // short-circuit multiple calls to load bookmarks
+      reloadBookmarks()
     }
   }, [])
 
@@ -54,11 +55,11 @@ const LinkableSearchTabs = function(props) {
   return (
     <div>
         <nav className="nav search-links" data-analytics-name="search" role="tablist">
-          <Link to={`${basePath}/app/studies${location.search}`}
+          <Link to={`${basePath}/app/studies${location.search}`} onClick={reloadBookmarks}
                 className={showGenesTab ? '' : 'active'}>
             <span className="fas fa-book"></span> Search studies
           </Link>
-          <Link to={`${basePath}/app/genes${location.search}`}
+          <Link to={`${basePath}/app/genes${location.search}`} onClick={reloadBookmarks}
                 className={showGenesTab ? 'active' : ''}>
             <span className="fas fa-dna"></span> Search genes
           </Link>
--- a/app/javascript/components/HomePageContent.jsx
+++ b/app/javascript/components/HomePageContent.jsx
@@ -1,5 +1,5 @@
 
-import React, { useContext } from 'react'
+import React, { useContext, useState, useEffect } from 'react'
 import { Router, Link, useLocation } from '@reach/router'
 
 import GeneSearchView from '~/components/search/genes/GeneSearchView'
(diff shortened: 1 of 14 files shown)


jlchang (changes requested):
Minor papercut that should be fixed before release.

If I bookmark a study search (example "Homo sapiens" and "B cell") and gene search "gapdh" and save it. Then change gene search to "mapk1" and save it. When I use the "gapdh" bookmark, the URL and plots all say "gapdh" but the gene search field still says "mapk1"...

Let me know if a screencapture would be more helpful.

Also, I don't have "strong" thoughts on this but should new bookmarks should show up at the bottom of the bookmarks list or the top?

bistline:
> Minor papercut that should be fixed before release.
> 
> If I bookmark a study search (example "Homo sapiens" and "B cell") and gene search "gapdh" and save it. Then change gene search to "mapk1" and save it. When I use the "gapdh" bookmark, the URL and plots all say "gapdh" but the gene search field still says "mapk1"...
> 
> Let me know if a screencapture would be more helpful.

I noticed this as well.  The issue is that there is no synchronization between this gene search bar and the URL, except on a form post.  It does do this correctly on study-based gene searches, so I'll see if I can copy the paradigm there.  It might be part of the `updateExploreParams` infrastructure though, so it possibly could be more trouble than it's worth.

> 
> Also, I don't have "strong" thoughts on this but should new bookmarks should show up at the bottom of the bookmarks list or the top?

Chrome and most browsers show bookmarks in order of oldest first.  It makes sense (at least to me) to mimic this behavior.

bistline:
@jlchang [fd052b9](https://github.com/broadinstitute/single_cell_portal_core/pull/2017/commits/fd052b97e1d5d7f08fcf4e728a62283e80a0ba5f) this should work properly now, can you give it a try?

app/javascript/components/HomePageContent.jsx
```diff
@@ -34,16 +35,19 @@ const LinkableSearchTabs = function(props) {
...
+            <span className="fas fa-book"></span> Search studies
+          </Link>
+          <Link to={`${basePath}/app/genes${location.search}`}
+                className={showGenesTab ? 'active' : ''}>
+            <span className="fas fa-dna"></span> Search genes
+          </Link>
+          <span id='home-page-bookmark'>
+            <BookmarkManager eagerLoad={true} />
+          </span>
+        </nav>
```
eweitz, line 50: Could you de-indent this block to use the standard SCP JS 2-space indent?  That'd denoise the changeset.

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -246,6 +246,31 @@ export default function BookmarkManager({bookmarks, studyAccession, clearExplore
   const [serverBookmarksLoaded, setServerBookmarksLoaded] = useState(false)
   const [showBookmarksModal, setShowBookmarksModal] = useState(false)
 
+  /** load bookmarks from server on initialization */
+  async function eagerLoadBookmarks() {
+    setHasEagerLoaded(true) // prevent thundering herd
```
eweitz, line 251: Could you elaborate a bit on how this would thundering-herd?

app/javascript/components/explore/ExploreDisplayPanelManager.jsx
```diff
@@ -492,13 +497,24 @@ export default function ExploreDisplayPanelManager({
...
+              <button onClick={copyLink}
+                      className="action action-with-bg"
+                      data-toggle="tooltip"
+                      title="Copy a link to this visualization to the clipboard">
+                <FontAwesomeIcon icon={faLink}/> Get link</button>
+              {exploreInfo?.bookmarks &&
+                <BookmarkManager
+                  bookmarks={exploreInfo.bookmarks}
+                  studyAccession={studyAccession}
+                  clearExploreParams={clearExploreParams}/>
```
eweitz, line 515: Potential lurking bug: `clearExploreParams` is no longer a parameter in the `BookmarkManager` function.

app/javascript/components/bookmarks/BookmarkManager.jsx
```diff
@@ -15,15 +13,17 @@ import { log } from '~/lib/metrics-api'
  * Component to bookmark views in Explore tab, as well as existing link/reset buttons
  *
  * @param bookmarks {Array} existing user bookmark objects
- * @param clearExploreParams
+ * @param studyAccession {String} currently loaded study, if present
+ * @param eagerLoad {Boolean} load bookmarks from server on initialization
```
eweitz, line 17: Can you help me see why `eagerLoad` is true on the Home page, and false on the Study Overview page?

eweitz (approved):
Code and UI look good overall!  My local manual test passed.

Some UI notes on the bookmark star, all non-blocking:

1. Showing the star _conditionally_ might be better.  It is currently shown in the default state, even when the user is viewing an unchanged page and has no saved bookmarks.  Displaying the star _only when the user is in a non-default state_ would A) likelier catch their attention, and B) reduce visual noise in the default state -- no changes, no other bookmarks -- as this feature doesn't apply there.  (Similar principles relate for "Reset view" and "Get link".)  This is a pre-existing matter, but incidentally relevant.

2. Having the star display beside the search type tabs ("Search studies", "Search genes") seems slightly out-of-place.  The bookmark star is a way to save queries.  Putting the bookmark star in the query summary (shown after executing a search) strikes me as more UX-coherent.  Alternatively and better than current, I think, would be putting the star to the right of the query input control row, i.e. to the right of "Search by text" / "Search by genes".  This suggestion is of minor importance and somewhat subjective, but struck me as useful food for thought.

To be clear, I'd be fine if this were merged as is, it's a good enhancement!

For convenient reference, here's how the current PR's UI looks on my local machine:

Bookmark_star_global_search__SCP_2024-04-18

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: This was a side effect of doing the loading via the `useEffect` with the empty dependency array - this lead to the async API call to the `/bookmarks` endpoint firing multiple times until the response came back and the state was set.  Having a preemptive short-circuit here cut that down to one and only one.  There might be a cleaner way of doing this, but I found this worked well enough.

app/javascript/components/explore/ExploreDisplayPanelManager.jsx
bistline: Nice catch!

app/javascript/components/bookmarks/BookmarkManager.jsx
eweitz: Thanks!

app/javascript/components/bookmarks/BookmarkManager.jsx
bistline: On the Study Overview, we get the bookmarks back from the `study_user_info` API call.  The `BookmarkManager` component isn't rendered until this comes back.  However, there's no analogous API call on the home page, so I added in the explicit call to load the bookmarks when the component is initialized.  This allows the state to set correctly here (e.g. showing if a query has already been bookmarked) without the user having to open the bookmarks modal first.

eweitz: Much appreciated, this is clear to me now.

bistline:
> 1. Showing the star _conditionally_ might be better.  It is currently shown in the default state, even when the user is viewing an unchanged page and has no saved bookmarks.  Displaying the star _only when the user is in a non-default state_ would A) likelier catch their attention, and B) reduce visual noise in the default state -- no changes, no other bookmarks -- as this feature doesn't apply there.  (Similar principles relate for "Reset view" and "Get link".)  This is a pre-existing matter, but incidentally relevant.

If I move it down to the query summary like you mentioned below, I think this automatically takes care of that problem.

> 2. Having the star display beside the search type tabs ("Search studies", "Search genes") seems slightly out-of-place.  The bookmark star is a way to save queries.  Putting the bookmark star in the query summary (shown after executing a search) strikes me as more UX-coherent.  Alternatively and better than current, I think, would be putting the star to the right of the query input control row, i.e. to the right of "Search by text" / "Search by genes".  This suggestion is of minor importance and somewhat subjective, but struck me as useful food for thought.

I _completely_ agree, and I tried at one point to render the bookmark icon to the right of the keyword searches, but then I had to include it _twice_, and their state was managed independently.  That ended up being very confusing.  Otherwise I'd have to thread it all the way down the stack, and I didn't want to add it as a parameter to every component on the way down.

That being said, putting it in the query summary like you suggested means it would end up in the `ResultsPanel` component.  That's shared between both searches, and I'd only need to thread it down a few levels.  Let me see if I can put it in there without having to tear up the stack too much, otherwise I'll ticket it as a future enhancement.

jlchang:
> @jlchang [fd052b9](https://github.com/broadinstitute/single_cell_portal_core/pull/2017/commits/fd052b97e1d5d7f08fcf4e728a62283e80a0ba5f) this should work properly now, can you give it a try?

It didn't seem to repopulate the gene search field when I tested. Curiously, if I changed the content of the gene search field in the cross-study search tab (aka "global" gene search), it would _sometimes_ modify the Bookmark name field. (Coincidentally, when that happens the star stays filled even though the new gene searched has not been bookmarked.

https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/67fea8db-f04f-4fb4-942c-0f4a9e8d096d

jlchang (approved):
Use case:  jump back and forth between two different cross study gene searches. 

There are two UI consistency papercuts:
- the state of the star 
- content in the gene search field

They don't affect the functionality of persisting the searches I want to compare so I don't consider them blocking for merging this PR. You can create a follow-on ticket to address the papercuts if that's desirable - I'm approving the PR so you can go that route.

The screencapture is long so I've noted down the relevant timepoints in the screencapture:

https://github.com/broadinstitute/single_cell_portal_core/assets/6005195/3b0757e3-970f-4e29-a5ce-bd671dfa27d4


0:55 create bookmark
Metadata contains (species: Mus musculus) AND (cell type: B cell) + gene APOC3
star fills, looks great.

1:09 maintain study search, change gene search to APOE
star unfilled, as expected

1:43 create bookmark
Metadata contains (species: Mus musculus) AND (cell type: B cell) + gene APOE
star fills, still looking good

1:46 use bookmarks to return to APOC3 cross-study gene search
star **unfilled** (this search is bookmarked, I used a bookmark to get here)
gene **search field still says apoe** (URL shows apoc3)

1:58 use clear all to reset URL (note: naive user may try to reload the page which will not clear the gene search)
At this point I'm a bit at a loss how to access my bookmarks (the downside of conditional display of the star, if I already have bookmarks already and I'm on the home page w/o a results summary to access My bookmarks...)

out-of-PR-scope suggestion: My Bookmarks, in the dropdown with My annotations & My studies? (maybe a non-starter because My Bookmarks isn't a page?)

2:12 Anyways so I flounder a bit before I go to a study, just to access My bookmarks

2:38 Visit My bookmarks to return to APOC3 cross-study gene search again
**star unfilled** 
gene search field says apoc3

3:02 Use My bookmarks to visit apoe cross-study gene search
star filled 
gene **search field still says apoc3** (URL shows apoe)

bistline:
@jlchang ok the gene search bar thing is bizarre because that _was_ working.  I'm not sure what changed but I can replicate your findings.  Regarding the bookmark icon - it seems I can trigger that happening the very first time as you showed, but after both bookmarks are saved and I navigate back to them, I can't trigger it again.  So something else state-wise is happening there.  I'll try to see if I can figure it out.

jlchang:
I noticed that when I used a bookmark to get to a bookmarked cross-study gene search and then switch to the study search tab, I can end up with a URL:

> https://localhost:3000/single_cell/app/studies?type=study&page=1&**genes=gapdh**&genePage=1&facets=species%3ANCBITaxon_9606%3Bcell_type%3ACL_0000236

Will that genes parameter in a study search URL cause anything unexpected?

bistline:
@jlchang [1c8b8ec](https://github.com/broadinstitute/single_cell_portal_core/pull/2017/commits/1c8b8ec57cc7881c4548120bf4a9d00d2017201b) both of those issues should be addressed now.

bistline:
> I noticed that when I used a bookmark to get to a bookmarked cross-study gene search and then switch to the study search tab, I can end up with a URL:
> 
> > https://localhost:3000/single_cell/app/studies?type=study&page=1&**genes=gapdh**&genePage=1&facets=species%3ANCBITaxon_9606%3Bcell_type%3ACL_0000236
> 
> Will that genes parameter in a study search URL cause anything unexpected?

It won't break anything, however it won't match an existing bookmark unless you save it in that state.  We can keep an eye on this to see if it happens more often.

jlchang (approved):
Thanks for addressing those two papercuts - I can confirm that they're both resolved.

PR #2019, opened 2024-04-24
Fixing bookmark icon positioning in spatial studies (SCP-5615)

bistline:
#### BACKGROUND & CHANGES
This fixes an existing issue where the bookmark icon line wraps when viewing spatial studies.  Now, the positioning of the bookmark icon mimics that of all other studies.  It is worth noting that at smaller viewports, the bookmark/get link icons will wrap automatically, but now the behavior for spatial & non-spatial studies is the same.

#### MANUAL TESTING
1. Boot as normal
2. In two tabs, load a spatial & non-spatial study
3. Confirm the position of the bookmark icon is the same, and that both wrap at the same viewport width.


eweitz (approved):
Code and local manual test look good!

### Before

Bookmark_star_spatial_study_before__SCP_2024-004-24

### After

Bookmark_star_spatial_study_after__SCP_2024-004-24

jlchang (approved):
Bookmarks icon looks fine. Sorry for the delay - took me a while to track down a spatial study in my local instance!

PR #2022, opened 2024-04-25
Removing differential expression from sync UX

bistline:
#### BACKGROUND & CHANGES
This update removes the file type `Differential Expression` from the sync page.  The reasoning is that the UX is complicated, and requires specialized handling done in the Upload wizard to gather specific author DE fields.  The lack of this information caused all sync actions of DE files to fail in an unrecoverable way, often requiring an administrator to intervene.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Open the Google bucket for any study (accessible from `My studies` > `Details` for any study) and upload a new file, like anything from `test/test_data`
3. Back on the `My studies` page, select `Sync workspace files` from the ellipsis menu for that study
4. Once the form renders, note that `Differential Expression` is no longer in the list

### DIFF
--- a/app/models/study_file.rb
+++ b/app/models/study_file.rb
@@ -22,7 +22,7 @@ class StudyFile
   mount_uploader :upload, UploadUploader, mount_on: :upload_file_name
 
   # constants, used for statuses and file types
-  STUDY_FILE_TYPES = ['Cluster', 'Coordinate Labels' ,'Expression Matrix', 'MM Coordinate Matrix', '10X Genes File',
+  STUDY_FILE_TYPES = ['Cluster', 'Coordinate Labels','Expression Matrix', 'MM Coordinate Matrix', '10X Genes File',
                       '10X Barcodes File', 'Gene List', 'Metadata', 'Fastq', 'BAM', 'BAM Index', 'Documentation',
                       'Other', 'Analysis Output', 'Ideogram Annotations', 'AnnData', 'Seurat',
                       'Differential Expression'].freeze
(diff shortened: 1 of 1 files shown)


jlchang (changes requested):
Confirmed that Differential Expression is no longer in the Sync dropdown. 

Since we can now upload bucket files using the wizard (for Reference AnnData files), we should seriously consider removing AnnData from the list since we can't do visualizable AnnData ingest via sync. Alternatively, we could change `AnnData` to `AnnData - other` but I suspect that'd may not be interpreted correctly by a new user.

bistline:
> Confirmed that Differential Expression is no longer in the Sync dropdown.
> 
> Since we can now upload bucket files using the wizard (for Reference AnnData files), we should seriously consider removing AnnData from the list since we can't do visualizable AnnData ingest via sync. Alternatively, we could change `AnnData` to `AnnData - other` but I suspect that'd may not be interpreted correctly by a new user.

That's a good point - let me see if I can figure out a good way to convey that in the Sync UX for AnnData.

bistline:
@jlchang I went this route since syncing a reference AnnData file is still a valid use case.  Instead of removing it, I added this disclaimer when you select `AnnData` as a file type.

Screenshot 2024-04-29 at 11 18 41 AM

eweitz (approved):
Code and UI in manual test look good!

jlchang (approved):
Thanks so much for considering the AnnData case. The notice seems like a good way to provide the necessary context.

PR #2023, opened 2024-04-25
Remove user annotations from global gene search requests (SCP-5617)

bistline:
#### BACKGROUND & CHANGES
This update removes all user-scoped annotations from the `Annotation` dropdown on global gene search requests.  The reasoning is that due to caching, these annotation lists are stored along with the plot data for performance.  However, user annotations have permissions associated with them, meaning the first user that requests a particular study/gene will then cache the list of annotations they can see of _all_ subsequent requests for that study/gene.  If the user is an admin, this means that _all_ users can see _all_ user annotations for that study/gene.  While some user annotations are made public, not all are, so this constitutes a small data leak.  Additionally, computing the list of available user annotations for admins is a costly operation, and removing these from global gene search has the added benefit of speeding up search requests for admin users.

#### MANUAL TESTING
1. Boot as normal and sign in with an admin account
2. Ensure that you have at least one saved user annotation on the `Human milk - differential expression` study
3. Run a search for that study's accession, and then click over to the `Search genes` tab
4. Search for a gene, and then in the `Annotation` dropdown, validate that you do not see any user annotations listed

### DIFF
--- a/app/controllers/api/v1/visualization/expression_controller.rb
+++ b/app/controllers/api/v1/visualization/expression_controller.rb
@@ -136,8 +136,7 @@ module Api
               cluster: @cluster,
               selected_annotation: @annotation,
               boxpoints: params[:boxpoints],
-              consensus: params[:consensus],
-              current_user: current_api_user
+              consensus: params[:consensus]
             )
             render json: render_data, status: :ok
           end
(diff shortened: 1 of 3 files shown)


jlchang (approved):
Confirmed that user annotations no longer show up as annotation options in cross-study gene search.

eweitz (approved):
Code looks good, kudos on an economical solution.

PR #2024, opened 2024-04-26
Cell filtering: fix panel state on cluster change, initial links (SCP-5591, SCP-5610)

eweitz:
This fixes issues with panel state and certain links, which manifest in some non-default cell filtering scenarios.

#### Fix panel state after "Clustering" change
Previously, in some studies, changing the "Clustering" menu would set the cell filtering panel into a subtly incorrect state.  Annotations in the default cluster -- but not the new cluster -- would still appear as available facets and filters in the panel.  Cell filtering in those residual annotations would hide the whole plot, or include an "undefined" trace after a clustering change.

Now, only annotations in the current cluster are shown in the cell filtering panel.  That prevents issues caused by errant residual annotations.  This fixes SCP-5591.

Here's a view of the problem, and its fix.

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/68f235e1-8cd5-4136-9ef7-2d5fa8943fe1

#### Fix initial cell filtering links
Previously, in #2009, changing your cell filtering selection updated your browser address bar with a URL that specified your selection for _all_ numeric facets -- even ones that were not changed.  Handling for those URLs assumed _all_ numeric facets would be so specified.  However, prior to PR 2009, _no_ numeric facets were specified -- only categorical facets were loadable via URL.  So those old-style URLs (with only categorical facets specified) broke the page when they were loaded.  

Although high severity when it occurred, this had very low incidence, in practice essentially only affecting views linked from a [feature announcement](https://singlecell.broadinstitute.org/single_cell/features/2024-01-04-filter-plotted-cells-for-a-deeper-dive-into-study-data) and [documentation](https://singlecell.zendesk.com/hc/en-us/articles/21415002866587-Filter-plotted-cells).  Incidence was 2-3 study views per day out of 1686, or 0.2% of study page views.  

Now, old cell filtering URLs work as expected.  URLs minted in the last two weeks work as well, including those with changed numeric facets.  This fixes SCP-5610.  These changes also fix two other `facets` URL loading UX bugs.  PR 2009 also broke sparkbars when when loading via `facets` URL parameter.  And the "reset all facets" button seems to have never worked in that particular scenario.  Those now also work seamlessly with URLs.

Here's how it looks:

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/090886ba-6223-4431-bdbc-7f86405f4209

### Test
If you'd like to manually verify these fixes, steps are below.

For panel state after "Clustering" change:
1.  Go to "Cellular and transcriptional diversity over the course of human lactation" study
2.  Click "Filter plotted cells"
3.  Deselect "Cell type: epithelial"
4.  Click "<-" button to exit panel
5.  Click "Filter plotted cells"
6.  Confirm "Cell type: epithelial" is selected, and plot shows default state
7.  Exit panel
8.  Select "Epithelial Cells UMAP" in "Clustering"
9.  Confirm no "Cell type" facet is not shown

For initial cell filtering links:
1.  Note the accession for your "Cellular and transcriptional diversity over the course of human lactation" study
2. In your browser address bar, paste: https://localhost:3000/single_cell/study/$YOUR_STUDY_ACCESSION/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation?genes=TNC&cluster=All%20Cells%20UMAP&spatialGroups=--&annotation=General_Celltype--group--study&subsample=all&hiddenTraces=--Filtered--&facets=milk_stage--group--study%3Alate_1%7Clate_2%7Clate_3%7Clate_4%7Cmature%7CNA%7Ctransitional%20%7Ctransitional#study-visualize
3. Confirm plots appear as shown in in 2nd image in [related documentation](https://singlecell.broadinstitute.org/single_cell/features/2024-01-04-filter-plotted-cells-for-a-deeper-dive-into-study-data)

This satisfies SCP-5591 and SCP-5610.

### DIFF
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -40,9 +40,8 @@ function getHasNondefaultSelection(selectionMap, facets) {
   const entries = Object.entries(selectionMap)
   for (let i = 0; i < entries.length; i++) {
     const [selectedFacet, selection] = entries[i]
-
     const facet = facets.find(f => f.annotation === selectedFacet)
-    if (!facet) {return false}
+    if (!facet) {continue}
     let normDefault = facet.defaultSelection
     let normSelection = selection
     if (facet.type === 'group') {
@@ -50,6 +49,7 @@ function getHasNondefaultSelection(selectionMap, facets) {
       normDefault = new Set(normDefault)
       normSelection = new Set(normSelection)
     }
+
     if (!_isEqual(normDefault, normSelection)) {
       return true
     }
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -47,6 +47,12 @@ export function getHasNondefaultSelection(selectionMap, facets) {
     if (facet.type === 'group') {
       // Normalize categorical filters, given order doesn't matter for them
       normDefault = new Set(normDefault)
+      if (normDefault.size === 1) {
+        // Skip considering ineligible categorical facets, like those with
+        // only 1 group.
+        continue
+      }
+
       normSelection = new Set(normSelection)
     }
 
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -36,7 +36,7 @@ export function CellFilteringPanelHeader({
 }
 
 /** Determine if user has deselected any filters */
-function getHasNondefaultSelection(selectionMap, facets) {
+export function getHasNondefaultSelection(selectionMap, facets) {
   const entries = Object.entries(selectionMap)
   for (let i = 0; i < entries.length; i++) {
     const [selectedFacet, selection] = entries[i]
--- a/app/javascript/components/explore/CellFilteringPanel.jsx
+++ b/app/javascript/components/explore/CellFilteringPanel.jsx
@@ -301,7 +301,7 @@ function CellFacet({
     return <></>
   }
 
-  const selection = selectionMap[facet.annotation]
+  const selection = selectionMap[facet.annotation] ?? facet.defaultSelection
 
   return (
     <div
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -135,7 +135,6 @@ function getCellFacetingData(cluster, annotation, setterFunctions, context, prev
             setCellFaceting(newCellFaceting)
           }
 
-
           // The cell filtering UI is initialized in batches of 5 facets
           // This recursively loads the next 5 facets until faceting is fully loaded.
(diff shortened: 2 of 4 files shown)


jlchang (approved):
I confess I expected a different UX. I was hoping that closing the filter plotted cells panel could now allow the filtered view to persist (this would allow users to bookmark the filtered state).  You can imagine that if you've built up a complex filter, it might be nice to be able to apply it to a different clustering (the issue there, for us, is when there are non-applicable filters remembered by the URL). I'd be happy to chat more about whether the bugfix should work to support bookmarking. I'm also okay with the UX that this bugfix will support if we're planning to elevate filtering to the top level soon anyways.

I initially thought the `initial cell filtering links` test was failing for me. Turns out you need to replace both the accession AND the study name if your local instance is named differently. If the study name is incorrect, the portal replaces your URL request with the default URL for the study. As discussed at triage recently, this is not the expected behavior, [SCP-5619](https://broadworkbench.atlassian.net/browse/SCP-5619) opened to track the issue.

eweitz:
> I confess I expected a different UX. I was hoping that closing the filter plotted cells panel could now allow the filtered view to persist (this would allow users to bookmark the filtered state).

That'd be a nice enhancement!  But it's separate, as you intuit.  These bugs would have existed in that UX without these fixes.  

I've added cross-panel persistence for cell filtering as [an engineering strategy topic](https://docs.google.com/document/d/11eNlwMjwKoUcx4TQflzxw4DuN9q03fNNLphUuAVNIc4/edit), as it would be useful but non-trivial.  

> Turns out you need to replace both the accession AND the study name if your local instance is named differently.

Good find, I've updated the instructions accordingly.  Thanks for opening that new issue.

bistline (changes requested):
I am unable to get the UX to behave correctly (see attached video).  While the state issues with overlapping `/facets` requests are pre-existing and documented ([SCP-5379](https://broadworkbench.atlassian.net/browse/SCP-5379)), this update seems to have exacerbated that problem somehow.  If I attempt to interact with the cell filtering UX at all while requests are still coming in, the UI quickly gets into an inconsistent state, and eventually to a point that seems unrecoverable.  While not shown in the below video, roughly half the time I ended up in an error state that blew up the UI.

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/e9f3aed4-a5e4-4116-9681-717876df23df

In the JS console, I see numerous instances of the same two errors/warnings:
```
@sentry_react.js?v=e0bc2670:4993 Uncaught TypeError: Cannot read properties of undefined (reading 'forEach')
    at cell-faceting.js:749:15
```
and 
```
@sentry_react.js?v=e0bc2670:774 Warning: A component is changing a controlled input of type checkbox to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component.
```

bistline (approved):
After conferring, I agree that the above issue is actual already present in `development` and as such is pre-existing.  Waiting for all `/facets/` requests to finish does let the UI behave normally.  Since this does address the actual issue, I'm approving now such that we can pivot to addressing state issues.

PR #2025, opened 2024-04-29
Reject XSS-based requests to avoid caching invalid requests (SCP-5620)

bistline:
#### BACKGROUND & CHANGES
Some new security scans of DSP-based applications have resulted in some visualization routes caching responses with parameters containing XSS-based values like `script` or `xssdetected`.  While this update doesn't address the underlying issue (which would entail a large amount of work as it involves using the [`require-trusted-types-for`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) CSP header and associated core JS package updates, like jQuery and React), this does prevent attempting to cache these obviously bogus requests.  The goal is to prevent incidences of the "[cache path bug](https://broad-institute.sentry.io/issues/5184436549/?project=1424198)" where percent-encoded values get broken by a `/` path terminator, causing errors any time a visualization cache is cleared.  Since this affects all users/studies portal-wide, preventing this error from occurring is paramount.

#### MANUAL TESTING
1. Boot as normal
2. Ensure caching is turned on for your local instance
3. Load the Swagger documentation for any visualization API endpoint that accepts parameters, like [loading cluster viz data](https://localhost:3000/single_cell/api/v1#/Visualization/study_cluster_path).
4. Enter a valid study accession, but for `cluster_name` use either `xssdetected` or `script` and then submit the request
5. Ensure you receive a `400: Bad Request` response, and in the server log output find the corresponding entry to note that neither `check_api_cache!` nor `write_api_cache!` are invoked:
```
Started GET "/single_cell/api/v1/studies/SCP13/clusters/xssdetected" for 127.0.0.1 at 2024-04-29 15:15:58 -0400
  Parameters: {"study_id"=>"SCP13", "cluster_name"=>"xssdetected"}
Filter chain halted as :validate_cache_request rendered or redirected
Completed 400 Bad Request in 41ms (MongoDB: 0.0ms | Allocations: 1645)
```
6. Enter a valid request (like for the `_default` cluster) and ensure you get a `200` response, and that either `:check_api_cache!` or `write_api_cache!` are invoked:
```
Started GET "/single_cell/api/v1/studies/SCP13/clusters/_default" for 127.0.0.1 at 2024-04-29 15:16:56 -0400
  Parameters: {"study_id"=>"SCP13", "cluster_name"=>"_default"}
Writing to API cache: _single_cell_api_v1_studies_SCP13_clusters__default_4803421a30c68e74735896abc8eeef5bf578466df8cd2ad940a9b39906615017
Completed 200 OK in 1066ms (Views: 18.4ms | MongoDB: 0.9ms | Allocations: 62097)
```
7. Enter a Rails console session, and confirm you can clear the caches for the above study without error:
```
CacheRemovalJob.new('SCP13').perform
=> 55

```

### DIFF
--- a/app/controllers/api/v1/visualization/annotations_controller.rb
+++ b/app/controllers/api/v1/visualization/annotations_controller.rb
@@ -94,9 +94,6 @@ module Api
             response 200 do
               key :description, annotation_description_doc
             end
-            response 400 do
-              key :description, 'Bad request'
-            end
             extend SwaggerResponses::StudyControllerResponses
           end
         end
@@ -144,9 +141,6 @@ module Api
             end
             response 200 do
               key :description, '2-column TSV of cell names and their values for the requested annotation.  Column headers are NAME (the cell name) and the name of the returned annotation'
-            end
-            response 400 do
-              key :description, 'Bad request'
             end
             extend SwaggerResponses::StudyControllerResponses
           end
@@ -240,9 +234,6 @@ module Api
                   end
                 end
               end
-            end
-            response 400 do
-              key :description, 'Bad request'
             end
             extend SwaggerResponses::StudyControllerResponses
           end
@@ -339,9 +330,6 @@ module Api
             end
             response 200 do
               key :description, '2-column TSV of column header names from the gene list file.  Column headers are NAME (the cell name) and the name of the gene list'
-            end
-            response 400 do
-              key :description, 'Bad request'
             end
             extend SwaggerResponses::StudyControllerResponses
           end
--- a/app/controllers/api/v1/visualization/annotations_controller.rb
+++ b/app/controllers/api/v1/visualization/annotations_controller.rb
@@ -14,6 +14,7 @@ module Api
         before_action :set_study
         before_action :check_study_view_permission
         # don't cache the annotation list, since it is user dependent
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/visualization/annotations_controller.rb
```diff
@@ -330,6 +340,9 @@ def facets
             response 200 do
               key :description, '2-column TSV of column header names from the gene list file.  Column headers are NAME (the cell name) and the name of the gene list'
             end
+            response 400 do
```
eweitz, line 343: Is there an easy way to abstract this duplicative code?  It'd help denoise.  If not, that's fine!

eweitz (approved):
Code looks good!

app/controllers/api/v1/visualization/annotations_controller.rb
bistline: Yeah there is, I was just being lazy...

jlchang (approved):
Confirmed that 'xss' or 'script' triggers `400: Bad Request` with 
`Filter chain halted as :validate_cache_request rendered or redirected` in logs.

PR #2030, opened 2024-05-02
Adding integration for Rawls#checkBucketReadAccess after study creation (SCP-5625)

bistline:
#### BACKGROUND
Terra uses groups for managing access to resources, including workspaces, buckets, etc.  An [upstream change](https://cloud.google.com/iam/docs/access-change-propagation) in how permissions propagate in Google meant that policies managed by groups - while always eventually consistent - could take up to several hours to take effect.  This led to users creating studies in non-default billing projects sometimes being unable to upload files for 3-5 hours (this appears to not be a problem in our [default billing project](https://app.terra.bio/#billing?selectedName=single-cell-portal&type=project) as the main service account is a direct owner and inherits that permission on all workspaces/buckets in that project).  Terra solved this problem by creating the FastPass service, which will add direct user grants to buckets upon request (this is what happens whenever a user signs into a workspace via the Terra UI).  These policies take effect within 2-7 minutes, which unblocks the user.  These policies are then eventually removed by a background polling mechanism.

#### CHANGES
Now when a user creates a study (regardless of the billing project), a background process will issue a request to this Rawls [API endpoint](https://rawls.dsde-prod.broadinstitute.org/#/workspaces/readBucket) that initiates the FastPass process.  If the response is `200`, this means that bucket access for the service account has already been established.  A `403` response indicates that the policy hasn't made it to the bucket yet, and Terra will add the service account directly as a user.  In either case, no further action is required, and a user will be able to add a file.  All other error scenarios follow standard reporting paths (Sentry, ingest auto-upload failure emails).

#### MANUAL TESTING
1. Boot all services (including DelayedJob) and sign in
2. Open `development.log` in the Console app
3. Create a new study in a non-default project where your service account is not a direct owner, like `single-cell-portal-test`
4. In the logfile, look for an entry like the following:
```
Checking bucket access on single-cell-portal-test/fastpass-test with tracking identifier: default-service-account@broad-singlecellportal-bistlin.iam.gserviceaccount.com

Permissions not yet synchronized for single-cell-portal-test/fastpass-test, FastPass request initiated
```
5. Open a new Rails console session and load the new study:
```
study = Study.last
```
6. Check the bucket access policy again, and confirm that you get `true` as the response (calling `study.check_bucket_read_access` with the suffix `_without_delay` will run this process in the foreground instead of the normal background process):
```
study.check_bucket_read_access_without_delay
=> true
```

### DIFF
--- a/app/models/fire_cloud_client.rb
+++ b/app/models/fire_cloud_client.rb
@@ -22,6 +22,7 @@ class FireCloudClient
   # base url for all API calls
   BASE_URL = 'https://api.firecloud.org'.freeze
   BASE_SAM_SERVICE_URL = 'https://sam.dsde-prod.broadinstitute.org'.freeze
+  BASE_RAWLS_SERVICE_URL = 'https://rawls.dsde-prod.broadinstitute.org'.freeze
   # default auth scopes for client tokens
   GOOGLE_SCOPES = %w[
     https://www.googleapis.com/auth/userinfo.profile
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good!  Kudos on having all studies ready for upload in minutes, not hours.

Relevant, but not necessarily in scope here: 

How feasible would it be to log issues like these as `ingest` failures in Mixpanel?  

These events technically occur between completion of file upload and instantiation of IngestJob, but they seem useful to group with "ingest" failures.  That'd help compare incidence to other [upload and ingest failures](https://mixpanel.com/project/2120588/view/19059/app/boards#id=734187), and intuitively answer questions like "Do we have any examples of this error from a while ago?" and "What happened around that time?"

Admin emails and Sentry give some insight for errors like this `Google::Cloud::PermissionDeniedError` case.  Mixpanel logging for these would significantly improve their observability.

bistline:
> Code looks good! Kudos on having all studies ready for upload in minutes, not hours.
> 
> Relevant, but not necessarily in scope here:
> 
> How feasible would it be to log issues like these as `ingest` failures in Mixpanel?
> 
> These events technically occur between completion of file upload and instantiation of IngestJob, but they seem useful to group with "ingest" failures. That'd help compare incidence to other [upload and ingest failures](https://mixpanel.com/project/2120588/view/19059/app/boards#id=734187), and intuitively answer questions like "Do we have any examples of this error from a while ago?" and "What happened around that time?"
> 
> Admin emails and Sentry give some insight for errors like this `Google::Cloud::PermissionDeniedError` case. Mixpanel logging for these would significantly improve their observability.

I thought about this for a bit... as discussed at standup, I'm going to attempt to create a two event funnel of capturing the initial `403` from the FastPass request to any downstream `Google::Cloud::PermissionDeniedError` instances for a given study.

bistline:
I've [ticketed](https://broadworkbench.atlassian.net/browse/SCP-5632) the work for Mixpanel logging as it is both more complicated than I thought, and also a potential opportunity to have better visibility into background errors in terms of user flow/impact.

jlchang (approved):
Functional review - didn't see the permission error but did see:

> Permissions not yet synchronized for single-cell-portal-test/pr2030, FastPass request initiated

Got bucket read access very quickly.

bistline:
> Functional review - didn't see the permission error but did see:
> 
> > Permissions not yet synchronized for single-cell-portal-test/pr2030, FastPass request initiated
> 
> Got bucket read access very quickly.

Actually, that's correct as far as logging goes - I just realized the comment was from earlier work that printed the upstream Google error.

PR #2031, opened 2024-05-06
Conditionally position bookmark icon (SCP-5630)

bistline:
#### BACKGROUND & CHANGES
The position of the bookmark star icon can be affected numerous ways depending on what the lowest element is in the "Options" panel.  This could lead to either the icon wrapping to another line, or being positioned too close/far to the last element in the panel.  This update adds conditional logic to determine what html/css markup is necessary so that the positioning of this and the "Get link" & "Reset view" links are consistent.  There are 3 main scenarios to account for:

1. There are buttons for cell filtering or differential expression shown
filter_cells

2. The subsampling menu is shown
subsample

3. Genes have been searched
color_scale

#### MANUAL TESTING
1. Boot as normal, sign in and load 3 studies in separate panels:
    * Human milk - differential expression
    * Any study with spatial data, or a study with cell filtering and differential expression disabled
    * Any study with subsampled clustering
2. Confirm that the bookmark icon position is consistent across all three studies
3. Search for one or more genes in all 3 studies and confirm positioning remains consistent

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
+++ b/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
@@ -335,6 +335,10 @@ export default function ExploreDisplayPanelManager({
     setRenderForcer({})
   }, 300)
 
+  // decide how to position the bookmark icon, depending on menu elements shown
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Functional review - icon in the right place, every time!

PR #2033, opened 2024-05-07
Add concurrency groups to deployment jobs (SCP-5633)

bistline:
#### BACKGROUND & CHANGES
This adds concurrency groups to both the [staging](https://github.com/broadinstitute/single_cell_portal_core/actions/workflows/deploy-to-staging.yml) & [production](https://github.com/broadinstitute/single_cell_portal_core/actions/workflows/deploy-to-production.yml) deployment workflows.  This will prevent concurrent runs of either workflow, as two simultaneous deployments to the same host will always cause the downstream one to fail.  Additionally, the concurrency group allows for only one pending job.  Any additional pending jobs will be cancelled.  Since the outcome of any deployment is identical - it pull the specified branch on the specified host and restarts the container - it will not matter that a later workflow is cancelled provided _any_ deployment runs after the final merge.

This also updates all of our Github Actions workflows to use the most current versions of dependent actions available.  This is in regards to the deprecation of Node.js v16 in GHA.  We still will run our version of Node inside the Docker container.

#### MANUAL TESTING
This can only be tested after this PR is merged, as any changes to the workflow YAML will not yet be integrated.  Once it is merged, then a manual deployment to staging should be pended until the first deployment finishes, and any subsequent merges/deployments will be cancelled until there are no other pending jobs.

### DIFF
--- a/.github/actions/docker-build-env-setup/action.yml
+++ b/.github/actions/docker-build-env-setup/action.yml
@@ -11,7 +11,7 @@ runs:
       uses: ./.github/actions/install-vault-and-utils
     - name: Detect changes to Dockerfile
       id: changed-dockerfile
-      uses: tj-actions/changed-files@v35.4.4
+      uses: tj-actions/changed-files@v44.3.0
       with:
         files: Dockerfile
     - name: Rebuild local image if Dockerfile changed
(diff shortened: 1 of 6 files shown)


jlchang (approved):
Makes sense to me!

eweitz (approved):

PR #2053, opened 2024-06-12
Filter IGV by cell annotations, tailor ATAC rendering (SCP-5612, SCP-5614)

eweitz:
This enables filtering IGV features by cell annotations, and refines how IGV shows ATAC.  It enriches multiome exploration.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/f8060e12-75d6-4a8a-89ea-f5b41b3bca2f

(Filtering is slow.  It's behind a flag; next steps include optimization.)

### Impact
This lets users see how chromatin accessibility differs in _custom-filtered_ cell populations, in genomic context.  So if your multiome data has e.g. cell type annotations, you can see how transcribable any gene of interest is in cells filtered to show only e.g. B cells or macrophages.  

Because it's multiome, these genomic insights can be related to expression insights for the same _filtered_ cell population, e.g. plots showing UMAP clusters and categorical distributions.

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2041), users could upload scATAC-seq fragments as BED files and view them in igv.js.  However, our IGV rendered ATAC data as generic BED, which obscured relevant scientific details.  It also didn't integrate with powerful SCP exploration features.  ATAC fragment BEDs in IGV integrated with gene search, but not cell filtering.

Now, multiome IGV has better default rendering, and integrates cell filtering from an earlier [proof of concept](https://github.com/broadinstitute/single_cell_portal_core/pull/2021).  This also removes the "Genome" tab from the unfiltered view, but retains it in the single-gene view.

#### Tailored rendering
Per [discussion](https://broadinstitute.slack.com/archives/CESEYJW9W/p1717099893194269) with the Pipeline Development team, instead of coloring ATAC fragments by BED "score" with a random palette, fragment colors are now meaningful.  A sequential palette colors fragments with higher "scores" increasingly bold red.  The popover shown on clicking ATAC features now labels field names more specifically: "Name" is "ATAC barcode", and "Score" is "Read count".  Other IGV feature properties are refined for easier exploration.

#### Cell filtering in IGV
Cell filtering for genomic features in IGV is also now available in the Explore UI, when the new `show_igv_multiome` feature flag is on.  It works at a rudimentary level.  The filtering uses cell name indexes that extend those for violin plots.  For multiome data, cell barcodes for ATAC fragments match the cell barcodes for GEX data, so applying filters now not only updates cells shown for GEX data in scatter plots and violin plots, but also cells shown for ATAC data in IGV.

[...]

PR #2068, opened 2024-06-20
Make IGV filtering instant (SCP-5662)

eweitz:
This drastically speeds up IGV filtering by cell annotations, making novel genome browser functionality a delight to use.

Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/c257354e-4bd9-4b67-804b-b71187daf90b

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2053), filtering in IGV was new and transformative, but painfully slow.  It took 10-30 seconds to filter.  That was drastically faster than writing code, running it, and uploading new tracks. But that's still much slower than users expect in a web app.

Now, IGV filtering is instant.  This is achieved by only applying filter logic to genomic features that are in-frame, i.e. in the current genomic window in igv.js.  These genomic features are typically fully contained in the frame, but can also merely partly overlap at left (i.e. upstream) or right (downstream) of the genomic coordinates in view.

### Next steps
Filtering is fast now, but still brittle.  For example, panning to left or right or zooming out will not show genomic features that should be there.  Searching for a new gene also still makes things go awry in the genome browser.  This PR doesn't fully satisfy SCP-5662, but helps show this prototype filtering to stakeholders.  Robustness improvements are next.

### Test
An automated test was updated to account for this change.  If you'd like to manually test:
1. Go to a multiome study
2. Search a gene
3. Apply filters
4. Confirm genomic features in IGV track update, and quickly

This relates to SCP-5662.

### DIFF
--- a/cell-filtering-igv/index.html
+++ b/cell-filtering-igv/index.html
@@ -12,10 +12,7 @@ More context: https://github.com/broadinstitute/single_cell_portal_core/pull/202
 
   /** Render update to reflect newly-selected features in IGV track */
   function updateTrack(trackIndex, filteredFeatures, igv, igvBrowser) {
-    // How many layers of features can be stacked / piled up.
-    // TODO (SCP-5662): eliminate this constraint
-    const maxRows = 20
-    igv.FeatureUtils.packFeatures(filteredFeatures, maxRows)
+    igv.FeatureUtils.packFeatures(filteredFeatures)
     igvBrowser.trackViews[trackIndex].track.featureSource.featureCache = new igv.FeatureCache(filteredFeatures, igvBrowser.genome)
     igvBrowser.trackViews[trackIndex].track.clearCachedFeatures()
     igvBrowser.trackViews[trackIndex].track.updateViews()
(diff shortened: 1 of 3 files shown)


cell-filtering-igv/index.html
```diff
@@ -37,26 +34,60 @@
   return originalChrFeatures
 }
 
-  function filterAtac(scoreSelection) {
-    const trackIndex = 4 // Track index
-    const igvBrowser = window.igvBrowser
+/** Determine if feature is in genomic frame */
+function getIsFeatureInFrame(feature, igvBrowser) {
```
bistline, line 38: Why is this function defined twice, both here and in `GenomeView`?

bistline (approved):
Great improvement!  Very exciting to see this productionized further.

cell-filtering-igv/index.html
eweitz: This area intended is as a 0-config, rapid prototyping environment.  It's super rough -- stuff is copied and pasted, poorly formatted, etc. -- but helps iterate 5-20x faster.

jlchang (approved):
Functional review - fragment filtering for CEBPB in 10x multiome study no longer needs page refresh to see different filtered states and filtering while visualizing SPIB seems more snappy.

PR #2069, opened 2024-06-20
Removing Google Analytics tracking (SCP-5670)

bistline:
#### BACKGROUND & CHANGES
This update removes Google Analytics integration and any associated methods/libraries.  Our GA tracking has been disabled since July of last year, and all associated data is scheduled to be purged on July 1, 2024.  Since the vast majority of analytics has moved to Mixpanel, GA is no longer strictly necessary.  We will lose historical data, but that has been exported from GA to our shared portal drive: https://drive.google.com/drive/u/0/folders/133c49urwFi8hCOf3ueD5dRR0Tn67N9IL

The only remaining utility in GA was being able to get counts on numbers of users in various countries for inclusion in proposals.  There is a ticket in the backlog (SCP-5134) to ask about adding this to Mixpanel user tracking.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Open the Dev Tools > Network tab and enter `google-analytics.com` in the filter box
3. Click around the application and confirm you do not see any requests to the `/collect` endpoint for GA
4. Confirm normal tracking to Mixpanel (filter by `event`) is still enabled, and that normal functionality (search, visualizations) is not impacted

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -651,20 +651,6 @@ function flipToggleSwitch(toggle) {
     jqToggle.html(newText)
 }
 
-// function to call Google Analytics whenever AJAX call is made
-// must be called manually from every AJAX success or js page render
-function gaTracker(id){
-    $.getScript('https://www.google-analytics.com/analytics.js'); // jQuery shortcut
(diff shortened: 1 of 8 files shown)


eweitz (approved):
Looking forward to less noisy error monitoring in Sentry!

jlchang (approved):

PR #2070, opened 2024-06-24
Fixing scatter plot legends in spatial UX (SCP-5626)

bistline:
#### BACKGROUND & CHANGES
This update fixes issues discovered with scatter plot legends when viewing multiple plots in the spatial UX.  Whenever a user selects a new spatial plot, every legend shown then updates the counts to use data from the newest spatial plot.  Additionally, if certain labels are missing from the newest plot, those legend entries are still shown but with a count of 0.

The cause was from a shared `useState` reference in `ExploreDisplayTabs` for setting label counts (this was done so it could be shared with various components in the differential expression UX).  Now, counts are managed at the `ScatterPlot` level instead of `ExploreDisplayTabs`.  Additionally, color assignments are now tracked using the main, non-spatial plot as the control.  Any additional spatial plots will use that color map to ensure that labels keep the same color assignment across all plots.

Note: you may see a brief flicker as colors are reassigned, this has to do with waiting for the main plot to finish rendering.

#### MANUAL TESTING
The easiest way to test this PR is to ingest data from [SCP1375](https://singlecell.broadinstitute.org/single_cell/study/SCP1375) (this is the study where the above bugs were found).  Only 4 files are needed for testing:
- [metadata.csv](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1375/integrative-in-situ-mapping-of-single-cell-transcriptional-states-and-tissue-histopathology-in-an-alzheimer-disease-model?filename=metadata.csv) (Metadata file)
- [top_level_umap.csv](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1375/integrative-in-situ-mapping-of-single-cell-transcriptional-states-and-tissue-histopathology-in-an-alzheimer-disease-model?filename=top_level_umap.csv) (Normal clustering)
- [spatial_8months-control-replicate_1.csv](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1375/integrative-in-situ-mapping-of-single-cell-transcriptional-states-and-tissue-histopathology-in-an-alzheimer-disease-model?filename=spatial_8months-control-replicate_1.csv) (Spatial transcriptomics)
- [spatial_13months-disease-replicate_2.csv](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1375/integrative-in-situ-mapping-of-single-cell-transcriptional-states-and-tissue-histopathology-in-an-alzheimer-disease-model?filename=spatial_13months-disease-replicate_2.csv) (Spatial transcriptomics)

[...]

PR #2073, opened 2024-07-08
Only use refColorMap in multi-plot context (SCP-5692)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug introduced in #2070 where color assignments becomes skewed when switching back and forth between clustering plots in a non-spatial context.  Additionally, colors did not map correctly in a single- or multi-gene view.  Now, color assignments always reset between plots and are consistent across all gene-based views.

Note: this does not fix and existing bug with correlation scatter plots where color assignments are not consistent.

#### MANUAL TESTING
1. Boot as normal and load the `Human milk - differential expression` example study
2. Search for `CLDN4`
3. Confirm that `LC1` and `LC2` are pink & grey respectively and match in the scatter & violin tabs
4. Switch to the `Epithelial cells UMAP` and confirm `LC1` and `LC2` are now red & blue in both tabs
5. (Optional) re-check the steps from #2070 and confirm that spatial plot color assignments are still correct

### DIFF
--- a/app/javascript/components/explore/ScatterTab.jsx
+++ b/app/javascript/components/explore/ScatterTab.jsx
@@ -25,6 +25,7 @@ export default function ScatterTab({
   function updateScatterColor(color) {
     updateExploreParamsWithDefaults({ scatterColor: color }, false)
   }
+  const hasMultipleRefs = exploreParamsWithDefaults?.spatialGroups?.length > 0
 
   // identify any repeat graphs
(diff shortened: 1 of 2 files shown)


jlchang (approved):
functional review - fixes observed coloring discrepancy between scatter and distribution plots.

eweitz (approved):
Code looks good, manual test passed!

PR #2072, opened 2024-07-03
Prevent rendering spatial plots with missing cluster-based annotations (SCP-5671)

bistline:
#### BACKGROUND & CHANGES
This fixes an issue where spatial plots will break when trying to load a cluster-based annotation that is available for the main scatter plot, but missing for the spatial plot(s).  Now, instead of an error, a warning is shown instead telling the user that the requested annotation is not available for the given spatial plot.

This also fixes an unrelated bug where the "Use bucket path" feature was not able to correctly validate/browse metadata files.

Quick demo:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/bdaee530-f377-4da6-8e0a-10da579f360c

#### MANUAL TESTING
The easiest way to test this is with two scatter plots from [SCP1756](https://singlecell.broadinstitute.org/single_cell/study/SCP1756) (you can also add the metadata but it is not strictly necessary):

- [umap_1month.txt](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1756/cortical-organoids-atlas?filename=umap_1month.txt) (normal clustering)
- [umap_1month-org1-Slide-seq.txt](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1756/cortical-organoids-atlas?filename=umap_1month-org1-Slide-seq.txt) (spatial)
- [meta_all.txt](https://singlecell.broadinstitute.org/single_cell/data/public/SCP1756/cortical-organoids-atlas?filename=meta_all.txt)

1. Create a new study with the above scatter plots, ensuring that `umap_1month-org1-Slide-seq.txt` is uploaded as spatial
2. Once ingested, go to the Explore tab for the new study and load the `umap_1month-org1-Slide-seq.txt` as an extra spatial scatter plot
3. Select the `CellType` cluster-based annotation and confirm that it loads correctly
4. Select the `Cluster` annotation and confirm that you see the warning that the spatial plot does not have the requested annotation

PR #2076, opened 2024-07-09
Using persistent color maps for group-based annotations (SCP-5694)

bistline:
#### BACKGROUND
Currently, SCP dynamically assigns colors when viewing group-based annotations in scatter/violin plots based on the labels represented.  These colors always start with red and cycle through a [fixed array](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/javascript/lib/plot.js#L5) of values.  This can lead to situations where a given label does not have the same color in different scatter/violin plots, as that clustering may not have all the same labels represented as others.  While the plots are internally consistent, this could be confusing to users if a population changes color across plots.

#### CHANGES
Now, all group-based annotations will have a consistent color scheme across scatter & violin plots, regardless of labels represented.  These colors are computed server-side and returned as part of the `annotParams` object.  User-defined custom colors can still override this in scatter plots.

Brief demo:

https://github.com/broadinstitute/single_cell_portal_core/assets/729968/436fc679-0571-413b-b7aa-03caa6241a8f

It is worth noting that this will change the appearance of some default plots used in normal testing, such as the `Human milk - differential expression` study.  In the default `All Cells UMAP`, the color for the `T cells` label used to be `#fc8d62`, but is now `#8da0cb` as the `removed` label shifts this down one value, even though that population is not included in the plots.

Note: when viewing spatial clusters using cluster-based annotations that share names & labels, situations can arise where colors are not the same across plots for a given label.  The plots will still be internally consistent, which is an improvement over #2072, where colors could be repeated inside a given plot.  This currently affects only 4 studies on production, so the scope is very limited.

#### MANUAL TESTING
1. Boot as normal and load the `Human milk - differential expression` study
2. Note the color for `LC1` and `LC2` are pink and grey, respectively
3. Load the `Epithelial cells UMAP` clustering and note the colors stay the same for `LC1` and `LC2`
4. Search for the gene `CLDN4`
5. Note the colors remain the same for `LC1` and `LC2` in the scatter & violin plots

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -249,11 +249,6 @@ export default function ExploreDisplayTabs({
   const [clusterCanFilter, setClusterCanFilter] = useState(true)
   const [filterErrorText, setFilterErrorText] = useState(null)
 
-  // map of colors for maintaining associations in spatial UX
-  const [refColorMap, setRefColorMap] = useState({})
-  // state tracker to ensure that spatial/secondary plots update with correct color map after main plot finishes
-  const [refClusterRendered, setRefClusterRendered] = useState(false)
-
   const {
     enabledTabs, disabledTabs, isGeneList, isGene, isMultiGene, hasIdeogramOutputs
   } = getEnabledTabs(exploreInfo, exploreParamsWithDefaults, cellFaceting)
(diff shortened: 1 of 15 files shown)


jlchang (approved):
Functional review - works as expected.

This is a great new (subtle) feature to enhance consistency across a study.

eweitz (approved):
Code and demo look good!  Kudos on the elegant solution.

PR #2078, opened 2024-07-11
Enabling cell filtering on subsampled data (SCP-5608)

bistline:
#### BACKGROUND
Currently, cell filtering only operates on full-resolution clustering & annotation data.  While this is fine for studies with around 100K cells and below, it quickly becomes unusable at higher cell counts, even causing crashes.  This is even more noticeable for violin plots over 100K cells.  The original reason for not using subsampled data was that there is no guarantee that the same cells were sampled across annotations, meaning that the same index would not necessarily be the same cell across annotations.

#### CHANGES
Now, subsampled data can be used during cell filtering.  The critical realization is that once the subsampled cell name indices are generated the 100K level for qualifying clusters/annotations, these indices could still be used with the full resolution annotation arrays.  There will only ever be one subsampled list of cells used at any given time - the cluster/annotation currently being viewed - and the data returned from the annotation facets endpoint is the index of a label for a given cell in the list of unique values for every annotation listed.  This means that the subsampled data still maps correctly to annotations, and the vast majority of our cell filtering logic remains unchanged.

Also, the backend automation for generating indices was refactored to generate all qualifying arrays after successful parsing/subsampling.  A backfill migration has also been added to generate these new indices for existing data.  Minor changes were made to the indexing logic to make it more performant (i.e. "all cells" arrays are shared between index computations rather than being reloaded every time).  A query on production found ~4200 qualifying annotations, but since all of these indexes will be single arrays, there hopefully will not be CPU saturation issues like with the initial backfill.  Test runs on staging should shed light on whether this will be a potential issue or not.

Note: due to [existing logic](https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/javascript/components/explore/plot-data-cache.js#L335) with plot caching, changing filters will cause subsampled plots to re-fetch data from the API.  This is expected.

#### MANUAL TESTING
If you do not have an existing study over 100K cells, SCP2247 is a good candidate as it only has one cluster of 155K cells, and a mix of cluster- and study-based annotations.  The expression data is not strictly necessary, but it does allow you to verify that expression-based plots work.
- [UMAP](https://singlecell.broadinstitute.org/single_cell/data/public/SCP2247/neuronal-diversity-in-the-human-hypothalamus?filename=KaZhouAll_mt10_SCP_UMAP_coords_meta.csv)
- [Metadata](https://singlecell.broadinstitute.org/single_cell/data/public/SCP2247/neuronal-diversity-in-the-human-hypothalamus?filename=KaZhouAll_mt10_SCP_metadata.csv)
- [Processed mtx](https://singlecell.broadinstitute.org/single_cell/data/public/SCP2247/neuronal-diversity-in-the-human-hypothalamus?filename=matrix.mtx)
    - [genes](https://singlecell.broadinstitute.org/single_cell/data/public/SCP2247/neuronal-diversity-in-the-human-hypothalamus?filename=genes.tsv)
    - [barcodes](https://singlecell.broadinstitute.org/single_cell/data/public/SCP2247/neuronal-diversity-in-the-human-hypothalamus?filename=barcodes.tsv)

1. Start all services, including DelayedJob
2. In a separate terminal, run the migration to backfill data (if you have a study with more that 100K cells) with `bin/rails db:migrate`
3. If using SCP2247 source data, upload the necessary files and wait for them to ingest (should take 5-10m)
4. Once subsampling completes, note the similar messages in `development.log`):
```
creating cell name index on SCP105:KaZhouAll_mt10_SCP_UMAP_coords_meta.csv with Study--group--cluster:100000
creating cell name index on SCP105:KaZhouAll_mt10_SCP_UMAP_coords_meta.csv with biosample_id--group--study:100000
creating cell name index on SCP105:KaZhouAll_mt10_SCP_UMAP_coords_meta.csv with celltype--group--cluster:100000
creating cell name index on SCP105:KaZhouAll_mt10_SCP_UMAP_coords_meta.csv with disease--group--study:100000
creating cell name index on SCP105:KaZhouAll_mt10_SCP_UMAP_coords_meta.csv with disease__ontology_label--group--study:100000
...
```
5. Open the explore tab for the study, load the `celltype` annotation and then click the `Filter plotted cells` button, noting that the subsample menu stays at `100000`
6. Scroll down to `Seurat clusters` and deselect the value `12`
7. Note the bottom-center teardrop-like population filters most of the bottom half like this:
Screenshot 2024-07-11 at 2 17 44 PM
8. Compare this to [SCP2247 on production](https://singlecell.broadinstitute.org/single_cell/study/SCP2247/neuronal-diversity-in-the-human-hypothalamus?cluster=KaZhouAll_mt10_SCP_UMAP_coords_meta.csv&spatialGroups=--&annotation=celltype--group--cluster&subsample=all&facets=seurat_clusters--group--cluster%3A12#study-visualize) and note the plots look the same
9. Reselect `12` to unfilter
10. If you uploaded expression data locally, query for `CFAP53` and note the elevated expression in the same area
11. Click to the `Distribution` tab and note the high expression for the `Ependymal` group
12. Re-click `12` in the cell filtering panel and note how the violin thins out, and the maximum expression went from `16` down to `10`.
Screenshot 2024-07-11 at 2 34 39 PM
13. Compare this against [SCP2247 on production](https://singlecell.broadinstitute.org/single_cell/study/SCP2247/neuronal-diversity-in-the-human-hypothalamus?genes=CFAP53&cluster=KaZhouAll_mt10_SCP_UMAP_coords_meta.csv&spatialGroups=--&annotation=celltype--group--cluster&subsample=all&tab=distribution&facets=seurat_clusters--group--cluster%3A12#study-visualize) and note the plots are the same (you may have to re-click `12` in the filtering panel as it sometimes does not apply)

### DIFF
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -446,7 +446,8 @@ class ClusterGroup
                         "#{subsample_annotation}:#{subsample_threshold}"
       create_cell_name_index!(study_cells, subsample_annotation:, subsample_threshold:)
     end
-    index_status = DataArray.where(index_query).any?
+    # determine if any indices
+    index_status = DataArray.any_of(index_query).any? || ClusterGroup.find(id).use_default_index
     update!(is_indexing: false, indexed: index_status)
   end
 
--- a/app/models/cluster_group.rb
+++ b/app/models/cluster_group.rb
@@ -386,7 +386,7 @@ class ClusterGroup
   # study_cells.index_with.with_index creates hash of cell names => array index position
   # averts CPU saturation of calling Array#index on very large arrays
   def cell_name_index(study_cells, subsample_annotation: nil, subsample_threshold: nil)
-    cluster_cells = concatenate_data_arrays('text', 'cells', subsample_annotation, subsample_threshold)
+    cluster_cells = concatenate_data_arrays('text', 'cells', subsample_threshold, subsample_annotation)
     # if cluster cells & study cells are identical, return empty array so we can set #use_default_index
     return [] if cluster_cells == study_cells
 
@@ -395,18 +395,16 @@ class ClusterGroup
   end
 
   # create all necessary data array entries for cell_name_index
-  def create_cell_name_index!(subsample_annotation: nil, subsample_threshold: nil)
-    study_cells = study.all_cells_array
+  def create_cell_name_index!(study_cells, subsample_annotation: nil, subsample_threshold: nil)
     cell_index = cell_name_index(study_cells, subsample_annotation:, subsample_threshold:)
     # if some cells are not indexed, don't create array as this will break things downstream
     index_count = cell_index.compact.size
     expected_cells = subsample_threshold || points
-    if cell_index.empty?
+    if cell_index.empty? && subsample_threshold.nil?
       Rails.logger.info "using default cell index for #{study.accession}:#{name}"
-      update!(is_indexing: false, indexed: true, use_default_index: true)
+      update!(use_default_index: true)
     elsif index_count != expected_cells
       Rails.logger.info "aborting cell index for #{study.accession}:#{name} - #{expected_cells - index_count} cells not found"
-      update!(is_indexing: false)
     else
       cell_index.each_slice(DataArray::MAX_ENTRIES).with_index do |slice, index|
         begin
@@ -416,7 +414,6 @@ class ClusterGroup
             subsample_annotation:, subsample_threshold:
           )
         rescue => e
-          update!(is_indexing: false)
           context = { job: :create_cell_name_index!, subsample_annotation:, subsample_threshold: }
           ErrorTracker.report_exception(e, nil, study, self, context)
         end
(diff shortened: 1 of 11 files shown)


bistline:
Build failure is same `FireCloudClientTest#test_create_and_manage_user_groups` test

app/models/cluster_group.rb
```diff
@@ -395,39 +395,69 @@ def cell_name_index(study_cells)
...
+          context = { job: :create_cell_name_index!, subsample_annotation:, subsample_threshold: }
+          ErrorTracker.report_exception(e, nil, study, self, context)
         end
       end
-      update!(is_indexing: false, indexed: true)
     end
   end
 
+  # create all necessary cell index arrays, both full-resolution and subsampled
+  def create_all_cell_indices!
```
eweitz, line 425: Beyond cell indexing, do we have other ingest tasks that aren't in Ingest Pipeline?  I don't see it as urgent, but having this processing on the app server strikes me as a lurking bottleneck for parallel ingest.  

As we increase subsampling thresholds for larger studies (e.g. 250K), this limitation seems like it would become more substantial.

I imagine porting this to Ingest Pipeline would be straightforward, not worth doing imminently, but worth keeping in mind.

eweitz (approved):
This is an effective and elegant way to scale cell filtering -- kudos!  Lossy-yet-representative results are much better than disabling cell filtering (or crashing) for large studies.

I note one strategic technical concern in the code.  It isn't urgent or blocking.

Beyond that, as we discussed, my review incidentally surfaced two issues:

* Colors sometimes cycle prematurely in cluster scatter plot (SCP-5703)
* Filter counts make it much easier to notice cell count discrepancies in groups when subsampled vs not.  This was basically a known issue, but the concern was more abstract before.  I've opened SCP-5704 to investigate how to improve proportionality in subsampling.

Neither of those is blocking, but they're germane here.

jlchang (approved):
Functional review - confirmed all except step 12. My local instance never redrew the page...

Everything looks good and I can re-check when the changes are on staging.

app/models/cluster_group.rb
bistline: Thats a decent point - I don't see this as being a large issue given the sparsity of study creation, but monitoring the backfill migration will give us a good indication of any bottlenecks.  Last time it rose CPU to ~80% without any other deleterious effects.  Also, while we could definitely do this on an ingest machine, I think it's a lot easier to do it in a background process on the app server since we can leverage classes/models.  But if we see lots of  CPU saturation we can move it off easily enough.

PR #2075, opened 2024-07-09
Robust IGV filtering: drag, zoom, search (SCP-5678)

eweitz:
This decreases brittleness in dynamic filtering in IGV tracks, maturing major new genome browser functionality.

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2068), applying a filter degraded subsequent locus changes in IGV.  After filtering IGV, any change in the genomic coordinate frame -- panning / dragging, zooming, searching another gene -- did not show expected genomic features in the filtered track.

Now, such locus changes after multiome filtering work as expected.  This hardens the UX for slicing-and-dicing ATAC fragment BED files.  That in turn accelerates insights on gene expression and epigenetics.

### Demo
Here's how it looks!

https://github.com/broadinstitute/single_cell_portal_core/assets/1334561/dd93199b-4823-428a-9870-7d8e0d58f4c0

### Next steps
The UX is more robust, but some papercuts remain.  A better "loading" indication for filtered results would be smoother.  Support for [IGV filtering via URL](https://broadworkbench.atlassian.net/browse/SCP-5673) would help share chromatin accessibility insights.

### Tests
Automated tests were updated to account for this refinement.  To manually test:
1.  Go to a study that has an ATAC fragment BED file, e.g. the 3k cell PBMC 10x multiome example dataset
* Staging: https://singlecell-staging.broadinstitute.org/single_cell/study/SCP395
* Production: https://singlecell.broadinstitute.org/single_cell/study/SCP2699
2.  Search gene SPIB
3. Go to "Genome" tab
4. Click "Filter plotted cells"
5. Click root checkbox in "Atac cluster" facet, to deselect all filters
6. Click filter for "5" 
7. Confirm ATAC fragment track updates
8. Drag IGV frame to the left
9. Confirm updated ATAC fragments still appear
10. Zoom out in IGV
11. Confirm updated ATAC fragments still appear
12. In SCP gene search box, query for "ANXA1"
13. Confirm IGV updates to ANXA1, and ATAC fragments appear
14. Change filter selection
15. Confirm ATAC fragment track updates

This satisfies SCP-5678.

PR #2079, opened 2024-07-12
Enable cell filtering in IGV via URL, fix a search case (SCP-5673)

eweitz:
This lets users link to filtered IGV states, which helps share epigenomic insights.  It also fixes a bug with search and IGV.

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2075), users could instantly filter tracks, but they could not persist or share those filtered results.  Also, a bug caused IGV to error when the user loaded IGV, then changed to a no-gene or multi-gene view, then returned to any single-gene IGV view.

Now, cell filtering results can be saved in IGV by simply copying the current URL.  Loading the URL restores that filtered IGV state.  This makes it easy to share chromatin accessibility insights with collaborators and research audiences.

### Video
Here's how it looks!

https://github.com/user-attachments/assets/bf01965f-0abd-4caa-a506-697f5055639e

### Test
To manually test, if you'd like:
1.  Go to a study that has an ATAC fragment BED file, e.g. the 3k cell PBMC 10x multiome example dataset
* Staging: https://singlecell-staging.broadinstitute.org/single_cell/study/SCP395
* Production: https://singlecell.broadinstitute.org/single_cell/study/SCP2699
2.  Search gene SPIB
3. Go to "Genome" tab
4. Click "Filter plotted cells"
5. Click root checkbox in "Atac cluster" facet, to deselect all filters
6. Click filter for "5" 
7. Confirm ATAC fragment track updates
8. Click the "<-" button to the right of the gene search box
9. In the default, 0-gene view, search for ANXA1
10. Confirm "Genome" tab is in view, and IGV shows your new gene and the ATAC track
11. Clear your filters
12. Confirm ATAC track updates
13. Apply a filter
14. Confirm ATAC track updates
15. Refresh page
16. Confirm ATAC track shows state from step 14 (unfiltered ATAC might very briefly appear; unideal but expected)

Also note that these changes required updating to the latest upstream IGV, which e.g. changes the genes track to show Ensembl transcript ID instead of gene name / symbol.  The familiar gene name is still visible in the search bar and related genes ideogram.  As a separate task, it might help to refine IGV or our UI so users can easily discern which genes are in the genomic frame, which becomes more pertinent if the user drags or zooms.

This satisfies SCP-5673.

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -623,6 +623,7 @@ export default function ExploreDisplayTabs({
                   trackFileName={exploreParams.trackFileName}
                   uniqueGenes={exploreInfo.uniqueGenes}
                   isVisible={shownTab === 'genome'}
+                  cellFilteringSelection={cellFilteringSelection}
                   queriedGenes={exploreParams.genes}
                   updateExploreParams={updateExploreParams}
                 />
--- a/app/javascript/components/explore/GenomeView.jsx
+++ b/app/javascript/components/explore/GenomeView.jsx
@@ -17,8 +17,11 @@ function GenomeView({
 }) {
   const [isLoading, setIsLoading] = useState(false)
 
-  const numFacets = Object.keys(cellFilteringSelection).length
-  const [hasAppliedInitFilters, setHasAppliedInitFilters] = useState(numFacets > 0)
+  let numFacets
+  if (cellFilteringSelection) {
+    numFacets = Object.keys(cellFilteringSelection).length
+  }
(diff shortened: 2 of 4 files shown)


bistline (approved):
Works as advertised.  However, I had some issues getting to work locally that were unrelated, but worth noting:

My locally configured reference data workspace had issues that were preventing the bucket permissions from propagating.  This led to the Ensembl genes track not loading with a `TypeError: igvxhr.loadByteArray is not a function` error.  This then caused cell filtering to stop working on the ATAC fragement BED file, though no other errors were emitted.  Once I was able to address the upstream issue, everything worked fine.  I don't think this is something that be an issue in a deployed environment, though me may want to keep an eye on it.

jlchang (approved):
Very excited to have URL persistence!

PR #2084, opened 2024-07-18
Match server- and client-side label sorting (SCP-5703)

bistline:
#### BACKGROUND & CHANGES
This update fixes a bug introduced in #2076 where the server-side sorting of group-based annotations differed from the sorting of those same labels client-side.  This led to corner cases where colors seems to repeat prematurely when handling complex label entries, like those from `biosample_id`.  The problem was that the sorting server-side was purely lexical, where as the client-side sorting used the JS method `localeCompare`, which handles numbers/punctation much more naturally.

Now, the `LabelSorter` class replicates the functionality of `localeCompare` by extracting numeric values contained inside labels and ignoring non-word characters, and then sorts these internal values naturally.  This in turn makes the order of color maps more predictable, and repeats of nearby labels are now less likely.  It is worth noting that _any_ annotation with more that 27 unique labels will repeat colors - this fix only makes the likelihood of those repeats being close or next to each other in the legend far less likely.

#### MANUAL TESTING
Using clustering &  metadata from [SCP2247](https://singlecell.broadinstitute.org/single_cell/study/SCP2247/neuronal-diversity-in-the-human-hypothalamus) makes this much easier to test.

1. Boot as normal and load your copy of SCP2247 with the `biosample_id` annotation
2. Confirm that the colors for `GW7_Lane1` and `GW7_Lane2` are brown and pink, respectively
3. Compare against the [default clustering on production](https://singlecell.broadinstitute.org/single_cell/study/SCP2247/neuronal-diversity-in-the-human-hypothalamus) and confirm the plots are identical

### DIFF
--- a/app/lib/annotation_viz_service.rb
+++ b/app/lib/annotation_viz_service.rb
@@ -222,7 +222,7 @@ class AnnotationVizService
   # this allows for labels to maintain the same color across multiple plots, regardless of whether or not the
   # group is represented
   def self.color_map(values)
-    values.sort_by(&:downcase).map.with_index do |value, index|
+    LabelSorter.natural_sort(values).map.with_index do |value, index|
       { "#{value}" => COLORBREWER_SET[index % COLORBREWER_SET.size] }
     end.reduce({}, :merge)
   end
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code and UI look good!  Another nice pre-deployment fix.

PR #2085, opened 2024-07-22
Restore customizing scatter plot colors (SCP-5112)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with saving custom scatter plot colors where updates never persist, despite telling the user that the save was successful.  This was an issue likely introduced [here](https://github.com/broadinstitute/single_cell_portal_core/commit/64f7de70e0d2ab5ff347f453064538cdf3835256) as a part of work to improve the AnnData upload UX, where the `custom_color_updates` key was being de-nested and not passed to the server.  This key is a controlled value required for proper handling of color objects, and the omission caused all updates to be discarded by the server before saving.

#### MANUAL TESTING
1. Boot as normal and sign in and either a) turn off all caching, or b) start DelayedJob to process cache clearing jobs
2. Load any normal scatter plot and click the `Customize colors` link in the legend once it loads
3. Select the first entry and change the color, then click `Save colors`
4. Confirm the color saves and you see the success dialog
5. Refresh the page and confirm that the updated color is still shown
6. Click `Customize colors` again and then `Reset to defaults`, confirming the colors reset
7. Refresh the page again and confirm that the colors are back to the default palette

### DIFF
--- a/app/javascript/components/upload/upload-utils.js
+++ b/app/javascript/components/upload/upload-utils.js
@@ -42,7 +42,7 @@ const PROPERTIES_NOT_TO_SEND = [
 ]
 
 const PROPERTIES_AS_JSON = ['custom_color_updates']
-const DEEPLY_NESTED_PROPS = ['data_fragments']
+const DEEPLY_NESTED_PROPS = ['custom_color_updates', 'data_fragments']
 
 /** gets an object representing a new, empty study file.  Does not communicate to server */
 export function newStudyFileObj(studyId) {
--- a/yarn.lock
+++ b/yarn.lock
@@ -309,6 +309,11 @@
     "@babel/helper-create-class-features-plugin" "^7.18.6"
     "@babel/helper-plugin-utils" "^7.18.6"
 
+"@babel/plugin-proposal-private-property-in-object@7.21.0-placeholder-for-preset-env.2":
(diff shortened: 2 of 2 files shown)


jlchang:
If you want to have some custom colors in your local instance, here are the custom colors for the human milk study
"{\"General_Celltype\":{\"B cells\":\"#7294D4\", \"CSN1S1 macrophages\":\"#E6A0C4\", \"dendritic cells\":\"#C6CDF7\", \"eosinophils\":\"#D9D0D3\", \"fibroblasts\":\"#5eb668\", \"GPMNB macrophages\":\"#EE46A6\", \"LC1\":\"#9986A5\", \"LC2\":\"#bb99ff\", \"neutrophils\":\"#D8A499\", \"T cells\":\"#81A88D\"}}"

jlchang (approved):
Glad the fix was not difficult. Functional review - works great! Happy to see `Customize colors` is working again :)

eweitz (approved):
Code looks good, and UI is much better!

Per chat, I encountered a problem where "Customize color" sometimes doesn't update color.  In my local environment it fails in ~10% of attempts.  Steps to reproduce and bug demo video are in a new ticket, SCP-5721.  This seems pre-existing, and these changes are a clear and substantial improvement, so I don't consider that bug a blocking issue.

PR #2086, opened 2024-07-23
Handle nil values in LabelSorter, refine ordering (SCP-5714)

bistline:
#### BACKGROUND & CHANGES
This fixes two issues recently discovered with `LabelSorter` and `--Unspecified--` values in group-based annotations:

First, there are some legacy studies on production ([SCP542](https://singlecell.broadinstitute.org/single_cell/study/SCP542/pan-cancer-cell-line-heterogeneity?cluster=tSNE%20Head%20and%20Neck%20Cancer&spatialGroups=--&annotation=Discrete_cluster_minpts5_eps1.8--group--study&subsample=all#study-visualize) & [SCP820](https://singlecell.broadinstitute.org/single_cell/study/SCP820/hbv-liverconsortium-smartseq2?cluster=All%20Cells&spatialGroups=--&annotation=sortDate--group--study&subsample=all#study-visualize)) that have invalid data in the unique values array for some group-based annotations, specifically `NaN`.  This was due to previous issues ([here](https://github.com/broadinstitute/scp-ingest-pipeline/pull/81) and [here](https://github.com/broadinstitute/scp-ingest-pipeline/pull/114)) in `scp-ingest-pipeline` where `NaN` was returned by `pandas` instead of blank strings (both studies predate these fixes).  The new natural sorting `LabelSorter` class throws an error when trying to call `downcase` on these values, as seen in the UI.

Secondly, for studies that have `--Unspecified--` labels, these are incorrectly sorted into the first position of the color map server-side, and then reassigned to the semi-transparent light grey color by the front end and shifted to the end of the legend.  This means that the first available color for group-based annotations (red) is never used in this case.

Now, these `NaN` values are correctly converted to blank strings, and any `--Unspecified--` entries are automatically moved to the end of the color map, making the red label color available again for the first entry.

#### MANUAL TESTING
The data used in the original PR to [fix cell filtering for blank labels](https://github.com/broadinstitute/single_cell_portal_core/pull/1926) works very well in this case, if you have a local copy (sourced from [SCP2407](https://singlecell.broadinstitute.org/single_cell/study/SCP2407/single-cell-atlas-of-the-liver-myeloid-compartment-before-and-after-cure-of-chronic-viral-hepatitis?cluster=all_tsne&spatialGroups=--&annotation=celltype_subcluster--group--study&subsample=all#study-visualize)).  However, any group-based annotation that has `--Unspecified--` values will work.
1. In a Rails console session, load the `celltype_subcluster` annotation from your copy of SCP2407 (if using):
```
study = Study.find_by(accession: <your accession>)
meta = study.cell_metadata.by_name_and_type('celltype_subcluster', 'group')
```
2. (Optional) if not using SCP2407, find any metadatum locally that has a blank value in the unique values array that has many labels:
```
meta = CellMetadatum.where(:values.in => ['']).sort_by { |m| m.values.count }.last
```
3. Change the blank entry to `NaN` to reproduce the data integrity error:
```
blank_idx = meta.values.index('')
meta.values[blank_idx] = Float::NAN
meta.save
```
4. Boot as normal and load the same study/annotation that you just edited
5. Confirm there is no error in the UI, and that the first label shown in the legend uses the color red
6. Back in the Rails console, revert the values array:
```
meta.values[blank_idx] = ''
meta.save
```

### DIFF
--- a/lib/label_sorter.rb
+++ b/lib/label_sorter.rb
@@ -5,8 +5,8 @@ class LabelSorter
   attr_reader :lowercase, :natural_types, :type_order
 
   def initialize(str)
-    @str = str
-    @lowercase = str.downcase
+    @str = str.to_s # safeguard against nil or non-string values
+    @lowercase = @str.downcase
     @natural_types = @lowercase.scan(/\d+|\D+/).map { |s| s =~ /\d/ ? s.to_i : s }
     @type_order = @natural_types.map { |el| el.is_a?(Integer) ? :i : :s }.join
   end
@@ -24,6 +24,11 @@ class LabelSorter
   end
 
   def self.natural_sort(values)
-    values.map { |v| new(v) }.sort.map { |el| el.to_s }
+    sorted = values.map { |v| new(v) }.sort.map { |el| el.to_s }
+    # move any blank/Unspecified entries to the end to allow use of first color for actual label
+    if sorted.first.blank? || sorted.first == AnnotationVizService::MISSING_VALUE_LABEL
+      sorted << sorted.shift
+    end
+    sorted
   end
 end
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good, nice catch!

jlchang (approved):
Wow, there was a lot of "under the hood" knowledge of SCP legend label handling in this bugfix! Functional review - fix returns red as the first assigned color even when unassigned values are present.

PR #2093, opened 2024-07-31
Omit subsample parameter if cluster is not subsampled (SCP-5438)

bistline:
#### BACKGROUND & CHANGES
This fixes another corner case with AnnData files not being subsampled where cell filtering does not work anymore now that the subsample selector has been hidden.  The issue is that the subsample parameter is still set to `100000` and is not correctly honoring the `isSubsampled` flag.  Now, it will always default to `null` if the cluster is not subsampled.

#### MANUAL TESTING
1. In a Rails console, find a cluster that is indexed and subsampled over 100K points and set `subsampled: false` :
```
cluster = ClusterGroup.where(indexed: true, subsampled: true, :points.gt => 100000).sample
cluster.update(subsampled: false)
```
2. Load the Explore tab for the above study and open the network tab in Chrome Dev Tools
3. Confirm that cell filtering works, and that the `facets` requests have a `subsample` value of `null`
4. Reset the cluster:
```
cluster.update(subsampled: true)
```

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayTabs.jsx
+++ b/app/javascript/components/explore/ExploreDisplayTabs.jsx
@@ -99,6 +99,14 @@ export function handleClusterSwitchForFiltering(cellFilteringSelection, newCellF
   }
 }
 
+/** get the appropriate subsampling value, accounting for number of points & subsampling status **/
+export function getSubsampleThreshold(exploreParams, exploreInfo) {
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good, nice quick fix!

jlchang (approved):
Thank you for the quick fix!

PR #2095, opened 2024-08-01
Prevent resetting parse status on update for "bucket path" files (SCP-5736)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with updating study files where any file that was created using the "bucket path" option will have the `parse_status` flag set to `unparsed`.  This can have side effects for AnnData or metadata files that will cause the "All Cells" array to not load.  This will break visualizations in some cases, depending on the path the user takes.  Now, files are only marked as "unparsed" if they haven't yet been ingested.

This also fixes a bug with updating existing clusterings in the AnnData UX where the file does not save correctly.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Find an AnnData file in the Rails console that has been uploaded and parsed using the "bucket path" option (or create a new one if you don't have any):
```
file = StudyFile.where(file_type: 'AnnData', parse_status: 'parsed', 'options.upload_trigger' => 'bucket', 'ann_data_file_info.reference_file' => false).sample
file.parsed?
=> true
```
3. Confirm that the "all cells" array loads for this file/study:
```
study = file.study
study.all_cells_array.count
=> 2638
```
4. Load this file in the upload wizard, update some attribute of the file (like the description) and click save
5. Back in the Rails console, reload the file/study and confirm it is still marked as "parsed" and that the "all cells" array loads:
```
file.reload
study.reload
file.parsed?
=> true
study.all_cells_array.count
=> 2638
```

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -438,7 +438,8 @@ module Api
       end
 
       def complete_upload_process(study_file, parse_on_upload)
-        study_file.update!(status: 'uploaded', parse_status: 'unparsed') # set status to uploaded on full create
+        # set status to uploaded on full create, unless file is parsed and this is just an update
+        study_file.update!(status: 'uploaded', parse_status: 'unparsed') unless study_file.parsed?
         if parse_on_upload
           if study_file.parseable?
             persist_on_fail = study_file.remote_location.present?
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Code looks good!

jlchang (approved):
Thanks for pulling in the fix to cluster name updates! I was able to manually test both file description and cluster name changes and they both no longer unset the parse_status.

PR #2100, opened 2024-08-06
Hide profile warning in public studies; log whether study public (SCP-5699)

eweitz:
This fixes an incorrect warning in IGV.  It also improves observability by logging `isPublic` for the study in Explore events.

Here's how the old bug and new fix look.

https://github.com/user-attachments/assets/a4139961-3aaf-48e4-87ab-c2c9413b23ce

### Test
1.  Go to a public study with igv.js multiome visualizations
2. Confirm no warning message appears below IGV

This satisfies SCP-5699.

### DIFF
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -159,7 +159,8 @@ module Api
             hasImageCache: study.cluster_groups.where(has_image_cache: true).pluck(:name),
             clusterPointAlpha: study.default_cluster_point_alpha,
             colorProfile: study.default_color_profile,
-            bucketId: study.bucket_id
+            bucketId: study.bucket_id,
+            isPublic: study.public
           }
         end
       end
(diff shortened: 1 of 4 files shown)


bistline (approved):

jlchang (approved):
Functional review - spurious warning is no longer visible. 

Non-blocking suggestion: The warning does not show up in private studies where the user has been granted edit access BUT it does show up for visitors to a private study that have only been granted view access. In these cases, **the warning is misleading** because registering for Terra will not allow them to see the visualization. This is definitely the case for reviewers. Is there a quick way to modify the warning so it shows for the "not registered for Terra" case but shows a more correct error like "Sorry, you don't have access to that data because this is a private study that requires edit permissions for visualization".

I'm looking forward to leveraging `isPublic` in Mixpanel!

eweitz:
Thanks for the review, and investigation!  Your findings noted above and at standup indicate complex unexpected behavior with this Terra profile error, even beyond edit permissions.  I've opened SCP-5757 to track that.  

Given this fixes the highest-incidence manifestation of that issue and doesn't degrade anything per discussion, I'll merge now and we can prioritize the deeper work separately.

PR #2102, opened 2024-08-07
Enable AnnData subsampling (SCP-5140)

bistline:
#### BACKGROUND & CHANGES
This applies fixes from https://github.com/broadinstitute/scp-ingest-pipeline/pull/355 to enable subsampling of AnnData files.  As previously noted, it was an improperly scoped query that resulted in incorrect IDs being reused that led to the `DataArray` index violations.  Since AnnData files can have multiple clusterings, this prevented subsampling in general until the issue was addressed.  Now, any AnnData file can be subsampled like other plain text SCP formatted files.  This also includes a backfill migration to identify any previously ingested AnnData files and subsample them, if necessary.

Most importantly, this update changes the behavior if a file has ingested successfully but then fails to subsample.  Previously, the failure would trigger a delete cascade that would remove any data that was ingested from that file.  In the case of AnnData files, this would mean _**all study data**_ is removed.  Since there are multiple AnnData studies on production that have already been successfully ingested that have visualizations enabled, this could potentially be highly disruptive to those study owners.  Now, only subsampled data will be removed if that job fails, and all other full-resolution data will be left if place since it has already been validated and can still be viewed.  This applies to normal cluster files as well.

#### MANUAL TESTING
The simplest way to test this is to run the backfill migration to add subsamples to any previously ingested AnnData files.  Once the subsampling jobs have completed, you can skip to step 5 below.  If you do not have any in your local instance:
2. Create a new study and chose the AnnData upload UX
3. Download [compliant_pbmc3K.h5ad](https://console.cloud.google.com/storage/browser/_details/fc-2f8ef4c0-b7eb-44b1-96fe-a07f0ea9a982/test_Data/ingest_manual_test/compliant_pbmc3K.h5ad;tab=live_object) and add it to the study, specifying two clustering objects with the `obsm_key_names` of `X_umap` and `X_tsne`
4. Wait for all ingest processes to complete, including subsampling (the entire process should take 7-10 minutes)
5. In a Rails console session, validate that the clusterings have been subsampled (depending on the file you used, your counts may not be the same but they should be greater than 0 and the same for both clusterings):
```
study = Study.find_by(accession: <accession of above study>)
study.cluster_groups.each do |cluster|
  puts "#{cluster.name} subsampling status: #{cluster.subsampled}"
  puts "#{cluster.name} number of subsampled arrays: #{cluster.find_subsampled_data_arrays.count}"
end
=>
UMAP subsampling status: true
UMAP number of subsampled arrays: 45        
tSNE subsampling status: true
tSNE number of subsampled arrays: 45
```

### DIFF
--- a/app/models/data_array.rb
+++ b/app/models/data_array.rb
@@ -30,7 +30,7 @@ class DataArray
   index({ study_id: 1, linear_data_id: 1}, { unique: false, background: true })
 
   validates_presence_of :name, :cluster_name, :array_type, :array_index, :values
-  validates_uniqueness_of :name, scope: [:study_id, :cluster_name, :linear_data_name, :linear_data_id, :array_type,
+  validates_uniqueness_of :name, scope: [:study_id, :cluster_name, :linear_data_type, :linear_data_id, :array_type,
                                          :array_index, :subsample_threshold, :subsample_annotation]
 
   # maximum number of entries for values array (to avoid MongoDB max document size problems)
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -374,25 +374,39 @@ class IngestJob
       log_error_messages
       log_to_mixpanel # log before queuing file for deletion to preserve properties
       # don't delete files or notify users if this is a 'special action', like DE or image pipeline jobs
-      subject = "Error: #{study_file.file_type} file: '#{study_file.upload_file_name}' parse has failed"
-      unless special_action?
-        create_study_file_copy
-        study_file.update(parse_status: 'failed')
-        DeleteQueueJob.new(study_file).delay.perform
-        unless persist_on_fail
-          ApplicationController.firecloud_client.delete_workspace_file(study.bucket_id, study_file.bucket_location)
-          study_file.bundled_files.each do |bundled_file|
-            ApplicationController.firecloud_client.delete_workspace_file(study.bucket_id, bundled_file.bucket_location)
-          end
-        end
-        user_email_content = generate_error_email_body
-        SingleCellMailer.notify_user_parse_fail(user.email, subject, user_email_content, study).deliver_now
-      end
+      handle_ingest_failure unless special_action?
       admin_email_content = generate_error_email_body(email_type: :dev)
       SingleCellMailer.notify_admin_parse_fail(user.email, subject, admin_email_content).deliver_now
     else
       Rails.logger.info "IngestJob poller: #{pipeline_name} is not done; queuing check for #{run_at}"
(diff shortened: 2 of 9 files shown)


eweitz (approved):
Great scaling improvement for AnnData!  Code looks good, and manual test passed.

jlchang (changes requested):
I didn't _try_ to cause a subsampling job to fail but, my first ingest of `compliant_pbmc3K.h5ad` failed to subsample the UMAP clustering. The failure was not detected (could be due to existing subsampling code).

In my `jobs:work` terminal I saw:

> [Worker(host:wm29a-efa pid:60707)] Job IngestJob#poll_for_completion (id=66ba25b4dc1a08e849e6a6a4) (queue=default) FAILED (0 prior attempts) with TypeError: no implicit conversion of nil into String
[Worker(host:wm29a-efa pid:60707)] Job IngestJob#poll_for_completion (id=66ba25b4dc1a08e849e6a6a4) (queue=default) FAILED permanently because of 1 consecutive failures

BUT I did not receive a subsampling failure email

From rails console:

> not really tSNE subsampling status: true
not really tSNE number of subsampled arrays: 45
not really UMAP subsampling status: false
not really UMAP number of subsampled arrays: 0

got a subsampling success email for "not really tSNE" but didn't get any subsampling notification (neither success nor failure) for "not really UMAP" (captured some logs in a [doc](https://docs.google.com/document/d/1CwDVK_0iguXNmjtXpjU-KfJFU_rtKOuPbBsHxFminc4/edit))

I tried with a new study, same h5ad. This time rails said:
my tSNE subsampling status: false
my tSNE number of subsampled arrays: 0
my UMAP subsampling status: false
my UMAP number of subsampled arrays: 0

jobs:work still saw:
> [Worker(host:wm29a-efa pid:62487)] Job IngestJob#poll_for_completion (id=66ba3788dc1a08f41796d69e) (queue=default) FAILED (0 prior attempts) with TypeError: no implicit conversion of nil into String
[Worker(host:wm29a-efa pid:62487)] Job IngestJob#poll_for_completion (id=66ba3788dc1a08f41796d69e) (queue=default) FAILED permanently because of 1 consecutive failures

This time got a subsampling success email for "my UMAP" but didn't get any subsampling notification for "my tSNE"... Which is even more wacky because rails says `subsampling status: false`

Feels like a race condition when an AnnData has more than one cluster specified for ingest? Happy to make my local instance available to take a closer look.

jlchang (approved):
Gotta keep better track of my docker images! Thanks for figuring out my oops, @bistline

PR #2106, opened 2024-08-13
Fix setting tab parameter for analytics (SCP-5634)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with setting the `tab` property on client-side Mixpanel events.  The issue was an improperly scoped regular expression that relied on the `tab` search parameter to be in the first position (i.e. directly after the `?` character).  Now, the native `URLSearchParams` class is used to parse the entire search string and determine if the `tab` parameter has been specified.  If not, it will default to the value of `window.location.hash`, if present.

#### MANUAL TESTING
1. Boot as normal and load any study that has clustering & expression data
2. Open the Chrome Dev Tools > Network tab
3. Query for a gene and then click into the `Distribution` tab and wait for the plot to load
4. In the network tab, find the entry for `plot:violin` and confirm that you see the tab parameter is correctly set:
Screenshot 2024-08-13 at 2 21 58 PM

(Note: if the violin plot renders before you click into the `Distribution` tab, this may be set to `scatter` or `#study-visualize`, depending on your URL state).

### DIFF
--- a/app/javascript/lib/metrics-api.js
+++ b/app/javascript/lib/metrics-api.js
@@ -385,11 +385,14 @@ function getAnalyticsPageName() {
  * gets the tab name for analytics
  */
 function getTabProperty() {
-  if (window.location.href?.match(/\?tab=/)) {
-    return window.location.href.split('?tab=')[1]
-  } else {
-    return window.location.hash?.replace(/#/, '')
+  let tabParam = window.location.hash?.replace(/#/, '')
+  if (window.location.search) {
+    const searchParams = new URLSearchParams(window.location.search)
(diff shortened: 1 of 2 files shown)


eweitz (approved):
A definite observability improvement!  Code looks good, manual tests passed.

> this may be set to `scatter` or `#study-visualize`, depending on your URL state

I wonder how feasible it'd be to refine that analytics rough edge.  The current URL-only approach precludes it, but perhaps an enhancement could account for the default Explore tab even when `tab` is absent in the URL.

jlchang (approved):
Passes functional review.

PR #2107, opened 2024-08-14
Replacing master with main (SCP-3446)

bistline:
This renames the master branch to main and replaces any references with the new name.  This will affect CD only for production deployments, but as the branch will be renamed before any deployments happen this should not matter.

You will need to run the following commands to update your local repository:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a 
```

### DIFF
--- a/.github/workflows/delete-branch-on-merge-to-master.yml
+++ b/.github/workflows/delete-branch-on-merge-to-master.yml
@@ -1,16 +1,16 @@
 name: Selectively delete feature branch after merge to master
 
-# Only run the workflow when a PR that targets master is closed
+# Only run the workflow when a PR that targets main is closed
 on:
   pull_request:
     types:
       - closed
     branches:
-      - 'master'
+      - 'main'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
-  
+
 jobs:
   merge_job:
     # this job will only run if the PR has been merged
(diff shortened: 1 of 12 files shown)


jlchang (approved):
Changes make sense to me! Thanks for transitioning us from `master` to `main`.

.github/workflows/delete-branch-on-merge-to-master.yml
eweitz, line 1: Renaming this file seems clearest.  Preserving Git history across renames is tricky; `git mv` is my go-to.  If history isn't preserved, I think that's OK here.  Preserving clarity via the rename seems more important.

`git mv .github/workflows/delete-branch-on-merge-to-master.yml .github/workflows/delete-branch-on-merge-to-main.yml`

.github/workflows/delete-branch-on-merge-to-master.yml
```diff
@@ -1,16 +1,16 @@
 name: Selectively delete feature branch after merge to master
```
eweitz, line 1:
```suggestion
name: Selectively delete feature branch after merge to main
```

eweitz (approved):
Nice update to modern defaults!  

I reviewed https://github.com/search?q=repo%3Abroadinstitute%2Fsingle_cell_portal_core%20master&type=code and found a few missed references that would help to update, one noted here and two in comments below.

* [Line 34 in search_facet.rb](https://github.com/broadinstitute/single_cell_portal_core/blob/db53124dd325b7047cefbb1c04b9c54f306d54cf/app/models/search_facet.rb#L34)

I also replaced `master` with `main` in the [SCP release engineer](https://docs.google.com/document/d/1rKcOQPtkbhZCbTqxTnaT-4E29VLzkJrm0kAoWjLx3QA/edit) docs.  The [SCP playbook](https://docs.google.com/document/d/1vJz6hXdjAu8LBAfIS4oMy7yLIUX177gU1H7BxTPPzAU/edit) has more references that would be good to review and change as warranted.

Do any of our Terraform integrations need to be updated?

PR #2099, opened 2024-08-05
Fix legend filter for labels with commas (SCP-5746)

eweitz:
This enables using the scatter plot legend to filter entries that have labels containing commas, as seen in e.g. CAS demo.

### Before
Previously, legend filters broke for entries that contained commas, e.g. "central memory CD4-positive, alpha-beta T cell".  Such labels are uncommon, but exist in standard cell type labels from Cell Ontology.  Upon clicking the label, instead of hiding the filter, nothing would happen.  Clicking "Hide all filters" would similarly do nothing for any labels with commas.

https://github.com/user-attachments/assets/7fd7318b-1e30-4318-a490-064c0f401163

### After
Now, that case works as expected.  This smoothens exploration for the CAS-SCP MVP, so that labels output by analysis seen in e.g. the [quickstart_tutorial.ipynb](https://github.com/cellarium-ai/cellarium-cas/blob/main/notebooks/quickstart_tutorial.ipynb) CAS notebook can be filtered in the legend as expected.  

https://github.com/user-attachments/assets/e2220564-92cd-4681-b1c4-44ce76136111

### Technical notes
The bug was that commas are used as delimiters in stringified `exploreParams` as written to the URL.  The fix was simply replacing the comma with a hyphen in some non-display frontend contexts.

### Deeper testing
I've ensured other less-common characters in Cell Ontology don't trigger this bug.  See the new `cluster_odd_labels.txt` for a more comprehensive dataset with unusual labels.  For reference, the only characters present in Cell Ontology labels _not_ among `A-Za-z0-9-_ ` are `.=/[:]–><';,()+`, per a simple script I put together at [uncommon_ontology_chars.py](https://gist.github.com/eweitz/ab559f0a5a32997b19f2c5e7498727b0#file-uncommon_ontology_chars-py).

### Test
A new automated test confirms this works.  If you'd like to manually test:
1.  Ensure you're running DelayedJob, i.e. `rails jobs:work`
2. Ensure you're not running Vite service worker cache
3.  Create a new study named e.g. "PBMC 4k 10x H5AD with CAS results"
4. Download [cas_annotated_trimmed_compliant__pbmc_10x_v3_4k.h5ad](https://console.cloud.google.com/storage/browser/_details/fc-2f8ef4c0-b7eb-44b1-96fe-a07f0ea9a982/test_Data/scp-ingest-pipeline/cas/cas_annotated_trimmed_compliant__pbmc_10x_v3_4k.h5ad)
[...]

PR #2109, opened 2024-08-15
Fix expression scatter plot upon legend filter (SCP-5762)

eweitz:
This fixes a bug in #2099 where applying a legend filter caused the gene expression scatter plot to break.

### Test
An automated test prevents regression.  To manually test:
* Go to a study with visualizations initialized
* Search for a gene
* Click on a legend entry beside the reference cluster scatter plot
* Confirm gene expression scatter plot does not updated

This satisfies SCP-5762.


jlchang (approved):
Functional tests look good. Thanks for the quick bugfix!

bistline (approved):

PR #2108, opened 2024-08-15
Update metadata explainer image for AnnData (SCP-5761)

jlchang:
Existing image for "Classic" metadata upload is slightly incompatible with AnnData upload needs. This updated image should convey the requirements for AnnData without the classic metadata-specific requirements muddying the waters. 
metadata-convention-explainer-anndata

This PR also adds a visual cue that `.obsm key name` is a file attribute required for upload.

MANUAL TESTING

- Review the image above and confirm that no "classic metadata"-specific requirements are noted in the figure
- In your local instance, visit the AnnData experience and verify an asterisk is shown next to `.obsm key name` to indicate a required field (the validation is already implemented, there's just no visual cue on the form).

Known issue
- if >1 cluster file, cluster forms are collapsed BUT collapsed forms are unnamed (in Classic mode, each collapsed cluster form shows the filename and filesize as a file download link; for AnnData, the collaped cluster form should probably show the cluster name).
Screenshot 2024-08-15 at 9 37 24 AM

PR #2111, opened 2024-08-15
UX refinements for AnnData upload (SCP-5727)

bistline:
#### BACKGROUND & CHANGES
This updates the 3 main steps in the AnnData upload UX.  For expression upload, an extra preamble has been added about the which data frames need to contain processed data, and where to store raw counts information:

Screenshot 2024-08-15 at 1 51 28 PM

The metadata "infographic" is updated with the new image that was added from this [PR](#2108).  Lastly, the clustering forms now will use the name of the clustering in place of the filename when the form is collapsed (this only applies once a file has been uploaded/saved):

Screenshot 2024-08-15 at 2 00 26 PM

#### MANUAL TESTING
1. Boot as normal and load the upload wizard for an existing AnnData study
2. Confirm you see the informational text from above displayed in the `Expression matrices` tab
3. In the `Metadata` tab, confirm the image displayed talks about `adata.obs` and does not have a `NAME` column or `TYPE` row
4. In the `Clustering` tab, confirm the name of the cluster is displayed for each form, even when collapsed


app/javascript/components/upload/MetadataStep.jsx
```diff
@@ -39,6 +40,7 @@ function MetadataForm({
   const userState = useContext(UserContext)
   const featureFlagState = serverState.feature_flags
   const conventionRequired = featureFlagState && featureFlagState.convention_required
+  const explainerImg = isAnnDataExperience ? annDataMetadataExplainerImage : metadataExplainerImage
```
jlchang, line 43: Great for me to see how this is done (I would have done something much less elegant/DRY).

jlchang (changes requested):
Changes look great! Since we're updating to make the UI more tailored to AnnData, we should change the statement "A metadata file lists all cells in the study" to "Uploaded data must conform to the SCP metadata convention by including  required metadata in adata.obs"
Screenshot 2024-08-15 at 2 47 22 PM

bistline:
@jlchang [2fcc001](https://github.com/broadinstitute/single_cell_portal_core/pull/2111/commits/2fcc00139f3dbac9f16a98cc8d01392edcc09057)

jlchang (approved):
Thanks for the quick update on that stray sentence!

app/javascript/components/upload/MetadataStep.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useContext } from 'react'
 import metadataExplainerImage from '~/../assets/images/metadata-convention-explainer.jpg'
+import annDataMetadataExplainerImage from '~/../assets/images/metadata-convention-explainer-anndata.jpg'
```
eweitz, line 3: Nit: non-photographic raster images tend to use WEBP or PNG instead of JPG, because the former can be lossless.  JPG is more common for photos, where lossy compression makes a lot of sense.  

The image looks good and this is definitely not a blocker, especially since the other image uses JPG.  But WEBP or PNG might be better if we ever recreate such images from scratch and want very crisp graphics.

eweitz (approved):
Code looks good, manual tests passed!  

I discuss image formats in a comment, as context for possible future graphics.

app/javascript/components/upload/MetadataStep.jsx
jlchang: Good point, @eweitz. I only generated jpg because the previous image was in that format.
I loaded [mc_compliant_metadata_image_full.psd ](https://github.com/broadinstitute/single_cell_portal/blob/master/site_figures/metadata_file/mc_compliant_metadata_image_full.psd) into GIMP and generated a PNG
mc_compliant_metadata_image_full
and a PNG for the AnnData version
anndata_compliant_metadata_image_full_edited
Are there any gotchas where a round trip through GIMP might have messed with the resolution of the image (that would not be detectable on my screen).

@bistline Not entirely sure if Github retains the full resolution image this way. Let me know if I should get the files to you another way.

eweitz: Thanks for the comparison!  The attached WEBP is somewhat smaller in byte size and looks equivalent in quality compared to the PNG.  GitHub retains the raw file.  

Compare_WEBP_PNG_JPG__2024-08-19

GitHub comments seem to prevent WEBP attachments, but WEBP files are allowed in GitHub repos.  Simply remove the .txt extension from the attached file to get original WEBP.  [WEBP support](https://caniuse.com/webp) is effectively universal.

PR #2120, opened 2024-09-04
Adding machine type & failure properties to ingestSummary event

bistline:
#### BACKGROUND & CHANGES
In an effort to make the `ingestSummary` Mixpanel event for AnnData ingest more useful, we are adding three new properties:

1. `machineType`: GCE instance type, e.g. `n2d-highmem-8`
2. `error`: ingest process that failed (if present), e.g. `ingest_cell_metadata`, `ingest_anndata`, etc.
3. `exitCode`: exit code from process that terminated the instance - `0` for success, non-zero for failures

Along with existing properties, these new fields will afford us better insight into how and why AnnData ingests either succeed or fail.  For instance, the `error` and `exitCode` fields will let us determine which type of extracted data fails most frequently, and if those failures are memory-related (`137` or `139`), or in the case of metadata files if they are JSON schema errors (`65`).  The `machineType` field will be useful in conjunction with memory-related exit codes as it will allow us to plot file size against machine types to better ascertain where we should be scaling up to the next available machine type.

There could be future work done in `scp-ingest-pipeline` to differentiate other failure states with unique exit codes, if desired.   Since many of our handled exceptions exit with `1`, we could chose to pick different codes for more common errors to allow us better visibility into how ofter these happen, which could then drive future development into CSFV to prevent these upload from proceeding to ingest.

#### MANUAL TESTING
2. Go to the Feature Flags control panel and deactivate `clientside_validation` for your user account (otherwise you will not be able to upload the invalid AnnData file later)
3. Create a new study and select the AnnData upload UX
4. Upload the AnnData file `test/test_data/anndata_test_bad_header_no_species.h5ad` and specify that you have clustering in the `X_umap` slot (this doesn't exist and will throw an error to be caught later)
5. After a few minutes, you should see the `ingestSummary` event in the logs with the following properties:
```
jobStatus: "failed", 
numFilesExtracted: 0, 
machineType: "n2d-highmem-4",
error: "ingest_anndata",
exitCode: 1
```

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -286,8 +286,6 @@ class IngestJob
     nil # catch-all
   end
 
-
-
   # Reconstruct the command line from the pipeline actions
   #
   # * *returns*
@@ -973,7 +971,8 @@ class IngestJob
   def anndata_summary_props
     pipelines = ApplicationController.life_sciences_api_client.list_pipelines
     previous_jobs = pipelines.operations.select do |op|
-      op.metadata.dig('pipeline', 'actions').first['commands'].detect { |c| c == study_file.id.to_s }
+      pipeline_args = op.metadata.dig('pipeline', 'actions').first['commands']
+      pipeline_args.detect { |c| c == study_file.id.to_s } && !pipeline_args.include?('--differential-expression')
     end
     # get total runtime from initial extract to final parse
     initial_extract = previous_jobs.last
@@ -985,10 +984,22 @@ class IngestJob
     file_type = study_file.file_type
     trigger = study_file.upload_trigger
     job_status = previous_jobs.map(&:error).compact.any? ? 'failed' : 'success'
+    error_action = nil
+    code = 0
+    if job_status == 'failed'
+      first_failure = previous_jobs.reverse.detect { |p| p.error.present? }
+      args = first_failure.metadata.dig('pipeline', 'actions').first['commands']
+      error_action = args.detect {|c| LifeSciencesApiClient::FILE_TYPES_BY_ACTION.keys.include?(c.to_sym) }
+      failure_events = first_failure.metadata['events'].sort_by! { |event| event['timestamp'] }
+      stopped_msg = failure_events.detect do |event|
+        event.dig('containerStopped', 'exitStatus').present?
+      end
+      code = stopped_msg.dig('containerStopped', 'exitStatus').to_i
+    end
     # count total number of files extracted
     num_files_extracted = previous_jobs.reject do |op|
-      op.metadata.dig('pipeline', 'actions').first['commands'].detect { |c| c == '--extract' } &&
-        op.error.nil?
+      op.metadata.dig('pipeline', 'actions').first['commands'].detect { |c| c == '--extract' } ||
+        op.error.present?
     end.count
     # event properties for Mixpanel summary event
     {
@@ -999,7 +1010,9 @@ class IngestJob
       studyAccession: study.accession,
       trigger:,
       jobStatus: job_status,
-      numFilesExtracted: num_files_extracted
+      numFilesExtracted: num_files_extracted,
+      error: error_action,
+      exitCode: code
     }
   end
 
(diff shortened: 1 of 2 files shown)


app/models/ingest_job.rb
```diff
@@ -999,7 +1010,10 @@ def anndata_summary_props
       studyAccession: study.accession,
       trigger:,
       jobStatus: job_status,
-      numFilesExtracted: num_files_extracted
+      numFilesExtracted: num_files_extracted,
+      machineType: params_object.machine_type,
+      error: error_action,
```
eweitz, line 1015: AIUI, this is equivalent to the `action` property in `ingest` events, which we use in non-AnnData contexts.

Logging this data via the same `action` property in AnnData contexts -- and logging it for failure (like here) _as well as_ success -- seems like it would ease analysis.  E.g. it would allow us to use the same property in Mixpanel queries about ingest, whether the ingest was AnnData or not.  That could help e.g. compare error causes for the same underlying classic file type (e.g. metadata, cluster) whether derived from AnnData or not.  It could also help us sum up ingests by classic file type, whether the parent ingest was AnnData or not.

(Intuitively, I'd expect `error` to log a message or a code, not the ingest action / job run type.)

Am I overlooking something?  The suggestion below is part of a batch.

```suggestion
      action: ,
```

app/models/ingest_job.rb
```diff
@@ -985,10 +984,22 @@ def anndata_summary_props
     file_type = study_file.file_type
     trigger = study_file.upload_trigger
     job_status = previous_jobs.map(&:error).compact.any? ? 'failed' : 'success'
+    error_action = nil
+    code = 0
+    if job_status == 'failed'
+      first_failure = previous_jobs.reverse.detect { |p| p.error.present? }
+      args = first_failure.metadata.dig('pipeline', 'actions').first['commands']
+      error_action = args.detect {|c| LifeSciencesApiClient::FILE_TYPES_BY_ACTION.keys.include?(c.to_sym) }
```
eweitz, line 992: See comment beginning "AIUI".

```suggestion
```

app/models/ingest_job.rb
```diff
@@ -985,10 +984,22 @@ def anndata_summary_props
     file_type = study_file.file_type
     trigger = study_file.upload_trigger
     job_status = previous_jobs.map(&:error).compact.any? ? 'failed' : 'success'
+    error_action = nil
```
eweitz, line 987: See comment beginning "AIUI".

```suggestion
    action = args.detect {|c| LifeSciencesApiClient::FILE_TYPES_BY_ACTION.keys.include?(c.to_sym) }
```

test/models/ingest_job_test.rb
```diff
@@ -412,7 +406,10 @@ class IngestJobTest < ActiveSupport::TestCase
       studyAccession: @basic_study.accession,
       trigger: ann_data_file.upload_trigger,
       jobStatus: 'success',
-      numFilesExtracted: 1
+      numFilesExtracted: 1,
+      machineType: metadata_params.machine_type,
+      error: nil,
```
eweitz, line 411: See comment beginning "AIUI".

```suggestion
      action: 'ingest_cluster',
```

test/models/ingest_job_test.rb
```diff
@@ -429,6 +426,39 @@ class IngestJobTest < ActiveSupport::TestCase
...
+        ingest_cell_metadata: true, cell_metadata_file:, ingest_anndata: false, extract: nil, obsm_keys: nil,
+        study_accession: @basic_study.accession
+      )
+      metadata_job = IngestJob.new(study: @basic_study, study_file: ann_data_file, user: @user,
+                                   action: :ingest_metadata, params_object: metadata_params)
+      props = metadata_job.anndata_summary_props.with_indifferent_access
+      assert_equal 3_600_000, props[:perfTime]
+      assert_equal 'failed', props[:jobStatus]
+      assert_equal 0, props[:numFilesExtracted]
+      assert_equal 'ingest_cell_metadata', props[:error]
```
eweitz, line 457: See comment beginning "AIUI".

```suggestion
      assert_equal 'ingest_cell_metadata', props[:action]
```

eweitz (approved):
Nice observability enhancements!  Code looks generally good.

I suggest some hopefully small, specific changes to generalize one of the new log properties, for easier and more powerful analysis.

PR #2118, opened 2024-08-28
Fast validation for remote AnnData headers (SCP-5770)

eweitz:
This extends up-front checks to cover adding an AnnData file from a bucket.  It makes authoring smoother, and thus faster.

### Impact
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2112), new AnnData validation caught errors upon selecting a local file.  Validating _before_ upload is a great idea.  It eliminates long wait times, potentially reducing a disjointed 15-45 minute process into a cohesive 10 second process.  But local AnnData validation doesn't help when the file is remote, i.e. in a Terra workspace GCP bucket.

Now, client-side file validation (CSFV) for AnnData headers also applies to remote files.  When adding AnnData via "Use bucket path", header validation finishes in < 10 seconds rather than 5-10 minutes like before.

### Video
Here's how it looks!

https://github.com/user-attachments/assets/360b2bfe-f744-4476-b439-d37fdabb8bf9

### Technical notes
This required fixing a bug in the external HDF5 library that prevented using it with files needing an OAuth token.  Details on that are in https://github.com/jrobinso/hdf5-indexed-reader/pull/2.  That HDF5 library is now also available as a [canonical NPM package](https://github.com/jrobinso/hdf5-indexed-reader/issues/3), which replaces our use of an org-scope package.

### Next steps
* Instant _content_ validation for AnnData files
* Show loading indicator to avoid confusion when validating remote files
* Fix how "Replace" keeps bad file name if good file selected before

This satisfies SCP-5770.

PR #2122, opened 2024-09-06
Fast validation of AnnData rows (SCP-5719)

eweitz:
This extends up-front checks to cover row-level data in an AnnData file. It makes authoring smoother, and thus faster.

### Impact
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2118), new AnnData validation caught errors in headers.  Fast validation for headers is very convenient, but superficial.  Most validation errors occur in the body of the AnnData file.

Now, deeper checks validate some _content_ of rows in an AnnData, for both local and remote HDF5 files.  A new basic rule examines the format of ontology IDs for rows of required metadata.  It covers up to [23% of file validation errors](https://mixpanel.com/project/2120588/view/19059/app/boards#id=734187&editor-card-id=%22report-28093580%22) in classic upload, so starting AnnData row validation with this relatively simple rule seems worthwhile.  

### Video

Here's how it looks!

https://github.com/user-attachments/assets/7f35cc3c-a3bd-4ea7-ba98-4ddec63cee0a

### Test
A new automated test verifies behavior.  Optional manual test:

1. Ensure Vite frontend service worker cache is disabled
2. Go to study "AnnData CSFV"
3. Select AnnData upload UX
4. In AnnData tab, click "Choose file"
5. In Finder, go to single_cell_portal_core/test/test_data/
6. Select anndata_test_invalid_disease.h5ad
7. Confirm error message is shown
8. Click "Delete" to remove errant file, then click "Choose file"
9. In same directory as before, select anndata_test.h5ad
10. Confirm no error message is shown, and "Save" button is enabled
11. Upload anndata_test_invalid_disease.h5ad and to your study's GCP bucket
12. Click "Use bucket path"
13. Enter "anndata_test_invalid_disease.h5ad" in text input
14. Click off input to blur / change browser focus
15. Confirm error message is shown is disabled

This satisfies SCP-5719.


bistline (changes requested):
Everything looks good and behaves as expected, however I decided to try with a known good AnnData file that was larger than the small test case files (compliant_pbmc3K.h5ad).  Both the local upload and remote validation return an error state despite the file being completely valid, and there is no indication of what the error is:

Screenshot 2024-09-09 at 10 25 48 AM

eweitz:
Good find!  Fixed.  It turns out this was caused by different internal file structures, not file size.

I've confirmed a compliant, remote 6 GB AnnData file validates in 10 seconds without detecting problems.  I also refined local error observability so we get some insight from the console when "Please reach out" problems occur.

Below are some details on the case you found.

#### Old-format AnnData
`compliant_pbmc3K.h5ad` uses an old version of the AnnData spec.  It stores categorical arrays in a separate `__categories` obs.  In new versions of the AnnData spec (since at least 2023-01-23, like a fair while before), categorical arrays are encoded as self-contained groups containing their own `categories` and `codes`.  https://github.com/scverse/anndata/issues/879 has an example with a problem reported using the old format (using `__categories`), and the project maintainer suggests using a newer version of the AnnData format.  It's unclear precisely when the new format was introduced.

I tried to make a minimal old-format test file from `compliant_pbmc3K.h5ad` using Scanpy.  However, it seems basic manipulations like `adata_subset.obs = adata.obs[:10].copy()` implicitly convert old-format AnnData `__categories` to new format.  So I've tested the negative case with old-format AnnData, i.e. I've tested that validation doesn't report errors for compliant files.  But I don't know a quick way to create positive cases for old-format AnnData.  

#### Future refinement
Do we have any old AnnData files with invalid ontology IDs from before 2023?  If not, I can make a ticket to create one with an old version of `scanpy`.  As is, we at least know this new validation no longer throws false-positives for old-format AnnData, which is indeed a blocking problem.  Creating a true-positive failure case for this validation in old-format AnnData would let us confirm this handling catches that class of problem.  As is, this new validation _probably_ handles that case, and if it doesn't it will be caught by ingest Python validation.

bistline:
> Good find! Fixed. It turns out this was caused by different internal file structures, not file size.
> 
> I've confirmed a compliant, remote 6 GB AnnData file validates in 10 seconds without detecting problems. I also refined local error observability so we get some insight from the console when "Please reach out" problems occur.

Interesting!  If this is indeed the case, then I would consider this to be a much lower-priority issue as the workaround is to make sure any packages used are updated.  I'll go ahead and approve now.

bistline (approved):
Thanks for the quick investigation!

jlchang (approved):
Thanks for the help troubleshooting, @eweitz! File validation works as described.

PR #2127, opened 2024-09-11
Create default `X_umap` cluster fragment for AnnData upload unless specified (SCP-5796)

bistline:
#### BACKGROUND & CHANGES
This update addresses a corner case in the AnnData upload UX where the user never clicks into the Clustering tab.  This results in that form never being initialized, and as such the required validations don't run.  This ends up creating an invalid Python command line for ingest where `obsm_keys` is not set.  Now, if the user does not interact with the Clustering tab, the app adds a clustering data fragment for the `X_umap` key by default.  Additionally, the `X_tsne` default has been removed.  If a user does interact with the clustering tab, then this default fragment is not created as they cannot submit the form until completing the required fields.  Overall, this makes uploading an AnnData file simpler as now the only required interactions are specifying a species, library preparation protocol, and then providing a file.

#### MANUAL TESTING
2. Create a new study, or select and empty one and load the AnnData upload UX
3. Select a species/protocol and the go directly to the AnnData tab _**without clicking the Clustering tab**_
4. Use the "Bucket path" option to specify a path to a remote file and then click Save
5. Confirm you see a success confirmation
6. Go back to the Clustering tab and confirm there is an entry with `X_umap` as the `obsm_key_name`

### DIFF
--- a/app/models/ann_data_file_info.rb
+++ b/app/models/ann_data_file_info.rb
@@ -202,7 +202,10 @@ class AnnDataFileInfo
     default_obsm_keys = AnnDataIngestParameters::PARAM_DEFAULTS[:obsm_keys]
     default_obsm_keys.each do |obsm_key_name|
       name = obsm_key_name.delete_prefix('X_')
-      data_fragments << { _id: BSON::ObjectId.new.to_s, data_type: :cluster, name:, obsm_key_name: }
+      fragment = {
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Nice simplification!  Code looks good.

jlchang (approved):
Thanks for figuring out yet another weird corner case (I wonder how many people _do_ skip clicking all the tabs. To be honest, I skip the metadata tab consistently for AnnData :)

Testing in a newly created study works as expected.

PR #2128, opened 2024-09-16
Handle missing cluster-based annotation in plot data cache (SCP-5799)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where a missing cluster-based annotation in spatial plots causes all expression-based scatter plots to throw an error.  This was due to the `scatter` object having a value of `undefined` for `annotParams` in the response when trying to populate the cache.  Now, the plot correctly shows the warning that the requested annotation does not exist on that plot.

Example:
 
Screenshot 2024-09-16 at 11 08 27 AM

This update also deals with some test instability issues in `ingest_job_test.rb` and `import_service_config/nemo_test.rb` due to recent updates, order of operations, etc.

#### MANUAL TESTING
2. If you do not have a copy of the `chick_raw_counts` synthetic study in your instance, you can load it from the Rails console with:
```
SyntheticStudyPopulator.populate('chicken_raw_counts')
```
3. If you do have a copy, go to the upload wizard for that study and upload `db/seed/synthetic_studies/chicken_raw_counts/cluster_odd_labels.txt` file as a **normal** cluster file
4. Once ingested, go to the Explore tab for this study and select `cluster_odd_labels`, and then `odd_labels` as the annotation
5. Load the spatial cluster `spatial_letters` to confirm you see the `"spatial_letters" does not have the requested annotation "odd_labels--group--cluster"` warning
6. Search for the gene `RAD51` and confirm you see the same warning in the `spatial_letters` expression plot

### DIFF
--- a/app/javascript/components/explore/plot-data-cache.js
+++ b/app/javascript/components/explore/plot-data-cache.js
@@ -289,7 +289,8 @@ export function createCache() {
     Fields.cellsAndCoords.merge(cacheEntry, scatter)
     // only merge in annotation values if the annotation matches (or the default was requested, so
     // we can then assume the response matches)
-    if (!requestedAnnotation.name || scatter.annotParams.name === requestedAnnotation.name) {
+    // annotParams may be undefined in spatial UX if a cluster-based annotation does not exist for the plot
+    if (!requestedAnnotation.name || scatter.annotParams?.name === requestedAnnotation.name) {
       Fields.annotation.merge(cacheEntry, scatter)
     }
     if (scatter.genes.length && scatter.genes.join('') === requestedGenes.join('')) {
--- a/app/javascript/components/visualization/ScatterPlot.jsx
+++ b/app/javascript/components/visualization/ScatterPlot.jsx
@@ -598,16 +598,14 @@ function RawScatterPlot({
   return (
     <div className="plot">
       { ErrorComponent }
+      <PlotTitle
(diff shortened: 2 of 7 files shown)


eweitz (approved):
Elegant fix for this edge case!  Code looks good.

jlchang (approved):
Functional testing confirms fix behaves as described.

Non-blocking suggestion: allow the plots that lack the requested annotations to continue to have their plot titles rendered. This gives context to the warning about lacking annotations. Otherwise the warning floats "in space" and we lose track of what data is missing the annotation.

Because the audience for spatial plots is small, I've made this a non-blocking suggestion. As long as it is not too time-consuming to implement, it increase clarity if the plot titles could be retained.

Attached is a screencapture showing how those titles disappear when the warning message is rendered.
https://github.com/user-attachments/assets/b38f57fc-fa77-4396-a72e-5b0542b5b20f

bistline:
> Non-blocking suggestion: allow the plots that lack the requested annotations to continue to have their plot titles rendered. This gives context to the warning about lacking annotations. Otherwise the warning floats "in space" and we lose track of what data is missing the annotation.

[90d4c34](https://github.com/broadinstitute/single_cell_portal_core/pull/2128/commits/90d4c34b441ae31ae3b3daa999daab0d913a954d)

I will note that the plot title is preserved in the warning message, but it's not very visible, so this suggestion is much more coherent.

PR #2133, opened 2024-09-18
Add spinner for remote validation, improved timing (SCP-5797, SCP-5776)

bistline:
#### BACKGROUND & CHANGES
This update improves several aspects of remote file validation when using the "bucket path" feature in the upload wizard:

* **Faster invocation**: validation is now invoked after a 300ms delay, removing the need to click outside the input field.  Additionally, this timeout is cleared on every change event, meaning that only one validation request will be issued.
* **File extension check**: validation is not invoked unless the user has provided a file extension (denoted by the existence of a . in the filename).  Once provided, this will be checked against the list of allowed file extensions like uploads are.
* **Clearing validations**: if a user empties the input field, all existing validations are cleared out, returning the control to the default state

Demo of updated functionality:

https://github.com/user-attachments/assets/fc69a2ea-4d6b-48b2-b585-b9693468ad46

#### MANUAL TESTING
1. Boot as normal and go to the upload wizard for a new/empty study
2. In any form, click the "Use bucket path" button and then "Browse bucket"
3. After authenticating, place both valid and invalid files in the bucket (e.g. for AnnData, use `test/test_data/anndata_test.h5ad` as a valid file and `test/test_data/anndata_test_invalid_disease.h5ad` as an invalid file)
4. Copy the GS URL of a valid file
5. Back in the upload wizard, paste this into the input field and confirm you see the `Validating...` message after a brief delay
6. Confirm there are no errors
7. Copy the GS URL of the invalid file and repeat, confirming that error messages are shown
8. Clear out the input and confirm that the error messages disappear and the input is back in the default state
9. Start typing a filename and confirm no validation executes until a period is typed

### DIFF
--- a/app/javascript/components/upload/FileUploadControl.jsx
+++ b/app/javascript/components/upload/FileUploadControl.jsx
@@ -104,11 +104,11 @@ export default function FileUploadControl({
     }
 
     // don't continue unless a dot is present (otherwise, no valid file extension)
-    if (trimmedPath.indexOf('.') < 0 ) { return false }
+    if (!trimmedPath.includes('.')) { return false }
 
     const fileType = file.file_type
     const fileExtension = `.${trimmedPath.split('.').slice(-1)[0]}`
-    if (!inputAcceptExts.includes(fileExtension)) {
+    if (fileExtension.length > 1 && !inputAcceptExts.includes(fileExtension)) {
       const invalidExt = {
         errors: [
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Much smoother UX!  Code looks good overall, and manual test passed.

I suggest a validation UX tweak, and a minor readability refinement.

PR #2142, opened 2024-09-25
Allow providing external link for AnnData clusters (SCP-5816)

bistline:
#### BACKGROUND & CHANGES
This fixes an issue with the AnnData upload UX where the external link feature for clustering data does not persist when saving the form.  This was due to how the data is validated and saved inside the `AnnDataFileInfo` object and the parameters not being allowed or handled correctly.  Now, these parameters are allowed and the data is saved correctly.  Additionally, this cleans up some tests where names of clustering in AnnData files were accidentally prepending `X_`.

This does bring up a frequent issue when adding attributes to models - the need to add them to the "strong parameters" block in associated controllers (e.g. `params.require(model_name).permit`).  I created SCP-5818 to track this.  It's not critical work, but would hopefully alleviate this issue.

#### MANUAL TESTING
You will need to turn off caching for this to work properly, or start DelayedJob.

1. Boot as normal and sign in
2. Load the upload wizard for any AnnData-based study
3. Go to the clustering tab and add an external link for one of the clusters
4. Open the explore tab for the above study/cluster and confirm the link is present above the scatter plot legend

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -749,7 +749,8 @@ module Api
           heatmap_file_info_attributes: [:id, :_destroy, :custom_scaling, :color_min, :color_max, :legend_label],
           cluster_form_info_attributes: [
             :_id, :name, :obsm_key_name, :description, :x_axis_label, :y_axis_label, :x_axis_min, :x_axis_max,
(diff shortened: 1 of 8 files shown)


jlchang (approved):
Thanks for helping me kick out the gremlins, @bistline! Functional review looks good (sans gremlins).

PR #2143, opened 2024-09-26
Instant validation for only-alphanumeric-underscore headers (SCP-5790)

eweitz:
This adds client-side checks for disallowed characters in annotation names.  It makes upload smoother, and thus faster.

### Impact
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2129), files added by users lacked quick validation that metadata annotation names only contained letters, numbers, and underscores.  Authors had to upload the file first.  Just yesterday, a user uploaded a 3 GB classic file, only to learn at the end they broke this simple rule -- their annotation name contained a period.  Similar violations account for [8% of slow validation failures](https://mixpanel.com/s/3q4t3s).  Error messaging was also fragmented; authors needed to monitor their email.

Now, this validation is instant.  It applies to classic metadata files and AnnData files.  Authors get feedback in < 1 second, not 5-45 minutes.  And messaging is done right in the app, contextually.

This better UX will help authors add more studies and, as we mature integration, engage with useful tools like [Cell Annotation Service](https://singlecell.broadinstitute.org/single_cell/features/2024-09-18-explore-cell-typing-with-ai-take-a-sneak-peek-at-cellarium-cell-annotation-service) (CAS).

### Demo
Here's how it looks!

https://github.com/user-attachments/assets/2ff967cc-4416-4410-936d-be343e4a8ce2

### Technical overview
This rule exists to simplify adding data to BigQuery, which requires quoting unless the field contains only alphanumeric characters and underscores.  BigQuery enables cross-study faceted search.  This rule could be loosened for annotations that are not globally queryable, but that's a potential refinement for the future.

These changes also fix false-positive test failures due to file renaming, and very old false-_negative_ test data that had characters which have been invalid since this underlying Ingest Pipeline validation rule was added a few years ago.

### Test
New automated tests verify these updates.  To optionally manually test:
1.  Go to AnnData upload UI in "AnnData CSFV" study
2. Select `test/test_data/anndata/invalid_annotation_name_period.h5ad`
3. Confirm error message appears
4. Go to classic upload UI
5. In "Metadata" tab, select `test/test_data/validation/metadata_invalid_annotation_name_period.tsv`
6. Confirm error message appears
7. In "Clustering" tab, select `test/test_data/validation/cluster_valid_annotation_name_period.txt`
8. Confirm error message does not appear

This satisfies SCP-5790.


bistline (approved):

jlchang (approved):
Non-blocking observation: the "only-alphanumeric-underscore headers" rule was never applied to cluster files, just metadata. Only metadata files are evaluated by jsonschema, which is where the (slow) error is detected. We don't store any cluster-specific annotations into BigQuery so we've never applied this rule to cluster files. 

One file you indicated as false _negative_ test data was a cluster file, where the rule was never applied and the other is a non-conventional metadata file which was never intended as test data for jsonschema assessment and would not fail this rule but would have failed many other validations that belong to jsonschema evaluation.

If it isn't onerous to apply the rule only to metadata files, I'd suggest we make that change but I don't consider the lack to be blocking.

eweitz:
> the "only-alphanumeric-underscore headers" rule was never applied to cluster files, just metadata

Great insight!  I've made that refinement per your suggestion, which definitely improves signal-to-noise ratio for validation.

PR #2144, opened 2024-09-26
Fixing issues with reporting ingestSummary events (SCP-5802)

bistline:
#### BACKGROUND & CHANGES
This update addresses some issues surrounding reporting the `ingestSummary` event for AnnData files.  These events are a rollup of the 3 main ingest processes for AnnData files - clustering, metadata, and processed expression.  Recent updates to how related jobs were identified proved to be prone to race conditions that meant jobs checking simultaneously would identify each other as "still processing" and both exit, leading to no summary being sent.  Now, the query specifically looks for jobs of the 3 main types that are not "done" according to their status in the LifeSciences API.  Furthermore, this check can happen after _any_ job that is submitted for this file, be it main ingest, differential expression, subsampling, etc.  Once it is determined that there are no more jobs running that are extracting data, a summary is sent, and then that file is marked as having submitted an `ingestSummary`.  No further summaries will be sent for that file.  

Additional clusterings or raw counts data can theoretically still be extracted, but neither will result in submitting an `ingestSummary` - only the standard `ingest` event will be reported.  Since the point of the `ingestSummary` is to gauge the amount of time it takes a user to go from initial upload to an "initialized" study and whether or not this was successful, this is acceptable.

This also addresses an unrelated bug with deleting AnnData files where DE output files could not be clean up due to missing annotation data (was being used only for logging).

#### MANUAL TESTING
This is complicated to test, especially since in normal development the race condition can never happen as it requires multiple concurrent job workers processing jobs for the same AnnData file.  However, you can start DelayedJob locally with multiple workers using the following command (once your environment is initialized):
```
bin/delayed_job start development --pool=default:6, --pool=cache:2
```
Note that once you are done testing, you will need to manually stop the workers as they are launched headlessly.  This can be done with `bin/delayed_job stop development`.

Once your workers are running, follow these steps:
1. Boot as normal, sign in, and create a new study
2. Select the AnnData UX and fill out the forms as normal (supplying multiple clusterings and raw counts helps, if possible)
3. Once the file is uploaded, look in the logs for the extraction job and copy both the filename and MongoDB ID
4. In a separate terminal, go to the project root and tail your `development.log` file to look for messages regarding the AnnData summary (your filename/id will be different but the wording will be exact).  It will take several minutes to appear, but while you may see multiple "Checking" messages you should only see one message that says "Sending AnnData summary" and no further messages after that:
```
$ LOG_MESSAGE="AnnData summary for compliant_pbmc3K.h5ad (66f5b7b38bdcc6e804f001cb)"
$ tail -f log/development.log | grep -e $LOG_MESSAGE
...
Checking AnnData summary for compliant_pbmc3K.h5ad (66f5b7b38bdcc6e804f001cb) after ingest_expression
Sending AnnData summary for compliant_pbmc3K.h5ad (66f5b7b38bdcc6e804f001cb) after ingest_expression
```
5. Go to the [Dev Mixpanel events list](https://mixpanel.com/project/2085496/view/19055/app/events) and query for your event (note it may take a minute or so for it to appear):
    * event: `ingestSummary`
    * studyAccession: `<your accession>`
    * fileName: `<your file>`
 
Screenshot 2024-09-26 at 4 49 46 PM

6. Your event should have the correct number for `numFilesExtracted`: 1 each for expression/metadata, 1 if you extracted raw counts, and 1 per clustering.

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -981,7 +981,7 @@ class IngestJob
     mixpanel_log_props = get_job_analytics
     # log job properties to Mixpanel
     MetricsService.log(mixpanel_event_name, mixpanel_log_props, user)
-    report_anndata_summary if study_file.is_viz_anndata? && action != :differential_expression
+    report_anndata_summary if study_file.is_viz_anndata? && action != :ingest_anndata
   end
 
   # set a mixpanel event name based on action
@@ -994,7 +994,9 @@ class IngestJob
     pipelines = ApplicationController.life_sciences_api_client.list_pipelines
     previous_jobs = pipelines.operations.select do |op|
       pipeline_args = op.metadata.dig('pipeline', 'actions').first['commands']
-      pipeline_args.detect { |c| c == study_file.id.to_s } && !pipeline_args.include?('--differential-expression')
+      op.done? &&
+        pipeline_args.detect { |c| c == study_file.id.to_s } &&
+        (!pipeline_args.include?('--differential-expression') && !pipeline_args.include?('--subsample'))
     end
     # get total runtime from initial extract to final parse
     initial_extract = previous_jobs.last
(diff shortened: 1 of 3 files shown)


jlchang (approved):
Functional review: Confirmed one IngestSummary event (note: `hidden event` in Mixpanel
`Sending AnnData summary for compliant_pbmc3K.h5ad (66fd3d54eb0c8dcba8f32c7a) after ingest_expression`

I extracted processed, metadata, X_tsne, X_umap and raw. The Mixpanel event indicates 4 files extracted - does `raw` not count?
Screenshot 2024-10-02 at 9 08 12 AM

I realize that `raw` doesn't generate a fragment file but it might count as an "extraction"? Would we want to reflect "raw" in the numFilesExtracted or leave numFilesExtracted as-is and have an array of detailed extraction types. Possible the detailed info isn't useful, don't have enough caffeine in me to have a opinion.

app/models/ingest_job.rb
```diff
@@ -994,7 +994,9 @@ def anndata_summary_props
     pipelines = ApplicationController.life_sciences_api_client.list_pipelines
     previous_jobs = pipelines.operations.select do |op|
       pipeline_args = op.metadata.dig('pipeline', 'actions').first['commands']
-      pipeline_args.detect { |c| c == study_file.id.to_s } && !pipeline_args.include?('--differential-expression')
+      op.done? &&
+        pipeline_args.detect { |c| c == study_file.id.to_s } &&
+        (!pipeline_args.include?('--differential-expression') && !pipeline_args.include?('--subsample'))
```
eweitz, line 999: Do we have any central / const-like variable that records "special actions" like jobs for DE or subsampling?  If we anticipate new such pipeline arguments for CAS or on-demand pairwise DE, then I could see not having that being a likely cause of issues down the line.

That's far from a blocking issue, just at least a note to keep in the back of our minds.

eweitz (approved):
Code looks good!  Glad we'll have more accurate metrics on AnnData ingest.

I ask a question about a potential maintainability refinement, no blockers.

app/models/ingest_job.rb
bistline: Sort of?  We have the below constant for "special actions", but it's used to determine which jobs don't email users:

```
SPECIAL_ACTIONS = %i[differential_expression render_expression_arrays image_pipeline].freeze
```

I think your comment still stands though, and I can rewrite this as an inclusive check for the main three ingest processes (expression, metadata, clustering).

bistline: [72544ff](https://github.com/broadinstitute/single_cell_portal_core/pull/2144/commits/72544ffdb774ef6f623567c15b2acd79be7eff70)

bistline:
> Functional review: Confirmed one IngestSummary event (note: `hidden event` in Mixpanel `Sending AnnData summary for compliant_pbmc3K.h5ad (66fd3d54eb0c8dcba8f32c7a) after ingest_expression`
> 
> I extracted processed, metadata, X_tsne, X_umap and raw. The Mixpanel event indicates 4 files extracted - does `raw` not count? Screenshot 2024-10-02 at 9 08 12 AM
> 
> I realize that `raw` doesn't generate a fragment file but it might count as an "extraction"? Would we want to reflect "raw" in the numFilesExtracted or leave numFilesExtracted as-is and have an array of detailed extraction types. Possible the detailed info isn't useful, don't have enough caffeine in me to have a opinion.

[631f415](https://github.com/broadinstitute/single_cell_portal_core/pull/2144/commits/631f41598fd7ab02f8aa69613cef9221a1ed285f)

PR #2145, opened 2024-09-27
Fix validation error when gzipped file lacks proper extension (SCP-5820)

eweitz:
This fixes a misformatted validation error message.  It also suggests a better way for the author to fix the problem.

### Impact
SCP requires gzipped files to have a certain file extensions -- either .gz, .tbi, or .bam.  If the gzip-compressed file lacks such an extension, a `encoding:missing-gz-extension` error is logged, and a human-friendly message should be displayed.

Previously, however, the error message was somewhat mangled.  The actionability of the error message also wasn't as useful as it could be.  These issues caused confusion -- a user who encountered this wrote into us yesterday for help.

Now, the error message is formatted properly, and the recommendation for the author to fix the problem is clearer.

### Demo
Here's how it looks.

https://github.com/user-attachments/assets/130115d1-5f76-4add-953e-e30132e0a784

### Test
An automated test protects against regression.  Optionally, to manually test:
1.  Go to classic upload UI for a study
2.  Select "test/test_data/validation/missing-gz-extension.txt` (which is gzipped, but lacks a ".gz" file extension)
3. Confirm error message appears:
> Only files with extensions ".gz", ".bam", or ".tbi" may be gzipped.  Please add a ".gz" extension to the file name, or decompress the file, and retry.

This fixes SCP-5820.  This is the 1st of 2 PRs related to a user-reported issue with CSFV, RDS files, and gzip.  The PRs are split because the fixes address independent aspects of the problem.


bistline (approved):

jlchang (approved):
Functional review: confirmed updated error message.

PR #2148, opened 2024-10-01
Dynamically hide cell filtering UX when not applicable (SCP-5819)

bistline:
#### BACKGROUND & CHANGES
This update changes the logic of when the cell filtering UX is enabled.  Previously, any cluster that met requirements (i.e. was indexed, annotations available) would enable the cell filtering button.  However, dot plots, heatmaps, and annotated & consensus scatter plots all do not support cell filtering.  The panel will load, but none of the controls affect the plot.  This is confusing to users, and can lead to spurious error reports.

Now, only the scatter & distribution tabs will allow the user to filter plots.  These tabs are only enabled in non-expression contexts, when a single gene has been queried, or when multiple genes have been collapsed by a consensus metric.  Any existing filtered state will be saved if the user enters a view that doesn't support cell filtering, and the panel will be hidden.  The `Filter plotted cells` button will now display `Filtering unavailable` with a message that it is only supported in the above contexts.  Once the user returns to a view that supports cell filtering, the filtering panel reopens with previous state already loaded.

Short demo of how this looks in normal usage:

https://github.com/user-attachments/assets/fb99cd0b-5de1-4001-bf7c-289136a0fc34

Also, this removes a tooltip from the cell filtering button about subsampling.  Since cell filtering is now available on subsampled data, this message no longer applies.

#### MANUAL TESTING
1. Boot as normal and load any study that has clustering, expression data, and allows cell filtering
2. Open the cell filtering panel and select any filter(s)
3. Query for one gene and confirm the filtering panel stays open and the filtered state persists
4. Query for another gene and confirm that the options panel is now shown, no cells are filtered and the filtering button is disabled with the tooltip `Cell filtering cannot be shown in this context (only scatter or distribution tabs)`
5. Use the "Collapse genes by" menu and confirm the cell filtering panel reopens with the previous filtered state preserved
6. Remove one of the genes and confirm the cell filtering panel remains open

### DIFF
--- a/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
+++ b/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
@@ -454,7 +454,7 @@ export default function ExploreDisplayPanelManager({
                     </div>
                   </>
                 }
-                { showCellFiltering && !clusterCanFilter &&
+                { showCellFiltering && !clusterCanFilter && allowCellFiltering &&
                   <>
                     <div className="row">
                       <div className={`col-xs-12 cell-filtering-button ${shownTab === 'scatter' && !studyHasDe ? 'create-annotation-cell-filtering' : ''}`}>
--- a/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
+++ b/app/javascript/components/explore/ExploreDisplayPanelManager.jsx
@@ -200,7 +200,7 @@ export default function ExploreDisplayPanelManager({
   showDifferentialExpressionPanel, setIsCellSelecting, currentPointsSelected, isCellSelecting, deGenes,
   setDeGenes, setShowDeGroupPicker,
   cellFaceting, cellFilteringSelection, cellFilterCounts, clusterCanFilter, filterErrorText,
-  updateFilteredCells, panelToShow, toggleViewOptions, canFilter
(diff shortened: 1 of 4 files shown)


jlchang (approved):
Functional review: looks great!

Thanks for fixing the double button. It all looks good (if I'm a bit more patient and give my local environment some slack for being more pokey than production)

app/javascript/components/explore/ExploreDisplayPanelManager.jsx
```diff
@@ -477,6 +469,21 @@ export default function ExploreDisplayPanelManager({
                     </div>
                   </>
                 }
+                { showCellFiltering && !allowCellFiltering &&
+                  <>
+                    <div className="row">
+                      <div className={`col-xs-12 cell-filtering-button ${shownTab === 'scatter' && !studyHasDe ? 'create-annotation-cell-filtering' : ''}`}>
+                        <button
+                          disabled="disabled"
+                          className={`btn btn-primary`}
```
eweitz, line 478: For analytics on whether users click this, which would signal interest in cell filtering in the disabled context:

```suggestion
                          className={`btn btn-primary cell-filtering-button-disabled`}
```

eweitz (approved):
Code looks good!  Nice UX polish.  I suggest a non-blocking observability refinement.

PR #2151, opened 2024-10-08
Enable canvas recording for Sentry replays (SCP-5826)

bistline:
#### BACKGROUND & CHANGES
This updates the Sentry SDK to `7.119.1` to enable recording of canvas & WebGL DOM components during Session Replays and also address an open [Dependabot alert](https://github.com/broadinstitute/single_cell_portal_core/security/dependabot/159).  This will allow us much greater insight into user behavior when triaging front-end errors.

Demo of new replays (fast-forward to ~6:15 to see the new plots): https://broad-institute.sentry.io/replays/d6ee1185376e4b70832e94a965801330/?environment=development&project=1424198&query=&referrer=%2Freplays%2F&statsPeriod=14d&yAxis=count%28%29

#### MANUAL TESTING
1. Boot as normal
2. In your IDE, open `app/javascript/lib/sentry-logging.js` and in `getIsSuppressedEnv()` to add `return false` at line 71
3. On line 122, set `replaysOnErrorSampleRate` to `1.0`
4. Back in a browser, load any study with fully initialized plots
5. Query for one gene, and then once the plot has loaded use the `Return to cluster view` button immediately to the right of the gene search bar:
Screenshot 2024-10-08 at 11 12 32 AM
6. This will throw an error that is reported to Sentry (you can confirm this in the JS console & network tab) that should also now include a replay
7. Go to [Sentry Replays](https://broad-institute.sentry.io/replays/?environment=development&project=1424198&statsPeriod=14d), make sure you have the `development` environment selected, and then find your replay, confirming you see the content of scatter plots


package.json
```diff
@@ -22,8 +22,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.12.0",
     "@fortawesome/react-fontawesome": "^0.1.8",
     "@reach/router": "^1.3.3",
-    "@sentry/react": "^7.54.0",
-    "@sentry/tracing": "^7.54.0",
+    "@sentry/react": "^7.119.1",
```
eweitz, line 25: First time I've seen a 3-digit minor version!

app/javascript/lib/sentry-logging.js
```diff
@@ -114,8 +114,8 @@ export function setupSentry() {
   Sentry.init({
     dsn: 'https://a713dcf8bbce4a26aa1fe3bf19008d26@o54426.ingest.sentry.io/1424198',
     integrations: [
-      new Sentry.BrowserTracing(),
```
eweitz, line 117: I assume keeping [browser tracing](https://docs.sentry.io/platforms/javascript/tracing/) needs more than merely updating `@sentry/tracing` to the same version as `@sentry/react`.  If so, this removal makes sense to me, because I can't think of any cases where we've gotten high-priority insight from the granular performance metrics it provides.

eweitz (approved):
Code looks good.  Nice observability improvement!  

Thanks for confirming this doesn't seem to affect rendering performance for users.

app/javascript/lib/sentry-logging.js
bistline: `@sentry/tracing` has been deprecated since `7.47`, and while upgrading I received several warnings to that effect.  Removing it seems to have no apparent consequences.

https://github.com/getsentry/sentry-javascript/blob/7.47.0/MIGRATION.md/#remove-requirement-for-sentrytracing-package-since-7460

jlchang (approved):
Excited that will get more insight from Sentry replays with this change.

PR #2152, opened 2024-10-08
Update filtering tooltip to indicate data is unfiltered (SCP-5819)

bistline:
#### BACKGROUND & CHANGES
This changes the tooltip when cell filtering has been unapplied due to UX state (e.g., multi-gene view) to indicate that the data has in fact reverted to the unfiltered state.

#### MANUAL TESTING
1. Follow instructions from #2148, but in step for confirm the tooltip now displays `Cell filtering cannot be applied in this context (only scatter or distribution tabs) - visible plots are unfiltered`


eweitz (approved):

jlchang (approved):
Looks good!

PR #2154, opened 2024-10-09
Automatic retries of OOM failures for ingest processes (SCP-5827)

bistline:
#### BACKGROUND
With the addition of AnnData parsing & automated differential expression calculation for qualifying studies, we have seen a rise of ingest processes failing with error codes of `137` or `139`, both of which indicate an out of memory exception.  Usually, these jobs can be re-rerun using larger GCE instances and will eventually successfully ingest.  However, this process is entirely manual and requires direct admin intervention which can be quite cumbersome as it requires reconstituting files/parameters from various sources.

#### CHANGES
Now, any ingest process that fails to ingest with either of the above error codes and uses a parameters class (i.e. `AnnDataIngestParameters`, `DifferentialExpressionParameters`) will automatically retry the job using the next available machine type.  For instance, if a file fails with `n2d-highmem-8`, it will automatically retry with `n2d-highmem-16`.  This process will continue all the way up to `n2d-highmem-64`, after which if the file fails to ingest, all retries cease and end user is finally notified.  Normal admin messaging & cleanup procedures still apply for all retries, ensuring visibility into the process.  Additionally, we could also set up Mixpanel alerts for any job failures with the corresponding error codes to further enhance visibility.

The reason for the requirement on the parameters class is that normal ingest jobs of "classic" SCP files do not employ dynamic scaling of machines, and would require more significant refactoring to enable.  Furthermore, since we starting tracking `exitStatus` for ingest jobs, there have been no instances of non-AnnData files failing due to OOM exceptions.

#### MANUAL TESTING
To properly test, this does require the use of a large AnnData file that is known to fail on the default `machine_type` used for its ingest.  Such a file can be found here if you do not have access to one locally: https://console.cloud.google.com/storage/browser/_details/broad-singlecellportal-staging-testing-data/SCP2807/HRCA_snRNA_AC.h5ad;tab=live_object?project=broad-singlecellportal-staging

1. Boot all services and create a new study, selecting the AnnData upload UX
2. Copy the above file to the study's bucket and save the file using the "bucket path" option
3. In `development.log`, look for the following message (the ID/accession will be different):
```
Retrying ingest_anndata after 137 failure for HRCA_snRNA_AC.h5ad:6706c4a794ec8f2272dc977f (SCP161) with machine_type: n2d-highmem-16
```
4. Note a new job is launched immediately afterwards with the correct `machine_type`:
```
Request object sent to Google Life Sciences API, excluding 'environment' parameters:
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 66d8bcf394ec8f692a4ebea8
  - "--study-file-id"
  - 6706c4a794ec8f2272dc977f
  - "--user-metrics-uuid"
  - c6a08597-c735-4c93-b069-4b4a74b708ee
  - ingest_anndata
  - "--ingest-anndata"
  - "--anndata-file"
  - gs://fc-b04eabc0-55c0-4047-9a02-5d0f98a8f528/HRCA_snRNA_AC.h5ad
  - "--obsm-keys"
  - '["X_umap"]'
  - "--extract"
  - '["cluster", "metadata", "processed_expression", "raw_counts"]'
  :image_uri: gcr.io/broad-singlecellportal-staging/scp-ingest-pipeline:1.35.0
  :labels: {}
  :timeout: {}
:resources:
  :regions:
  - us-central1
  :virtual_machine:
    :boot_disk_size_gb: 300
    :labels:
      :study_accession: scp161
      :user_id: 63e27b72f39faa00560c1e25
      :filename: hrca_snrna_ac_h5ad
      :action: ingest_pipeline
      :docker_image: scp-ingest-pipeline
      :docker_tag: '1_35_0'
      :environment: development
      :file_type: anndata
      :machine_type: n2d-highmem-16
      :boot_disk_size_gb: '300'
    :machine_type: n2d-highmem-16 <==== now on 16 instead of 8
...
```

Note: you do not need to let the file complete the ingest process as this will take quite a while - it is still running 90 min into the retry for my manual test.

### DIFF
--- a/app/models/concerns/compute_scaling.rb
+++ b/app/models/concerns/compute_scaling.rb
@@ -4,11 +4,11 @@ module ComputeScaling
 
   # GCE machine types and file size ranges
   # produces a hash with entries like { 'n2-highmem-4' => 0..16.gigabytes }
-  # change GB_PER_CORE in parent class to adjust scaling (1-8, higher number means more aggressive scaling)
+  # change GB_PER_CORE in parent class to adjust scaling (1-8, lower number means more aggressive scaling)
   def scaled_machine_types
     gb_per_core = defined?(self.class::GB_PER_CORE) && self.class::GB_PER_CORE || 4
     num_cores = [4, 8, 16, 32, 48, 64]
-    ram_per_core = num_cores.map { |core| (core * gb_per_core).gigabytes }.freeze
+    ram_per_core = num_cores.map { |core| (core * gb_per_core).gigabytes }
     num_cores.map.with_index do |cores, index|
       floor = index == 0 ? 0 : ram_per_core[index - 1]
       limit = index == num_cores.count - 1 ? ram_per_core[index] * 2 : ram_per_core[index]
@@ -36,7 +36,7 @@ module ComputeScaling
   # find the next largest machine to use for an ingest process
   def next_machine_type
     machine_names = scaled_machine_types.keys
-    # return if there are no larger machines available
+    # nil return used as escape clause for stopping retries after largest machine fails
     return nil if machine_type == machine_names.last
 
     current_machine_idx = machine_names.index(machine_type)
--- a/app/models/concerns/compute_scaling.rb
+++ b/app/models/concerns/compute_scaling.rb
@@ -32,4 +32,14 @@ module ComputeScaling
     scaled_machine = scaled_machine_types.detect { |_, mem_range| mem_range === file_size }&.first
     scaled_machine || default_machine_type
   end
+
+  # find the next largest machine to use for an ingest process
+  def next_machine_type
+    machine_names = scaled_machine_types.keys
+    # return if there are no larger machines available
+    return nil if machine_type == machine_names.last
+
+    current_machine_idx = machine_names.index(machine_type)
+    machine_names[current_machine_idx + 1]
+  end
 end
--- a/app/models/concerns/compute_scaling.rb
+++ b/app/models/concerns/compute_scaling.rb
@@ -4,9 +4,9 @@ module ComputeScaling
 
   # GCE machine types and file size ranges
   # produces a hash with entries like { 'n2-highmem-4' => 0..16.gigabytes }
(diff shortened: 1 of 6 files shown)


app/models/concerns/compute_scaling.rb
```diff
@@ -4,11 +4,11 @@ module ComputeScaling
 
   # GCE machine types and file size ranges
   # produces a hash with entries like { 'n2-highmem-4' => 0..16.gigabytes }
-  # change GB_PER_CORE in parent class to adjust scaling (1-8, higher number means more aggressive scaling)
+  # change GB_PER_CORE in parent class to adjust scaling (1-8, lower number means more aggressive scaling)
```
eweitz, line 7: "More aggressive scaling" means "more RAM per core", right?  If so, higher (not lower) number means more aggressive scaling.
```suggestion
  # change GB_PER_CORE in parent class to adjust scaling (1-8, higher number means more aggressive scaling)
```

eweitz (approved):
Code looks good!  Looking forward to more effectively scalable ingest.

I note a seeming docstring bug that could be confusing.

app/models/concerns/compute_scaling.rb
bistline: No, I originally thought the same but realized it really is the inverse that is true, and I think because the name is misleading.  The GB_PER_CORE number is used as the multiplier when determining ranges for machine types, _not_ the actual number of GB being allocated per core.  The latter is constant at 8 GB.  Instead, setting GB_PER_CORE lower causes the ranges between machines to be smaller, meaning that as file size increases, the machine type scales more rapidly.

I think I'm going to change the name of that constant to be less confusing.

bistline: [51b193a](https://github.com/broadinstitute/single_cell_portal_core/pull/2154/commits/51b193a17318ef9358042388b3f7c0fb49ce9797)

jlchang (approved):
Code looks good. Looking forward to trying it out!

PR #2155, opened 2024-10-16
Adding high-level metadata summary to search results (SCP-5845)

bistline:
#### BACKGROUND
A recent [Zendesk request](https://broadinstitute.zendesk.com/agent/tickets/322279) was submitted by a user looking for ways to export or preview high-level information about studies in search results.  Currently, the only way to filter studies is through the search interface, but only on the filter values selected.  There is no way to glean additional high-level information (e.g. species, disease, etc.) about search results without loading each study individually and inspected the contents of said annotations.  The requesting user referenced the [Genomic Data Commons cohort builder](https://portal.gdc.cancer.gov/analysis_page?app=CohortBuilder&tab=general) as an example.

#### CHANGES
This update now adds a table to all search results (both SCP- and HCA-based) that contains the entries from 5 metadata convention based annotations: `disease__ontology_label`, `organ__ontology_label`, `species__ontology_label`, `library_preparation_protocol__ontology_label`, and `sex`.  This data is now present in the search API response for any request.  Each study will show up to 2 values per annotation, with the rest hidden behind a `{n} more...` label with a tooltip containing the rest of the labels.  Any studies lacking the information for a given column will display `unspecified`.  These result tables are shown both in study- and gene-based searches.

Examples of results:
Screenshot 2024-10-21 at 3 25 12 PM

Screenshot 2024-10-21 at 3 24 43 PM

`more...` tooltip
Screenshot 2024-10-21 at 3 27 02 PM

Example in global gene search:
Screenshot 2024-10-21 at 3 25 59 PM

This can also be viewed as potential preparatory work for a "cohort-builder" type interface, though this would only be for study-level data.  Cell-level cohorts would likely have more power and utility, but SCP lacks a true cell-level search interface, though it would be possible for metadata convention compliant studies.

#### MANUAL TESTING
1. Boot as normal and confirm default results contain metadata information
2. Run any search (facet- or text-based) and confirm that normal functionality is not affected
3. Run a global gene search and confirm the metadata table renders above the violin plot as well


app/javascript/components/search/results/ResultMetadataTable.jsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// return a table with the 5 top-level metadata entries and their values for a SCP study
+export default function ResultMetadataTable({study}) {
+  const headers = []
+  const studyValues = []
+  const tableId = `${study.accession}-cohort-metadata`
+
+  Object.entries(study.metadata).map((entry, index) => {
+    const header = entry[0].replace(/_/g, ' ')
```
eweitz, line 48: I like that this makes the annotation names human friendlier.  And I'd go further by applying [parseAnnotationName](https://github.com/broadinstitute/single_cell_portal_core/blob/1a111d6cb0fd678989c83ec9876e854e0940635f/app/javascript/components/explore/FacetComponents.jsx#L116).  That would sentence-case these strings, and apply other text polishing for easy readability.

app/javascript/components/search/results/ResultMetadataTable.jsx
```diff
@@ -0,0 +1,69 @@
...
+  })
+  return returnLabels.join('|')
+}
+
+// show list of metadata values as badges, truncating as needed
+function metadataList(values, accession, header) {
+  const moreValues = values.splice(maxMetadataEntities)
+  const list = values.map((val, i) => {
+    return <span key={`${accession}-${header}-entry-${i}-val`}
+                 className="label label-primary study-metadata-entry">{sanitizeMetadataLabel(val)}</span>
```
eweitz, line 25:
```suggestion
                 className="badge badge-secondary study-metadata-entry">{sanitizeMetadataLabel(val)}</span>
```

Applying a visually "primary" style to so many elements is definitely prominent, but also perhaps noisy.  

Would a less prominent style suffice?  Something like `badge badge-secondary` might fit.  The green text and white background would be lower contrast but still very legible.

Its green color would also be consistent with the "metadata matches" badges shown when applying a cross-dataset search filter.  The different _position_ seems sufficient to distinguish that existing "metadata match" badge from these new "metadata available" badges.

Perhaps in later work in another scope, we could further unify these "match" and "available" search result UI elements by e.g. using a more "primary" green badge (green _background_, white _text_ for higher contrast) when an available annotation value is also matched by a selected filter.

I don't feel strongly about that.  If you opt to keep the prominent blue, I'd recommend using `$action-color` (#3D5A87) as the background color.  That's the more SCP-conventional primary blue.

eweitz (approved):
Great search results UI enhancement!

I also like the "preparatory work" idea, which AIUI would make these new UI elements clickable / selectable from the _result_ panel and thus update the _query_.  I can't identify an analogous UX aspect in the [GDC Cohort Builder](https://portal.gdc.cancer.gov/analysis_page?app=CohortBuilder&tab=general), so the idea strikes me as novel and nonetheless quite useful. 

The search query UI enrichments we discussed [in #scp-ui-design](https://broadinstitute.slack.com/archives/C024HV2FANL/p1729108938484429?thread_ts=1729098550.136679&cid=C024HV2FANL) strike me as closer to enabling something like a conventional cohort builder UI.  Cohort builder UIs I've seen (including GDC's) seem essentially be faceted search query UIs.  Extending our cross-dataset faceted search query UI with filter counts (ideally _responsive_ / reactive counts with sparkbars, like cell filtering) would get us there.

In any case, this is a nice enhancement!  I suggest some non-blocking, minor text and style refinements for consideration.

app/javascript/components/search/results/ResultMetadataTable.jsx
bistline: I like the idea of visually linking it to the metadata look/feel, so I'm going to go with the inverse styling (green background, white text).  I'm using the same green but slightly transparent (.85) to mute it every so slightly.  Also, I love the idea of making these clickable eventually - that's a really nice enhancement.

jlchang (approved):

Can't wait to see it in production and get feedback!

PR #2161, opened 2024-10-23
update to ingest 1.36.1

jlchang:


eweitz (approved):

bistline (approved):

jlchang:
Build failure was due to OOM error that we're still troubleshooting but does not relate to the change in this PR which was just changing the ingest docker image id.

PR #2166, opened 2024-11-06
Autocomplete on gene symbol & id (SCP-5853)

bistline:
#### BACKGROUND & CHANGES
This update allows the autocomplete feature in the study gene search to also accept gene IDs, when provided by the study owner.  This does not appear to noticeably impact autocomplete performance - on average there are ~30K genes to a study, so going from 30K to 60K doesn't appear to slow down autocomplete.

#### MANUAL TESTING
The 10x ATAC demo study has both gene names and gene IDs, so this makes a good testing candidate.  If you do not have that in your local instance, you'll need to ingest expression data either in MTX or AnnData form that has both.
1. Boot as normal and load a study that supports both gene names and IDs
2. Run as search for `ENSG` and see that results are shown
3. Run a normal search on a gene symbol and see that it also is shown

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -1322,7 +1322,8 @@ class Study
 
   # get all unique gene names for a study; leverage index on Gene model to improve performance
   def unique_genes
-    Gene.where(study_id: self.id, :study_file_id.in => self.expression_matrix_files.map(&:id)).pluck(:name).uniq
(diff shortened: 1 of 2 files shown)


jlchang (approved):
Functional review - searching on gene_id works! So glad it wasn't too hard to get working.

PR #2168, opened 2024-11-12
Extending cache filtering of malicious requests (SCP-5967)

bistline:
#### BACKGROUND & CHANGES
This extends work started in #2025 to filter out more obvious malicious requests from API caching.  Currently, the default behavior in loading annotations is to grab the "first found" annotation if a request specifies a non-existent value.  This provides a backstop that prevents a visualization request from returning `404` when there is data that could be rendered.  This unfortunately means that during a security scan, a large volume of malicious requests in a short time period can result in the app overloading as it attempts to load default responses for a huge number of studies.  This update now automatically blocks many of those requests and responds `400` to prevent this from happening.  This is done at the routing level before any queries are run, so there is no load on the app/database.

It is worth noting that this does not address the underlying issue - there will be updates to these scans resulting in more exploits that will then need to be blocked in a similar fashion.  A better fix would be to _not_ load default clusters/annotations in these scenarios and let the app respond `404`.  However, there is a fair amount of functionality that depends on this default response, and therefore that approach is not a simple refactor.

#### MANUAL TESTING
1. Boot as normal and load the Swagger doc for [cluster visualization](https://localhost:3000/single_cell/api/v1#/Visualization/study_cluster_path) (note: it helps to load a smaller study with less than 5K points)
2. Enter a valid study/cluster/annotation combo and confirm you get a `200 OK` response
3. Change any value in the params (like `cluster_name` or `annot_name`) to one of these values: `.git`, `NULL OR 1`, `CODE_POINTS_TO_STRING`, `UPDATEXML`, or `$%7Benv` and submit
4. Confirm you get a `400 (Bad Request)` response

### DIFF
--- a/app/controllers/api/v1/concerns/api_caching.rb
+++ b/app/controllers/api/v1/concerns/api_caching.rb
@@ -3,7 +3,11 @@ module Api
     module Concerns
       module ApiCaching
         extend ActiveSupport::Concern
+
+        # regexes for blocker types of malicious requests
+        # this prevents the app from being flooded during security scans that can cause runaway viz caching
         XSS_MATCHER = /(xssdetected|script3E)/
+        SCAN_MATCHER = /(\.(git|svn|php)|NULL(%20|\+)OR(%20|\+)1|CODE_POINTS_TO_STRING|UPDATEXML|\$%7Benv)/
 
         # check Rails cache for JSON response based off url/params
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good!

bistline:
Closing in favor of #2169

PR #2169, opened 2024-11-12
Better handling of non-existent annotations (SCP-5867, SCP-5868)

bistline:
#### BACKGROUND & CHANGES
Currently, if the specified annotation is not found in a visualization request, the API will fall back to the default (or first available) annotation for that study.  This has several unintended consequences, one of them being that normal penetration scans that look for security vulnerabilities can result in memory/CPU saturation for a deployed instance as multiple bogus requests result in loading default data.  Now, if an annotation is specified and not found, the API will correctly respond `404`.  Normal "empty" requests that do not specify an annotation will still load the default annotation for a study, if one is available.

Additionally, known malicious requests that can lead to potential caching issues are blocked at the routing level with `400`.  While not strictly necessary, this will further reduce load by exiting before checking the cache on requests that are part of a security scan.  Since these scans are ever-evolving, the `404` approach is the more complete solution.

#### MANUAL TESTING
2. Confirm the empty default view renders without issue
3. Open the Network tab in Chrome Dev Tools
4. Select a different annotation to populate parameters into the URL
5. In the URL, change the name of the annotation to `not-found`.  For example, if the `annotation` parameter is `cell_type__ontology_label--group--study`, change it to `not-found--group--study` and submit the request.
6. Confirm that you see the error message `Error: Annotation "not-found" could not be found` and that the associated request in the network tab has a status of `404`.
7. Select a different annotation and confirm the plot is restored
8. Repeat step 5, but use `UPDATEXML` as the annotation name
9. Confirm the plot shows an error with the message `Bad request` and the associated request in the network tab shows a status of `400`

### DIFF
--- a/app/controllers/api/v1/visualization/annotations_controller.rb
+++ b/app/controllers/api/v1/visualization/annotations_controller.rb
@@ -100,6 +100,10 @@ module Api
 
         def show
           annotation = self.class.get_selected_annotation(@study, params)
+          if annotation.nil?
+            head :not_found and return
+          end
+
           render json: annotation
         end
 
@@ -148,6 +152,10 @@ module Api
 
         def cell_values
           annotation = self.class.get_selected_annotation(@study, params)
+          if annotation.nil?
+            head :not_found and return
+          end
+
           cell_cluster = @study.cluster_groups.by_name(params[:cluster])
           if cell_cluster.nil?
             cell_cluster = @study.default_cluster
@@ -364,9 +372,6 @@ module Api
         # parses the url params to identify the selected cluster
         def self.get_selected_annotation(study, params)
           annot_params = get_annotation_params(params)
-          if annot_params[:name] == '_default'
-            annot_params[:name] = nil
-          end
           cluster = nil
           if annot_params[:scope] == 'cluster'
             if params[:cluster].blank?
@@ -384,7 +389,7 @@ module Api
         def self.get_facet_annotations(study, cluster, annot_param)
           annotations = annot_param.split(',').map { |annot| convert_annotation_param(annot) }
           annotations.map do |annotation|
-            AnnotationVizService.get_selected_annotation(study, cluster:, fallback: false, **annotation)
+            AnnotationVizService.get_selected_annotation(study, cluster:, **annotation)
           end.compact
         end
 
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/visualization/annotations_controller.rb
```diff
@@ -100,6 +100,10 @@ def index
 
         def show
           annotation = self.class.get_selected_annotation(@study, params)
+          if annotation.nil?
+            head :not_found and return
+          end
```
eweitz, line 105: Kudos on the key insight to prevent server resource flooding by failing fast here, and equivalent areas.

eweitz (approved):
Code looks good!  This nicely refines #2168.  

For our future reference, more context is [in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1731418200513639).

jlchang (changes requested):
Works as described and I think a small tweak might be useful.

If I test as described and use URL
https://localhost:3000/single_cell/study/SCP94/pairwise-test-1?cluster=UMAP&spatialGroups=--&annotation=UPDATEXML--group--study&subsample=all#study-visualize

 I see the expected `Bad request` message:
Screenshot 2024-11-13 at 4 11 36 PM

BUT if I use URL https://localhost:3000/single_cell/study/SCP94/pairwise-test-1?cluster=UMAP&spatialGroups=--&annotation=UPDATEXML&subsample=all#study-visualize
Notice that URL has bogus <annotation> instead of <annotation>--group--study 

Then I see the pre-PR result, a warning emitted by [ScatterPlot.jsx](https://github.com/broadinstitute/single_cell_portal_core/blob/59ad7bf7dbe6af7bd65570dd754a8fa283e81628/app/javascript/components/visualization/ScatterPlot.jsx#L607) : 
Screenshot 2024-11-13 at 4 10 47 PM
Since this reflects pre-PR behavior, I'm guessing this URL change bypasses the short circuit we want in place.

Because it seems much more likely that bogus URLs will replace the value for `annotation` wholesale (and not preserve the \<annotation\>--group--study syntax), I'm requesting a tweak to ensure a shortciruit occurs in the wholesale replacement of  `annotation` case.

jlchang (approved):
Thanks for making non-existent {name} requests shortcircuit, @bistline.

Everything looks good to me!

bistline:
Test failure is known wonkiness in `AbTest` distribution calculation

PR #2170, opened 2024-11-13
Allow setting of expression axis label in AnnData studies (SCP-5856)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where studies using the AnnData upload UX do not propagate the "expression axis label" field to any plots in the Explore tab.  The value was being saved but not used - now it is wired in the same as in the "Classic" upload UX.  Additionally, a migration has been added to backfill any existing data.

#### MANUAL TESTING
1. Initialize your environment and run the migration to set the axis label on any existing studies
2. In `development.log`, you will see a message if any studies were updated:
```
Setting expression axis label to log(TPM) expression in SCP150
```
3. Go to the Explore tab for any matching studies and validate that the axis is now labeled
4. If you do not have any matching studies, go to the upload wizard for any AnnData study and specify a label in the expression form, then repeat step 3

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -392,9 +392,14 @@ module Api
 
         if ['Expression Matrix', 'MM Coordinate Matrix'].include?(study_file.file_type) && !safe_file_params[:y_axis_label].blank?
           # if user is supplying an expression axis label, update default options hash
-          options = study.default_options.dup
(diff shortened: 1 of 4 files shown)


bistline:
Build failure seems to be related to an upstream error in dev Bard.  Requests are timing out and it's causing weird behavior in places.

jlchang (approved):

So glad to see AnnData obtain parity on `expression axis label`!

eweitz (approved):

PR #2171, opened 2024-11-13
Bypass CSFV for reference AnnData files (SCP-5824)

bistline:
#### BACKGROUND & CHANGES
Recent updates to CSFV for AnnData files have greatly improved the user experience for initializing studies with a single AnnData file.  However, an unintended consequence of this was that reference AnnData files (i.e. not parsed for visualization) were now subject to validation.  While not in and of itself a bad thing, it is misleading in that the files will not be parsed and as such would not be validated server-side.

Now, any reference AnnData files can be uploaded without invoking CSFV, and will only be opened and confirm that data can be read via existing ingest pipeline validation.  Visualizable AnnData files are still validated as normal.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Go to an empty/new study and select the classic upload UX
3. Go down to the AnnData tab and upload `test/test_data/anndata/invalid_disease_label.h5ad`
4. Confirm you get not validation error
5. Click "use bucket path" and upload that file to the bucket
6. Use the bucket path feature back in the upload wizard and confirm there is still no error
7. Switch to the AnnData UX, and confirm the file fails CSFV with an appropriate error message

### DIFF
--- a/app/javascript/components/upload/FileUploadControl.jsx
+++ b/app/javascript/components/upload/FileUploadControl.jsx
@@ -72,7 +72,7 @@ export default function FileUploadControl({
     }
 
     setFileValidation({ validating: true, issues: {}, fileName: selectedFile.name })
-    const [issues, notes] = await ValidateFile.validateLocalFile(selectedFile, file, allFiles, allowedFileExts)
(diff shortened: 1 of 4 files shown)


eweitz (approved):
Code looks good!  Relaxing validation here is a useful UX refinement, indeed.

PR #2104, opened 2024-08-13
Pathway diagrams via related genes ideogram (SCP-5751)

eweitz:
This lets users see how their searched gene interacts in cell circuits, and overlays expression data on pathway diagrams.

https://github.com/user-attachments/assets/2e905a25-512d-4e48-873f-c33afcb9f4cc

### Impact
Previously, users got a brief text summary of how their searched gene interacts with adjacent genes in a biochemical pathway.  For example, "NR3C1 acts on CSN2" in a pathway like "Pregnancy and lactation".  This was useful -- some 5-8K single-gene searches per month (> 10%) derive from related genes ideogram -- but limited.  Genes work in directed graphs; the text only summarizes one adjacent gene among many connected genes that ultimately cause a biological condition.

Now, the pathway diagram is shown right in the embedding page, and it's interactive.  The searched gene and interacting gene are highlighted in the pathway while expression data loads.  The overlaid expression data comes from a dot plot.  Expression for any one annotation group can be depicted.  Like dot plots, genes with high mean scaled expression are colored red, medium purple, and low blue.  Percent of cells expression is conveyed via opacity: lower percent, lower opacity.

This gives deep, specific functional insight for transcriptomic data.

### Technical notes
The pathway diagrams are enabled by an upstream update in Ideogram.js.  Those retrieve diagram SVG from the WikiPathways GitHub repo by way of Pvjs, a pathway viewer JS component.  Pvjs also enables panning, zooming, and some diagram DOM API accessibility.

The expression overlay is enabled by SCP API endpoints for dotplots.  The default ideogram pathway diagram header is locally extended here to include an annotation label menu, and an info icon that briefly explains gene coloring and opacity in the pathway expression overlay.

A feature flag lets us disable the expression overlay in the pathway diagram, if we want.  Usage is already softly gated because pathway diagrams are only available when fairly deep into exploration: in a study, search 1 gene, hover over a purple gene marker, click a pathway.  This lets especially engaged users discover the new pathway expression functionality.  So keeping the functionality available by default seems to make sense.

### Known issues, limitations, and next steps
1. Around 2% of pathways do not render diagrams when clicked in the ideogram gene popover.
2. The framing UX seems viable, but a fuller integration into the many-gene display (e.g. 3+, with dot plots as default enabled tab) would be better.
[...]

PR #2172, opened 2024-11-14
Scalable expression overlays for pathway diagrams (SCP-5710)

eweitz:
This enables pathway visualizations for arbitrarily many genes.  That means more causal insight for scientists.

### Background
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/2104), pathway diagram expression overlays were novel, useful, and limited.  They give researchers dataset-specific expression summaries for genes in a biochemical pathway, e.g. a gene regulatory network for human lactation, an immune signaling pathway, etc.

### Problem
Users could only view expression for up to 50 genes in a pathway.  That derived from a simple way to prevent overloading the SCP app server.  However, roughly 30% of pathways have > 50 genes, and 15% of pathways have over 100 genes.  So a basic implementation of the 50-gene maximum really limited the causal mechanistic insight available to researchers.

### Solution
Now, expression data is fetched for _all_ genes in a pathway in _serial batches_ of up to 50 genes at a time.  For example, a pathway with 300 unique genes will get expression metrics loaded 50 genes at a time, one batch starting only after the previous batch completes, until all genes in the pathway are colored in.  

This greatly mitigates the risk of exhausting server CPU, RAM, and threads.  It does that by splitting any oversized queries into multiple smaller queries, and stream-processing them one at a time.

The updates here also pull in upstream improvements in Ideogram.js, to better account for pathway genes that interact as a member of a _gene group_ (i.e., a gene family or complex), and to use updated pathway data.

### Video
Here's how this robustification looks through an engineering lens, and how it yields richer results for big pathways.

https://github.com/user-attachments/assets/eb871da1-c89b-4085-9794-d194140e5a8c

### Next steps
* Apply this streaming-batch-query robustness technique to dot plots / heatmaps, to eliminate _from a user's perspective_ the max limit of number of genes that can be searched, while essentially preserving this limit _from the server's perspective_.
* Diagram nodes indicate they're clickable due to Pvjs, but clicking them does nothing. Ideally the cursor would be changed, or clicking would e.g. show expression metrics, gene information, etc.
* The framing UX seems viable, but a fuller integration into the many-gene display (e.g. 3+, with dot plots as default enabled tab) would be better.

### Test
A new automated test confirms batching behavior.  To manually verify:
1.  `yarn install --force`
2.  Start server without Vite, or clear your frontend service worker cache via Chrome DevTools
3. Go to "Cellular and transcriptional diversity over the course of human lactation"
4. Search CSN2
5. Open DevTools, go to Network panel
6. In related genes ideogram, hover over GSK3B
7. Click "IL8 signaling"
8. In Network panel, confirm several requests to Morpheus get sent, one after another, never overlapping in time.  See 0:07 - 0:20 in video above for reference.  Your local requests will take longer at least the first time, as they aren't even cached in Rails.  Rails-uncached requests can take > 20 seconds each in a local SCP instance.  If you see _no_ Morpheus requests in the Network panel, ensure your Vite server is not running or that you've deleted any frontend service worker cache entries for "morpheus" in the "scp-..." entry in "Cache storage" in the Application panel of Chrome DevTools.
9. Close DevTools
10. Confirm the vast majority of genes in the IL18 signaling pathway diagram have at least some color filled in.  Compare to video for reference of expected colors.

This satisfies SCP-5710, a Developer's Choice Days project.


bistline (approved):
Looks good!  I think this refinement makes a lot of sense.

jlchang (approved):

Looking forward to seeing this new robustness improvement in production!

PR #2175, opened 2024-11-26
Allow updates in AnnData expression form, extract raw counts if specified (SCP-5871)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug where an AnnData file that was previously uploaded without raw counts information cannot be updated using the Expression form in the upload wizard.  It was attempting to enforce the `raw_counts_required_frontend` feature flag when AnnData files should be exempt from this.  Now, updates may be applied regardless of the raw counts state.

Additionally, users may now go back to existing AnnData files and indicate whether or not they have raw counts information.  Upon doing so, this will launch an extraction of _only_ the raw cell names and no other downstream extractions.  By doing so, AnnData files that were previously exempted will now be included in automated differential expression calculations.

#### MANUAL TESTING
**IMPORTANT**
The easiest way to test is to use a previously ingested AnnData file that has not had raw cell names extracted.  If you do not have an ingested AnnData file, you will need to upload one, but do **not** specify that it has raw counts information until the 3 primary data types have been ingested.  If you have an AnnData file that has extracted raw cells, you can remove those arrays with the following queries:
```
study = Study.find_by(accession: '<accession of study>')
study_file = study.study_files.last
filename = "h5ad_frag.matrix.raw.mtx.gz"
query = {
  name: "#{filename} Cells", cluster_name: filename, array_type: 'cells', linear_data_type: 'Study',
  linear_data_id: study.id, study_file_id: study_file.id, cluster_group_id: nil, subsample_annotation: nil,
  subsample_threshold: nil
}
arrays = DataArray.where(query)
arrays.count => you should get a number other than 0 here
arrays.delete_all    
```
You can now proceed with the test. 

1. In a Rails console session, load you study & file and confirm there are no raw counts cells
```
study = Study.find_by(accession: '<accession of study>')
study_file = study.study_files.last
study.expression_matrix_cells(study_file, matrix_type: 'raw')
=> []
```
2. In a separate terminal, boot all services including `Delayed::Job` and sign in
3. Go to the upload wizard for your AnnData study
4. In the Expression tab, indicate that your file has raw counts (if the value is false), make another innocuous change (like adding an expression label) and click "Save" in the expression tab
5. Confirm the save succeeded, and in `development.log` you see an extraction job is launch for only raw counts:
```
Request object sent to Google Life Sciences API, excluding 'environment' parameters:
---
:actions:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 66bb673094ec8fd54fc5c77b
  - "--study-file-id"
  - 66d08390b886513501dd7a26
  - "--user-metrics-uuid"
  - c6a08597-c735-4c93-b069-4b4a74b708ee
  - ingest_anndata
  - "--ingest-anndata"
  - "--anndata-file"
  - gs://fc-b2dac18a-a985-40f6-9d7d-7e611e3353ed/compliant_pbmc3K.h5ad
  - "--extract"
  - '["raw_counts"]'
  :image_uri: gcr.io/broad-singlecellportal-staging/scp-ingest-pipeline:1.37.0
  :labels: {}
  :timeout: {}
...
```
6. Once you see that the ingest has succeeded, reload your study back in the Rails console and confirm you see cells:
```
study.reload
study.expression_matrix_cells(study_file, matrix_type: 'raw')
=>
=> 
["AAACATACAACCAC-1",                                                                                          
 "AAACATTGAGCTAC-1",  
...
```
7. You should also see differential expression jobs being launched in `development.log`, depending on the availability of clustering/annotation data:
```
SCP157 is eligible for differential expression, launching available jobs
SCP157 has annotations eligible for DE; validating inputs
Checking DE job for SCP157: UMAP (louvain--group--study)
==> DE job for SCP157: UMAP (louvain--group--study) successfully launched
Checking DE job for SCP157: tSNE (louvain--group--study)
==> DE job for SCP157: tSNE (louvain--group--study) successfully launched
SCP157 yielded 2 differential expression jobs; 0 skipped
```

### DIFF
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -112,8 +112,8 @@ class FileParseService
             )
           elsif study_file.needs_raw_counts_extraction?
             params_object = AnnDataIngestParameters.new(
-              anndata_file: study_file.gs_url, extract: %w[raw_counts], extract_raw_counts: true,
-              obsm_keys: nil, file_size: study_file.upload_file_size
+              anndata_file: study_file.gs_url, extract: %w[raw_counts], obsm_keys: nil,
+              file_size: study_file.upload_file_size
             )
           else
             params_object = AnnDataIngestParameters.new(
--- a/app/lib/file_parse_service.rb
+++ b/app/lib/file_parse_service.rb
@@ -110,6 +110,11 @@ class FileParseService
               anndata_file: study_file.gs_url, extract: %w[cluster], obsm_keys: [obsm_key],
               file_size: study_file.upload_file_size
             )
+          elsif study_file.needs_raw_counts_extraction?
+            params_object = AnnDataIngestParameters.new(
+              anndata_file: study_file.gs_url, extract: %w[raw_counts], extract_raw_counts: true,
+              obsm_keys: nil, file_size: study_file.upload_file_size
+            )
           else
             params_object = AnnDataIngestParameters.new(
               anndata_file: study_file.gs_url, obsm_keys: study_file.ann_data_file_info.obsm_key_names,
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -841,6 +841,7 @@ class IngestJob
       # if this was only a raw counts extraction, update parse status
       if params_object.extract == %w[raw_counts]
         study_file.update(parse_status: 'parsed')
+        launch_differential_expression_jobs
       else
         # unset anndata_summary flag to allow reporting summary later unless this is only a raw counts extraction
         study_file.unset_anndata_summary!
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -838,8 +838,13 @@ class IngestJob
         job.delay.push_remote_and_launch_ingest
       end
 
-      # unset anndata_summary flag to allow reporting summary later
-      study_file.unset_anndata_summary!
+      # if this was only a raw counts extraction, update parse status
+      if params_object.extract == %w[raw_counts]
+        study_file.update(parse_status: 'parsed')
+      else
+        # unset anndata_summary flag to allow reporting summary later unless this is only a raw counts extraction
+        study_file.unset_anndata_summary!
+      end
     end
   end
 
@@ -1017,7 +1022,7 @@ class IngestJob
     mixpanel_log_props = get_job_analytics
     # log job properties to Mixpanel
     MetricsService.log(mixpanel_event_name, mixpanel_log_props, user)
-    report_anndata_summary if study_file.is_viz_anndata? && action != :ingest_anndata
+    report_anndata_summary if study_file.is_viz_anndata? && !%i[ingest_anndata differential_expression].include?(action)
   end
 
   # set a mixpanel event name based on action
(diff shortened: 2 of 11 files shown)


app/lib/file_parse_service.rb
```diff
@@ -110,6 +110,11 @@ def self.run_parse_job(study_file, study, user, reparse: false, persist_on_fail:
               anndata_file: study_file.gs_url, extract: %w[cluster], obsm_keys: [obsm_key],
               file_size: study_file.upload_file_size
             )
+          elsif study_file.needs_raw_counts_extraction?
+            params_object = AnnDataIngestParameters.new(
+              anndata_file: study_file.gs_url, extract: %w[raw_counts], extract_raw_counts: true,
```
eweitz, line 115: It seems like it ought to be redundant to specify both `extract: %w[raw_counts]` and `extract_raw_counts: true`.  

Reviewing the code I see why it's currently necessary as a practical matter.  But if add equivalent calls elsewhere or refine AnnDataIngestParameters in the future, then setting `extract_raw_counts: true` if `@extract.include` etc. -- somewhat like [`append_raw_counts_extract`](https://github.com/broadinstitute/single_cell_portal_core/blob/cbefd8b5d56df76839728053fb6ce57e6767725f/app/models/ann_data_ingest_parameters.rb#L82C7-L82C32) -- might be cleaner.

app/models/ingest_job.rb
```diff
@@ -838,8 +838,14 @@ def launch_anndata_subparse_jobs
         job.delay.push_remote_and_launch_ingest
       end
 
-      # unset anndata_summary flag to allow reporting summary later
-      study_file.unset_anndata_summary!
+      # if this was only a raw counts extraction, update parse status
+      if params_object.extract == %w[raw_counts]
+        study_file.update(parse_status: 'parsed')
```
eweitz, line 843: If a user uploads a big AnnData file, then _while that large ingest is ongoing_ indicates in the UI that the file has raw counts, will this line falsely suggest their original larger file is parsed when it hasn't yet finished?

That seems like a plausible scenario.  I'm less confident about severity, but I suspect it'd at least be confusing while the larger ingest is ongoing.  In any case, impact seems low at worst, and think the current change is net benefit.

So I don't consider this a blocker, just worth keeping in mind.

eweitz (approved):
Code looks good!

I suggest a potential future maintainability refinement, and speculate about a potential (but low-impact) lurking bug.

app/lib/file_parse_service.rb
bistline: Actually on reflection you're correct - it isn't necessary.  As written it only requires `extract: %w[raw_counts]` and the command line will be correctly formatted.

app/models/ingest_job.rb
bistline: That's a decent point... the save buttons will be disabled for the most part, but the one in the AnnData tab will still work.  In theory they could use that to save.  The problem is that it won't actually do the raw counts extraction because it will first check to see if it's already parsed, and since this will be false it will exit silently while still saving the update.  I coded it this way because I didn't want the opportunity of concurrent extractions potentially failing and causing downstream issues.  If we don't check the `parse_status` flag, then any update while the extraction is going would spawn a new ingest process, and then they'd all fail.  I'm loathe to add more status flags - we're already replete with them, and they're not terribly reliable, especially in an asynchronous background process. 
 Given that the save button in the expression form is disabled until the main extractions are done, I don't think this is a large issue, but we can keep an eye on it to see if it becomes a problem.

PR #2181, opened 2024-12-11
Allow specifying order of expression scatter plot sorting (SCP-5885)

bistline:
#### BACKGROUND
By default, SCP sorts all expression-based scatter plots by the observed expression for each cell from lowest to highest and then plots in that order.  This has the visible effect of bringing the cells with the highest expression to the front, as earlier cells are plotted underneath those later in the array.  While this does surface significant expression values for users, it is biased towards highly expressed cells.  This may not be what a user is intending or expecting to find, and can be confusing.  Furthermore, this feature is not documented anywhere and is completely opaque to users.

#### CHANGES
Now, all expression-based scatter plots will render a control labeled `Order expression by` with three values: `high`, `low`, and `unsorted`.  The default is `high` which behaves exactly as plots do now.  Selecting `low` will invert this behavior with a special caveat - any cells with no observed expression (i.e. a value of `0.0`) will still render in the background.  This will surface cells with low observed expression which are normally hidden by highly expressed cells.  Finally, selecting `unsorted` will turn off this behavior and plot cells in the order in which they appear in the clustering file.  This last option gives study owners the ability to control rendering order explicitly for expression scatter plots.

Also, there is now a study default option that maps to this control.  It can be configured in either the Study settings > View options tab or in the Study details page.  More information about how SCP sorts scatter plot data has also been added to a new guide page: https://singlecell.zendesk.com/hc/en-us/articles/31772258040475-How-SCP-renders-data-in-scatter-plots (you must log in as scp-zendesk to view as it is unpublished).

#### MANUAL TESTING
1. Boot as normal and sign in
2. Load the `Human milk - differential expression` study
3. Query for `CLDN4` and confirm the default rendering appears as it normally does with highly expressed cells in the `LC1` region
Screenshot 2024-12-11 at 2 51 54 PM
4. In the options panel, change the sorting to `low` and confirm the plot updates accordingly:
Screenshot 2024-12-11 at 2 53 38 PM
5. Repeat with the `unsorted` option and confirm the display updates:
Screenshot 2024-12-11 at 2 53 50 PM
6. Hover over the info icon next to the select menu and confirm you can click the link to the documentation page
7. Go into the Settings > View options panel and select `low` as the default sorting and save
8. Go back to the explore tab, click "Reset view" (**this is important!**) and then re-query for `CLDN4` and confirm the sorting is set to `low` without the URL parameter being specified

### DIFF
--- a/app/controllers/api/v1/visualization/explore_controller.rb
+++ b/app/controllers/api/v1/visualization/explore_controller.rb
@@ -158,6 +158,7 @@ module Api
             differentialExpression: AnnotationVizService.differential_expression_menu_opts(study),
             hasImageCache: study.cluster_groups.where(has_image_cache: true).pluck(:name),
             clusterPointAlpha: study.default_cluster_point_alpha,
+            expressionSort: study.default_expression_sort,
             colorProfile: study.default_color_profile,
             bucketId: study.bucket_id,
             isPublic: study.public
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -728,7 +728,7 @@ class SiteController < ApplicationController
                                   :default_options => [:cluster, :annotation, :color_profile, :expression_label,
                                                        :deliver_emails, :cluster_point_size, :cluster_point_alpha,
                                                        :cluster_point_border, :precomputed_heatmap_label,
-                                                       override_viz_limit_annotations: []],
+                                                       :expression_sort, override_viz_limit_annotations: []],
                                   study_shares_attributes: [:id, :_destroy, :email, :permission],
                                   study_detail_attributes: [:id, :full_description],
                                   reviewer_access_attributes: [:id, :expires_at],
--- a/app/controllers/studies_controller.rb
+++ b/app/controllers/studies_controller.rb
@@ -1041,7 +1041,8 @@ class StudiesController < ApplicationController
   def default_options_params
     params.require(:study_default_options).permit(:cluster, :annotation, :color_profile, :expression_label,
                                                   :cluster_point_size, :cluster_point_alpha, :cluster_point_border,
-                                                  :precomputed_heatmap_label, override_viz_limit_annotations: [])
+                                                  :precomputed_heatmap_label, :expression_sort,
+                                                  override_viz_limit_annotations: [])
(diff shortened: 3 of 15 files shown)


bistline:
Build failure is due to an upstream `503` when trying to fetch the Allen Brain Map data for the `organ_region` ontology.
```
Executing query: SELECT DISTINCT cell_type AS id, cell_type__ontology_label AS name FROM alexandria_convention WHERE cell_type IS NOT NULL ORDER BY LOWER(cell_type__ontology_label)
Unable to fetch JSON from https://api.brain-map.org/api/v2/structure_graph_download/1.json: RestClient::ServiceUnavailable: 503 Service Unavailable
```

Making a `GET` on the file works, so this appears to just be a blip.  If this persists we should look more closely at why it fails.

jlchang (approved):
Manual review - works as described.

Suggestion: in `View Options` I'd suggest the name "Default expression plot order" instead of "Expression ordering" (or is "Default expression plotting order" more grammatic, I'm honestly not sure)

eweitz (approved):
Code looks good!  I suggest a non-blocking maintainability refinement.

The draft [documentation](https://singlecell.zendesk.com/hc/en-us/articles/31772258040475-How-SCP-renders-data-in-scatter-plots) also generally looks good.  

Some potential documentation improvements, all non-blocking:

* Add the screenshots in the PR description to the documentation.  Let's also link to the highlighted study view ([CLDN4 in SCP1671](https://singlecell.broadinstitute.org/single_cell/study/SCP1671/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation?genes=CLDN4#study-visualize)) from there.

* Explain in more biological detail how this helps.  The "why" is briefly covered  -- "to address issues surrounding over-plotting" -- but I suspect we can safely offer some additional context like: 

> **Low**
<Current content> Sometimes, you want to see cells that are expressing a gene _least_, while still expressing it.  The `Low` sorting option helps distinguish such cell populations that cluster with highly-expressing cells.  <Current content> 

> **Unsorted**
This can help you show your data in SCP as it is represented in your publication or other figures.

(That content or something contextual like it would be added to the current prose, not replace it.)

PR #2196, opened 2025-01-29
Reduce redundant automated DE jobs (SCP-5905)

bistline:
#### BACKGROUND
Currently, any study eligible for differential expression will automatically launch jobs upon successful ingest of clustering, metadata, or any expression file (raw or processed).  There are checks in place to skip any results that already exist, however if those results are currently processing, this will return false.  This can lead to duplicate jobs being launched after every qualifying ingest until those result files are created.  This doesn't have any negative downstream effect other than needless cloud spend on ingest jobs and duplicate notifications - the result files will be identical, so no data is clobbered by the duplicates - but there could be dozens of redundant jobs launched depending on how many eligible clusters/annotations are present in a study.  AnnData studies are particularly prone to this as their primary ingest jobs are launched within seconds of each other.

#### CHANGES
Now, the `DifferentialExpressionService` will check to see if there are any matching jobs currently running in the Batch API before submitting a new job.  If a job is found, the service will log a message like the following and then exit silently:

```
Checking DE job for SCP167: umap (cell_type__ontology_label--group--study)
Found 1 running DE jobs: projects/broad-singlecellportal-bistlin/locations/us-central1/jobs/job-3eef7a9a-f447-4086-a0af-ab8166d9ee44
Matching these parameters: python ingest_pipeline.py --study-id 67914d3c94ec8f858cdf09ba --study-file-id 679a4c68aafb37e8782f4cb8 --user-metrics-uuid c6a08597-c735-4c93-b069-4b4a74b708ee differential_expression --study-accession SCP167 --annotation-name cell_type__ontology_label --annotation-type group --annotation-scope study --annotation-file gs://fc-4b11a7be-c5d0-40da-8d42-53f58a5a1294/_scp_internal/anndata_ingest/SCP167_679a4c68aafb37e8782f4cb8/h5ad_frag.metadata.tsv.gz --cluster-file gs://fc-4b11a7be-c5d0-40da-8d42-53f58a5a1294/_scp_internal/anndata_ingest/SCP167_679a4c68aafb37e8782f4cb8/h5ad_frag.cluster.X_umap.tsv.gz --cluster-name umap --de-type rest --matrix-file-path gs://fc-4b11a7be-c5d0-40da-8d42-53f58a5a1294/HTAPP-330-SMP-1082_compliant.h5ad --matrix-file-type h5ad --differential-expression
Exiting without queuing new job
```

New jobs using the referenced parameters will be blocked for submission unless the previous run fails or an admin manually launches a new job from the Rails console.

It should be noted that this isn't perfect - it is still subject to a race condition as new jobs do not appear instantly in the Batch API jobs list after submission.  If two qualifying files in the same study finish ingesting within seconds of each other, the upstream check will likely not see the newly created jobs and then submit duplicates.  The window for this is small - usually only about 10-15 seconds - so this drastically reduced the incidence of this issue.

Note: you may have issues testing this manually until [OLS addresses an issue](https://github.com/EBISPOT/ols4/issues/834#issuecomment-2621685051) with some ontology queries.  If you are seeing metadata validation fail, you can comment out the following lines in in `BatchApiClient#format_command_line` (in `app/models/batch_api_client.rb`, staring on line 520):

```
if study_file.use_metadata_convention
  command_line += [
    '--validate-convention', '--bq-dataset', CellMetadatum::BIGQUERY_DATASET,
    '--bq-table', CellMetadatum::BIGQUERY_TABLE
  ]
end
``` 

#### MANUAL TESTING
2. Create a new study, and upload a valid AnnData file (like [compliant_liver.h5ad](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/tests/data/anndata/compliant_liver.h5ad)), ensuring that you specify you have raw counts
3. Once the cluster/metadata/expression jobs finish, look in `development.log` for a message like the above one.  Note that you potentially could get one set of duplicates, but at least the final ingest should detect the upstream jobs and block submission

### DIFF
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -164,7 +164,9 @@ class DifferentialExpressionService
     return true if dry_run # exit before submission if specified as annotation was already validated
 
     # check if there's already a job running using these parameters and exit if so
-    job_params = ['--study-file-id', study_file.id.to_s] + params_object.to_options_array
+    job_params = ApplicationController.batch_api_client.format_command_line(
+      study_file:, action: :differential_expression, user_metrics_uuid: user.metrics_uuid, params_object:
+    )
     running = ApplicationController.batch_api_client.find_matching_jobs(
       params: job_params, job_states: BatchApiClient::RUNNING_STATES
     )
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -169,9 +169,10 @@ class DifferentialExpressionService
       params: job_params, job_states: BatchApiClient::RUNNING_STATES
     )
     if running.any?
-      log_message "Found #{running.count} running DE jobs using these params: #{running.map(&:name).join(', ')}"
-      log_message "Params: #{job_params.join(' ')}"
+      log_message "Found #{running.count} running DE jobs: #{running.map(&:name).join(', ')}"
+      log_message "Matching these parameters: #{job_params.join(' ')}"
       log_message "Exiting without queuing new job"
+      false
     elsif params_object.valid?
       # launch DE job
       job = IngestJob.new(study:, study_file:, user:, action: :differential_expression, params_object:)
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -163,7 +163,16 @@ class DifferentialExpressionService
     params_object.machine_type = machine_type if machine_type.present? # override :machine_type if specified
     return true if dry_run # exit before submission if specified as annotation was already validated
 
-    if params_object.valid?
+    # check if there's already a job running using these parameters and exit if so
+    job_params = ['--study-file-id', study_file.id.to_s] + params_object.to_options_array
+    running = ApplicationController.batch_api_client.find_matching_jobs(
+      params: job_params, job_states: BatchApiClient::RUNNING_STATES
+    )
+    if running.any?
+      log_message "Found #{running.count} running DE jobs using these params: #{running.map(&:name).join(', ')}"
+      log_message "Params: #{job_params.join(' ')}"
+      log_message "Exiting without queuing new job"
+    elsif params_object.valid?
       # launch DE job
       job = IngestJob.new(study:, study_file:, user:, action: :differential_expression, params_object:)
       job.delay.push_remote_and_launch_ingest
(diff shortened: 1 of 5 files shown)


eweitz (commented):
> Now, the DifferentialExpressionService will check to see if there are any matching jobs currently running in the Batch API before submitting a new job. If a job is found, the service will log a message like the following and then exit silently

This seems like it could preserve inaccurate DE results, and thus yield scientific bugs.

If DE is computing for old data, and new GEX data or cell type annotations (with the same file names as old data) is added to the study before the old DE job completes, then this seems like it would skip computing DE on the new data.  

That would keep DE results for old data.  So it seems the SCP DE UI would show the old DE data alongside new -- and perhaps substantially different -- underlying GEX or annotation data.  Those DE results would be incorrect.

Potential options that would avoid such data integrity issues:
1.  Keep prior solution.  It's more expensive, but more correct than PR's current approach.  (It's still vulnerable to race conditions, but correct more often as it starts jobs for newer data later.)
2.  [Cancel](https://cloud.google.com/batch/docs/cancel-job) matching DE jobs and rollback any DB writes and bucket files from ongoing DE job for old data, then launch DE job for new data.  So only 1 job would run at a time for a given match (i.e., CLI signature).  This is less expensive than Option 1, and most correct.
3.  Option 2, but augment `find_matching_jobs` to account for MD5 hash or CRC32C hash in the [object metadata](https://cloud.google.com/storage/docs/viewing-editing-metadata#view) of underlying files.  If find_matching_jobs and hash match are true, then skip like current PR approach.  If false, then apply Option 2.  Less expensive than Option 2, tied for most correct.

eweitz (approved):
I overlooked that the CLI signature includes unique file IDs, in addition to files names.  So my prior concern doesn't apply.  Thanks to Jean for noting in standup.  Code looks good!

bistline:
> I overlooked that the CLI signature includes unique file IDs, in addition to files names. So my prior concern doesn't apply. Thanks to Jean for noting in standup. Code looks good!

Yes, that should address the concerns you outlined.  Even if a user replaces data with files of the exact same name, the file IDs would be different.  In that scenario it would launch new jobs that would replace the ones still being processed.  There is a _small_ chance of a collision if they both complete around the same time, but this would be a massive confluence of race conditions and I don't think it's worth guarding against.  We can always manually rerun jobs if we think this might have happened.

jlchang (approved):
Functional review: my attempt didn't run into redundant jobs.

PR #2197, opened 2025-01-29
Enhance DE selection UI

eweitz:
This improves the differential expression picker to show more groups, ease user interaction, and display useful DE state.

### Overview
[Previously](https://github.com/broadinstitute/single_cell_portal_core/pull/1806), the DE picker used drop-down menus, and omitted groups that were eligible for DE but had no comparisons available.  Virtually all DE was SCP-computed and thus only had one-vs-rest comparisons available.  A small number of studies had author DE, and a subset of those had pairwise DE available.  That old DE selection UI strongly assumed users wouldn't have pairwise DE comparisons, or would add them through the upload UI view.

Now, the DE picker uses lists that show all groups at a glance.  It uses color and hover text to distinguish which comparison groups have DE results available vs. not.  This will let users see one-vs-rest DE like before, while also having an intuitive way to enable pairwise DE comparisons.  

Here's how it looks!

https://github.com/user-attachments/assets/f37711b5-4d4a-4a62-a375-0b6c3b76c252

Later UI work will integrate downstream views for DE-available and DE-unavailable.

### Test
A new automated test verifies behavior.  To manually test:
1.  Run migration: `bin/rails db:migrate`
2.  Set `default_pairwise_de_ui` feature flag to `true`
3.  Go to study "Cellular and transcriptional diversity over the course of human lactation"
4.  Click "Differential expression" button
5.  Confirm two lists of groups appear, and 2nd list is disabled
6.  Pick a group in 1st list, e.g. "B cells"
7.  Confirm 2nd list becomes enabled
8.  Confirm 1st row in 2nd list shows "rest" in green
9.  Confirm all groups _except_ picked group from 1st list (e.g. "B cells") appear in yellow
10.  Pick a different group in 1st list
11.  Confirm picked group in 1st list doesn't appear in 2nd list

### More details
This satisfies SCP-5846.  Further context is available in [2025-01-28 SCP demo](https://drive.google.com/file/d/1TcxKURnfq1F2T-BDv1nmDvwEnMCjvHOa/view?t=14).


app/javascript/components/visualization/controls/DifferentialExpressionGroupPicker.jsx
```diff
@@ -7,6 +7,10 @@ import { fetchBucketFile } from '~/lib/scp-api'
 import PlotUtils from '~/lib/plot'
 const { getLegendSortedLabels } = PlotUtils
 
+
+// https://stackoverflow.com/questions/37909134/nbsp-jsx-not-working
+const blankSpace = '\u00A0' // How React JSX does &nbsp;
```
bistline, line 12: Gross, but good to know!

db/migrate/20250127092542_add_flag_for_default_pairwise_de_ui.rb
```diff
@@ -0,0 +1,11 @@
+class AddFlagForDefaultPairwiseDeUi < Mongoid::Migration
+  def self.up
+    FeatureFlag.create!(name: 'default_pairwise_de_ui',
+                        default_value: false,
+                        description: 'show pairwise differential expression UI by default')
+  end
+
+  def self.down
+    FeatureFlag.find_by(name: 'default_pairwise_de_ui')&.destroy
```
bistline, line 9: I know this is our SOP in migrations (as is evidenced by pretty much all other FF-related ones), but we ideally should be calling `FeatureFlag.retire_feature_flag('default_pairwise_de_ui')` here instead.  That also cleans up any existing flags that have been set for users/studies.  Really not a big deal, just a note for future "us".

bistline (approved):
Looks good so far!

db/migrate/20250127092542_add_flag_for_default_pairwise_de_ui.rb
eweitz: Nice refinement, done in 6ca4b6f.

eweitz:
The build failures are non-overlapping and inconsistent, and seem due to upstream issues.  So they're false positives insofar as these changes are concerned.

jlchang (approved):
Informational: the message "Select a group in other menu" is just long enough to cause a visual paper cut if my viewport is just a little more narrow than usual:
https://github.com/user-attachments/assets/bb1ab15f-06b6-48ac-a192-a302cbe96707

Some non-blocking UI message suggestions to consider:
Select a group in other menu -> Select from left column first
Rationale - "left" and "first" provide a bit more specificity

All other groups -> Compare to to combined groups
Rationale - emphasizes the selection is for comparison

Pick to enable -> Pick for pairwise comparison
Rationale - notes comparison is pairwise

Thought for future implementation: pairwise results are currently "directional" (ie, if Bcell vs Tcells is calculated, we'd only have "Bcell vs Tcells" and not "Tcell vs Bcells"). We can either 1. know in the UI to make both options available and if the "opposite direction" is chosen, flip the sign for the size metric and figure out the correct row order presentation) or indicate in the UX that they may want to try both combinations before concluding that the data has not yet been generated.

eweitz:
Thanks for the review!  I've opened a ticket to improve the note text responsiveness, and incorporated some suggestions.

PR #2198, opened 2025-01-31
Workaround for false positive Rawls healthcheck issue (SCP-5916)

jlchang:
## Issue: false positives checking on Rawls health

Starting 2025-01-31, Rawls health from https://api.firecloud.org/status was not "ok". Discussion with #dsp-core-services confirmed that the GoogleGenomics service is irrelevant now. 
Current output from `/status` endpoint:
`{"ok":false,"systems":{"Thurloe":{"ok":true},"Sam":{"ok":true},"Rawls":{"messages":["BillingProfileManager: CloudSQL: (ok: true, message: none), Sam: (ok: true, message: none)","GoogleGenomics: Timed out after 1 minute waiting for a response from GoogleGenomics"],"ok":false},"Agora":{"ok":true},"GoogleBuckets":{"ok":true},"LibraryIndex":{"ok":true},"OntologyIndex":{"ok":true}}}`

## External mitigation - Rawls to remove GoogleGenomics check

#dsp-core-services is removing that check from Rawls but recommend not using the /status endpoint to check for Rawls health but rather to look for `500` errors from actual requests. (see also [SCP-5647](https://broadworkbench.atlassian.net/browse/SCP-5647))

## Short-term mitigation for SCP - this PR
IF it takes a while for the Rawls change to take effect, we can use this PR to bypass rawls health checks until the change from  #dsp-core-services is in production. If the status endpoint has updated (ie. you can access study settings tab in production), this PR is no longer needed.

#### MANUAL TESTING
2. Visit a study where you have edit privileges and confirm the "Settings" tab is active (currently in production, the tab is greyed out and you'll see a tooltip with "Study workspaces are currently unavailable - please try again later".
3. Click on the "+ Create study" link and confirm you're directed to the study creation form.

[SCP-5647]: https://broadworkbench.atlassian.net/browse/SCP-5647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -666,7 +666,7 @@ class SiteController < ApplicationController
       if api_status.is_a?(Hash)
         system_status = api_status['systems']
         sam_ok = system_status.dig(FireCloudClient::SAM_SERVICE, 'ok') == true # do equality check in case 'ok' node isn't present
-        #  2025-01-31 quick and dirty workaround for false negative Rawls availability issue
+        #  2025-01-31 quick and dirty workaround for false positive Rawls availability issue
         #  long term mitigation exploration in SCP-5647        
         # rawls_ok = system_status.dig(FireCloudClient::RAWLS_SERVICE, 'ok') == true
         rawls_ok = true
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -666,7 +666,10 @@ class SiteController < ApplicationController
       if api_status.is_a?(Hash)
         system_status = api_status['systems']
         sam_ok = system_status.dig(FireCloudClient::SAM_SERVICE, 'ok') == true # do equality check in case 'ok' node isn't present
-        rawls_ok = system_status.dig(FireCloudClient::RAWLS_SERVICE, 'ok') == true
+        #  2025-01-31 quick and dirty workaround for false negative Rawls availability issue
+        #  long term mitigation exploration in SCP-5647        
+        # rawls_ok = system_status.dig(FireCloudClient::RAWLS_SERVICE, 'ok') == true
+        rawls_ok = true
         buckets_ok = system_status.dig(FireCloudClient::BUCKETS_SERVICE, 'ok') == true
         @allow_downloads = buckets_ok
         @allow_edits = sam_ok && rawls_ok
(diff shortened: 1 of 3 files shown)


app/controllers/site_controller.rb
```diff
@@ -666,7 +666,10 @@ def set_firecloud_permissions(study_detached)
       if api_status.is_a?(Hash)
         system_status = api_status['systems']
         sam_ok = system_status.dig(FireCloudClient::SAM_SERVICE, 'ok') == true # do equality check in case 'ok' node isn't present
-        rawls_ok = system_status.dig(FireCloudClient::RAWLS_SERVICE, 'ok') == true
+        #  2025-01-31 quick and dirty workaround for false negative Rawls availability issue
```
eweitz, line 669: The current phrasing is subtly but importantly incorrect.  For reference:
* False positive: non-problem reported as problem
* False negative: problem reported as non-problem
* True positive: problem reported as problem
* True negative: non-problem reported as non-problem

The issue is that Rawls is effectively OK but is reported as not-OK, so the status report is a false positive.

```suggestion
        #  2025-01-31 quick and dirty workaround for false positive Rawls availability issue
```

app/controllers/studies_controller.rb
```diff
@@ -1148,7 +1148,10 @@ def check_edit_permissions
 
   # check on FireCloud API status and respond accordingly
   def check_firecloud_status
-    unless ApplicationController.firecloud_client.services_available?(FireCloudClient::SAM_SERVICE, FireCloudClient::RAWLS_SERVICE)
+    #  2025-01-31 quick and dirty workaround for false negative Rawls availability issue
```
eweitz, line 1151:
```suggestion
    #  2025-01-31 quick and dirty workaround for false positive Rawls availability issue
```

eweitz (approved):
Code looks good!  My local manual tests passed.

I recommend a correction in some code comments, nothing blocking.

app/controllers/site_controller.rb
jlchang: Thanks @eweitz. I always think this through the wrong way -  ie. Rawls was reporting it was "not OK" (ie. a negative that was incorrect)and so I end up using the wrong term.  Updated in `1cbcc606f`.

app/controllers/studies_controller.rb
jlchang: Updated in `1cbcc606f`.

jlchang:
#dsp-core-services updated Rawls to stop checking GoogleGenomics which resolves our issue so we don't need to consider this workaround anymore.

PR #2207, opened 2025-02-25
Updating to scp-ingest-pipeline:1.40.1 (SCP-5902)

bistline:
#### BACKGROUND & CHANGES
This updates to `scp-ingest-pipeline:1.40.1` to allow logging to `STDOUT` from ingest pipeline.  This will enable real-time logs in the Batch API for running jobs.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -29,7 +29,7 @@ module SingleCellPortal
     config.middleware.use Rack::Brotli
 
     # Docker image for file parsing via scp-ingest-pipeline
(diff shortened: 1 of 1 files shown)


eweitz (approved):

bistline:
@eweitz I just noticed that CI didn't run on this PR... I'm going to wait before merging this to see if it eventually runs or try pushing a no-op commit to force it to retry.  I'm sure it's fine but I'd still like to ensure normal CI works.  All our other PRs ran tests so I'm puzzled why this one didn't.

PR #2211, opened 2025-02-26
Removing pre-flight Terra status checks (SCP-5647)

bistline:
#### BACKGROUND
SCP has a practice of using pre-flight status checks for Terra API calls to guard against outages/latency in the event of upstream instability.  While this usually results in a much smoother UX where errors are prevented by checking various Terra services in the `/status` information, false positives can lead to situations where users are blocked from normal operations like uploading files or editing studies because a Terra service reports a problem when in fact there is not one.  Sometimes these false outage persist for hours, leading to widespread user impact.

This update removes almost all pre-flight checks and instead allows API calls to proceed and move through normal error handling.  This will all but eliminate the prior false positive issue.  The one place where a check was left in place was when checking whether a [user has access to a study via a group share](https://broadworkbench.atlassian.net/browse/SCP-2325).  This caused study overview pages to hang in the event of upstream outages, so an explicit check was put in place to guard against this.  All other service checks have been removed.

It should be noted that this could raise the incidence of 500-level errors that result in users seeing the "We're sorry, something went wrong" page.  However, the risk is quite low, and any such errors would be reported upstream to Sentry and can be triaged appropriately.

#### MANUAL TESTING
2. Load a study with visualizations and confirm nothing is affected and all tabs are enabled
3. Download a file and confirm it succeeds
4. Make an edit in the study settings page (like to the visualization defaults) and confirm there are no issues
5. Go to the upload wizard and add a file, confirming there are no issues
6. To simulate an outage, on line 237 of `app/models/fire_cloud_client.rb`, add the following call to throw an error:
```
raise 'internal error'
response = RestClient::Request.execute(method: http_method, url: path, payload: payload, headers: headers)
...
```
7. Repeat steps 2-4 to confirm that there is no interruption to normal usage
8. Attempt to share the study with another user and confirm in `development.log` you see a similar error, but the UI does not break and you can continue to use the site:
```
2025-02-26 13:48:01 -0500: Creating FireCloud ACLs for study Parse Flag Logic - share jonbistline@gmail.com, permission: View
FireCloud API request (PATCH) https://api.firecloud.org/api/workspaces/single-cell-portal-development/parse-flag-logic/acl?inviteUsersNotFound=true with tracking identifier: 63e27b72f39faa00560c1e25
Unknown error: RuntimeError - internal error
Completed 500 Internal Server Error in 1137ms (MongoDB: 0.7ms | Allocations: 153746)


internal error excluded from capture: DSN not set
  
RuntimeError (internal error):
  
app/models/fire_cloud_client.rb:237:in `process_firecloud_request'
app/models/fire_cloud_client.rb:445:in `update_workspace_acl'
app/models/study_share.rb:213:in `set_firecloud_acl'
app/controllers/site_controller.rb:112:in `update_study_settings'
app/controllers/application_controller.rb:82:in `set_current_user'
app/controllers/concerns/real_ip_logger.rb:15:in `log_real_ip'
```
Note: the modal will not close because we did not explicitly handle this exception.  Normally, a `RestClient::Exception` would be thrown and handled, but simulating those types of errors is complicated and unnecessary for this test.

### DIFF
--- a/app/controllers/analysis_configurations_controller.rb
+++ b/app/controllers/analysis_configurations_controller.rb
@@ -3,7 +3,6 @@ class AnalysisConfigurationsController < ApplicationController
                                                     :submission_preview, :load_study_for_submission_preview,
                                                     :update_analysis_parameter]
   before_action :set_analysis_parameter, only: [:update_analysis_parameter]
-  before_action :check_firecloud_status, only: [:new, :create]
   before_action do
     authenticate_user!
     authenticate_admin
@@ -176,17 +175,5 @@ class AnalysisConfigurationsController < ApplicationController
                                                                                            :attribute_value, :association_source,
                                                                                            :association_method, :association_data_type,
                                                                                            :_destroy])
-    end
-
-    def check_firecloud_status
-      unless ApplicationController.firecloud_client.services_available?(FireCloudClient::RAWLS_SERVICE)
-        alert = 'The Methods Repository is temporarily unavailable, so we cannot complete your request.  Please try again later.'
-        respond_to do |format|
-          format.js {render js: "$('.modal').modal('hide'); alert('#{alert}')" and return}
-          format.html {redirect_to merge_default_redirect_params(studies_path, scpbr: params[:scpbr]),
-                                   alert: alert and return}
-          format.json {head 503}
-        end
-      end
     end
 end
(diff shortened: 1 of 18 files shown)


jlchang (approved):
Functional review - manual test worked as described

eweitz (approved):
Code looks good!  

Insightful descriptive of the UX and reliability trade-off here -- this is a net benefit.

PR #2213, opened 2025-03-06
Fixing query logic & order of operations for automated AnnData DE jobs (SCP-5946)

bistline:
#### BACKGROUND & CHANGES
This update addresses two corner cases when processing differential expression jobs from AnnData files:

1. The query to find the associated matrix file (i.e. the AnnData file) can fail if the `ExpressionFileInfo` document is not properly initialized.  There was only one recorded instance of this, so it is unclear if there is a related bug or this is simply user error, but it causes `DifferentialExpressionResult` objects to not properly save, even though the output files successfully generated.  Jobs will continually retry on this file every time a qualifying ingest event checks if the study is eligible for DE.  Now, instead of directly querying on that nested document, the convenience method `study_file.is_raw_counts_file?` is used as this is more robust across different file types.  Also, the filename query was updated to only reference `upload_file_name` as this value is always set, regardless of how the file was uploaded.

2. If raw counts data is extracted after the initial file upload by the user or an admin specifying via the upload wizard, automated downstream DE jobs do not properly launch because the check for eligibility happens before the file is updated to indicate it had raw counts data extracted.  Now, the order of these operations is switched and DE eligibility is checked last.

#### MANUAL TESTING
To test the fix for the file query issue:
1. Open a Rails console session and find an AnnData file that has raw counts data:
```
study_file = StudyFile.where(file_type: 'AnnData').detect(&:is_raw_counts_file?)
```
2. Make a copy of the `ExpressionFileInfo` document and then remove it from the source file
```
exp_info = study_file.expression_file_info
study_file.update(expression_file_info: nil)
=> true                                               
```
3. Confirm the file still registers as a raw counts file even without the nested document:
```
study_file.reload
study_file.expression_file_info
=> nil
study_file.is_raw_counts_file?
=> true
```
4. Restore the file to its original state:
```
study_file.update(expression_file_info_attributes: exp_info.attributes)
=> true
study_file.expression_file_info
=> #<ExpressionFileInfo _id: 67bddff294ec8f57097e0cd5, ...
```
To test the DE job launch issue:
1. Find a DE result where the source file was AnnData:
```
result = DifferentialExpressionResult.where(is_author_de: false).detect { |r| r.matrix_file && r.matrix_file.is_viz_anndata? }
```
2. Remove the extract raw count cell names and delete the DE reults:
```
study = study_file.study
query = { name: "h5ad_frag.matrix.raw.mtx.gz Cells", cluster_name: filename, array_type: 'cells', linear_data_type: 'Study', linear_data_id: study.id, study_file_id: study_file.id, cluster_group_id: nil, subsample_annotation: nil, subsample_threshold: nil }
DataArray.where(query).delete_all
=> # will return info on documents processed, likely only 1
result.destroy
=> true
```
3. Load the same study in the upload wizard, and click `Save` on the expression tab form.  This should launch a job to re-extract the raw counts info in `development.log`:
```
Request object sent to Google Batch API, excluding 'environment' parameters:
---
:task:
  :commands:
  - python
  - ingest_pipeline.py
  - "--study-id"
  - 67bddf8394ec8f57097e0cc5
  - "--study-file-id"
  - 67bddfd89b706436d767c635
  - "--user-metrics-uuid"
  - e20680f6-bcfb-4d37-b66a-be2c9c7c5ebd
  - ingest_anndata
  - "--ingest-anndata"
  - "--anndata-file"
  - gs://fc-cb8ee4bd-e427-4ef5-803d-b3a0e3a6fc8c/compliant_liver.h5ad
  - "--extract"
  - '["raw_counts"]'
```
4. Once this job completes, confirm that you see automated DE jobs launching in `development.log` (there should be at least one, but it will depend on eligible annotations/clusters in this study):
```
IngestJob poller: projects/broad-singlecellportal-bistlin/locations/us-central1/jobs/job-2502cbed-cbbb-4b3d-b2d2-47544d1e9fb2 is done!
IngestJob poller: projects/broad-singlecellportal-bistlin/locations/us-central1/jobs/job-2502cbed-cbbb-4b3d-b2d2-47544d1e9fb2 status: SUCCEEDED
SCP168 is eligible for differential expression, launching available jobs
SCP168 has annotations eligible for DE; validating inputs
Checking DE job for SCP168: umap (author_cell_type--group--study)
==> DE job for SCP168: umap (author_cell_type--group--study) successfully launched
Checking DE job for SCP168: umap (cell_type__ontology_label--group--study)
==> DE job for SCP168: umap (cell_type__ontology_label--group--study) successfully launched
SCP168 yielded 2 differential expression jobs; 0 skipped
```

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -514,9 +514,9 @@ class IngestJob
     when :image_pipeline
       set_has_image_cache
     when :ingest_anndata
+      set_anndata_file_info
       launch_anndata_subparse_jobs if study_file.is_viz_anndata?
       launch_differential_expression_jobs if study_file.is_viz_anndata?
-      set_anndata_file_info
     end
     set_study_initialized
   end
@@ -741,10 +741,10 @@ class IngestJob
     Rails.logger.info "Creating differential expression result object for annotation: #{annotation_identifier}"
     cluster = ClusterGroup.find_by(study_id: study.id, study_file_id: study_file.id, name: params_object.cluster_name)
     matrix_url = params_object.matrix_file_path
-    matrix_filename = matrix_url.split("gs://#{study.bucket_id}/").last
-    matrix_file = StudyFile.where(study: study, 'expression_file_info.is_raw_counts' => true).any_of(
-      { upload_file_name: matrix_filename }, { remote_location: matrix_filename }
-    ).first
+    bucket_path = matrix_url.split("gs://#{study.bucket_id}/").last
+    matrix_file = StudyFile.where(
+      study:, :upload_file_name.in => [bucket_path, bucket_path.split('/').last]
+    ).detect(&:is_raw_counts_file?)
     de_result = DifferentialExpressionResult.find_or_initialize_by(
       study: study, cluster_group: cluster, cluster_name: cluster.name,
       annotation_name: params_object.annotation_name, annotation_scope: params_object.annotation_scope,


eweitz (approved):

jlchang (approved):
Functional review: manual tests worked as described

PR #2222, opened 2025-03-20
Correctly identify bundled files in upload wizard (SCP-5941)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug in the upload wizard where bundled sequence index files were not correctly identified.  The parent files was instead being supplied to the index form as it was matching on the `study_file_bundle_id` first before finding the index file.  Now, an added check to ignore the file by its ID allows the index to be found.  This appears to have affect BAM files as well, but given their sparsity in SCP it wasn't identified until now due to the fragment BED file work.

#### MANUAL TEST
1. Boot as normal and load the upload wizard for any study that has BED or BAM files
2. Confirm the index file is correct for the parent sequence file

### DIFF
--- a/app/javascript/components/upload/upload-utils.js
+++ b/app/javascript/components/upload/upload-utils.js
@@ -111,7 +111,7 @@ export function findBundleChildren(file, files) {
       f.study_file_bundle_id
     ]
     return parentFields.includes(file._id) ||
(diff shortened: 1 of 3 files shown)


eweitz (approved):

jlchang (approved):
Functional review - Sequence file form now displays index file name as expected (thanks for figuring out my vite issue @bistline!)

PR #2224, opened 2025-03-24
Fix nginx logging, stringio version (SCP-5964)

bistline:
#### BACKGROUND & CHANGES
This fixes an issue in deployed environments where the version of `stringio` was out of sync with the loaded version.  Somehow version `3.1.2` was being loaded but `3.1.0` was declared in `Gemfile.lock`.  Updating to the latest version addresses this issues. 

Relatedly, Nginx changed how error logs are configured, so this moves the `error_log` directive to the root level.  Not having this in place swallowed the 500-level error, making it impossible to know that the above version conflict was happening.  Manual tests of re-building the image on staging proved that both these changes had the desired effect.  Also, the declared version of Ruby for passenger was incorrect and has been updated to use the new default 3.4 installation.

#### MANUAL TESTING
N/A - this only affects deployed environments, so the validation will be that once this merges staging can re-deploy without error.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -38,6 +38,7 @@ env ORCH_SMOKE_TEST;
 env TZ=America/New_York;
 daemon off;
 
+error_log /var/log/nginx/error.log;
 include /etc/nginx/main.d/*.conf;
 
 events {
@@ -70,7 +71,6 @@ http {
 
 	log_format custom '$proxy_add_x_forwarded_for - $remote_user [$time_local] "$request" $status $body_bytes_sent "$http_referer" "$http_user_agent"';
 	access_log /var/log/nginx/access.log custom;
-	error_log /var/log/nginx/error.log;
 
 	##
 	# Gzip Settings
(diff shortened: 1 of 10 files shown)


nginx.conf
```diff
@@ -115,6 +115,7 @@ http {
 	passenger_pool_idle_time 300;
     passenger_max_pool_size 10;
     passenger_min_instances 1;
+    passenger_log_level 3;
```
eweitz, line 118: Even though `passenger_log_level 3` is [default](https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_log_level), having this configuration explicit seems helpful.  I wouldn't know off-hand about this option, and I can see it be useful to adjust when diagnosing an issue.

eweitz (approved):
Code looks good.  Glad staging and logging are back!  More context is [in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1742502764466159).

nginx.conf
bistline: For reference, I got this from here: https://stackoverflow.com/questions/59167793/passenger-not-logging-ruby-stdout-stderr

This example had set it explicitly to `2`, and I figured declaring it as `3` would help insulate us from other upstream changes.

PR #2228, opened 2025-03-26
Make updating existing DE result more robust (SCP-5952)

bistline:
#### BACKGROUND & CHANGES
This update makes the process of updating existing `DifferentialExpressionResult` objects more robust.  Previously, there were many corners where finding the associate matrix file or clustering object didn't work properly, which would cause results creation to fail validation because associations weren't set, or it attempted to create a new result object when one already existed.  Now, rather than finding associations by matching names, `cluster_group_id` and `matrix_file_id` enable direct association mapping without ambiguity.  Convenience methods of `cluster_group` and `matrix_file` also allow loading the associated objects easier.  There are also new convenience methods for finding existing DE results.  Lastly, in the case of an existing result where the associated clustering object has been renamed, the cached name (i.e. `result.cluster_name`) is passed to ingest so that all output files for an existing result are consistent.  These updates together simplify the logic in `IngestJob#create_differential_expression_results`.

#### MANUAL TESTING
2. Go to any study that has existing one-vs-rest DE results
3. Open the upload wizard for this study and rename the cluster that has existing results (e.g. `Epithelial Cells UMAP` to. `Epithelial Cells`)
4. Open the [DE submission swagger form](https://localhost:3000/single_cell/api/v1#/Site/site_study_submit_differential_expression_path) and request new pairwise calculations for the existing result
5. In `development.log`, find the corresponding Batch API call and note that the original name for the cluster was passed for `cluster_name` (this makes output filenames match the existing results)
6. Once the job succeeds, confirm you can load the new pairwise results (you will need to reload the page as the `/explore` response is cached)


eweitz (approved):
Code looks good!

jlchang (approved):

Glad to see `DifferentialExpressionResult.find_existing_result`, should be quite useful!

PR #2230, opened 2025-03-27
Stop forcing isAuthorDe on pairwise results (SCP-5967)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug with where any existing result DE result that has pairwise comparisons will evaluate as having been author supplied.  This causes a UI bug where the column indices are skewed by one, which renders incorrect values for log fold change and adjusted p value.  The expression data rendered is still correct, but none of the genes shown by default are in fact differentially expressed, leading to a confusing UX.

#### MANUAL TESTING
1. Load any existing DE result that has pairwise comparisons
2. Download the result file locally and open in a text editor
3. Confirm the order matches in the file and the table, and that the values for log fold change and adjusted p value match

### DIFF
--- a/app/javascript/components/explore/DifferentialExpressionPanel.jsx
+++ b/app/javascript/components/explore/DifferentialExpressionPanel.jsx
@@ -762,6 +762,7 @@ export default function DifferentialExpressionPanel({
           countsByLabelForDe={countsByLabelForDe}
           deObjects={deObjects}
           setDeFilePath={setDeFilePath}
+          isAuthorDe={isAuthorDe}
           deGroupB={deGroupB}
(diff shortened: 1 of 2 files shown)


eweitz (commented):
Code looks good!  [SCP435 on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP435/human-milk-differential-expression-mar-2025-update) will be a good test case there.

eweitz (approved):

jlchang (approved):
Functional review - confirmed that DE output for pairwise results in portal match values for log fold change and adjusted p value match in file.

PR #2232, opened 2025-04-01
Adding `raw_location` for AnnData files (SCP-5961)

bistline:
#### BACKGROUND & CHANGES
This update adds client- and server-side integration of the `raw_location` attribute for visualization AnnData files.  This allows users to specify where they have stored count information.  Users can specify either `.raw` or the name of a layer in the `layers` section (such as `counts`).  This value is then stored inside the `AnnDataFileInfo` object and passed to ingest both for AnnData extraction and DE calculation.  The form input contains a tooltip with links to the AnnData doc pages for both [.raw](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.raw.html) and [layers](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.layers.html).

Screenshot of the new form input:
raw-location-field

Additionally, there is a migration to set this value on existing AnnData files that have raw counts.  Since we only supported `.raw` up until now, any parsed AnnData file in SCP with count data would use that slot.

#### MANUAL TESTING
1. Boot all services
2. In a separate console, run the migration and then enter a Rails console and validate that `.raw` has been set for existing AnnData files with raw counts:
```
files = StudyFile.where(file_type: 'AnnData', 'ann_data_file_info.reference_file' => false, parse_status: 'parsed').select(&:is_raw_counts_file?)
=> [#<StudyFile _id: 66b12ac32361e48b542bc7cd, ... ]
files.map {|f| f.ann_data_file_info.raw_location }
=> [".raw", ".raw", ".raw", ".raw", ".raw"]
```
3. In a browser tab, sign in and create a new study, then upload an AnnData file that contains count information, specifying the location in the expression form
4. Confirm that extraction completes without error (this validates the presence of the count data) and parsing continues all the way through DE

### DIFF
--- a/app/javascript/components/upload/ExpressionFileForm.jsx
+++ b/app/javascript/components/upload/ExpressionFileForm.jsx
@@ -8,6 +8,9 @@ import ExpandableFileForm from './ExpandableFileForm'
 
 import { TextFormField } from './form-components'
 import { findBundleChildren, validateFile } from './upload-utils'
+import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons'
+import { OverlayTrigger, Popover } from 'react-bootstrap'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 
 const REQUIRED_FIELDS = [{ label: 'species', propertyName: 'taxon_id' },
   { label: 'Biosample input type', propertyName: 'expression_file_info.biosample_input_type' },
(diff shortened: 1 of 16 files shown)


app/javascript/components/upload/ExpressionFileForm.jsx
```diff
@@ -61,9 +71,45 @@ export default function ExpressionFileForm({
     setShowRawCountsUnits(rawCountsVal)
   }
 
+  /** create the tooltip and message for the .obsm key name section */
+  function rawSlotMessage() {
+    const rawSlotToolTip = <span>
+      <OverlayTrigger
+        trigger={['hover', 'focus']}
+        rootClose placement="top"
+        delayHide={1500}
```
eweitz, line 80: Nice UX consideration.

eweitz (approved):

PR #2238, opened 2025-04-15
Release 1.93.1

eweitz:


bistline (approved):

eweitz:
The build failure is a true positive, but will have no direct user impact.  The failing `run_differential_expression_on_default` test is a regression in an a convenience method we added for ad-hoc, manual CLI calls.  It's only run by staff in special circumstances, and has a work around, and does not affect uploads or DE in any way that would degrade the UX.

PR #2242, opened 2025-04-15
Extending ontology-based CSFV to all metadata files (SCP-5814)

bistline:
#### BACKGROUND & CHANGES
This continues work extending minified ontology usage for convention validations in CSFV.  Previously, only AnnData files could perform ontology-based validation client-side.  Now, all metadata files can use minified ontologies to validate terms in the client.  Additionally, this now covers all ontology-based metadata columns as defined in version `3.0.0` of the Alexandria convention.

It is worth noting that while the ontologies are the same, and some code is reused from the AnnData validation module, the overall strategy is very different.  AnnData files can essentially be validated by column, as their file structure allows for extracting unique values from a single dataset column.  Classic flat files have to be validated line-by-line as this is the only efficient way to read them.  Errors are stored at the `parseMetadataFile` level, and as each validation failure is found, an identifier is stored to ensure that errors are only reported once.

Example of CSFV results for classic files:

https://github.com/user-attachments/assets/360107ed-df4f-44c5-b054-ddf030963cd2

#### MANUAL TESTING
1. Boot as normal and go to the upload wizard for an empty study
2. Choose "classic" mode and upload the metadata file `test_data/validation/metadata_bad_v3-0-0.tsv`
3. Ensure you get the following errors:
```
Invalid ethnicity label "invalid label". Valid labels for HANCESTRO_0005: European ancestry, Caucasian, European, white
Invalid cell_type label "not the label". Valid labels for CL_0000066: epithelial cell, epitheliocyte
```
4. Remove the file, then upload `test_data/validation/metadata_good_v3-0-0.tsv` and ensure no errors are present
5. Switch to AnnData mode, and upload the file `test_data/anndata/valid.h5ad` and confirm there are no errors
6. Remove the file, then upload `test_data/anndata/invalid_disease_label.h5ad` and confirm you see the following error:
```
Invalid disease label "foo". Valid labels for MONDO_0018076: tuberculosis, tuberculosis disease, active tuberculosis, Kochs disease, TB
```


app/javascript/lib/validation/validate-file-content.js
```diff
@@ -232,12 +237,87 @@ export async function parseMetadataFile(chunker, mimeType, fileOptions) {
...
+    const validLabels = ontology[ontologyId]
+
+    if (!(validLabels.includes(label))) {
+      errorIdentifier = `${ontologyId}-label-lookup-error`
+      // Register invalid ontology label
+      const prettyLabels = validLabels.join(', ')
+      const validLabelsClause = `Valid labels for ${ontologyId}: ${prettyLabels}`
+      const msg = `Invalid ${prop} label "${label}".  ${validLabelsClause}`
+      issue = [
+        'error', 'ontology:label-lookup-error', msg,
```
eweitz, line 308: Refining this error type seems [consistent with validation in Ingest Pipeline](https://github.com/broadinstitute/scp-ingest-pipeline/blob/8de8eca097fc3a9e37868142acb57ccb91cbe944/ingest/validation/validate_metadata.py#L1185).

```suggestion
        'error', 'ontology:label-not-match-id', msg,
```

eweitz (approved):
Code looks good overall -- I suggest a non-blocking refinement.

This will likely reduce slow validation failures by ~34% by instantly catching `ontology:label-lookup-error` and `ontology:label-not-match-id` error types.  

That's a big UX improvement!  Kudos.

PR #2243, opened 2025-04-17
Report ingestSummary even if AnnData extraction fails (SCP-5973)

bistline:
#### BACKGROUND & CHANGES
This fixes a corner case in Mixpanel reporting where AnnData files that fail during the initial extraction phase do not report an `ingestSummary` event.  These events are the main indication SCP has in Mixpanel for AnnData ingest as only one event is ever sent per AnnData file.  The lack of these extraction events means we are underreporting on AnnData ingest attempts.  Now, if the initial extraction fails and is not going to be retried due to an OOM exception, the summary is sent.

Also, this fixes a regression introduced in #2242 where metadata properties in AnnData files that have names ending in values that match ontology-based properties (e.g. `author_cell_type` matching to `cell_type`) incorrectly attempt to validate that column.  Now, exact name matches are enforced rather than relying on `endsWith()`.

#### MANUAL TESTS
1. Boot all services, sign in, and go to the upload wizard for a new/empty study and select the AnnData UX
2. Upload the [compliant_liver.h5ad](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/tests/data/anndata/compliant_liver.h5ad) AnnData file
    * Specify `raw_location: does_not_exist` in the Expression tab to ensure the extraction fails
3. Complete the upload and wait for the job to fail
4. Go the the [Dev Mixpanel events queue](https://mixpanel.com/project/2085496/view/19055/app/events) and search for an `ingestSummary` event
5. Confirm you see your upload with `jobStatus: failed` and `numFilesExtracted: 0`

### DIFF
--- a/app/models/ingest_job.rb
+++ b/app/models/ingest_job.rb
@@ -1044,7 +1044,7 @@ class IngestJob
     mixpanel_log_props = get_job_analytics
     # log job properties to Mixpanel
     MetricsService.log(mixpanel_event_name, mixpanel_log_props, user)
-    report_anndata_summary if study_file.is_viz_anndata? && !%i[ingest_anndata differential_expression].include?(action)
+    report_anndata_summary if study_file.is_viz_anndata?
   end
 
   # set a mixpanel event name based on action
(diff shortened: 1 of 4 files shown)


app/models/ingest_job.rb
```diff
@@ -1115,10 +1115,20 @@ def extracted_raw_counts?(job)
     extract_params.include?('raw_counts')
   end
 
+  # determine if this job qualifies for sending an ingestSummary event
+  # will return false if summary exists, this is a DE job, or
+  # a successful AnnData extract (meaning downstream jobs are running)
```
eweitz, line 1120: Helpful note!  Looking only at the clause in the code, I was confused.

eweitz (approved):
Code looks good!  Looking forward to more accurate AnnData ingest metrics.

PR #2251, opened 2025-05-07
Adding morphology and electrophysiology facets (SCP-5980)

bistline:
#### BACKGROUND & CHANGES
This adds two new search facets to the "more facets" menu on the homepage: `has_morphology` and `has_electrophysiology`.  Both of these search facets are MongoDB-based, meaning they do not source data from BigQuery for either filter values or search results.  Additionally, these facets have a new property called `is_presence_facet` which means that no exact filter matches are made, but rather only the presence of a given metadatum name within a study.  This update also lays the groundwork for retiring BigQuery entirely for study-based search.  More work is needed, but the main foundation of being able to source filter labels and perform searches directly in MongoDB is now in place.

#### MANUAL TESTING
1. Run the migration to create the new facets with `rails db:migrate`
2. Boot all services and sign in
3. Download the metadata file from [SCP2674](https://singlecell.broadinstitute.org/single_cell/study/SCP2674/phenotypic-variation-of-transcriptomic-cell-types-in-mouse-motor-cortex?scpbr=brain-multi-modal#study-download) and upload it to a new study
4. Once the metadata ingest completes, go back to the home page and click "more facets"
5. Confirm both new facets are present, and that they both return the new study you created when used

### DIFF
--- a/app/controllers/api/v1/search_controller.rb
+++ b/app/controllers/api/v1/search_controller.rb
@@ -238,10 +238,20 @@ module Api
         if @studies.count > 0 && @facets.any?
           sort_type = :facet
           @studies_by_facet = {}
-          @big_query_search = self.class.generate_bq_query_string(@facets)
-          logger.info "Searching BigQuery using facet-based query: #{@big_query_search}"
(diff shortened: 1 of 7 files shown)


eweitz (approved):
Code looks good, and my local manual test passed!

Screenshot 2025-05-08 at 12 01 41 PM

PR #2256, opened 2025-05-19
Updating to sentry-ruby gem for error reporting (SCP-5260)

bistline:
#### BACKGROUND & CHANGES
This update migrates SCP onto the newer `sentry-ruby` and `sentry-rails` gems instead of the deprecated `sentry-raven`.  This will restore server-side stack traces in Sentry for all errors emanating from Rails.

This also addresses a search-related bug found during triage where empty `type` parameters throw an error instead of defaulting to study-based search.

#### MANUAL TESTING
1. Set the value for `SENTRY_DSN` in a terminal session:
```
gcloud secrets versions access latest --secret scp-config-json --project broad-singlecellportal | export SENTRY_DSN=$(jq -r .SENTRY_DSN)
```
2. Initialize your dev environment as normal and boot the server
3. In `app/lib/azul_search_service.rb` (line 40), add `raise 'this is an error'` to simulate a server-side error
4. Sign in and then run any faceted search, confirming you see an error in the UI 
5. Sign into Sentry, load the development environment from the dropdown menu, and confirm you see a new instance of `RuntimeError`
6. Confirm you see a stack trace like below:
Screenshot 2025-05-19 at 10 17 00 AM

7. (Optional): Run a search via `cURL` with no `type` parameter and confirm you get a valid response with a type of `study`:

```
curl -X GET "https://localhost:3000/single_cell/api/v1/search" -H  "accept: application/json" 
...
{"type":"study","terms":null,"term_list":null,"current_page":1,"total_studies":40, ...
```


bistline:
> Code looks good, and my manual test passed. Glad our Ruby error observability will be restored!
> 
> For my manual test, I had to slightly refine the instructions:
> 
> * Use `export SENTRY_DSN=$(gcloud secrets versions access latest --secret scp-config-json --project broad-singlecellportal | jq -r .SENTRY_DSN)`
> * Comment out `development` in line 7 of `error_tracker.rb`

Ah yes, good catch!

PR #2267, opened 2025-06-02
Adding study-creation and study-state Mixpanel events

bistline:
#### BACKGROUND & CHANGES
This update adds two new Mixpanel events: `study-creation` and `study-state`.  These events track when new studies are created, and whenever a study is published or initialized, giving us robust longitudinal data moving forward regarding study state management.  Each event will contain the following properties:

- studyAccession
- created_at: study creation date
- domain: email domain of the user
- numCells
- public: current state of public flag
- last_public_date: date of latest change to public state
- initialized: current state of initialized flag
- last_initialized_date: date of latest change to initialized state

#### MANUAL TESTING
1. Boot as normal and sign in
2. In a separate tab, open the [Dev Mixpanel Events](https://mixpanel.com/project/2085496/view/19055/app/events) feed
3. Back in SCP, create a new study, marking it as private
4. In Mixpanel, find the `study-creation` event and confirm that `public` and `initialized` are both `false` and neither has a date
5. In the Rails console, load your new study
6. Sequentially update both `public` and `initialized` to `true` and confirm you see a `study-state` event with the correct values
7. (Optional): reset `public` and `initialized` to `false` and confirm there are no Mixpanel events

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -724,7 +724,7 @@ module Api
 
           safe_params[param] = val if StudyFile.fields[param.to_s].present?
         end
-        safe_params
+        safe_params.with_indifferent_access
       end
 
       private
--- a/app/lib/metrics_service.rb
+++ b/app/lib/metrics_service.rb
@@ -123,7 +123,6 @@ class MetricsService
       headers: headers,
       payload: post_body
     }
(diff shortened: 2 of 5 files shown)


eweitz (approved):
Code looks good.  Glad we'll get more detailed longitudinal study analytics in Mixpanel!

PR #2275, opened 2025-06-16
Better sorting/uniquifying of facet filters (SCP-6023)

bistline:
#### BACKGROUND & CHANGES
This update fixes a new regression where it is more likely to see duplicate or improperly sorted facet filters since the switch to MongoDB as the data source.  Now, there is more aggressive pruning of duplicates, and facets are sorted case-insensitively.

#### MANUAL TESTING
1. Open a Rails console session and update all search facet filters (there's no migration for this as it will happen automatically whenever a metadata file is updated)
```
SearchFacet.update_all_facet_filters
```
2. Boot as normal and sign in
3. Open any search facet and confirm there are no duplicate filter values, and all filters are sorted alphabetically regardless of case

### DIFF
--- a/app/models/search_facet.rb
+++ b/app/models/search_facet.rb
@@ -402,7 +402,7 @@ class SearchFacet
     end
     return filter_map if is_numeric?
 
-    filter_map.uniq { |filter| [filter[:id].downcase, filter[:name].downcase] }.reject do |filter|
+    filter_map.uniq { |filter| [filter[:id]&.downcase, filter[:name]&.downcase] }.reject do |filter|
       filter[:id].blank? || filter[:name].blank?
     end
   end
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Code looks good!  Does this fix the 2-3 duplicates for "Homo sapiens" and "Mus musculus" filters in the home page's "species" facet?

bistline:
> Code looks good! Does this fix the 2-3 duplicates for "Homo sapiens" and "Mus musculus" filters in the home page's "species" facet?

It should, yes.  There's already code in place to deal with the id format issue (i.e. `:` vs `_` as the delimiter), and this deals with the case sensitivity problem as well.  We should now only have one entry for each.

PR #2285, opened 2025-07-17
Fixing filter search state bug when IDs are duplicated (SCP-6028)

bistline:
#### BACKGROUND & CHANGES
This fixes a bug that was only observed in deployed environments where duplicate facet filter IDs caused search problems leading to orphaned filter values in the UI.  This could be cleared by closing and reopening the facet, but this solution only works until the search is re-run, at which point the bug reappears.  Now, `SearchFacet#get_unique_filter_values` only uses the filter ID for getting unique values.  Additionally, the filter search API call has been replaced with client-side filtering as the round-trip API call is unnecessary, and progressive search is a more common UX for filtering.  As such, the FiltersSearchBar component has been removed entirely.

#### MANUAL TESTING
1. Open a Rails console session and load the `disease` facet:
```
facet = SearchFacet.find_by(identifier: 'disease')
```
2. Make a copy of the public filters, then add in a duplicate value (in this case, tuberculosis):
```
original_filters = facet.public_filters.dup
last_filter = facet.public_filters.last
=> {"id" => "MONDO_0018076", "name" => "tuberculosis"}
facet.public_filters << last_filter
facet.save
=> true
```
3. Boot as normal and run a search in the `disease` facet for `normal`, confirming that filters update in real time but you see the extra value for `tuberculosis` persist
4. Back in the console, reset the facet
```
facet.update(public_filters: original_filters)
=> true
```
5. Refresh the page and confirm the search for `normal` only returns one filter now


app/javascript/components/search/controls/FiltersBoxSearchable.jsx
```diff
@@ -42,14 +43,19 @@ export default function FiltersBoxSearchable({ facet, selection, setSelection, s
    * For example, among the many filters in the "Disease" facet, search
    * for filters matching the term "tuberculosis".
    */
-  async function searchFilters(terms) {
-    const apiData = await fetchFacetFilters(facet.id, terms)
-    const matchingFilters = apiData.filters
-    const hasResults = apiData.query !== '' && matchingFilters.length > 0
+  function searchFilters(terms) {
+    // const apiData = await fetchFacetFilters(facet.id, terms)
+    // const newFilters = apiData.filters
```
eweitz, line 48:
```suggestion
```

eweitz (approved):

PR #2286, opened 2025-07-21
Adding search results text export (SCP-6026)

bistline:
#### BACKGROUND & CHANGES
This update adds a text export option for search results.  This enables a user to export a snapshot of studies containing all the high-level information displayed in the search UI along with direct links to studies both in SCP and HCA.  This feature was requested by a user as a way to evaluate search query results in an offline fashion.

#### MANUAL TESTING
1. Boot as normal and run any search that returns results
2. Confirm you see the `Export` button, and that clicking it exports a text file with your results
3. Run another search with a random nonsense text string and confirm the `Export` button is disabled


eweitz (approved):
Nice enhancement!  Code looks good.

PR #2291, opened 2025-07-29
Adjusting size/spacing of study viz settings menus (SCP-6032)

bistline:
#### BACKGROUND & CHANGES
This update adjusts the size and spacing of the new cluster order menus to be both larger and more consistent with the existing annotation override menu.  Now both menus show a max of 5 items, and each item is the same size across all menus.  Below are some screenshots:

Chrome
chrome

Firefox
firefox

#### MANUAL TESTING
1. Boot as normal and sign in
2. Load any study with multiple clusterings and confirm the menus look identical.

### DIFF
--- a/app/javascript/styles/_forms.scss
+++ b/app/javascript/styles/_forms.scss
@@ -200,14 +200,14 @@ hr.divider {
         margin: 0 4px 0 0;
       }
     }
-    li:hover {
-      background-color: #f4f4f4;
-    }
   }
 }
 
 li.cluster-order {
-  cursor: pointer;
+  cursor: grab;
+}
+li.cluster-order:active {
+  cursor: grabbing;
 }
 
 .large-checkbox {


bistline:
> Code, screenshots, and [demo](https://drive.google.com/file/d/1vUBZHvNtpgJDtPMRXwbQywv8qU6bi8VK/view?t=1m45s) look good!
> 
> I suggest a trivial, non-blocking UX refinement. Also, do we track these drag-and-drop events in Mixpanel?
> 
> Potential opportunities for future enhancement:
> 
> 1. "Move to top / bottom"  E.g. show up / down arrow icon on hovering row, with brief text shown on icon hover.  This could also be done with "right click" [context menus](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus), albeit less discoverably.
> 2. Add a "Sort" ascending / descending icon.  Bonus: asc / desc sort by cell counts.
> 3. More intelligent default sorting.  The demo shows an example of this being used with time-based annotation group labels, which we've noticed are often sorted sub-optimally.  We can likely improve default sorting on 80% of such cases with modest engineering for heuristics to infer day (d) / week (w / wk) / month (m) / year (y).
> 
> All that is beyond scope here, requiring varying amounts of additional engineering. This initial version is a big improvement over the prior state!

I think the first can be done easily.  The second less so, especially by cell count - that information simply isn't available in this view, and getting it in would be tricky.  The last makes sense as well, but it's also more work, and I told the user this would be available tomorrow.  I'll do the first and see if I can easily add the sort buttons as well.

PR #2295, opened 2025-07-31
Update study default cluster orders when CRUDing clusters (SCP-6037)

bistline:
#### BACKGROUND & CHANGES
This addresses a regression introduced in #2289 where once the order has been set for a clustering menu, it will not update further as clusters are added/updated/deleted.  Now, any CRUD action on a cluster will cause an update to the corresponding menu: new clusters are appended to the end, existing clusters that are renamed are replaced, and deleted clusters are removed.  Caches are also cleared asynchronously to avoid UI blocking.

#### MANUAL TESTING
Note: you may need to wait a few seconds in between each update for DelayedJob to clear caches if you have caching enabled
1. Boot all services including DelayedJob and sign in
2. Load any study with multiple clusterings (Chicken raw counts is a good candidate), go to the Settings > View options tab and set the order of one of the menus, then save
3. Go to the upload wizard for this study and rename one of the clusters
4. Go back to the Explore tab and confirm it has been renamed correctly in both the menu and the View options tab
5. Back In the upload wizard, delete a cluster
6. Back in the Explore tab, ensure the cluster has been removed from all menus

### DIFF
--- a/app/controllers/api/v1/study_files_controller.rb
+++ b/app/controllers/api/v1/study_files_controller.rb
@@ -385,7 +385,6 @@ module Api
               study.save
             end
             cluster.update(name: study_file.name)
-            cluster.reload
           end
         end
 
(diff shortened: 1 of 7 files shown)


eweitz (approved):
Code looks good!

PR #2305, opened 2025-09-29
Properly scope existing results query to given cluster (SCP-6053)

bistline:
#### BACKGROUND & CHANGES
This update fixes a bug with DE automation where existing results from other clusters prevent new clustering objects from calculating their own results.  This appears to work fine if a user adds several clusterings all at once, but if there are enough such that downstream parses end after the first batch of results come in, then all downstream DE jobs are blocked for that annotation.  Now, the query for existing results is scoped to the cluster in question.  

This was found by the owner of SCP2817 & SCP3055.  That study has ~20 clusterings each with over 50 eligible annotations in each.  It is worth noting that this would have resulted in thousands of jobs that potentially could have clogged up ingest since we have a quota in the total number of CPUs available at any given point in GCE, and DE jobs are the most compute-intensive jobs.  Luckily, the clusterings are tiny and each job completed in a matter of minutes, so the total batch of jobs would have completed in a few hours.  A manual backfill ran to add results for the cell type annotation in both studies (other annotations were skipped for now).  

However, we may see that the number of DE jobs increases greatly with this bugfix, and we may want to add sufficient guardrails to prevent this, or at least throttle them such that they do not block new file uploads.  This could probably be achieved by putting a cap on the total number of concurrent DE jobs (say, 50) and then adding logic to all DE job submissions to re-queue a job if the cap has been met.  There is already automation to skip submitting new jobs if a matching job exists, so this could be extended to include any DE job and to re-queue jobs when under load.

#### MANUAL TESTING
1. Boot all services as normal and sign in
2. Go to any study with existing DE results
3. Create a copy of one of the clustering files (with a new name) and upload this file as a new cluster
4. Once ingest has completed, confirm that you see new DE jobs launching on existing annotations

### DIFF
--- a/app/lib/differential_expression_service.rb
+++ b/app/lib/differential_expression_service.rb
@@ -64,7 +64,7 @@ class DifferentialExpressionService
                                               skip_existing: false)
     study = Study.find_by(accession: study_accession)
     validate_study(study)
-    eligible_annotations = find_eligible_annotations(study, skip_existing:)
+    eligible_annotations = find_eligible_annotations(study)
     raise ArgumentError, "#{study_accession} does not have any eligible annotations" if eligible_annotations.empty?
 
     log_message "#{study_accession} has annotations eligible for DE; validating inputs"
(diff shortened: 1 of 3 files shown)


eweitz (approved):
Nice robustness fix.

> ~20 clusterings each with over 50 eligible annotations in each

Whoa!  Another benefit of horizontal compute.

PR #2307, opened 2025-10-02
Adding search options for data types (SCP-6051)

bistline:
#### BACKGROUND & CHANGES
This update adds three new search options to filter results by available data types: raw counts, differential expression, and spatial clustering.  This is in response to a user request to allow filtering results by the presence of raw count matrices.  These options are different from the `has [modality]` search facets as we are looking for the presence of files of a certain type (or in the case of DE, results objects) rather than by the contents of the metadata file.  As such, these were put in the `Options` control to avoid having to refactor how `SearchFacet` objects work.

These options will apply to any study-based search request (it doesn't make sense for global gene search).  Additionally, these are all merged into the same search parameter of `data_types` to avoid having to maintain extra search parameters for each.

Additionally, this update fixes a regression in `HcaAzulClient` where `GET` requests no longer work with the `Content-Type` header.  This header is now added dynamically on all non-GET requests.

#### MANUAL TESTING
1. Boot as normal and run a search for `species:Homo sapiens;cell_type:B cell`
2. Confirm you see the `Human milk - differential expression` study
3. Click the data type options for raw counts and differential expression and confirm the search parameter is merged as `data_types:raw_counts,diff_exp`
4. Confirm that you still get the above study in both requests
5. Click the option for spatial and confirm the above study is no longer found

### DIFF
--- a/test/api/search_controller_test.rb
+++ b/test/api/search_controller_test.rb
@@ -226,7 +226,7 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
                  "Did not return correct array of matching accessions, expected #{expected_accessions} but found #{matching_accessions}"
 
     assert_equal @random_seed, json['studies'].first['term_matches'].first
-    assert_equal 2, json['match_by_data']['numResults:scp:text']
+    assert_equal expected_accessions.count, json['match_by_data']['numResults:scp:text']
 
     # test exact phrase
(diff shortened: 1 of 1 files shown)


eweitz (commented):
Being able to filter to studies that have raw counts, differential expression, or spatial data is quite a nice new feature!  Code looks good.

PR #2369, opened 2026-03-26
Enable pathway diagrams to use precomputed dot plot data (SCP-6119)

bistline:
#### BACKGROUND & CHANGES
This update enables pathway diagrams to leverage precomputed dot plot data from `DotPlotService`.  Since pathways often contain large numbers of genes, using precomputed data will vastly improve render times and reduce load on the server when trying to transmit large amounts of dot plot data.

Below is a comparison of a study with 38K cells running in development with caching turned off.
Regular data (~1 min to load all data for single annotation):

https://github.com/user-attachments/assets/d18bd3d4-75b7-4f88-b51c-80e222164401

Precomputed data (each annotation loads in ~5-10s):

https://github.com/user-attachments/assets/95a4b62c-ad78-40c3-8e6c-258f28d6b82d

In production, these render times will be even faster for precomputed data (usually < 1s).

#### MANUAL TESTING
1. Boot as normal and load any study with precomputed dot plot data
2. Query for pathways and confirm data loads and diagrams overlay expression
3. (Optional): Turn off `dot_plot_preprocessing_frontend` feature flag and confirm data still loads for same study

### DIFF
--- a/app/javascript/lib/dot-plot-precompute-patch.js
+++ b/app/javascript/lib/dot-plot-precompute-patch.js
@@ -76,8 +76,17 @@
         // Series 1: percent expressing (for size) - will be scaled to 0-100
         // Data format: values[0] = mean_expression, values[1] = percent_expressing
         geneNames.forEach((gene, i) => {
-          const geneData = data.genes[gene]
-
+          // gotcha for pathway diagrams and precomputed dot plots
+          // this is because the pathway diagram may contain genes that were not included in the dot plot data
+          // (e.g. because they were not expressed)
+          let geneData = data.genes[gene]
+          if (!geneData) {
+            geneData = []
+            // If gene is missing from data, fill with zeros
+            for (let j = 0; j < nCols; j++) {
+              geneData.push([0.0, 0.0])
+            }
+          }
           // Map data from original order to sorted order
           geneData.forEach((values, originalIndex) => {
             const sortedIndex = originalToSortedIndex[originalIndex]
--- a/app/javascript/lib/dot-plot-precompute-patch.js
+++ b/app/javascript/lib/dot-plot-precompute-patch.js
@@ -82,9 +82,9 @@
           let geneData = data.genes[gene]
           if (!geneData) {
             geneData = []
-            // If gene is missing from data, fill with zeros
+            // If gene is missing from data, fill with NaN and 0 cells expressing
             for (let j = 0; j < nCols; j++) {
-              geneData.push([0.0, 0.0])
+              geneData.push([NaN, 0])
             }
           }
           // Map data from original order to sorted order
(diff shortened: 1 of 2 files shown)


app/javascript/lib/dot-plot-precompute-patch.js
```diff
@@ -76,8 +76,17 @@
...
-
+          // gotcha for pathway diagrams and precomputed dot plots
+          // this is because the pathway diagram may contain genes that were not included in the dot plot data
+          // (e.g. because they were not expressed)
+          let geneData = data.genes[gene]
+          if (!geneData) {
+            geneData = []
+            // If gene is missing from data, fill with zeros
+            for (let j = 0; j < nCols; j++) {
+              geneData.push([0.0, 0.0])
```
eweitz: Does pushing something like `[null, null]` or `[undefined, undefined]` or `[-1, -1]` work?  

A downside of `[0.0, 0.0]` is that it conveys "assayed and found to be not expressed in any cells" but that's also there used for "not assayed".  In other words, this conflates "gene is in matrix and reported as no expression" with "gene is not in matrix".

To be clear, this downside is much better than the current downside of broken pathway overlays for preprocessed data!

But if we could easily preserve semantic clarity, that'd be ideal.

eweitz (approved):
Kudos!  Thank you for supercharging pathway expression overlays.

I note an issue, but it doesn't strike me as a blocker.

app/javascript/lib/dot-plot-precompute-patch.js
bistline: The reason I chose `[0.0, 0.0]` is because of this code here: https://github.com/broadinstitute/single_cell_portal_core/pull/2369/changes#diff-90b848ec8be7e00df37e1c2edb2cf8e8ed40db7e6d31aa9bc4b885b010fcfc44L89

Since the legacy handling will always convert a `nil` into `0.0` when loading gene expression data from Rails, it seemed safest to copy that behavior so that the overlay will convert that to `NaN`.  I suppose we could also just pass `[NaN, 0]` as well... thoughts?

eweitz: That might work!  

I think a good test case is e.g. the "Nav1.6" gene in the pathway "Chronic hyperglycemia impairment of neuron function" (WP5283), [live example here](https://singlecell.broadinstitute.org/single_cell/study/SCP1671/cellular-and-transcriptional-diversity-over-the-course-of-human-lactation?pathway=WP5283).  WikiPathways sometimes (non-optimally) labels human genes with various non-HGNC names.

Currently, hovering over e.g. "Nav1.6" shows no tooltip.  My (admittedly somewhat speculative) concern is that `[0.0, 0.0]` here would show a tooltip explicitly reporting no expression in any cells.  If `[NaN, 0]` (or `[null, null]` or `[undefined, undefined]` or even `[-1, -1]`) works here, that'd be preferable.  The intent would be to help preserve the distinction between "not found" and "found, 0" in cases like "Nav1.6" above.

The distinction is already lacking in the graphic itself -- both cases show no color for such gene nodes.  So if these ideas aren't trivial or don't work out of the box, then I support merging.

bistline: [2b61f08](https://github.com/broadinstitute/single_cell_portal_core/pull/2369/commits/2b61f08952fdf9d11b13b48586368552010a477d)

PR #2399, opened 2026-08-28
Fixing uncaught regression with datetime to_s, duplicate workspace creation (SCP-6156)

bistline:
#### BACKGROUND & CHANGES
This fixes a newly found regression with calling DateTime#to_s.  This has been replaced with `to_fs` and has the same signature as before.  This also fixes another regression with new studies that have shares created inline causing duplicate workspace creation when callbacks are called on associated records (like study_shares).  This now disables those nested validations during study creation only.

Additionally, the google-cloud-storage gem (and dependencies) were updated to remove deprecation warnings from a transitive dependency in `multi_json`.

#### MANUAL TESTING
1. Boot as normal and sign in
2. Pull up the study details page for any study and confirm it loads without error


eweitz (approved):

bistline:
Closing this for now as there are too many issues to deal with quickly, so reverting the upstream PR that caused the problems in the first place.

PR #2405, opened 2026-09-01
Addressing Vite, eslint security updates (SCP-6154)

bistline:
#### BACKGROUND & CHANGES
This update addresses multiple security updates surrounding Vite and its dependencies.  Additionally, this updates `eslint` and associated dependent packages.  This also removes an earlier framework defaults file that  mistakenly got restored in #2400.


bistline:
Build failure seems to be because of a momentary blip on OLS (SearchFacetPopulatorTest which calls out to get ontology names/descriptions).

eweitz (approved):

PR #2416, opened 2026-09-17
Updating to rails-baseimage:3.1.1, fixing Azul 400 errors (SCP-6172, SCP-6171)

bistline:
#### BACKGROUND & CHANGES
This updates to the latest version of `rails-baseimage`, which includes security patches for both the base operating system and for Passenger and nginx.  Additionally, this updates to Ruby 3.4.10 and the most recent stable version of `bundler`.

Lastly, this fixes the new regression from Azul resulting in `BadRequest` responses for project-based queries.  `MAX_RESULTS` is now 50, down from 75.


eweitz (approved):

bistline:
Build failure looks like a race condition with permission delegation in GCS, not reproducible

